PR #85, opened 2020-03-11
Fix repository URL into package.json

antleblanc:
Related to #84

Signed-off-by: Antoine Leblanc <ant.leblanc@gmail.com>


hsablonniere:
Good catch, I went to fast while doing in on 3 different projects :-(

PR #87, opened 2020-03-14
Replace `<env-var-full>` with updated `<env-var-form>` and `<env-var-linked-services>`

hsablonniere:


hsablonniere:
This branch was merged too soon by mistake, we reverted the change.

PR #88, opened 2020-03-14
Replace `<env-var-full>` with updated `<env-var-form>` and `<env-var-linked-services>` (the good one)

hsablonniere:


hsablonniere:
FYI:

* whitelisting `<br>` was not done in the right commit
* a type fix for i18n fr `cc-elasticsearch-options.description` was not done in the right commit

PR #98, opened 2020-04-16
Add <cc-tcp-redirection>

### LINKED ISSUE #90: Create a new component for TCP redirections
## Context

We need a UI component so a user can manage the TCP redirections of an app.

## User actions

For a given app:

* A user can see the list of existing TCP redirections
* A user can create a new TCP redirection
* A user can delete an existing TCP redirection

## Rules and notes

(some rules may evolve later)

* A user can only create one TCP redirection per namespace for a given app.
  * The namespace must be linked to / authorized for the app's organisation.
* The public TCP port is selected by the system and presented to the user.
* The inner TCP port is always `4040`.
* We can assume a given organisation won't have access to lots of namspaces (at least 2, max would be around 5/7).
* WORDING: We should always mention TCP in upper case.
* WORDING: We should always use `redirections` and not `redirs`

## Input requirements

* For the list of existing redirections, we need:
  * the name of the namespace
  * the port number
  * we may need some metadata about the namespace (is it related to cleverapps.io? Is it related to the custom DNS?)
  * we may need the list of domains (favorite?, cleverapps.io? ...) so we can display help/hint about a redirection.
* For the redirection creation, we need the list of namespaces and if a redirection already exists for this namespace (and this app).
* API for errors to be clarified

## Output requirements

* Event on creation
* Event on deletion

## UI ideas

* As always, we'd like to avoid "DB driven UI design", so if we can do better than a list with remove buttons and a form to add a line in the list, it would be great.
* The "only one redir per namespace" and the "namespace list will be `2 <= count <= ~7` should help us to find a better idea.

## Stories

* Skeleton (no data yet)
  * I don't think we need partial skeleton states
* Empty
  * Data loaded but no redirections yet
  * Data loaded  but no redirections yet and no namespace linked to the organisation
* Data loaded with a redirection on cleverapps.io
* Data loaded with a redirection on a custom domain
* Data loaded with a redirection on a "dedicated namespace"
* Data loaded with 10 redirections (10 namespaces)
* Error while loading data
* Error while adding a new redirection
* Error while removing an existing redirection

NOTES:

* We don't need a readonly state yet.

## Questions

* One internal state and one update button (like env-vars) vs. each button (create/delete) triggers an event (to trigger an API call)?
  * I think the second option fits the model better API wise
* Do we want to let users interact with the form while we're waiting for the API over a creation or deletion?
  * It would be way easier for us to go this way in terms of state management outside of component and regarding error management and display.

## Sub-components?

Depending on the different cases we have in the display of a given redirection, we may want to create a sub component for one redirection like we did with `<env-var-input>`.

## Unknowns

* We don't know much about how this will be billed yet

urcadox:
Closes #90


hsablonniere:
Feedbacks from the team:

* No confirm on deletion, what about a delay on the button?
* Why is cleverapps.io in italics?

hsablonniere:
We forgot to handle the **empty** story.

urcadox:
A suggestion for the description (in French):

```
Un espace de nommage correspond à un groupe de frontaux. Il y a deux groupes de frontaux publics :

    "cleverapps" pour les frontaux servant les domaines en cleverapps.io
    "default" pour les frontaux servant vos domaines personnels (exemple.com)

et si l'orga est premium :

Vous avez également accès aux namespaces liés à vos frontaux privés :

    namespace 1
    namespace 2
```

urcadox:
An idea was to specify if a namespace is public / private and to specify the CNAME record so people identify better what we are talking about.

Here are my thoughts.

The public/private idea is fine, it helps to understand that a private namespace is indeed, well, private.

The CNAME one though just does not work in my opinion, mainly because of the shared default and cleverapps namespace across zones. Specifying it only for the private namespaces would be fine but I'm not sure that it helps in any way, this should just be added in the domain names section (it's not at the moment).

Instead, we should write a rather extensive documentation with examples and link to this documentation from this UI. It allows us to make something simpler, less convoluted and less scary.

judu:
I agree with the CNAME and avoiding the "seems to be only available for domain X" risks.

Extensive documentation is obviously mandatory, like for everything in Clever Cloud.

What about a simple example like www.whatsfordinner.com VS whatsfordinner.cleverapps.io? Like with that exact domain, something that says "it is an example, but you get it!". Or does that adds too much garbage and complexity to the interface?

hsablonniere:
I think we won't be able to explain everything in the block and we need to refer to this: https://www.clever-cloud.com/doc/clever-tools/tcp-redirections/

hsablonniere:
## `<cc-tcp-redirection>`

### UI

* The delete button should have a delay (3 secs should be enough)
* If we do this, the delete button will be almost 3 times wider than the create button (in french). We may want to force equal width somehow.
* :ballot_box_with_check: 
* WORDING: other sites seems to say `noms de domaine` and not `noms de domaines`.
* :ballot_box_with_check: Rest of the wording is nice
* :ballot_box_with_check: Error state with retry button is nice
* I have an idea to improve waiting state
* :ballot_box_with_check: Story names and structure is nice

### Code

* :ballot_box_with_check: Docs is nice
* else in `_getHelpText` can be removed
* `isRedirectionDefined = this.sourcePort != null` is done multiple times, we should greate a private method (or getter)
* else in `_getHelpTextAddendum` can be removed
* else if can be just a if and else can be removed in `_getIconUrl` 
* can we move the i18n for the button in a getter?
* :ballot_box_with_check: CSS is nice

## `<cc-tcp-redirection-form>`

### UI

* :ballot_box_with_check: Wording is nice
* :ballot_box_with_check: Empty state is nice
* :ballot_box_with_check: Simulations are awesome, the cc-expand helps a lot with the height jumps

### Code

* DOCS: Type def for `Redirection` is missing `private` and `error`
* DOCS: We can remove the warning in `## Details` about errors
* :ballot_box_with_check: HTML and CSS are nice
* STORY: You don't need to reset `componentError.error = true;` each time in the simulation.

hsablonniere:
OK, everything is ready, thank you @urcadox for this work, this is great!!

PR #121, opened 2020-05-03
cc-env-var-*: rename all i18n keys (forgot to do it before)

hsablonniere:


hsablonniere:
Grrr, it was not a FF...

Issue #140, opened 2020-08-18
Add-on options


BlackYoup:
Our add-ons will have more and more options over time. Currently, we did the `<cc-elasticsearch-options>` component to handle Elastic Stack options. We now have the `encrypt at rest` option that will be available for other add-ons like PostgreSQL, MySQL, MongoDB, ...

The idea is to have the options system for add-ons a bit more decoupled. Here are the components will we create, from bottom to top:

## `<cc-addon-option>`

`<cc-addon-option>` will be the container of the option. It will have:
- 2 attributes / properties: `title` and `logo`. Title is the option's name, logo is its logo.
- A default slot which will have the "description" of the option.
- Emit an event when the enable/disable toggle is hit.

## `<cc-addon-option-form>`

`<cc-addon-option-form>` will be the container of the possible multiple `<cc-addon-option>`. It will have:
- 2 attributes / properties: `title`, `options`. Title is the title of the add-on options. Currently for elastic it's `Options for the Elastic Stack`. Options is an array of options, see below. The component will iterate over this options array and pass the properties to the `<cc-addon-option>` components.
- Emit an event when the `confirm options` button is hit with the options states (enabled / disabled)

## `<cc-addon-elasticsearch-options>, <cc-addon-postgresql-options>, <cc-addon-mysql-options>, <cc-addon-mongodb-options>`

Those components will be the global components to display those options. They have the `i18n` for each add-on and instantiate a `<cc-addon-option-form>` with the correct properties. It will have:
- 1 attribute / property: `options`. Options is an array of the available options for this add-on. See below for the format.

### Options

It will be an array of objects where each object has:
- A logo (String). e.g: APM Logo
- A title (String). e.g: APM
- description (DOM): a Lit Element dom element with the option's description
- name (String): Internal to the `<cc-addon-option-form>` to identify the option's event.

hsablonniere:
Reading this again, I wonder if we should/could remove the `-addon` for the ES, PG, MS and mongo components.

BlackYoup:
I don't think we should drop it. If we ever need another specific component for those add-ons, it would be weird to have `<cc-addon-postgresql>` and `<cc-addon-postgresql-whatever>`. When we will have to redo all the add-ons dashboard, we might need such component.

BlackYoup:
I'm closing this one for now, I'll open a new issue for the encryption option for some of the add-ons.

PR #142, opened 2020-08-21
Add <cc-addon-option>, <cc-addon-option-form>, <cc-addon-elasticsearch-options> and remove <cc-elasticsearch-options>

BlackYoup:
This PR adds three new components to handle add-ons options. It's aim is to make it easier to add options for add-ons. I replaced the current `<cc-elasticsearch-options>` component with the new one, `<cc-addon-elasticsearch-options>` that will use the new components.

Part of #140


src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * @slot - The content of the option's description (text or HTML).
+ */
+export class CcAddonOption extends LitElement {
+
+  static get properties () {
+    return {
+      title: { type: String, attribute: 'title' },
+      logo: { type: String, attribute: 'logo' },
+      enabled: { type: Boolean, attribute: 'enabled' },
```
hsablonniere, line 39: Specifying the attribute name is only required if the property name is camel case (to have a kebab case attribute). When it's a single word, you can just omit `attribute`.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  _onToggleOption () {
+    this.enabled = !this.enabled;
+    dispatchCustomEvent(this, 'input', this.enabled);
+  }
+
+  render () {
+    const logo = this.logo;
+    const title = this.title;
```
hsablonniere, line 57: You should use `this.logo` and `this.title` directly in your template.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+    const choices = [{
+      label: i18n('cc-addon-option.disabled'), value: false,
+    }, { label: i18n('cc-addon-option.enabled'), value: true }];
+
+    return html`
+      <div class="option ${classMap({ 'option--enabled': this.enabled })}">
```
hsablonniere, line 63: If you add `reflect:true` for `enabled` in `static get properties()`, you'll be able to style without needing this. It's not a 100% strict rule but in this case it works well I think.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+      </div>
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          /* You may use another display type but you need to define one. */
```
hsablonniere, line 79: Remove this ;-)

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        .option-main {
+          display: grid;
+          grid-gap: 0.5rem;
+        }
+
+        .option-name {
+          font-weight: bold;
+          line-height: 1.6rem;
```
hsablonniere, line 95: Try to use unit-less line-height.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/addon/cc-addon-option.js)
+ *
+ * ## Details
+ *
+ * * If `enabled` is set to true, the option will be enabled by default. Otherwise, it will be disabled.
+ * * Emits an `input` event when the option is enabled or disabled.
+ *
+ * ## Technical details
+ * * If you want to display a warning in the option, add the `option-warning` class to an HTML element. It will be displayed
```
hsablonniere, line 22: Needs a blank like between title and list.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+ * A component that allows to enable or disable an add-on option.
+ *
+ * * 🎨 default CSS display: `block`
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/addon/cc-addon-option.js)
+ *
+ * ## Details
+ *
+ * * If `enabled` is set to true, the option will be enabled by default. Otherwise, it will be disabled.
+ * * Emits an `input` event when the option is enabled or disabled.
```
hsablonniere, line 19: I'm not sure we need those details. The `enabled` is already documented in properties and the `cc-addon-option:input` is also documented in events.

src/addon/cc-addon-option.js
```diff
@@ -0,0 +1,139 @@
...
+    this.enabled = !this.enabled;
+    dispatchCustomEvent(this, 'input', this.enabled);
+  }
+
+  render () {
+    const logo = this.logo;
+    const title = this.title;
+    const choices = [{
+      label: i18n('cc-addon-option.disabled'), value: false,
+    }, { label: i18n('cc-addon-option.enabled'), value: true }];
```
hsablonniere, line 60:
```suggestion
    const choices = [
      { label: i18n('cc-addon-option.disabled'), value: false },
      { label: i18n('cc-addon-option.enabled'), value: true },
    ];
```

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * A component that displays a form of `<cc-addon-option>`.
+ *
+ * * 🎨 default CSS display: `block`
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/addon/cc-addon-option-form.js)
+ *
+ * ## Details
+ *
+ * * Emits a `submit` event when the form is submited. The emitted event is an object of options state (enabled / disabled).
```
hsablonniere, line 17: The event is already documented below.

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * ## Type definitions
+ *
+ * ```js
+ * interface Option {
+ *   title: String,
+ *   logo: String,
+ *   description: DOM,
```
hsablonniere, line 25: Could also be just a string without HTML so I would put `HTMLElement | string`

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+ *   name: String,
+ *   enabled: Boolean
+ * }
+ *
+ * interface Options {
+ *   optionName1: Boolean,
+ *   optionName2: Boolean,
```
hsablonniere, line 32: I think there's another way to describe this in TypeScript interfaces. Here it seems like a strict `optionName1`.

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+ * @event {CustomEvent<Options>} cc-addon-option-form:submit - Fires when the form is submitted.
+ *
+ * @slot description - The description of the add-on and available options.
+ */
+export class CcAddonOptionForm extends LitElement {
+
+  static get properties () {
+    return {
+      title: { type: String, attribute: 'title' },
+      options: { type: Array, attribute: 'options' },
```
hsablonniere, line 48: Same as other component, the attribute is optional in this case.

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+ *   optionName2: Boolean,
+ * }
+ * ```
+ *
+ * @prop {String} title - Title of the whole options form.
+ * @prop {Array<Option>} options - List of Option object to render.
+ *
+ * @event {CustomEvent<Options>} cc-addon-option-form:submit - Fires when the form is submitted.
+ *
+ * @slot description - The description of the add-on and available options.
```
hsablonniere, line 41: Do we need a named slot in this case?

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+      </cc-block>
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          /* You may use another display type but you need to define one. */
```
hsablonniere, line 107: Remove this.

stories/addon/cc-addon-option-form.stories.js
```diff
@@ -0,0 +1,62 @@
...
+  description: htmlExample,
+  enabled: false,
+  name: 'kibana',
+}, {
+  title: 'APM',
+  logo: 'https://static-assets.cellar.services.clever-cloud.com/logos/elasticsearch-apm.svg',
+  description: htmlExample,
+  enabled: true,
+  name: 'apm',
+}];
```
hsablonniere, line 33:
```suggestion
const optionsExamples = [
  {
    title: 'Kibana',
    logo: 'https://static-assets.cellar.services.clever-cloud.com/logos/elasticsearch-kibana.svg',
    description: htmlExample,
    enabled: false,
    name: 'kibana',
  },
  {
    title: 'APM',
    logo: 'https://static-assets.cellar.services.clever-cloud.com/logos/elasticsearch-apm.svg',
    description: htmlExample,
    enabled: true,
    name: 'apm',
  }
];
```

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+    super();
+    this.title = null;
+    this.options = [];
+    this._optionsStates = {};
+  }
+
+  _onSubmit () {
+    // If some options were not changed, fill them here
+    this.options.forEach((option) => {
+      if (this._optionsStates[option.name] === undefined) {
```
hsablonniere, line 62: Please always use `foobar == null` instead of `foobat === undefined`

src/addon/cc-addon-option-form.js
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    dispatchCustomEvent(this, 'submit', this._optionsStates);
+  }
+
+  _onOptionToggle (optionName) {
+    return ({ detail }) => {
+      this._optionsStates[optionName] = detail;
+    };
+  }
```
hsablonniere, line 74: In other components we don't return an arrow function in an event handler, we pass the event object directly in the template.

The handler would look like this:

```js
_onOptionToggle ({ detail }, optionName) {
  this._optionsStates[optionName] = detail;
}
```

and in the template, you do this:

```
@cc-addon-option:input=${(e) => this._onOptionToggle(e, option.name)}
```

src/addon/cc-addon-elasticsearch-options.js
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * A component that displays the available options of an elasticsearch add-on.
+ *
+ * * 🎨 default CSS display: `block`
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/addon/cc-addon-elasticsearch-options.js)
+ *
+ * ## Details
+ *
+ * * Emits a `submit` event when the form is submited. The emitted event is an object of options state (enabled / disabled).
```
hsablonniere, line 20: Not necessary.

src/addon/cc-addon-elasticsearch-options.js
```diff
@@ -0,0 +1,146 @@
...
+      </cc-addon-option-form>
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          /* You may use another display type but you need to define one. */
```
hsablonniere, line 138: We can remove this ;-)

hsablonniere (changes requested):

src/addon/cc-addon-option-form.js
BlackYoup: I felt like it needed a name just to know what it was about because the options will also be rendered in this component.
```
<cc-addon-option-form .options=${options}>
  my html
</cc-addon-option-form>
```

I found that this layout doesn't really describe what's going on. But I can remove it if you prefer to only have named slots when needed.

hsablonniere: I wanted to always use the default when we just need one slot but what you're saying makes sense. Let's keep it like that and we'll see in the future...

hsablonniere (approved):

hsablonniere:
:clap:

PR #147, opened 2020-08-27
cc-addon-option: fix styling

BlackYoup:
This fixes some missing styles (toggle color and border radius) for the `<cc-addon-option>` component

### DIFF
--- a/src/addon/cc-addon-option.js
+++ b/src/addon/cc-addon-option.js
@@ -7,7 +7,7 @@ import { i18n } from '../lib/i18n.js';
 /**
  * A component that allows to enable or disable an add-on option.
  *
- * * 🎨 default CSS display: `block`
+ * * 🎨 default CSS display: `grid`
  * <br>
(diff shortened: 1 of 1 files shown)


src/addon/cc-addon-option.js
```diff
@@ -7,7 +7,7 @@ import { i18n } from '../lib/i18n.js';
 /**
  * A component that allows to enable or disable an add-on option.
  *
- * * 🎨 default CSS display: `block`
+ * * 🎨 default CSS display: `flex`
```
hsablonniere, line 10: You used grid in the code.
BTW, I have a branch to automatically get this from code but it's fragile so for now, we need to be thorough.

hsablonniere (changes requested):

src/addon/cc-addon-option.js
BlackYoup: I was sure it was flex..

PR #149, opened 2020-08-27
Add encryption option and components for PostgreSQL, MySQL, MongoDB

### LINKED ISSUE #145: Encryption at-rest add-on option
Some add-ons will need an encryption at-rest option. Those add-ons are, for now:
- PostgreSQL
- MySQL
- MongoDB

## Encryption at-rest option

The encryption at-rest option is an option to encrypt at-rest the disk of the add-on.

### Description

- What it does
- The price (Free for now. 10% of the plan price when we will start to bill it)
- A link to our documentation, for example https://www.clever-cloud.com/doc/addons/postgresql/#encryption-at-rest

## Create the corresponding `<cc-addon-postgresql-options>`, `<cc-addon-mysql-options>`, `<cc-addon-mongodb-options>`

Each add-on will have the same option, only the documentation link will change.

The option will be disabled by default.

BlackYoup:
Fixes #145

### DIFF
--- a/src/addon/cc-addon-mongodb-options.js
+++ b/src/addon/cc-addon-mongodb-options.js
@@ -1,0 +1,91 @@
+import './cc-addon-option-form.js';
+import '../molecules/cc-error.js';
+import { css, html, LitElement } from 'lit-element';
+import { dispatchCustomEvent } from '../lib/events.js';
+import { i18n } from '../lib/i18n.js';
+import { ccAddonEncryptionAtRestOption } from '../templates/cc-addon-encryption-at-rest-option.js';
+
+const ENCRYPTION_DOCUMENTATION_URL = 'https://www.clever-cloud.com/doc/addons/mongodb/#encryption-at-rest';
+
+/**
+ * A component that displays the available options of a MongoDB add-on.
+ *
+ * * 🎨 default CSS display: `block`
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/addon/cc-addon-mongodb-options.js)
+ *
+ * * ## Type definitions
+ *
+ * ```js
+ * interface Option {
+ *   name: string,
+ *   enabled: boolean,
+ *   // Option specific params
+ *   price: number, // for "encryption" option
+ * }
+ * ```
+ *
+ * ```js
+ * interface Options {
+ *   encryption: boolean,
+ * }
+ * ```
+ *
+ * @prop {Option[]} options - List of options for this add-on.
+ *
+ * @event {CustomEvent<Options>} cc-addon-mongodb-options:submit - Fires when the form is submitted.
+ */
+export class CcAddonMongodbOptions extends LitElement {
+
+  static get properties () {
+    return {
+      options: { type: Array },
+    };
+  }
+
+  constructor () {
+    super();
+    this.options = [];
+  }
+
+  _onFormOptionsSubmit ({ detail }) {
+    dispatchCustomEvent(this, 'submit', detail);
+  }
+
+  _getFormOptions () {
+    return this.options
+      .map((option) => {
+        switch (option.name) {
+          case 'encryption':
+            return ccAddonEncryptionAtRestOption(ENCRYPTION_DOCUMENTATION_URL, option);
+          default:
+            return null;
+        };
+      })
+      .filter((option) => option !== null);
+  }
+
+  render () {
+    const options = this._getFormOptions();
+    const title = i18n('cc-addon-mongodb-options.title');
+
+    return html`
+      <cc-addon-option-form title="${title}" .options=${options} @cc-addon-option-form:submit="${this._onFormOptionsSubmit}">
+        <div slot="description">${i18n('cc-addon-mongodb-options.description')}</div>
+      </cc-addon-option-form>
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-addon-mongodb-options', CcAddonMongodbOptions);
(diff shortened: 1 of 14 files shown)


src/templates/cc-addon-encryption-at-rest-option.js
```diff
@@ -0,0 +1,25 @@
+import '../molecules/cc-error.js';
+import { html } from 'lit-element';
+import { assetUrl } from '../lib/asset-url.js';
+import { i18n } from '../lib/i18n.js';
+
+const LOGO = assetUrl(import.meta, '../assets/encryption-at-rest.svg');
```
hsablonniere, line 6: The current convention is `camelCaseNameOfTheImageSvg` with a `Svg` suffix.

src/addon/cc-addon-postgresql-options.js
```diff
@@ -0,0 +1,83 @@
...
+
+  _onFormOptionsSubmit ({ detail }) {
+    dispatchCustomEvent(this, 'submit', detail);
+  }
+
+  _getFormOptions () {
+    return this.options.map((option) => {
+      switch (option) {
+        case 'encryption': return ccAddonEncryptionAtRestOption(ENCRYPTION_DOCUMENTATION_URL, this.encryptionPrice);
+        default: return null;
```
hsablonniere, line 55: Not a great fan of one liners like this. Is ESLint OK with that? If so, let's keep it and I'll update the ESLint config :p

src/addon/cc-addon-postgresql-options.js
```diff
@@ -0,0 +1,83 @@
...
+      };
+    }).filter((option) => option !== null);
```
hsablonniere, line 57: In those situations, you can align the `.map` and `.filter`

src/addon/cc-addon-mysql-options.js
```diff
@@ -0,0 +1,83 @@
...
+ * @prop {Number} encryptionPrice - Price of the encryption option.
+ * @prop {String[]} options - List of options to enable for this selection.
+ *
+ * @event {CustomEvent<Options>} cc-addon-mysql-options:submit - Fires when the form is submitted.
+ */
+export class CcAddonMysqlOptions extends LitElement {
+
+  static get properties () {
+    return {
+      options: { type: Array, attribute: 'options' },
```
hsablonniere, line 35: No need for `attribute` there.

src/addon/cc-addon-mongodb-options.js
```diff
@@ -0,0 +1,83 @@
...
+ * @prop {Number} encryptionPrice - Price of the encryption option.
+ * @prop {String[]} options - List of options to enable for this selection.
+ *
+ * @event {CustomEvent<Options>} cc-addon-mongodb-options:submit - Fires when the form is submitted.
+ */
+export class CcAddonMongodbOptions extends LitElement {
+
+  static get properties () {
+    return {
+      options: { type: Array, attribute: 'options' },
```
hsablonniere, line 35: No need for `attribute` here.

src/addon/cc-addon-postgresql-options.js
```diff
@@ -0,0 +1,83 @@
...
+ * @prop {Number} encryptionPrice - Price of the encryption option.
+ * @prop {String[]} options - List of options to enable for this selection.
+ *
+ * @event {CustomEvent<Options>} cc-addon-postgresql-options:submit - Fires when the form is submitted.
+ */
+export class CcAddonPostgresqlOptions extends LitElement {
+
+  static get properties () {
+    return {
+      options: { type: Array, attribute: 'options' },
```
hsablonniere, line 35: No need for `attribute` here.

hsablonniere (changes requested):
I haven't looked at the result in a browser yet.

src/addon/cc-addon-postgresql-options.js
BlackYoup: Yes, eslint is ok :p

src/templates/cc-addon-encryption-at-rest-option.js
BlackYoup: Should be fixed now

src/addon/cc-addon-postgresql-options.js
BlackYoup: Updated, please double check to see if I correctly understood you

src/addon/cc-addon-postgresql-options.js
hsablonniere: OK so let's keep it like that, we'll update ESLint later.

src/addon/cc-addon-postgresql-options.js
hsablonniere: This is what I expected.

hsablonniere:
I didn't understood at first that options were opt-in with `options`.
We need a story to see what the component looks like when `options` is an empty array.
Will we need an empty message?

I see `encryptionEnabled` in:

* the JSDoc header
* the `static get properties () {}`
* the `constructor`

but it's never used in the component template and never used by the story.

BlackYoup:
> I didn't understood at first that options were opt-in with options.

Yes, options might be different across add-ons plans

> We need a story to see what the component looks like when options is an empty array.

Yeah, I didn't make one because I assumed that you need to provide at least one option, otherwise this option selection shouldn't even be displayed. I can indeed add a `there are no options` text if you think it's worth adding such a case.

I fixed the `encryptionEnabled` boolean not correctly displayed and added a story with the , sorry for the silly mistake.

hsablonniere:
> This option is currently free but will then be billed 10.0% of the plan's price. If this option is enabled, it will be billed €10.00 per month.

I still think it's a bit confusing.

hsablonniere:
About the translation, @urcadox suggests:

> This option is currently free. In the future, it will be billed 10% of the plan price, which amounts to €10 per month here.

hsablonniere:
I think we should move `options` from `String[]` to `Option[]` where `Option` has a `name`, a `enabled` boolean and some custom options like `price` for encryption or `flavor` for apm and kibana.

PR #154, opened 2020-09-07
Refactor map components and introduce zones

hsablonniere:


src/maps/cc-logsmap.js
```diff
@@ -83,15 +86,84 @@ export class CcLogsMap extends LitElement {
...
+    // Add point now
+    if (this._pointsByCoords[coords] == null) {
+      this._pointsByCoords[coords] = [];
+    }
+    this._pointsByCoords[coords].push(newPoint);
+
+    this._updatePoints();
+
+    // Schedule delete point
+    setTimeout(() => {
```
BlackYoup, line 125: Don't you want to abort the timeout when the component is "dropped"?

src/zones/cc-zone-input.js
```diff
@@ -0,0 +1,370 @@
...
+ * A input component to select a zone with a map and a list.
+ *
+ * * 🎨 default CSS display: `grid`
+ * <br>
+ * 🧐 [component's source code on GitHub](https://github.com/CleverCloud/clever-components/blob/master/src/zones/cc-zone-input.js)
+ *
+ * ## Details
+ *
+ * * When `zones` is nullish, a skeleton screen UI pattern is displayed (loading hint).
+ * * Zones are sorted in the list using `tags`. Clever Cloud, then private, the alphanum sort on the city name.
```
BlackYoup, line 25: did you mean `then alphanum`?

src/zones/cc-zone-input.js
```diff
@@ -0,0 +1,370 @@
...
+      this.shadowRoot.querySelector(`input[id=${name}]`).scrollIntoView({
+        behavior: 'smooth',
+        block: 'nearest',
+      });
+    }, 200);
+  }
+
+  _renderZoneInput (zone) {
+
+    // Simplified template withou label/input when skeleton is enabled
```
BlackYoup, line 198:
```suggestion
    // Simplified template without label/input when skeleton is enabled
```

src/maps/cc-map.js
```diff
@@ -223,109 +201,128 @@ export class CcMap extends withResizeObserver(LitElement) {
...
-
-    // Compute total count of the points at those coordinates
-    // We don't use this value for the tooltip for now
-    const totalCount = this._points
-      .filter((p) => p.coords === point.coords)
-      .map((p) => p.count)
-      .reduce((a, b) => a + b, 0);
+    const obsoleteIds = new Set(Object.keys(this._pointsCache));
+
+    // Create new markers or update them if there's already a one with the same tag at the same coordinates
```
BlackYoup, line 212:
```suggestion
    // Create new markers or update them if there's already one with the same tag at the same coordinates
```

src/zones/cc-zone-input.js
hsablonniere: I meant "classic alphanumeric sort". I fixed that and also replaced the "the" with "then".

src/maps/cc-logsmap.js
hsablonniere: You mean if the component is removed from the DOM?
It's the code we had in `<cc-map>`, as is, I just move it.
Maybe we could investigate this later...

https://github.com/CleverCloud/clever-components/issues/158

hsablonniere:
@BlackYoup Thanks for the review :+1:

src/maps/cc-logsmap.js
BlackYoup: > You mean if the component is removed from the DOM?

Yes. I don't like having timeouts that won't do anything anymore running in the background. Sure, let's check it later :+1:

Issue #186, opened 2020-12-28
cc-addon-backups: update component to handle PostgreSQL, MySQL, MongoDB and Redis backups


BlackYoup:
Currently this component can only display Elasticsearch backups, it should be able to display others, including those in the title.

BlackYoup:
Closing this one as the linked PR did them all.

PR #188, opened 2020-12-28
cc-addon-backups: add backups for PostgreSQL, MySQL, MongoDB and Redis

BlackYoup:
This adds support for PostgreSQL, MySQL, MongoDB and Redis backups to the `<cc-addon-backups>` component.

All of these backups do not yet have a way to be deleted so if the `deleteCommand` is null, we don't display the `delete` link.

At least redis (and there will be Jenkins too in a near future) can't be restored by the users. So if the `restoreCommand` is null, the `restore` link is still displayed but the modal then informs the user to open a ticket to the support team.

WIP while I test how it goes integrating this and to see if I can manage adding Jenkins at the end.

Part of #186


hsablonniere:
My review:

* [ ] UI/UX
* [ ] i18n
* [ ] code
* [ ] story

hsablonniere:
Improvement idea: should/could we link to the tools' docs we're refering to:

* cURL
* pg_dump / pg_restore
* mysqldump / mysql
* mongodump / mongorestore

hsablonniere:
In the docs comments, we should update `interface Backup` because `deleteCommand` is now optionnal.

```
deleteCommand?: string,
```

hsablonniere:
NOTE: We could refactor/group stuffs in the story but clearly, I don't really mind. It works, let's keep it like that.

hsablonniere (changes requested):
Small docs adjustments

BlackYoup:
> Improvement idea: should/could we link to the tools' docs we're refering to:
> 
>     * cURL
> 
>     * pg_dump / pg_restore
> 
>     * mysqldump / mysql
> 
>     * mongodump / mongorestore

Yeah, that could be a good idea, I'll try to add some links

BlackYoup:
I amended the commit to add documentation links for the tools and I also fixed the doc definition

hsablonniere:
LGTM :+1:

hsablonniere (approved):

PR #193, opened 2021-01-22
Encryption at rest: update documentation link

BlackYoup:
There is now a dedicated documentation for this feature instead of
having it in each of the add-ons.

I guess this is a breaking change.


hsablonniere:
@BlackYoup LTGM, you could ammend the commit to update the CHANGELOG before merging ;-)

I'm not sure if it's really a breaking change.
The component's API is still the same so maybe a minor or patch.
It's a bit like you fixed the URL.

PR #197, opened 2021-01-28
cc-pricing-table: init component

### LINKED ISSUE #196: Implement cc-pricing-table
A component that displays "items" with features and pricies in a table (or compact list when narrow).

TODO:

* [x] Add a story to show an example with a runtime (ex: Ruby)
* [x] Change the API for `price` on an `item` from `number` to `{ monthly: number, daily: number }`
* [x] Fix alignement in narrow mode
* [x] Change `_onChangeDataState()` function to not modify the DOM ourselves 
* [x] Think about how we could handle alignement of header (feature names)
* [x] Think about how and when we decide the array is not wide enough
* [x] Update docs
* [x] Fix i18n (plan)
* [x] plus and toggle btn => use new circle cc-button with image
  * [x] Cleanup unused CSS
* [x] Cleanup story
  * Refactor postgres with postgresDollar
* [x] Remove `case 'object': return 'test-backup';`
* [x] I think the `number-align` class in `_renderSmallItems` and mode is useless
* [x] Add some space on the right of the feature-value in small mode (expanded)

Galimede:
Fixes #196


hsablonniere:
Let's try again with #249

PR #218, opened 2021-03-17
<cc-input-number>

### LINKED ISSUE #211: Implement cc-input-number
## Context

We need an `<input type=number>` with -/+ buttons.
We won't be using the existing `<cc-input-text>` because it already does a lot of things and they're not really useful for input number.

## Requirements

```
-----------------
| - |    10 | + |
-----------------
```

* Needs to look like `<cc-input-text>` (same borders, spacing...)
* `<input type=number>` under the hood
* Text aligned to the right by default (we may consider a CSS custom prop)
* We need to hide the arrow keys that some browser display for such fields
* (-) minus button on the left of the field
* (+) plus button on the right of the field

## Inputs (props)

* label
* value
* min
* max
* step
* readonly
* disabled

## Outputs (events)

* value

Galimede:
Fixes #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-...
-
+* New component : 
+  * `<cc-input-number>` 
+  
 ## 6.0.1 (2021-03-15)
 
 * `<cc-addon-linked-apps>`: fix CSS alignment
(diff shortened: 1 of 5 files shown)


hsablonniere:
TODO:

:tada: This is awesome !!!

* Merge commits into one
* Name the commit `cc-input-number: init component`
* Add a line in CHANGELOG.md in Unreleased
* Sort props in JSDoc by prop name
* Can we specify the values of min and max in the label so it's a bit cleared in the story?
* Can you add a story with the step

PR #230, opened 2021-04-08
cc-input-number: controls mode disabled opacity

### LINKED ISSUE #229: cc-input-number: disable buttons when necessary
We need to set `disabled` on the minus/plus buttons:

* `disabled` => both buttons are visible and disabled
* `readonly`  => both buttons are visible and disabled
* `value` === `min` => minus button is disabled
* `value` === `max` => plus button is disabled

We should style these disabled state with some kind of `opacity: 0.25`.

Galimede:
Fixes #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,15 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-* cc-input-number
+* `<cc-input-number>`
   * When in controls mode and in a disabled state the buttons also appear in a disabled state 
     with a slight opacity change
   * Fixed a bug that allowed to change the value of the input in controls mode 
     when the component was in a readonly state
+  * The buttons in controls mode now get the state of disabled when min or max value is reached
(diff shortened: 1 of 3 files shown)


hsablonniere:
@Galimede I guess my sentence was not clear, sorry:

> We need the minus and plus buttons to be disabled when not possible

I meant, when `min=0`, disable the minus button if value is `0`.
Same behaviour for the plus button when value is equal to max.

About the CHANGELOG, I think we should refer to components with MD backticks and HTML tag markers like this: `<cc-input-number>` and not like this: cc-input-number.

I failed to do that in the 3 latest changelog entries, sorry.
Can you update those in a separate commit?

Galimede:
Here you go @hsablonniere😉

hsablonniere:
@Galimede Awesome, can you add a commit where you rewrite the last CHANGELOG entries where I forgot to correctly refer to components with `<cc-the-component>`?

Galimede:
Sure 😉

hsablonniere:
Let's go then ;-)

PR #232, opened 2021-04-09
adjust spacing on cc-invoice and add responsive mode to cc-invoice-table

hsablonniere:


Galimede:
I have checked both the code and what it looks like on the storybook and it looks good to me. :smile:

PR #274, opened 2021-06-28
Replace WCA with CEM analyzer (and custom plugins)

hsablonniere:


hsablonniere:
* add `cssParts` to sort-items

PR #292, opened 2021-07-24
Wrong documentation language

grallc:
Changed 2 links that were pointing to the french documentation


hsablonniere:
Wow, thanks, good catch 😍🤗

Issue #225, opened 2021-03-31
cc-block: refactor and introduce slot stack


hsablonniere:
## Context
We want to refactor the `cc-block` component with a new structure based on slot stack (see #151 and #108) . 

Here is a list of all the components we want to use the new `cc-block` in.

### DONE (or almost done)

- [x] cc-env-var-form
- [x] cc-header-addon
- [x] cc-header-orga
- [x] cc-ssh-key
- [x] cc-tcp-redirection-form
- [x] cc-addon-admin
- [x] cc-addon-credentials
- [x] cc-addon-linked-apps
- [x] cc-addon-option-form
- [x] cc-addon-features
- [x] cc-elasticsearch-info
- [x] cc-email-list
- [x] cc-heptapod-info
- [x] cc-invoice-list
- [x] cc-invoice
- [x] cc-jenkins-info
- [x] cc-matomo-info
- [x] cc-orga-member-list
- [x] cc-grafana-info
- [x] cc-header-app 
- [x] cc-addon-backups
- [x] cc-domain-management

hsablonniere:
https://github.com/CleverCloud/clever-components/issues/151

Issue #151, opened 2020-08-28
Investigate how so called <slot> stacks could simplify stuffs


hsablonniere:
See https://www.youtube.com/watch?v=MS7y2K7tZto&t=51m09s

Ideas:

* Simple border cc-block
* Remove slotchange hacks in cc-block
* Maybe auto cc-error in cc-addon-option

hsablonniere:
We used them on `<cc-pricing-product>` :tada:

hsablonniere:
I'm closing this. We know the pattern now. This will be used intensively in #225.

PR #294, opened 2021-07-30
cc-block: add ribbon and no-head property

### LINKED ISSUE #293: cc-block: add a ribbon property
* if property is not empty add the text in a ribbon (top left corner) and add a left margin
* use with cc-jenkins-info
* use with cc-elasticsearch-info

Galimede:
- documentation
- split commits for cc-block and commit for elastic (don't forget changelog)
- naming: If we go this way go for override-foobar
- noTitle (boolean: false by default)

Fixes #293, #225, #151


hsablonniere:
@Galimede hey

* I forced pushed a few formatting details
* I changed padding to 1rem around and how we define it (with a reflected attribute)

Issue #288, opened 2021-07-21
Update cc-env-var-* components to help users with env vars specific rules for Java


hsablonniere:
Right now, our env vars system allows standard POSIX env vars + some custom stuffs for Java so we can inject them as Java properties.

* We need to inform the user that non standard env vars won't be injected as env var
  * but will be injected as properties in java contexts
* We may need a way to prevent users to create them when we know it's not a java app

See https://github.com/CleverCloud/clever-client.js/issues/46

hsablonniere:
We may try to add a small Java logo next to Java variables... Not convinced it's a good idea but we'll try.

Issue #97, opened 2020-04-16
Add a JSON mode to `<cc-env-var-form>`


hsablonniere:
* [ ] Extract JSON validation from the CLI https://github.com/CleverCloud/clever-tools/pull/409 and move it to the client https://github.com/CleverCloud/clever-client.js/issues/28
* [x] Create a `<cc-env-var-editor-json>`
* [x] Use `<cc-env-var-editor-json>` in `<cc-env-var-form>`

Galimede:
- Delete simple quote from placeholder
- Should we make an empty array when there's no env vars

PR #297, opened 2021-08-02
cc-env-var-form: add JSON mode & strict mode

### LINKED ISSUE #94: cc-env-var-form: behaviour when switching between modes
Let's assume the following state in simple mode:

* `ONE` => `1`
* `TWO` => `2`
* `THREE` => `3`

In this simple mode, you are prevented from adding a variable named `THREE` a second time and it's good.

If you switch to expert mode: 

```
ONE="1"
THREE="3"
TWO="2"
```

you be able to change the some values but also to type `THREE` a second time:

```
ONE="11"
THREE="3"
TWO="22"
THREE="333"
```

In this case, an error will be displayed:

```
line 4: be careful, the name THREE is already defined
```

If you switch back to simple mode, you can see that nothing has changed, even `ONE="11"` or `TWO="22"` are not changed.

QUESTION 1: Should we keep this behaviour?

QUESTION 2: Should we prevent the user from switching modes if there is an error?

If you switch again from simple mode to expert mode, you will still see this:

```
ONE="11"
THREE="3"
TWO="22"
THREE="333"
```

You won't see the error message :-(

QUESTION 3: Should we display back the error message? => YES!

If you switch to simple, edit `ONE` to `111` and go back to expert mode, you will see this:

```
ONE="11"
THREE="3"
TWO="2"
```

The previous entry is lost.

QUESTION 4: Is this what we want?

Galimede:
Fixes #94, #97, #288


hsablonniere:
* The "learn more" links don't seem to be up to date
* I would have used the same example between expert and JSON
  * ` Format : NOM_DE_LA_VARIABLE="valeur de la variable" `
  * ` Format : { "name": "NOM_VAR", "value": "la valeur" } `

hsablonniere:
* Why do we have a `mode` on `cc-env-var-create` but a `parserOptions` on others?
* The alignment on the CSS is broken (cc-env-var-editor-json)
* You forgot to remove the `strictMode` in the cc-env-var-form
* You can merge the last two commits so the update of the client is in the same commit as the strict mode

PR #314, opened 2021-09-07
Modernize code with safari 14.1 release

hsablonniere:
Fixes #217

PR #315, opened 2021-09-07
refactor: move sub render methods `_renderFoo()` below the main `render()` method

### LINKED ISSUE #278: Propose a new order between class elements
We have a few private methods called `_renderSomething` that act as sub parts of the main `render` method. This is very convenient when we have repetitions.

I propose that we move them below `render` and before `styles`.

@Galimede What do you think?

hsablonniere:
Fixes #278 (must merge #314 before)


Galimede:
LGTM 👍

PR #325, opened 2021-10-07
Add temporality on pricing components

hsablonniere:
This PR also introduces an env param to disable the auto generation of the docs.
Sometimes it's better because it does not always refresh the whole page and use the HMR.

I also wanted to organize the stories in those components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ title: Changelog
 ## Unreleased (????-??-??)
 
 * `<cc-pricing-table>`: add `temporality` feature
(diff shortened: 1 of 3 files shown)


Galimede:
Here's a short review:
- There's some errors and warnings with check-lit and the linter.
-  Quotes are not formatted correctly here as mentioned here: https://github.com/CleverCloud/clever-components/pull/325/files#r724177037
- I think we should use "day" and "thirty-days" as translation keys instead of "daily", "monthly" for the temporality translations to keep it consistent, WDYT ?

Other than that, that's great :hugs: :+1:

Issue #334, opened 2021-10-17
Bug: with-resize-observer doesn't apply style when changing width in story


pdesoyres:
When changing size, we get an error:
```
Uncaught TypeError: action(...) is not a function
    onResize http://localhost:6006/stories/mixins/with-resize-observer.stories.js?story=true:34
    _onResize http://localhost:6006/src/mixins/with-resize-observer.js:7
    ro http://localhost:6006/src/mixins/with-resize-observer.js:37
```

the `action` function should return another function

PR #335, opened 2021-10-17
with-resize-observer: fix story when changing width (#334)

pdesoyres:


hsablonniere:
Rebased and merged manually :wink:  Thanks

PR #350, opened 2022-01-10
(STANDBY) Add `<cc-input-duration>` to select a duration in Years/Months/Days

judu:
- [ ] Clean CSS.
- [x] Fix the layout shifting issues.
- [x] Review the CSS solution to remove the inner rings is clean.
- [ ] Find a better btn to switch between modes.


judu:
We should have a quick talk to find the right button :)

hsablonniere:
Closing for now...

PR #358, opened 2022-02-04
(STANDBY) cc-tile-metrics: init component

### LINKED ISSUE #240: cc-tile-metrics: init small metrics tile component for overview
## Display

We need a small tile component to display some rough metrics (CPU and RAM).

* The tile will have the same design as the status codes and requests tiles (title, info button, chart)
* We'll display two lines with
  * Label in upper case and bold (CPU/RAM)
  * Blue spark line
  * Value in % in big bold too
* We use a % value instead of a raw value for RAM to help users identify if the scalability is badly configured
* We may take this opportunity to increase the height of this first line of tiles
* We may take this opportunity to reorder the tiles (scalab, instances, metrics, last deploy)

image

### Details about the chart

* Will use charts.js (https://www.chartjs.org/docs/latest/charts/line.html)
* Needs 10-20 points (to be defined)
* Display a dot at the end of the line with the current value
* A horizontal window of one hour should be good
* A vertical window between 0% and 100% is better
  * it's the best way to identify a bad scalability config
  * a vertical window between min and max would show peaks even when just moving from a few %
* We still need to define what needs to be displayed on hover
* If the scalability config has changed in the last hour, we'll try to break the chart in multiple lines

## Inputs

* The input is basically a list of points for CPU and a list of points for RAM.
* For a point, we need:
  * timestamp
  * % value
  * a way to know if a series of points belongs to the same instances config

## Outputs

None

## Wording / i18n

* RAM and CPU are OK in english *and* french
* Just like the other charts, we'll need an info message briefly describing the data

## Data & smart component integration

* Data comes from warp10
* ownerId, appId
* Do we have a way to get a warp10 token for those metrics?
* Update every 30 seconds
* We may need to ask in the past like for the dotmap

Galimede:
Fixes #240


BlackYoup:
Just a thought: what happens when we start a new hour and the metric aren't yet available? Like, if it's `12:00:00`, you will most probably won't have a metric point for that hour yet (last point might be `11:59:50`). Will it display an empty bar? Or do you only draw bars for which you have data points?

hsablonniere:
@BlackYoup That's a very good feedback, thanks. I guess we need to figure this out. For the requests bar chart, we only display "complete hours", we'll have to discuss if we need to do the same.

PR #375, opened 2022-02-25
Web Component Guide: small improvements and typo fix

florian-sanders-cc:
- Added some links to illustrate some of the guidelines,
- Fixed a small typo


hsablonniere:
I'm merging this but we'll have to be careful with the line reference, this will break easily.

PR #379, opened 2022-03-03
cc-matomo-info: init component

### LINKED ISSUE #372: cc-mat-info: init component
A bit like the cc-jenkins-info, we need 3 sections:

* doc
* link
* context + details about PHP+redis+mySQL

hsablonniere:
Fixes #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ title: Changelog
 
 * `<cc-input-number>`: fix controls mode not working properly from types introduction
 * `<cc-pricing-header>`: fix `skeleton` animation still displaying on the select on Safari after loading of data
+* New component:
+  * `<cc-matomo-info>`
 
 ## 7.9.1 (2022-03-15)
 
--- a/src/addon/cc-matomo-info.js
+++ b/src/addon/cc-matomo-info.js
@@ -1,0 +1,149 @@
+import '../atoms/cc-img.js';
+import '../molecules/cc-block-section.js';
+import '../molecules/cc-block.js';
+import '../molecules/cc-error.js';
+import { css, html, LitElement } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map.js';
+import { i18n } from '../lib/i18n.js';
+import { skeletonStyles } from '../styles/skeleton.js';
+import { ccLink, linkStyles } from '../templates/cc-link.js';
+
+const infoSvg = new URL('../assets/info.svg', import.meta.url).href;
+
+const MATOMO_LOGO_URL = 'https://assets.clever-cloud.com/logos/matomo.svg';
+const PHP_LOGO_URL = 'https://static-assets.cellar.services.clever-cloud.com/logos/php.svg';
+const MYSQL_LOGO_URL = 'https://static-assets.cellar.services.clever-cloud.com/logos/mysql.svg';
+const REDIS_LOGO_URL = 'https://static-assets.cellar.services.clever-cloud.com/logos/redis.svg';
+const MATOMO_DOCUMENTATION = 'https://www.clever-cloud.com/doc/deploy/addon/matomo/';
+
+/**
+ * A component to display various informations (Documentation, access, links, ...) for a Matomo service.
+ *
+ * @cssdisplay block
+ */
+export class CcMatomoInfo extends LitElement {
+
+  static get properties () {
+    return {
+      error: { type: Boolean },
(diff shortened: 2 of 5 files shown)


Galimede:
* In the English translation there is `organisation` in the Access section and `organizations` on the About section. I don't know what is the one we use the most on our components, but we should decide what is the one we want to use and maybe use this opportunity to choose if we'd rather use US or EN in our components ?

florian-sanders:
# English Version
- “using the Clever Cloud SSO (Single Sign-On)” ⇒ could we say “using your Clever Cloud Account” ? Seems more straightforward.
- organisation vs organization ⇒ we need to choose whether we use US or british English. I don’t think it is that big of a deal but since we also are working on org member management, now is a good time to do so ? Both choices are fine I think so we just need to pick one and stick with it.
- “meta package” ⇒ this term is very unclear for me. The following sentence (”It will automatically...”) explains it very well so it could definitly remain this way.
- I think we should avoid using the future tense as much as possible to be more straigthforward (which is generally the English way of doing things) :
    - It automatically provides you ... They appear in your organization... You can still configure ... This add-on is free but its dependencies are billed based on their consumptions

# French Version proposition
L’add-on Matomo inclut toutes les dépendances nécessaires à son bon fonctionnement. (<= could be shortened by removing everything after "nécessaires")

## Accéder à Matomo
Accédez à Matomo en utilisant votre compte Clever Cloud (<= **see note about SSO**). Tous les membres de l’organisation peuvent également accéder au service grâce à leur propre compte.

Accéder à Matomo

## Documentation
Consultez notre documentation pour en apprendre plus sur l'utilisation ou la configuration de votre Matomo et de ses dépendances. (<= if you find a shorter and more straightfoward sentence, be my guest !)

Accéder à la documentation

## A propos
L’add-on Matomo inclut des dépendances indispensables à son bon fonctionnement. Il est accompagné d’une application PHP, d’un add-on MySQL et d’un add-on Redis.

Ces dépendances sont affichées dans votre organisation comme n’importe quelle autre application ou add-on. Vous pouvez les configurer comme bon vous semble. Vous pouvez modifier le domaine de l’application PHP ou encore opter pour un autre plan d'add-on MySQL.

Cet add-on est gratuit mais ses dépendances sont facturées en fonction de leur ???.

# Sidenotes & questions:
* French version headings - do we use verbs or nouns. Infinitive / imperative ?
* do we address users directly like "votre compte Clever Cloud", "Accédez" or do we use impersonal verbs ? (I'm in favor of the first option)
* "consumption plan", for the last sentence. I cannot find any good translation here... "Tarification" ? "Tarif", "Consommation" (this one might be confusing though as its not the actual consumption right ?) or just "plan" like we use in other pages but i find it very unclear.
* On a sidenote, as a customer I find it confusing to learn that the add-on is free but its dependencies are charged without knowing the minimum cost of these dependencies ? I don’t quite know what to expect 🙄.
For instance, do these dependencies all start with the same plan or does Clever Cloud select a plan depending on several factors ? If they all start with the same plan, then we could mention PHP is an XS plan by default, etc.

Galimede:
Here are my question(s) for the french version: 
In the `A propos`  section, there is:
> L’add-on Matomo inclut des dépendances indispensables à son bon fonctionnement. Il est accompagné d’une application PHP, d’un add-on MySQL et d’un add-on Redis.

where in the `Accéder à Matomo` we already said: 
> L’add-on Matomo inclut toutes les dépendances nécessaires 
 
do we want to repeat it ?

hsablonniere:
@Galimede 

> In the English translation there is organisation in the Access section and organizations on the About section. I don't know what is the one we use the most on our components, but we should decide what is the one we want to use and maybe use this opportunity to choose if we'd rather use US or EN in our components ?

We need to chose one and start this dictionnary we've been talking about.
In the meantime, I propose we go for `organisation` as it's what in the already existing translations (except 1).

hsablonniere:
@florian-sanders-cc 


> “using the Clever Cloud SSO (Single Sign-On)” ⇒ could we say “using your Clever Cloud Account” ? Seems more straightforward.

We already have this from the Jenkins component. We change change both, I need to check if it's OK.

> “meta package” ⇒ this term is very unclear for me. The following sentence (”It will automatically...”) explains it very well so it could definitely remain this way.

Maybe we could say meta add-on instead.

> I think we should avoid using the future tense as much as possible to be more straigthforward (which is generally the English way of doing things) :

Thanks, I'll update it following your feedbacks.

florian-sanders-cc:
:+1: All good

PR #387, opened 2022-03-09
cc-tcp-redirection-form: add smart definition

### LINKED ISSUE #344: cc-tcp-redirection-form: create smart definition

hsablonniere:
Fixes #344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,11 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-...
+* `<cc-tcp-redirection-form>`: add smart definition
 
 ## 7.8.0 (2022-03-10)
 
 * `<cc-env-var-*>`: fix links to the doc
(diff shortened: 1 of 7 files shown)


florian-sanders-cc:
:wave: here is my feedback:

- I wonder if it is really necessary to specify "au travers des frontaux" / "via the load balancers" (when I read it, it makes me think the whole thing is complicated :roll_eyes: )
- In the English version, sentence we could be a litte more straightforward by removing "you can" in "you can find out more...".
- In the English version, there is a small typo: "documenation" missing a "t".
- I find the "default" namespace thing a little bit confusing because we associate "default" with "cleverapps.io" and then "default" with "custom domains" (that is to say, not "cleverapps.io". Maybe remove the "default" from the sentence about namespaces just to avoid confusion ?

Other than that I found it very clear (even the part about 1 redirection by namespace and per application :smile: ) :+1:

PR #388, opened 2022-03-10
cc-select: init component

### LINKED ISSUE #378: cc-select: init component
# Main objectives

- provide a styled native select
- provide a label linked to the select element (for + matching id)

This component should provide both a better accessibility support as well as better responsive behavior.

florian-sanders-cc:
Fix #378


florian-sanders-cc:
@hsablonniere ready for review :hugs:

src/invoices/cc-invoice-list.js
```diff
@@ -140,6 +155,22 @@ export class CcInvoiceList extends LitElement {
...
+          display: none;
+        }
+
+        cc-select {
+          display: flex;
+          gap: 0.5em;
+          margin-left: auto;
+          white-space: nowrap;
+          width: max-content;
+        }
```
hsablonniere, line 169: This relies on the fact that you know the `cc-select` has the label and input as direct children. Because there's also a third child, the gap introduces a space on the right and breaks alignment.

I think this should be handled by `cc-select` component directly.

We could add a `inline` property maybe, this needs to be discussed briefly.

hsablonniere (changes requested):

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+      disabled: { type: Boolean, reflect: true },
+      /** @required */
+      label: { type: String },
+      name: { type: String, reflect: true },
+      options: { type: Array },
+      placeholder: { type: String },
+      required: { type: Boolean },
+      value: { type: String },
+      _uniqueName: { type: String, attribute: false },
+      _uniqueId: { type: String, attribute: false },
```
hsablonniere, line 32: Sorry, I think my last comment about those was confusing.

I think:
* `_uniqueName` should be `_uniqueId` (used for the component and its label)
* `_uniqueId` should be `_uniqueErrorId` (used for the error message linked to the component)

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+    /** @type {string|null} Sets the selected value of the element. */
+    this.value = null;
+
+    // use this unique id for isolation (Safari seems to have a bug)
+    /** @type {string} */
+    this._uniqueId = Math.random().toString(36).slice(2);
+
+    // use this unique name for isolation (Safari seems to have a bug)
+    /** @type {string} */
+    this._uniqueName = Math.random().toString(36).slice(2);
```
hsablonniere, line 66: Same comment: please rename them as described above.

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  _onSelectInput (e) {
+    this.value = e.target.value;
+    dispatchCustomEvent(this, 'input', this.value);
+  }
+
+  render () {
+    return html`
+      <label for=${this._uniqueName}>
```
hsablonniere, line 83: => `_uniqueId`

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+        ${this.label}
+        ${this.required ? html`
+          <span class="required">${i18n('cc-select.required')}</span>
+        ` : ''}
+      </label>
+      <div class="selectWrapper ${classMap({ disabled: this.disabled })}">
+        <select
+          id=${this._uniqueName}
```
hsablonniere, line 91: => `_uniqueId`

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+          ?disabled=${this.disabled}
+          aria-describedby=${this._uniqueErrorId}
```
hsablonniere, line 93: This was kept as `_uniqueErrorId` which would fail in the current state of the code. If we rename the other parts as described, this can be kept as is.

src/atoms/cc-select.js
```diff
@@ -0,0 +1,217 @@
...
+            <option
+              value=${option.value}
+              ?selected=${option.value === this.value}
+            >
+              ${option.label}
+            </option>
+          `)}
+        </select>
+      </div>
+      <div id=${this._uniqueId}>
```
hsablonniere, line 109: => `_uniqueRandomId`

hsablonniere (changes requested):

hsablonniere:
@florian-sanders-cc

* The first commit still has a few details to fix
* The other ones are perfect, I see that your git skills are :+1:
* The formatting of HTML is :+1:

stories/atoms/cc-select.stories.js
```diff
@@ -0,0 +1,119 @@
...
+);
+
+export const errorMessage = makeStory(conf, {
+  items: [
+    {
+      label: 'Favourite artist',
+      placeholder: '-- Select an artist --',
+      required: true,
+      options: baseOptions,
+      innerHTML: `<p slot="error">Veuillez saisir une valeur</p>`,
```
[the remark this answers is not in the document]
florian-sanders-cc: @hsablonniere This also brings interesting questions about how we write error messages like this in French and English.

I don't have the time to think about it right now so I put it the best way I could but I'll create an issue about this subject so we don't forget.

florian-sanders-cc:
@hsablonniere last check unless I messed something up again :roll_eyes:

PR #408, opened 2022-03-24
cc-warning-payment: init component

### LINKED ISSUE #402: cc-payment-warning: init component
## Context

We need a component to display warnings when some payment method are not "valid".
We have 3 types of payment method error:

* no payment method
* no default payment method
* default payment method is expired

The clever-client handles this...

We need 3 variants of this component:

* one for `home` with links to payment pages of orgas with errors
* one for the `orga-overview` with link to payment page
* one for the `orga-billing`

The user of the component cannot use it if there are no errors.

## Input

* mode: `"home" | "overview" | "billing"`
* errors: `Array<{ type: number, orgaName: string, orgaBillingLink: string }>`

## Output

none

## Stories

* mode home, one error (one instance for each error type)
* mode home, multiple errors (all 3 types)
* mode overview (one instance for each error type)
* mode billing (one instance for each error type)

hsablonniere:
Fixes #402

Preview : https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-payment-warning/index.html?path=/story/%F0%9F%9B%A0-notices-cc-warning-payment--default-story

### DIFF
--- a/src/notices/cc-warning-payment.js
+++ b/src/notices/cc-warning-payment.js
@@ -1,0 +1,127 @@
+import { ERROR_TYPES } from '@clevercloud/client/esm/utils/payment.js';
+import { css, html, LitElement } from 'lit-element';
+import { i18n } from '../lib/i18n.js';
+import { defaultThemeStyles } from '../styles/default-theme.js';
+import { linkStyles } from '../templates/cc-link.js';
+
+const warningSvg = new URL('../assets/warning.svg', import.meta.url).href;
+
+/**
+ * @typedef {import('./types.js').PaymentMethodError} PaymentMethodError
+ * @typedef {import('./types.js').PaymentWarningModeType} PaymentWarningModeType
+ */
+
+/**
+ * A component to display a warning block with details about payment methods errors.
+ *
+ * @cssdisplay grid
+ */
+export class CcWarningPayment extends LitElement {
+
+  static get properties () {
+    return {
+      errors: { type: Array },
+      mode: { type: String },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {PaymentMethodError[]} Sets the list of payment method errors with type (and optional orga name and billing link). */
+    this.errors = [{ type: ERROR_TYPES.NO_PAYMENT_METHOD }];
+
+    /** @type {PaymentWarningModeType} Sets the mode, depending on where the warning is used, the level of details is not the same. */
+    this.mode = 'billing';
+  }
+
+  render () {
+    return html`
+      <img class="icon" src="${warningSvg}" alt="">
+      <div>
+        ${this.mode === 'home' ? html`
+          <div>${i18n('cc-payment-warning.home', { orgaCount: this.errors.length })}</div>
+          <ul>
+            ${this.errors.map((error) => html`
+              <li>${this._renderHomeItem(error)}</li>
+            `)}
+          </ul>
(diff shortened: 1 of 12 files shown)


src/notices/cc-warning-payment.js
```diff
@@ -0,0 +1,127 @@
...
+      errors: { type: Array },
+      mode: { type: String },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {Array<PaymentMethodError>} */
+    this.errors = [{ type: ERROR_TYPES.NO_PAYMENT_METHOD }];
```
Galimede, line 32: * We do use: `type[]` instead of `Array<type>` in our other components we do need to change it otherwise type definitions plugin doesn't recognize it. 
* Type lacks a description
* You need to export your interface in your type file

src/notices/cc-warning-payment.js
```diff
@@ -0,0 +1,127 @@
...
+
+    /** @type {PaymentWarningModeType} */
```
Galimede, line 34: * Type lacks a description

src/translations/translations.fr.js
```diff
@@ -446,6 +446,19 @@ export const translations = {
   'cc-matomo-info.open-matomo.text': `Vous pouvez accéder à votre Matomo en utilisant votre compte Clever Cloud. Tous les membres de l'organisation peuvent également accéder au service grâce à leur propre compte.`,
   'cc-matomo-info.open-matomo.title': `Accéder à Matomo`,
   //#endregion
+  //#region cc-payment-warning
+  'cc-payment-warning.billing-page-link': ({ orgaName, orgaBillingLink }) => sanitize`<a href="${orgaBillingLink}" aria-label="Aller à la page de facturation - ${orgaName}">Aller à la page de facturation</a>`,
```
florian-sanders-cc, line 450: Could replace "Aller à" with "Se rendre sur" as we discussed but it is also ok as it is :+1:

florian-sanders-cc (commented):
Looks so much better than what it did before :heart_eyes:

Galimede (changes requested):
# Remarks 

* Watch out you left the types you made when you were in `overview/types.d.ts` they are still in the file
* You have pasted all the types of overview plus your types in your `notices/types.d.ts`without removing the one you didn't need

# Browser support 

- Chrome: :heavy_check_mark: 
- Firefox: :heavy_check_mark: 
- Safari: :heavy_check_mark: 

# Responsive ratio
Looks good :+1: 
However, sometimes with long text the link doesn't fit on one line, I wonder if it would be better to always have a line break between the text and the link when we are in the case where there are some.

Other than that : **GG** :clap:

hsablonniere:
@florian-sanders-cc @Galimede thanks for the feedbacks on the copy/paste error with `types.d.ts`

@Galimede 

>  However, sometimes with long text the link doesn't fit on one line, I wonder if it would be better to always have a line break between the text and the link when we are in the case where there are some.

I tried but in narrow screens, in french, the link will be multiline so I'm keeping it like this right now.

src/notices/cc-warning-payment.js
hsablonniere: Good catch.

src/notices/cc-warning-payment.js
hsablonniere: Good catch.

PR #420, opened 2022-04-13
Form controls: Help texts and error messages

florian-sanders-cc:
# Changelog:
## cc-input-text & cc-input-number
* Add help text and error message support as well as their related stories.
* Remove label stories.
* Add label inside all stories.

## cc-select
* Add help text support as well as its related story.
* Tweaked label CSS to make it simpler.

# Sidenote:
- Code format: hope prettier and I did not mess things up @hsablonniere :angel: 
- [Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-input/required-and-help-text/index.html)

TODO before merging:
- [x] Check preview on Safari :dizzy_face: 
- [x] Run lint fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,13 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-* colors: use variables to specify colors and swap old colors for new contrasted ones.
+* colors: use variables to specify colors and swap old colors for new contrasted ones. Add ADR about colors and Design Tokens.
 * all-form-controls story: display disabled and readonly form controls next to each other to make sure they look the same.
 * New component:
   * `<cc-warning-payment>`
+* cc-input-text: add help text and error message support as well as their related stories. Remove label stories and add label inside all stories.
+* cc-input-number: add help text and error message support as well as their related stories. Remove label stories and add label inside all stories.
+* cc-select: add help text support as well as its related story.
 
 ...
 
--- a/src/atoms/cc-input-number.js
+++ b/src/atoms/cc-input-number.js
@@ -1,6 +1,7 @@
 import { css, html, LitElement } from 'lit-element';
 import { classMap } from 'lit-html/directives/class-map.js';
 import { dispatchCustomEvent } from '../lib/events.js';
+import { i18n } from '../lib/i18n.js';
 import { defaultThemeStyles } from '../styles/default-theme.js';
 import { skeletonStyles } from '../styles/skeleton.js';
 
@@ -21,6 +22,9 @@ const decrementSvg = new URL('../assets/decrement.svg', import.meta.url).href;
  * @event {CustomEvent} cc-input-number:requestimplicitsubmit - Fires when enter key is pressed.
  *
  * @cssprop {Align} --cc-input-number-align - Change the alignment of the number present in the input (defaults: `right`).
+ *
+ * @slot error - The error message to be displayed below the `<input>` element or below the help text. Please use a `<p>` tag.
+ * @slot help - The help message to be displayed right below the `<input>` element. Please use a `<p>` tag.
  */
 export class CcInputNumber extends LitElement {
 
@@ -33,10 +37,14 @@ export class CcInputNumber extends LitElement {
       min: { type: Number },
       name: { type: String, reflect: true },
       readonly: { type: Boolean, reflect: true },
+      required: { type: Boolean },
       skeleton: { type: Boolean, reflect: true },
       step: { type: Number },
       value: { type: Number },
       _invalid: { type: Boolean, attribute: false },
+      _uniqueErrorId: { type: Boolean, attribute: false },
+      _uniqueHelpId: { type: Boolean, attribute: false },
+      _uniqueInputId: { type: Boolean, attribute: false },
     };
   }
 
@@ -64,6 +72,9 @@ export class CcInputNumber extends LitElement {
     /** @type {boolean} Sets `readonly` attribute on inner native `<input>` element. */
     this.readonly = false;
 
+    /** @type {boolean} Sets the "required" text inside the label */
+    this.required = false;
+
     /** @type {boolean} Enables skeleton screen UI pattern (loading hint). */
     this.skeleton = false;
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,8 +6,11 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-* colors: use variables to specify colors and swap old colors for new contrasted ones.
+* colors: use variables to specify colors and swap old colors for new contrasted ones. Add ADR about colors and Design Tokens.
```
florian-sanders-cc, line 9: Not sure this can be in the same commit and PR but since we're not going to create a commit and PR for this only, I figured I might as well add it here :roll_eyes:

src/atoms/cc-input-number.js
```diff
@@ -21,6 +22,9 @@ const decrementSvg = new URL('../assets/decrement.svg', import.meta.url).href;
  * @event {CustomEvent} cc-input-number:requestimplicitsubmit - Fires when enter key is pressed.
  *
  * @cssprop {Align} --cc-input-number-align - Change the alignment of the number present in the input (defaults: `right`).
+ *
+ * @slot help - The help message to be displayed right below the `<input>` element. Please use a `<p>` tag.
```
florian-sanders-cc, line 26: does the "Please use a `<p>` tag" read as a little bit passive aggressive or is it ok ? :sweat_smile:

src/atoms/cc-input-number.js
```diff
@@ -187,6 +208,14 @@ export class CcInputNumber extends LitElement {
           </button>
         ` : ''}
       </div>
+
+      <div id=${this._uniqueHelpId}>
+        <slot name="help"></slot>
+      </div>
```
florian-sanders-cc, line 214: Slots vs props: for error and help messages, I chose to use a slots because it allows devs to easily insert HTML into it (`<br>` mainly, but could be links also ?). I am not sure it's the best move as it means we have 2 empty `<div>` elements below fields. This does not cause any visual or accessibility issue but is it bad practice ?

src/atoms/cc-input-number.js
```diff
@@ -201,11 +230,31 @@ export class CcInputNumber extends LitElement {
         }
 
         label {
+          align-items: flex-end;
           cursor: pointer;
-          display: block;
+          display: flex;
+          gap: 2em;
```
florian-sanders-cc, line 236: this is to make sure the "required" mention and the label text are always at least separated by 2em. Could be set as a margin on `.required` instead but I found it simpler this way.

src/atoms/cc-input-number.js
```diff
@@ -201,11 +230,31 @@ export class CcInputNumber extends LitElement {
...
+          color: var(--color-text-light);
+          font-size: 0.9em;
+          margin: 0.3em 0 0 0;
+        }
+        
+        slot[name='error']::slotted(*) {
+          color: var(--color-text-danger);
+          margin: 0.5em 0 0 0;
+        }
+
```
florian-sanders-cc, line 257: all the CSS from `label` to here is exactly the same for the 3 form controls (text, number and select). Should we move this into a shared style ?
To me, these elements should always look the same across these 3 form controls so I don't see why not.

stories/atoms/cc-input-text.stories.js
```diff
@@ -154,7 +172,9 @@ export const allFormControls = allFormControlsStory;
 
 enhanceStoriesNames({
   defaultStory,
-  label,
+  required,
+  helpMessage,
+  errorMessage,
```
florian-sanders-cc, line 177: Not sure about how to order stories :roll_eyes: 
error and help messages should be together but should they appear as one of the first stories like here or as one of the last instead ?

src/atoms/cc-input-number.js
hsablonniere: Nope, it's fine :+1:

src/atoms/cc-input-number.js
hsablonniere: It's always the problem with "how can I know if a slot is used" :-(

I think it will be fine like this...

src/atoms/cc-input-number.js
hsablonniere: We used to avoid gap on flexbox because of lack of support for Safari but we're one year after the 14.1 and more that 2 versions where released recenlty. That's fine like this :+1:

src/atoms/cc-input-number.js
hsablonniere: I would say no for now but maybe add a comment on each 3 form controls. This way we know it's duplicated and we know we could address it when it's the right time.

stories/atoms/cc-input-text.stories.js
hsablonniere: It's fine like this :+1

hsablonniere:
@florian-sanders-cc 

* I have no idea why (yet) but on Safari, the skeleton rectangle for cc-input-text is weird (see capture)
* I think we should have a story with just a error message but without a help message, what do you think?

src/atoms/cc-input-number.js
```diff
@@ -21,6 +22,9 @@ const decrementSvg = new URL('../assets/decrement.svg', import.meta.url).href;
  * @event {CustomEvent} cc-input-number:requestimplicitsubmit - Fires when enter key is pressed.
  *
  * @cssprop {Align} --cc-input-number-align - Change the alignment of the number present in the input (defaults: `right`).
+ *
+ * @slot help - The help message to be displayed right below the `<input>` element. Please use a `<p>` tag.
+ * @slot error - The error message to be displayed below the `<input>` element or below the help text. Please use a `<p>` tag.
```
hsablonniere, line 27: Alphabetical order on slots :pray:

src/atoms/cc-input-number.js
hsablonniere: Note that you're using a gap between a text node (the label) and the span required. We should add a span on the `${this.label}` so it's really a flex item (see other comment).

src/atoms/cc-input-number.js
```diff
@@ -151,7 +166,12 @@ export class CcInputNumber extends LitElement {
     return html`
 
       ${this.label != null ? html`
-        <label for=${this._uniqueName}>${this.label}</label>
+        <label for=${this._uniqueInputId}>
+          ${this.label}
```
hsablonniere, line 170: Because `<label>` is a flex container, it's better for its direct children to be DOM elements and not text nodes. For this, I think `${this.label}` should be inside a span `<span>${this.label}</span>`

src/atoms/cc-input-number.js
```diff
@@ -37,6 +41,9 @@ export class CcInputNumber extends LitElement {
       step: { type: Number },
       value: { type: Number },
       _invalid: { type: Boolean, attribute: false },
+      _uniqueErrorId: { type: Boolean, attribute: false },
+      _uniqueHelpId: { type: Boolean, attribute: false },
+      _uniqueInputId: { type: Boolean, attribute: false },
```
hsablonniere, line 46: You forgot to define `required`

src/atoms/cc-input-number.js
```diff
@@ -76,9 +83,17 @@ export class CcInputNumber extends LitElement {
...
+
+    // use this unique id for isolation (Safari seems to have a bug)
+    /** @type {string} used by the `aria-describedby` attribute on the `<input>` element and the `id` attribute on the help text container */
+    this._uniqueHelpId = Math.random().toString(36).slice(2);
+
     // use this unique name for isolation (Safari seems to have a bug)
-    /** @type {string} */
-    this._uniqueName = Math.random().toString(36).slice(2);
+    /** @type {string} used by the for/id relation between `<label>` and `<input>` */
+    this._uniqueInputId = Math.random().toString(36).slice(2);
```
hsablonniere, line 96: You forgot to define `required`

hsablonniere (changes requested):

PR #426, opened 2022-04-15
cc-article-card: init component

Galimede:
There are a few things that I am m not quite sure of: 

* The error message and its position in responsive 
* If you put a custom width as its done on the `cc-article-list` story and as it will be in the console it seems to break responsive I don't really know why :thinking: 
*  I put a big title and  a big description in one of the card in the `cc-article-list` story and as you will see the description is not aligned with the others => Should we try to put them on the same line and if yes how should we do it?


Galimede:
Thanks for the review ! :smiley:

PR #432, opened 2022-04-22
cem: fix type document generation

Galimede:
# Context

The problem was when there were two types with the same name but belonging to two different component categories only one of them was kept and used for both of them. 

Example: both Addon and Atoms component categories had an `Option` type but for the component in the atoms category that used his own `Option` type the one of addon was used instead. 

# What caused the bug?

The bug was caused from the cache only taking the type name so if you had multiple types in all the components categories that had the same name (e.g: Option) the cache would be overwritten everytime we found another one of the same name and keep only one instead of having all of them.

# How it was fixed?

Instead of setting the cache with the type name, the cache was set to `typename-filename` so that it can't be ambiguous anymore.


hsablonniere:
:clap: Well done @Galimede

PR #447, opened 2022-05-02
cc-link: small fixes following color update (visited color, duplicate styles)

### LINKED ISSUE #443: cc-link: visited link color is different from the default link color
Viisited link color is different from the default link color. 

Screenshot 2022-05-02 at 14-08-45 Console - Clever Cloud

This was not the case before the color update.
Since this does not cause any accessibility issue and this is not an actual UX issue, we prefer to revert back to how links behaved before the color update:
same color whatever their status may be (active, visited, etc.)

### LINKED ISSUE #444: cc-link: duplicate styles
Since the color update, styles are imported twice in `cc-link`.

Remove the unnecessary addition.

florian-sanders-cc:
Fixes #443 
Fixes #444 

- Forced `:visited` state to use the same color as links by default (`--color-text-primary-highlight`).
- Removed `defaultThemeStyles` insertion inside `cc-link` CSS.
- Added `defaultThemeStyles` import in components that use `cc-link`.
- Updated `cc-link` story / doc following the colors update (new contrast ratio + added `defaultThemeStyles` in the example)


hsablonniere:
@florian-sanders-cc LGTM, you can merge this (after rebase...) :wink:

PR #448, opened 2022-05-02
all-form-controls: add support and story for horizontal form layout

### LINKED ISSUE #446: all form controls: add support and doc for horizontal form layout
With clever component form controls next to each other with a `<cc-button>` at the end of the line, the button does not align the rest of the components when there is at least one error / help message displayed.

image

We need to find a robust way to align all the elements even when there are error / help messages displayed and however long they may be.

florian-sanders-cc:
Fixes #446 

- Added a new section in the `all-form-controls` story to showcase all form controls horizontally aligned with help messages below each form control.
- Specified line-height on the `<label>` (`cc-input-*` + `cc-select`)/ `<legend>` (`cc-toggle`) element of all the form controls.
- Aligned all the form controls from top (`align-items: start`).
- Added a margin variable in `defaultThemeStyles` to be used as `margin-top` value on a `cc-button` so that it is aligned with the rest of form controls (specifies an offset equal to `label` height (line-height + padding = `1.6em` )
- Added small paragraph to try and explain all of this :sweat_smile:.

### DIFF
--- a/src/styles/default-theme.js
+++ b/src/styles/default-theme.js
@@ -159,5 +159,18 @@ export const defaultThemeStyles = css`
     /* Usage: with default background for an "outlined" style or warning background for "plain" style" ? */
     --border-warning: 1px solid var(--color-yellow);
     /*endregion*/
+      
+    /*region Margin Decisions*/
+    /* Usage: 
+    *   - in an horizontal form
+    *   - use flex and align-items: start; on the form
+    *   - use margin-top: var(--margin-top-btn-horizontal-form); on the <cc-button> element so that it aligns 
+    *   horizontally even when error / help messages are displayed. 
+    *  See the All Controls Story - Horizontal form section
+    *  
+    *  The 1.6em value comes from the label line-height (1.25em) and bottom padding (0.35em)
+    */
+    --margin-top-btn-horizontal-form: 1.6em;
+    /*endregion*/
   }
 `;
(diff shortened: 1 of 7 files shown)


src/styles/default-theme.js
```diff
@@ -159,5 +159,18 @@ export const defaultThemeStyles = css`
...
+    /* Usage: 
+    *   - in an horizontal form
+    *   - use flex and align-items: start; on the form
+    *   - use margin-top: var(--margin-top-btn-horizontal-form); on the <cc-button> element so that it aligns 
+    *   horizontally even when error / help messages are displayed. 
+    *  See the All Controls Story - Horizontal form section
+    *  
+    *  The 1.6em value comes from the label line-height (1.25em) and bottom padding (0.35em)
+    */
+    --margin-top-btn-horizontal-form: 1.6em;
```
florian-sanders-cc, line 173: really not sure about this naming :roll_eyes:

stories/atoms/all-form-controls.js
```diff
@@ -3,52 +3,84 @@ import '../../src/atoms/cc-input-number.js';
...
-      align-items: flex-end;
-      border-bottom: solid 1px gray;
-      display: flex;
-      gap: 1em;
-      margin-bottom: 2em;
-      padding-bottom: 2em;
-    }
-  `,
+  css: [
+    defaultThemeStyles,
```
hsablonniere, line 11: The `css` used by `makeStory` expects a string, it is directly used under the hood to create a `<style>` tag. It doesn't work like lit so the array doesn't work and the result of lit's `css` function doesn't work either.

hsablonniere:
```
defaultThemeStyles.toString(),
```

```
[
/* ... */
].join('\n'),
```

hsablonniere (changes requested):

PR #452, opened 2022-05-04
cc-article-list (smart): fix error mode

Galimede:


hsablonniere:
* can you rename the commit into `cc-article-list: fix error mode and reset in smart definition`?
* you also need to reset the list of articles to null and error back to false before fetching

Galimede:
Sure, changes done  @hsablonniere :wink:

PR #455, opened 2022-05-06
cc-doc-card: init component

### LINKED ISSUE #433: cc-doc-card: init component
## Context

We need a component to display a card that contains information of a product and a link to our documentation that concerns this product.

## Input

* title: string
* desc: string
* icons: string[] (links to the icons)
* link: string

## Output

none

## Stories

 * defaultStory
 * skeleton
 * dataLoaded
 *  simulations

Galimede:
Fixes #433

### DIFF
--- a/src/homepage/cc-doc-card.js
+++ b/src/homepage/cc-doc-card.js
@@ -1,0 +1,142 @@
+import '../atoms/cc-img.js';
+import { css, html, LitElement } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map.js';
+import { fakeString } from '../lib/fake-strings.js';
+import { i18n } from '../lib/i18n.js';
+import { defaultThemeStyles } from '../styles/default-theme.js';
+import { skeletonStyles } from '../styles/skeleton.js';
+import { linkStyles } from '../templates/cc-link.js';
+
+const SKELETON_INFO = {
+  title: fakeString(6),
+  description: fakeString(110),
+};
+
+/**
+ * A component displaying basic information of a product with a link to redirect to its documentation in a card.
+ *
+ * @cssdisplay grid
+ */
+export class CcDocCard extends LitElement {
+
+  static get properties () {
+    return {
+      description: { type: String },
+      icons: { type: Array },
+      link: { type: String },
+      title: { type: String },
+    };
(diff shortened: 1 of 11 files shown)


hsablonniere:
* can we add border radius on icons?
* icon gap => `0.5em`
* cc-doc-list => there's a weird difference in the number of columns between skeleton and loaded
* UX writing for list error
  * `An error occurred while loading documentation`
  * `Une erreur est survenue pendant le chargement de la documentation`

src/translations/translations.en.js
```diff
@@ -195,6 +195,13 @@ export const translations = {
   'cc-datetime-relative.distance': ({ date }) => formatDistanceToNow(date),
   'cc-datetime-relative.title': ({ date }) => formatDate(lang, date),
   //#endregion
+  //#region cc-doc-card
+  'cc-doc-card.link': ({ link, product }) => sanitize(`<a href=${link} aria-label="Read the ${product} documentation">Read the documentation</a>`),
```
florian-sanders-cc, line 199: the `aria-label` value must start with the visible text (word for word):
"Read the documentation - ${product}"

src/homepage/cc-doc-card.js
```diff
@@ -0,0 +1,142 @@
...
+    const description = this.description ?? SKELETON_INFO.description;
+
+    return html`
+        <div class="images">
+          ${skeleton ? html`
+            <cc-img></cc-img>
+          ` : ''}
+          ${!skeleton ? html`
+            ${this.icons.map((icon) => html`
+              <cc-img src=${icon}></cc-img>
```
florian-sanders-cc, line 60: interesting case. These icons typically convey redundant information but in some cases like "Java", they convey additional info compared to the context.
In the example below, these images tell me that the doc is also related to "mvn" for instance.
image

The best solution (to avoid adding alt values that would be redundant in most cases) would be to add this piece of info in the description text but is it something we can do ?

florian-sanders-cc: nevermind, we can assume these are just random, not exhaustive examples. Let's not overthink it.
It's good as it is ;)

hsablonniere:
LGTM :+1:

PR #468, opened 2022-05-19
cc-orga-member-*: init components

florian-sanders-cc:
:wave: Following our discussions about button colors and styles, I updated the component:

- [Preview with plain primary validate and edit buttons](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member/blue-submit/index.html?path=/story/%F0%9F%9B%A0-organisation-cc-orga-member-list--default-story)
- [Preview with outlined primary validate and edit buttons](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member/outlined-btns/index.html?path=/story/%F0%9F%9B%A0-organisation-cc-orga-member-list--default-story)
- If you want to compare with the previous version: [Previous preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member-invite/init/index.html) (tricky one to pronounce :smirk:)

Don't forget to check the responsive views since the components change quite a lot.


florian-sanders-cc:
# Badges - primary-light color
image

florian-sanders-cc:
# Badges - neutral-alt color (slightly darker than default neutral)
image

florian-sanders-cc:
# Badges - neutral-alt color (slightly darker than default neutral)
image

roberttran-cc:
UI feedback:
- badges for roles & 2FA are really nice!
- maybe worth to try a light background when a member is in edit mode;
- i would horizontally center badges but it seems quite complicated.

Role badge color:
- my initial understanding was one color per role but, on second thoughts, it would probably make the UI too rainbow-ish;
- that being said, i vote for the primary-light color!

Your own card (current user):
- highlighting the current user card may not be a must-have but @pdesoyres-cc proposal seems nice and easy to implement.

Misc:
- the icon homogeneity should be addressed globally, not sure if it's worth the hassle to make it right beforehand imho (but a thicker "leave" icon would look nice).

pdesoyres-cc:
This icon is ok for me. But more generally, what do you think of picking icons from https://remixicon.com/ collection each time we look for a new icon?

roberttran-cc:
> on "(me)" next to current user name (from Pierre)

Sounds good! Maybe the left margin is a bit too much (`0.25em` or no left margin)...

> on leave icon thickness (from Pierre)

It's better: gg!

> on background color when in edit mode (from me)

I think it's more readable: let's see what the others think...

> on centered role badges (from me)

I like it better: again, let's see what the others think...

> on distinct color per badge role (from me)

A little rainbow-ish but i guess it depends on the list. Not sure we should keep it but thanks for the effort!

> on the primary-light background color on badge role (instead of light grey) (from me)

I like it better!

> on user fallback image (from me)

Very nice!

pdesoyres-cc:
I think that we should stick to one single color for the role badges for the following reasons:

* too many colors kill the color
* which color will we use if we add a new role kind?

I agree with @roberttran-cc: I prefer the primary-light color for the role badge.

+1 for the centered alignment.

florian-sanders-cc:
:wave: I updated the preview based on your feedback. Thanks a lot <3 

You can check out the updated version based on your feedback using this [Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member-invite/init/index.html).

# Main improvements: 

- fixed responsive layouts (icons, picture and badge alignment),
- added a badge for "your account" to identify your own card,
- changed icons (edit and profile picture fallback. Now using remix icons),
- removed `cc-orga-member-card` padding (spacing is now set by `cc-orga-member-list`. Used box shadow to fake a padding for the card being edited),
- changed background to primary light for role badges,
- added missing i18n strings and fixed typos,
- removed max-width from `cc-orga-member-card`.

# Notes
@hsablonniere a few notes about some points in your feedback:
> I'm not a big fan of the horizontally centered role badge
> - I doesn't work well between 660px and 800px
> - Same problem for 645/650px

I think this was mainly due to small CSS issues that I did not want to fix earlier because I was constantly moving things around between reviews.
These should be fixed now.
Badges are now only centered when viewport is large. Does it still trouble you ?

> When I click on the edit button, should we force the focus on the role select? Is is good practice?

This is actually mandatory (Pierre also pointed this out when we did the first reviews on Slack). I did not do anything about focus management because things did not feel stable enough until now (for instance, I had to move the whole "toggle edit mode" logic from `cc-orga-member-card` to the list, which also means managing focus is a bit trickier). I also need to figure out a good way to handle focus after API calls...

hsablonniere:
> This is actually mandatory (Pierre also pointed this out when we did the first reviews on Slack). I did not do anything about focus management because things did not feel stable enough until now (for instance, I had to move the whole "toggle edit mode" logic from cc-orga-member-card to the list, which also means managing focus is a bit trickier). I also need to figure out a good way to handle focus after API calls...

We can figure this out together in a pair programing session.

hsablonniere:
About the new preview:

* the responsive feels nice
* the not centered badge are OK I guess
* I would reduce the vertical gap between card inner parts
* The icons seem better

roberttran-cc:
Design-wise, nothing to add for me! Good stuff incoming! :1st_place_medal:

florian-sanders-cc:
:wave: Following our discussions about button colors and styles, I updated the component:

- [Preview with plain primary validate and edit buttons](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member/blue-submit/index.html?path=/story/%F0%9F%9B%A0-organisation-cc-orga-member-list--default-story)
- [Preview with outlined primary validate and edit buttons](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-orga-member/outlined-btns/index.html?path=/story/%F0%9F%9B%A0-organisation-cc-orga-member-list--default-story)

Don't forget to check the responsive views since the components change quite a lot.

roberttran-cc:
Same, I vote for the outline!

pdesoyres-cc:
I prefer the outlined version as it is in line with the principle agreed upon yesterday `buttons for a list of items => outline`.

Maybe, once we have one item in edition, the validate button could be in _plain_ mode:

* validation becomes more important than cancelation
* Only one item can be edited at a time so we will have only one _main_ action displayed at a time.

florian-sanders-cc:
@florian-sanders-cc test ping to see if I get GH notif

florian-sanders:
@florian-sanders-cc hello

florian-sanders:
@florian-sanders-cc @Galimede does it work ?

florian-sanders:
@florian-sanders-cc @Galimede one more ping test

florian-sanders:
@Galimede @florian-sanders-cc one more

florian-sanders:
@Galimede please answer me !

florian-sanders:
test

florian-sanders:
test @Galimede

florian-sanders:
test

hsablonniere:

* I think we'll be able to flatten the state of the orga member type on the item component
* Why is `isEditing` a separated boolean?
  * It seems like you're only using the state when it concerns API calls
  * It think editing could also be a state
* As we mentioned it, the UI component should carry the translated messages and only expose ways to trigger message A, B or C.
  * i18n outside of UI components will probably only be used for toasts
* I think we can get rid of some private state and rely on the DOM (it should simplify the code)
* You asked how we can prevent code duplicated between partial updates and global updates
  * Let's start with duplicated code and then find out how we can group stuffs
* We don't really have a loading state of the item component but I think I'd rather define a loading stage in the types and have nothing in the render with a comment saying we don't need this state instead of having null.

pdesoyres-cc:
## state model

### MemberCardState

* I would prefer not to repeat the 9 properties for each state using either
  * inheritance
  * type intersection (with the & operator)
  * internal data property
* I'm not sure the `hasError?: false,` is necessary for 'updating' and 'deleting' states.
* I would rename `mfa` into `isMfaEnabled`

### MemberListState

* `StateMemberListLoaded` contains a `memberInEditing` property and `MemberCard` has a `isEditing` property. I fill like we should not have both
* Shouldn't we share the `MemberCard` member value definition with the one used in `MemberCardState`?
* I don't understand why the `MemberCard` has a member state as it is always 'loaded'.

### StateMemberInvite
* I like the modeling

florian-sanders-cc:
# Why we chose to move the filters into the list property

image
the filter fields were synced with two internal properties:
`_filterIdentity` &  `_showDisabledMfaOnly`.

This meant that we couldn't reset their values from the smart component when the context changed.

The first solution that comes to mind is to make these properties public.
This would work but it's easy to forget to reset them since they are not part of the `OrgaMemberListState`.

Since we want to reset them every time the context is updated, we wanted to make these part of the `members` property (`OrgaMemberListState`).

If we follow our guidelines, since the filters are form fields, here is what we obtain:
```ts
interface OrgaMemberListStateLoaded {
  state: 'loaded';
  value: OrgaMemberCardState[];
  filters: {
    state: string;
    identity: {
      value: string;
      error: errorType;
    };
    mfa: {
      value: boolean;
      error: errorType;
    }
  }
}
```

These form fields cannot have errors so the `error` part is unnecessary. The form itself cannot contain any error either and its only state value would be `idle` (no `waiting`, no `error`) so let's remove these parts:
```ts
interface OrgaMemberListStateLoaded {
  state: 'loaded';
  value: OrgaMemberCardState[];
  filters: {
    identity: {
      value: string;
    };
    mfa: {
      value: boolean;
    }
  }
}
```

Now that error properties are gone, it feels weird to have objects with just a `value` property containing a string. We can easily make it flat:

```ts
interface OrgaMemberListStateLoaded {
  state: 'loaded';
  value: OrgaMemberCardState[];
  filterIdentity: string;
  filterMfaDisabledOnly: boolean;
}
```

This state is simpler and it allows us to reset the filters at the same time as the list. It makes sense because the list and the filters work together.
Making the filters part of the interface of the list makes it impossible to forget to reset the filters when resetting the list (when the context changes).

Reset from the smart component:
```js
updateComponent('members', {
  state: 'loaded',
  value: memberList.map((member) => ({state:'loaded',...member})),
  filterIdentity: '',
  filterDisabledMfaOnly: false,
});
```

## Exceptions to the state guidelines
This means we are introducing some exceptions to our guidelines about states:

* form fields should be typed as objects with a `value` and an `error` property unless they cannot contain errors, in which case we can use strings instead of objects directly.
* the filters are updated directly by the UI component (`cc-orga-member-list`). When we do this, we generally need to warn the parent components that the props have been updated by the component by dispatching an event ("props down, events up"). Since the smart component does not need to keep track of the values updated by the component, it only resets them, we do not need to dispatch an event from `cc-orga-member-list` in this specific case.

## Other benefits

Moving from internal properties to this also allows us to create stories showcasing the filters with values already set.
With the `cc-orga-member-list` component, this is only useful for stories, but with other components using filters, this technique might also be useful if you need to filter the component upon loading it (for instance depending on some query params, etc.).

PR #473, opened 2022-05-24
cc-email-list: init component

### LINKED ISSUE #435: cc-email-list: init component
## Context

A component that displays the email addresses associated with a user account.

The component shows two things:

* the primary email address
* the secondary email addresses

### The primary email address

* The component displays a description of what this email address is used for.
* The component displays whether the primary email address has been verified. If not, the user has the ability to resend the confirmation email that will let him finish the verification process.

### The secondary email addresses

* The component displays a description of what these email addresses are used for.
* The user can add a new secondary email address.
* A secondary email address can be removed.
* A secondary email address can be marked as primary.

### "email" vs "email address"

In this component, we talk about two different things but the wording can be confusing. I'm talking about the email address versus the email sent for verification. We need to think about a good wording that will tackle ambiguity between these two notions. Both in the localized texts and in the properties and event namings.

Should we rename the whole component into `cc-email-address`?

### verification email

How do we call the email that will let the user finalize the process?
- verification email
- confirmation email
- validation email

How do we call the action for finalizing the process?
- verify email address
- validate email address

As for now, 
- API uses "confirmation_email".
- The email uses 
  - title: "Please validate you address… " (notice the typo :roll_eyes:)
  - body: "verify your email address" and "verify your address".
- The console uses "confirmation email", "validate this email", "resend account confirmation email".

### Notification messages

When the user clicks on the "resend confirmation email" button, we should notify him that the email has been sent and he needs to check his mailbox. Who is responsible for displaying this notification?

### Attributes/properties

| `model` | `EmailModel` | `null` | `false` | Model containing primary email address, secondary email addresses and state |
| `_addAddressInputValue` | `string` | `''` | `false` | Email address form input value |
| `_addAddressInputError` | `'empty'\|'invalid'\|'already-defined'\|'used'\|null` | `null` | `false` | Email address form input error key |

## Outputs

| `send-confirmation-email` | `String` | Triggered when the user asks for sending a new confirmation email |
| `add` | `String` | Triggered when the user adds a new secondary email address |
| `delete` | `String` | Triggered when the user deletes a secondary email address |
| `mark-as-primary` | `String` | Triggered when the user marks a secondary email address as primary |

## Type Definitions

```typescript
export type PrimaryModelErrorType = "sendingConfirmationEmail";
export type PrimaryModelStateType = "sendingConfirmationEmail";

export type SecondaryModelErrorType = "adding";
export type SecondaryModelStateType = "adding";

export type SecondaryAddressErrorType = "markingAsPrimary" | "deleting";
export type SecondaryAddressStateType = "markingAsPrimary" | "deleting";

export interface EmailModel {
  primary: null | 'loadingError' | PrimaryModel;
  secondary: null | 'loadingError' | SecondaryModel;
}

export interface PrimaryModel {
  address: EmailAddress;
  state: PrimaryModelStateType;
  error: PrimaryModelErrorType;
}

export interface SecondaryModel {
  addresses: SecondaryAddressModel[];
  state: SecondaryModelStateType;
  error: SecondaryModelErrorType;
}

export interface SecondaryAddressModel {
  address: EmailAddress;
  state: SecondaryAddressStateType;
  error: SecondaryAddressErrorType;
}


export interface EmailAddress {
  value: string;
  verified: boolean;
}

```

## Data/APIs
* get the primary email address: `GET https://api.clever-cloud.com/v2/self`
* Send confirmation email: `GET https://api.clever-cloud.com/v2/self/confirmation_email`
* list secondary email addresses: `GET api.clever-cloud.com/v2/self/emails`
* add secondary email address: `PUT api.clever-cloud.com/v2/self/emails/{emailAddress}`
* delete secondary email address: `DELETE api.clever-cloud.com/v2/self/emails/{emailAddress}`
* mark secondary email address as primary: `PUT /v2/self/emails/<email>`

### Handling verified/unverified state

As for now, we cannot display the verified/unverified state because the API doesn't give us this information. In fact, we can only get verified email addresses from the API (unverified email addresses are never returned by the API).

We need the API to support:
- list all secondary email addresses (verified and unverified)
- each email address returned should have a `verified` boolean property.

### Resending confirmation email

As for now, the only way to resend a confirmation email is to submit the add form again. There is no explicit action for resending the confirmation email. And this is for the same reason as above: unverified email addresses are never returned by the API. Also, the API for such an action is missing.

See https://gitlab.corp.clever-cloud.com/core-team/cc-api/-/issues/743

## Stories & states

* Skeleton/loading
* Empty
* Loaded
* Loaded with a huge email addresses
* Waiting/sendingConfirmationEmail
* Waiting/adding
* Waiting/deleting
* Waiting/markingAsPrimary
* Error/loading
* Error/sendingConfirmationEmail
* Error/adding
* Error/deleting
* Simulation

## UX writing & i18n

* title: Secondary email addresses
* description: _TODO_
* delete button: just the icon without any label is ok.
* add button: Ajouter, Add
* error.add
* error.delete

_TBD_

### LINKED ISSUE #628: cc-button: display waiting state when in link mode
When cc-button is in link mode, the waiting state is not materialized.

pdesoyres-cc:
Fixes #435

Fixes #628

Preview: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-email/index.html?path=/story/%F0%9F%9B%A0-email-list-cc-email-list--default-story

- [x] remove mock API


roberttran-cc:
Not much to say about the UI, look rather clean to me!

Some feedback:
- i'd always put the confirmation mail error message under the resend button;
- multiple component in the skeleton story needed?

And about the verification email wording, i vote for 'confirmation'! :+1:

florian-sanders-cc:
Hey !

I think the UI looks good as it is :+1:.

A few details:

- we could try bringing the email icon closer to the email address, I feel like 1em is a bit too much, maybe 0.5em ? It's just a suggestion though, it's definitely up to you :+1: 
image
image
- "adding secondary" story, the email address field is empty and it would be a little more realistic if it contained a value (it's pretty insignificant though :joy:) 
- "simulations (secondary)" story, the error below the field should be displayed before setting components in waiting mode since we don't send the API call as long as the form contains an error.
- I find it disturbing that there are two block in stories, maybe a story for verified vs unverified ?

hsablonniere:

About the smart stuffs:

* console.warn when context is empty, I need to think about this and see if and when it's necessary
* I'm wondering if we really need the two separated functions `initComponent` and the other
  * It adds separation so maybe it's clearer but it also adds indirection
* I need to dig the `.ctrl.js` with the code but my first impression was "it seems too complex without much gain"
  * It's just a first impression :hugs:
  * I'm wondering if we're really going to write a `.ctrl.js` file for each component
  * I don't really understand how it simplify things or how it helps us to prevent errors
  * It's imported by the UI component and it relies on immer so technically, now all our components rely on immer
  * In a MVC or MV* world, I would argue that the controller is what we have in the smart definition, what you put in the `.ctrl.js` feels like both Model and Controller
  * I'm often in favor of WET, Write Everything Twice before diving into common/refactored helpers
    * It feels like the StateController, FormStateController and GenericStateController are premature optimization

About the state stuff:

* The work you've done on this is impressive :+1:
  * It's very close to what I want in the end, with the generics and common stuffs that impose consistency
* There's a bit too much nesting to my taste, especially when the names are too generic because of common/factorized types but I think it's close to our goal
* Many abstract/sub types mean the documentation from the outside could be very complex, we need to find the right balance

roberttran-cc:
Well, overall agree with @hsablonniere last comment! :wink: 

Really impressive stuff: implementing things this thoroughly is so nice to make decisions, thanks so much. :bow:
On both parts (state and smart), imho and from a developer point of view, it brings a legitimate sense of safety but at the cost of being a little difficult to understand when discovering the code at first (and if I were to implement it).

:+1: on the need to find the right balance collectively!

pdesoyres-cc:
PR is ready for review.

* Note1: it still contains the smart component in the md file so that you can test it easily.
* Note2: the APIs are mocked so that you won't be banned after several interactions.

Both are to be removed once the PR is approved.

src/components/cc-email-list/cc-email-list.stories.js
```diff
@@ -0,0 +1,424 @@
...
+          secondaryAddresses,
+        },
+      },
+    },
+  ],
+});
+
+export const skeleton = makeStory(conf, {
+  items: [
+    {},
```
roberttran-cc, line 43: Should we be more explicit with:
```suggestion
    {
      emails: {
        state: 'loading',
      },
    },
```

(and same in `simulationsWithPrimary` & `simulationsWithSecondary`)

src/components/cc-email-list/cc-email-list.stories.js
```diff
@@ -0,0 +1,424 @@
...
+export const dataLoadedWithHugeEmail = makeStory(conf, {
+  items: [
+    {
+      emails: {
+        state: 'loaded',
+        value: {
+          primaryAddress: { state: 'idle', address: HUGE_EMAIL_ADDRESS, verified: true },
+          secondaryAddresses: [
+            { state: 'idle', address: HUGE_EMAIL_ADDRESS, verified: true },
+            { state: 'idle', address: HUGE_EMAIL_ADDRESS, verified: true },
```
roberttran-cc, line 122: If I understand correctly, `verified` property on secondary addresses are not needed... am I right? :thinking: 

(here and in other stories below)

src/translations/translations.fr.js
```diff
@@ -234,6 +234,38 @@ export const translations = {
...
+  'cc-email-list.secondary.action.delete.success': ({ address }) => sanitize`L'adresse e-mail secondaire <strong>${address}</strong> a été supprimée avec succès.`,
+  'cc-email-list.secondary.action.mark-as-primary.accessible-name': ({ address }) => `Marquer l'adresse ${address} comme primaire`,
+  'cc-email-list.secondary.action.mark-as-primary.error': ({ address }) => sanitize`Une erreur est survenue pendant le marquage en tant qu'adresse e-mail primaire <strong>${address}</strong>.`,
+  'cc-email-list.secondary.action.mark-as-primary.name': `Marquer comme primaire`,
+  'cc-email-list.secondary.action.mark-as-primary.success': ({ address }) => sanitize`L'adresse e-mail <strong>${address}</strong> a bien été marquée comme primaire.'`,
+  'cc-email-list.secondary.address-input.error.already-defined': `Cette adresse e-mail vous appartient déjà`,
+  'cc-email-list.secondary.address-input.error.empty': `Veuillez saisir une adresse e-mail`,
+  'cc-email-list.secondary.address-input.error.invalid': () => sanitize`Format d'adresse e-mail invalide.<br>Exemple: john.doe@example.com.`,
+  'cc-email-list.secondary.address-input.error.used': `Cette address e-mail ne vous appartient pas`,
```
roberttran-cc, line 262:
```suggestion
  'cc-email-list.secondary.address-input.error.used': `Cette adresse e-mail ne vous appartient pas`,
```

roberttran-cc (changes requested):
Wow, we're reaaaaaaaally close to the end, that's so cool. :sunglasses: Some minor feedback nevertheless!

And a random idea (but I'd don't think we'll implement it): making email addresses clickable with `mailto:`. :eyes:

src/translations/translations.fr.js
```diff
@@ -234,6 +234,38 @@ export const translations = {
   'cc-elasticsearch-info.link.kibana': `Ouvrir Kibana`,
   'cc-elasticsearch-info.text': `Cet add-on fait partie de l'offre Suite Elastic. Vous pouvez retrouver la documentation ainsi que les différents services liés ci-dessous.`,
   //#endregion
+  //#region cc-email-list
+  'cc-email-list.loading.error': `Une erreur est survenue pendant le chargement des adresses e-mail.`,
+  'cc-email-list.primary.action.resend-confirmation-email': `Envoyer un nouvel e-mail de confirmation`,
+  'cc-email-list.primary.action.resend-confirmation-email.error': ({ address }) => sanitize`Une erreur est survenue pendant l'envoie de l'e-mail de confirmation à l'adresse <strong>${address}</strong>.`,
```
roberttran-cc, line 240:
```suggestion
  'cc-email-list.primary.action.resend-confirmation-email.error': ({ address }) => sanitize`Une erreur est survenue pendant l'envoi de l'e-mail de confirmation à l'adresse <strong>${address}</strong>.`,
```

roberttran-cc (changes requested):

florian-sanders-cc (changes requested):
:clap: Gg ! 
The component looks good, it's easy to use and it's well documented :+1: 

Just a few notes apart from the minor comments I've added:

# Email form with small viewport

I think we can improve how the form looks when the viewport is small:
image

We need to make the button wrap below the input field:
```css
form {
    align-items: start;
    display: flex;
    gap: 0 1em; // cancel the row gap
    flex-wrap: wrap; // wrap the button below
    justify-content: end; // align the button on the right like we do in other components
}

form > cc-input-text {
  flex: 1 1 19em; // make sure the input width is at least 19em (pick the value you think is appropriate)
}
```
image

If you think there is still too much space between the `cc-input-text` and `cc-button`, we can cancel its top margin with negative bottom margins on both these elements + bottom padding on the form element but I don't I don't like messing around with negative margins unless we really have to :roll_eyes:. 

I think we need to improve our solution for horizontal forms (`var(--cc-margin-top...`) because this is not the only case where this problem exists. I'll open an issue for this.

# Stories

It might be voluntary but there are no stories showing error messages for `already-defined` / `used` cases.
It would be easier to check the error message content if we had these but it's really not essential so I'll let you decide :+1:

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,400 @@
...
+        .address-line {
+          align-items: center;
+          display: flex;
+          flex-direction: row;
+          flex-wrap: wrap;
+          gap: 1em;
+        }
+        
+        .address-line .address.loading {
+          opacity: 0.6;
```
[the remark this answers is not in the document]
florian-sanders-cc: Ah alright, I ran the a11y tests on `cc-orga-member` and realised that with `0.6` it passes the color contrast ratio criterion and with `0.5` it doesn't (probably why you went for `0.6` in the first place).

Considering this, should we switch to `0.6` everywhere ?

src/components/cc-email-list/cc-email-list.types.d.ts
```diff
@@ -0,0 +1,50 @@
...
+//#endregion
+
+//#region data model
+export interface EmailList {
+  primaryAddress: PrimaryAddressState;
+  secondaryAddresses: SecondaryAddressState[];
+}
+
+export interface PrimaryAddressState extends EmailAddress {
+  state: "idle" | "sending-confirmation-email"
```
Galimede, line 25: There's a missing semicolon at the end.

Galimede (changes requested):
Hiya, 
GG @pdesoyres-cc that's great and as @florian-sanders-cc that's well documented :clap: :heart_eyes: 
I've listed a few notes/typos below :wink:

src/translations/translations.en.js
```diff
@@ -221,6 +221,38 @@ export const translations = {
...
+  'cc-email-list.secondary.action.delete.error': ({ address }) => sanitize`Something went wrong while deleting secondary email address <strong>${address}</strong>.`,
+  'cc-email-list.secondary.action.delete.name': `Delete`,
+  'cc-email-list.secondary.action.delete.success': ({ address }) => sanitize`Secondary email address <strong>${address}</strong> deleted successfully.`,
+  'cc-email-list.secondary.action.mark-as-primary.accessible-name': ({ address }) => `Mark email address ${address} as primary`,
+  'cc-email-list.secondary.action.mark-as-primary.error': ({ address }) => sanitize`Something went wrong while marking secondary email address <strong>${address}</strong> as primary.`,
+  'cc-email-list.secondary.action.mark-as-primary.name': `Mark as primary`,
+  'cc-email-list.secondary.action.mark-as-primary.success': ({ address }) => sanitize`Secondary email address <strong>${address}</strong> has been successfully marked as primary.`,
+  'cc-email-list.secondary.address-input.error.already-defined': `This email address already belongs to you`,
+  'cc-email-list.secondary.address-input.error.empty': `You must enter an email address`,
```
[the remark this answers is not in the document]
roberttran-cc: Agree to sync on this! Good catch.

src/components/cc-email-list/cc-email-list.stories.js
pdesoyres-cc: You're right. It is an anticipation of the future.
We have only one `EmailAddress` interface for the primary address and the secondary addresses. I did not want to create two different interfaces (one with a `verified` property and another one without).
In order to respect this interface I have to specify a `verified` property even if not used.
I'll let this like that and very soon, the property will be used once we have the right API.

src/components/cc-email-list/cc-email-list.js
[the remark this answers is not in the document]
pdesoyres-cc: Yes, we should consider switching to 0.6 everywhere... using a CSS var?

src/components/cc-email-list/cc-email-list.stories.js
roberttran-cc: Nice, thanks for the explanation! :+1:

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,334 @@
...
+            inputError = 'invalid';
+          }
+          else if (error.id === 101) {
+            inputError = 'already-defined';
+          }
+          else if (error.id === 1004) {
+            inputError = 'used';
+          }
+
+          if (inputError) {
```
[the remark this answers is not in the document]
pdesoyres-cc: I will use `if (inputError != null)`

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,400 @@
...
+ * * This validation is triggered whenever the add button is clicked.
+ * * If the validation doesn't succeed, an error message is displayed bellow the text input. Otherwise, the custom event `cc-email-list:add` is fired.
+ * * The validation handles only two cases:
+ *   - the input is empty
+ *   - the input is not a valid e-mail address.
+ * * For the other error cases, you'll need to pass the right `addEmailForm` property manually.
+ *
+ * ## Marking as primary
+ *
+ * * Unlike secondary email address deletion, marking as primary is exclusive: Only one email address can be marked as primary at a time.
```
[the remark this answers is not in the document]
pdesoyres-cc: I will use `email` everywhere

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,400 @@
...
+
+  _renderAddEmailForm () {
+    const isAdding = this.addEmailForm.state === 'adding';
+
+    return html`
+      <form>
+          <cc-input-text
+            label="${i18n('cc-email-list.secondary.address-input.label')}"
+            required
+            value="${live(this.addEmailForm.address.value)}"
```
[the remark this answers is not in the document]
pdesoyres-cc: good catch. thank you

src/components/cc-email-list/cc-email-list.js
```diff
@@ -117,6 +117,21 @@ export class CcEmailList extends LitElement {
...
+
+  /**
+   * @return {SecondaryAddressState[]}
+   */
+  _getSecondaryAddresses () {
+    if (this._isLoading() || this.emails.value.secondaryAddresses == null) {
+      return [];
+    }
+    return this.emails.value.secondaryAddresses
+      .sort((a1, a2) => a1.address.localeCompare(a2.address));
```
roberttran-cc, line 131: Actually, we've got a little utility for this one!

```js
import { sortBy } from '../../lib/utils.js';
//...
return [...this.emails.value.secondaryAddresses].sort('address');
```

roberttran-cc (approved):

src/translations/translations.fr.js
```diff
@@ -234,6 +234,38 @@ export const translations = {
...
+  'cc-email-list.secondary.action.delete.accessible-name': ({ address }) => `Supprimer - ${address}`,
+  'cc-email-list.secondary.action.delete.error': ({ address }) => sanitize`Une erreur est survenue lors de la suppression de l'adresse e-mail secondaire <strong>${address}</strong>.`,
+  'cc-email-list.secondary.action.delete.name': `Supprimer`,
+  'cc-email-list.secondary.action.delete.success': ({ address }) => sanitize`L'adresse e-mail secondaire <strong>${address}</strong> a été supprimée avec succès.`,
+  'cc-email-list.secondary.action.mark-as-primary.accessible-name': ({ address }) => `Définir comme primaire - ${address}`,
+  'cc-email-list.secondary.action.mark-as-primary.error': ({ address }) => sanitize`Une erreur est survenue pendant le marquage en tant qu'adresse e-mail primaire <strong>${address}</strong>.`,
+  'cc-email-list.secondary.action.mark-as-primary.name': `Définir comme primaire`,
+  'cc-email-list.secondary.action.mark-as-primary.success': ({ address }) => sanitize`L'adresse e-mail <strong>${address}</strong> a bien été définie comme primaire.'`,
```
roberttran-cc, line 258:
```suggestion
  'cc-email-list.secondary.action.mark-as-primary.success': ({ address }) => sanitize`L'adresse e-mail <strong>${address}</strong> a bien été définie comme primaire.`,
```

roberttran-cc (changes requested):

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,242 @@
...
+defineSmartComponent({
+  selector: 'cc-email-list',
+  params: {
+    apiConfig: { type: Object },
+  },
+  onContextUpdate ({ component, context, onEvent, updateComponent, signal }) {
+    updateComponent('emails', { state: 'loading' });
+    updateComponent('addEmailForm', CcEmailList.ADD_FORM_INIT_STATE);
+
+    const api = getApi(context.apiConfig, signal);
```
hsablonniere, line 22: Interesting... I'm not sure about this yet but I like the let's try to improve all this verbosity.

I'd like the opinion of the team on this and if positive maybe discuss doing this elsewhere.

Note that the API calls that modify stuffs should not use the signal as we still want them to go through even if there's a navigation. I'll add notes below.

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,242 @@
...
+      api.sendConfirmationEmail(address)
+        .then(() => notify(component, {
+          intent: 'info',
+          title: i18n('cc-email-list.primary.action.resend-confirmation-email.success.title'),
+          message: i18n('cc-email-list.primary.action.resend-confirmation-email.success.message', { address }),
+          options: {
+            timeout: 0,
+            closeable: true,
+          },
+        }))
```
hsablonniere, line 71: Can you use a body for the arrow function?

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,242 @@
...
+        .then(sendToApi({ apiConfig, signal }));
+    },
+
+    markSecondaryEmailAddressAsPrimary (address) {
+      return addEmailAddress({ email: address }, {
+        // eslint-disable-next-line camelcase
+        make_primary: true,
+      })
+        .then(sendToApi({ apiConfig, signal }));
+    },
```
hsablonniere, line 240: We don't need the signal on those as they are update calls and we want them to go through even if there's a navigation.

src/translations/translations.en.js
```diff
@@ -221,6 +221,38 @@ export const translations = {
...
+  'cc-email-list.secondary.address-input.error.invalid': () => sanitize`Invalid email address format.<br>Example: john.doe@example.com.`,
+  'cc-email-list.secondary.address-input.error.used': `This email address does not belong to you`,
+  'cc-email-list.secondary.address-input.format': `name@domain.com`,
+  'cc-email-list.secondary.address-input.label': `Email address`,
+  'cc-email-list.secondary.description': `Unlike the primary email address, no notifications are sent to these addresses. You can be invited inside an organization with any of your secondary email addresses.`,
```
hsablonniere, line 252: :gb:  organisation with an S

hsablonniere (changes requested):
Impressive work :clap: 

* I think I would have aligned the mark as primary and delete button with the email (and not the icon)
  * Same with the verified label on the primary email with long address
  * Maybe we already discussed this, it's something up to discussion, but both are OK with me ;-)
* I'm not sure about the simulations as we won't add a secondary address directly in the list without a validation first, right?

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,242 @@
...
+        return { self, secondary };
+      });
+    },
+
+    fetchPrimaryEmailAddress () {
+      return Promise.resolve({
+        method: 'get',
+        url: `/v2/self`,
+        headers: { Accept: 'application/json' },
+      })
```
hsablonniere, line 206: Sad but it seems the clever-client does not provide such a call.

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -0,0 +1,242 @@
...
+      return getEmailAddresses()
+        .then(sendToApi({ apiConfig, signal }));
+    },
+
+    sendConfirmationEmail () {
+      return Promise.resolve({
+        method: 'get',
+        url: `/v2/self/confirmation_email`,
+        headers: { Accept: 'application/json' },
+      })
```
hsablonniere, line 220: This one is available on `user` as `todo_getConfirmationEmail`. Can you use it with an alias?

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,404 @@
...
+    if (this.addEmailForm.address.error == null) {
+      dispatchCustomEvent(this, 'add', address);
+    }
+  }
+
+  updated (_changedProperties) {
+    if (this.addEmailForm.address.error != null) {
+      setTimeout(() => this._focusAddressInput(), 0);
+    }
+  }
```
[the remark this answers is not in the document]
pdesoyres-cc: With the `update` method, the focus behavior is entirely driven by the data model. When the error comes from the server (from the smart component) I don't need to focus the input with a public method explicitly. Instead, I just set the right error code in the data model, and the component focuses the text input automatically.

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,404 @@
...
+
+  _getVerifiedTagLabel () {
+    const verifiedLabel = i18n('cc-email-list.primary.email.verified');
+
+    if (this._isLoading()) {
+      return fakeString(verifiedLabel.length);
+    }
+    return this.emails.value.primaryAddress.verified ? verifiedLabel : i18n('cc-email-list.primary.email.unverified');
+  }
```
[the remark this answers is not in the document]
pdesoyres-cc: you're right

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,404 @@
...
+    const addresses = this._getSecondaryAddresses();
+    const markingAsPrimary = addresses.some((item) => item.state === 'marking-as-primary');
+
+    return html`
+      <cc-block-section>
+        <div slot="title">${i18n('cc-email-list.secondary.title')}</div>
+        <div slot="info">${i18n('cc-email-list.secondary.description')}</div>
+
+        <ul class="secondary-addresses">
+          ${addresses.map((secondaryAddress) => {
```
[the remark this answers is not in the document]
pdesoyres-cc: true, We don't really need it here

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,404 @@
...
+    return html`
+      <cc-block>
+        <div slot="title">${i18n('cc-email-list.title')}</div>
+
+        ${when(this.emails.state === 'error', () => html`
+          <cc-error>${i18n('cc-email-list.loading.error')}</cc-error>
+        `, () => html`
+          ${this._renderPrimarySection()}
+          ${this._renderSecondarySection()}
+        `)}
```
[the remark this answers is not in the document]
pdesoyres-cc: much cleaner. Thank you

src/components/cc-email-list/cc-email-list.js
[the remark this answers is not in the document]
pdesoyres-cc: I would say: If it's a big list with many refresh => use `repeat`. Otherwise, you can stick on `map`.
This being said, when using react you just don't have a choice: you must use a repeat-like function to display lists.

src/components/cc-email-list/cc-email-list.js
```diff
@@ -0,0 +1,404 @@
...
+          flex-wrap: wrap;
+          gap: 1em;
+        }
+
+        .address-line .address.loading {
+          opacity: var(--cc-opacity-when-disabled);
+        }
+
+        .address-line .address span {
+          min-width: 15em;
```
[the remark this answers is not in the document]
pdesoyres-cc: In fact, I just want this for secondary addresses to avoid this:
image
By the way, if you have a better solution to handle that, I'd be glad to hear about it.

src/components/cc-email-list/cc-email-list.smart.js
pdesoyres-cc: Some arguments for this:

* improve verbosity
* in the code, you manipulate a variable named `api` which guides the reader about what is involved
* If you want a temporary mocked `api` it may be easier to write.

src/translations/translations.en.js
pdesoyres-cc: There is another occurrence of `organization` on `cc-header-orga.error`. If it's ok for you, I'll fix it with this PR

src/components/cc-email-list/cc-email-list.smart.js
pdesoyres-cc: thanks

pdesoyres-cc:
First of all, I want to thank the team for the reviews. I really like our process: it makes my day-to-day work very satisfying :heart: 

About the email icon, several people don't like the `envelope` icon from remix icon (personally, I don't like it either). So I tried to modify the icon to make it more rectangular:

image
becomes
image

Please let me know if your ok with allowing some customization when we don't like the icon proposed by remix.

> I think I would have aligned the mark as primary and delete button with the email (and not the icon)
> * Same with the verified label on the primary email with long address
> * Maybe we already discussed this, it's something up to discussion, but both are OK with me ;-)

I can't remember if we talked about that. Here is a preview so that we can vote:
* :lion: [align with icon](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-email/index.html?path=/story/%F0%9F%9B%A0-profile-cc-email-list--data-loaded-with-huge-email)
* :zebra: [align with email](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-email-align/index.html?path=/story/%F0%9F%9B%A0-profile-cc-email-list--data-loaded-with-huge-email)

However, I don't understand your point about the alignment for the primary email. As of now, the verified label wraps under the address.

>    I'm not sure about the simulations as we won't add a secondary address directly in the list without a validation first, right?

Your right @hsablonniere, the simulations are useless: I removed them

florian-sanders-cc (changes requested):
Small detail because I forgot to check the part about `cc-button` link style.
Apart from this, all good, Gg ! :clap:

src/components/cc-button/cc-button.js
```diff
@@ -460,10 +460,15 @@ export class CcButton extends LitElement {
...
 
         .outlined progress,
-        .outlined progress::-webkit-progress-bar {
+        .outlined progress::-webkit-progress-bar{
           background-color: var(--btn-color);
         }
 
+        .cc-link progress,
+        .cc-link progress::-webkit-progress-bar {
+          background-color: var(--cc-color-text-strong);
```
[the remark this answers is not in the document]
pdesoyres-cc: Good catch!

pdesoyres-cc:
Found a very weird bug on Safari and only when integrated into the Console. Ths skeleton animation doesn't stop immediately after the cc-badge 'skeleton' property is switched to false:
- the component loads: the badge is in skeleton mode
- the API is called: the badge is still in skeleton mode
- the API responds: the badge is no more in skeleton mode
  => :bug: the badge is well displayed BUT the skeleton animation is still played for a certain amount of time.

When the primary email is validated:
cc-email-list-safari-bug-1

When the primary email is not validated (here the email also still has the animation running):
cc-email-list-safari-bug-2

After struggling, I found that feeding the cc-badge slot with a TextNode is something Safari doesn't like. The workaround I found is as follows:
```
<cc-badge>Verified</cc-badge>
```
becomes
```
<cc-badge><span>Verified</span></cc-badge>
```

To track the reason for this workaround, I will add a comment in the code with a link to this PR comment.

PR #474, opened 2022-05-25
cc-toast and cc-toaster: init components

### LINKED ISSUE #374: cc-toast and cc-toaster: init components
## Context

We want to display a message to the user after he performs an action on the console. These messages must be non-disruptive. They are displayed in a non-modal box that disappears automatically after a certain time.

Related issues:
#424 

To do that, we implement two components:

* a cc-toast for the box containing the message
* a cc-toaster for managing the toasts: their alignment, the animation played when they appear or disappear, etc.

### Buttons and accessibility

Do we allow buttons inside a toast? there is an accessibility issue with that:

* https://www.scottohara.me/blog/2019/07/08/a-toast-to-a11y-toasts.html

### Close action

* Do we have a `close` action? is it an option? what may be the default value? does it depend on the intent?

## cc-toast component

The component displays:

* a message
* an icon
* an optional title
* a close button
* a progress bar showing the timeout progress

The component has an `intent` property that drives:

* the color of the box
* the icon

After a certain amount of time, a dismiss event is sent. The cc-toaster listens to these events and reacts to them by removing the toast from the DOM.

#### Attributes/properties

| `key` | `string` | `null` | `true` | Identifier of the component that will be sent along with the dismissed event |
| `message` | `string` | `null` | `true` | The message to display |
| `title` | `string` | `null` | `false` | The message to display |
| `intent` | `Intent` | `'info'` | `false` | Email address form input error key |
| `timeout` | `number` | `5000` | `false` | Amount of time after the dismiss event is closed. `0` for no timeout. |
| `closeable` | `boolean` | `false` | `false` | Whether to show a close button |
| `showProgress` | `boolean` | `false` | `false` | Whether to display a progress bar showing the timeout progress |

### Outputs

| `dismiss` | `DismissEventDetail` | Triggered when the cc-toast is dismissed (by user action on the close button or by the timeout expiration) |

### Type Definitions

```typescript
export type Intent = 'info'|'success'|'warning'|'danger';

export interface DismissEventDetail {
  key: string;
  timeoutExpired: boolean;
}

```

### Stories & states

* With title but no message
* With message but no title
* With message and title
* With all different intents
* With timeout set to 0
* With closeable set to true

---

## cc-toaster

The cc-toaster component is responsible for managing the toasts lifecycle.

Two alternatives to display a new toast:

* use a public method #show
* dispatch a custom event `cc-toast:show` along with an object describing the message to display (see below the `Toast` interface).

#### Attributes/properties

| `position` | `ToastPosition` | `'top-right'` | `false` | The position in the screen where the toaster is placed. This property doesn't place the toaster on the screen. It only indicates where it is, and this helps to apply the right alignment and the right ordering of the toasts |
| `maxToasts` | `number` | `null` | `false` | The maximum number of toasts that can be displayed at a time. If a new toast is to be displayed, the last one is dismissed automatically |
| `animation` | `ToastAnimation` | `'fade'` | `false` | The animation to be played when the toast appears |
| `toastDefaultConfig` | `ToastConfig` | `'fade'` | `false` | The default configuration to apply to every cc-toast |

### Outputs

| `dismiss` | `ToastDismissEventDetail` | Triggered when a `cc-toast` is dismissed by the public `dismiss` method |

### Type Definitions

```typescript
export type ToastIntent = 'info'|'success'|'warning'|'danger';

export type ToastPosition = 'top'|'top-left'|'top-right'|'bottom'|'bottom-left'|'bottom-right';

export type ToastAnimation = 'fade'|'slide'|'fade-and-slide';

export interface Toast {
    key?: string;
    message?: string;
    title?: string;
    intent?: ToastIntent;
    config?: ToastConfig,
    _dismissing?: boolean;
}

export interface ToastConfig {
    timeout?: number;
    closeable?: boolean;
    showProgress?: boolean;
}

export interface ToastDismissEventDetail {
    key: string;
    timeoutExpired: boolean;
}
```

### public methods

* `show(toast: Toast): string`: displays a new toast and returns the `key` of the toast. If not provided, a new one is created. This `key` is the unique identifier of the toast that will be passed along with the `dismiss` event.
* `dismiss(key: string)`: dismisses the toast identified by the given `key`.

### Stories & states

* One single story with a control panel for toasting

pdesoyres-cc:
Fixes #374

* [Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-toast/index.html?path=/story/%F0%9F%9B%A0-toast-cc-toast--default-story)


florian-sanders-cc:
:wave: Thanks for the preview, especially the one with the console screenshot and the dashboard to control the toaster positioning, it's perfect :+1: 

My favorites:

1. Preview 3 :heart_eyes: 
2. Preview 2.

roberttran-cc:
Thanks a lot for the many previews: they give a lot of helpful insights. And I'll vote for the number 3 as well!

florian-sanders-cc:
:wave: definitely the horizontal one for me :+1:

roberttran-cc:
Just another thought...

If there's two toasts at the same time, let's say on top side of the screen, and we hover the lower one:
- the lower one is paused while the upper one's timer still goes on until it disappears,
- as the layout shifts, the cursor is not on the previously-lower-now-upper-and-only toast anymore and its timer resumes until it disappears!

It's a bit far-fetched but I don't know if we should deal with it (like pausing all visible toasts) as the user intent is not maintained in time. :thinking: 

---

I tried to illustrate it with the following recording (my cursor is not visible but assume it is on the blue toast on which i select the text).

Screencast from 06-10-2022 04_03_22 PM

pdesoyres-cc:
Good catch @roberttran-cc! From my point of view, I don't think we should handle this because:

* I don't think that having multiple toasts at the same time will happen a lot
* I don't think that the user will often need to copy the message

It is possible to implement that but with a big cost:

* complexity of code
* maintainability

Again, it's only my point of view, so if you really think this is worth doing, I can try.

florian-sanders-cc:
:wave: Here is some feedback !

Overall I think the UI looks good.

- the `tabindex="0"` attribute should only be present if the `timeout` prop value is `true`. We only need to let people focus the element if there is a timeout to pause.
- I don't really like the 0 border radius on right sides, but it really is a personal opinion so if it's only me, you can ignore this feedback ;) 
- I did not look much at the code but I just tested with increased font-size in the browser and I think the progress bar height should be set with em (since we are ditching rem :cry:). I would also set border radius with relative units like 0.3em so that it scales a little bit better with the font size but this is very much a detail ;)
- I just realized that the icon actually conveys information (the intent of the toast). Its alt attribute value should reflect that. Maybe something like `alt="error"` for danger, etc. (the simpler the better).

roberttran-cc:
Quick review code-wise: everything looks really clean, and the comments are really nice, great effort!

One thing: maximum number of toasts does not seem to work... Or I didn't figure how it is supposed to work? :thinking:

hsablonniere:
## UI review

This is great, I'm really looking forward to have this in the console :heart: 

* I really like the UI
* Having the possibility to pause and copy-paste on hover is so good
* I'm also a bit puzzle with the lack of border radius on the right but it's a personal opinion and small detail
* I love the animations, especially the slide + fade
* I love the toaster story/simulator
* As @rob said it, the max number of toasts doesn't seem to work
* Same feedbacks as @flo without looking a the code => please use ems for the progress bar and border rardius
* About the alt="error" etc... I'm OK with these but we'll need to translate them.

### About the pause and multiple items

I love that @rob tried this tiny detail but I think I agree with @pier on the it's a lot of code for a situation that is not so common.
I propose that we move forward without and see what people say.

hsablonniere:
## Code review

* I like the order of the commits <3 (add one thing and then use it)

### cc-toast

* Why is the class `CcToastComponent` not named `CcToast`?
* We could move the Timer to a different file (in the lib directory) and maybe create unit tests (not always easy with timers)
* Timer: I don't understand the onClear, onResume
  * Looking at the usage, I think we can simplify with only one callback
* I'm also considering inlining the Timer as we just need one per toast and it would simplify the code
* I'm sorry but I'm not a great fan of achieving the "subtle close button" with two CSS parts that open way too many customizations
  * We can either use a `<button>` directly like we do in the `<cc-input-*>`
  * or we can create a new mode for the `<cc-button>` for this purpose
* I would move this sentence in details and keep a one-line description
  * "This component is not to be used directly but should be used through the `cc-toaster` component."
* The docs in details is really nice and detailed :+1:
* The `@cssdisplay block` JSDoc should be the first (before event)
* I'm not sure we really need those CSS custom properties for customization
* What's the purpose of the `key` property?
  * I think I got the intent but I wish we could do without
  * I also wish it could be an implementation detail
* Why an object with `DEFAULT_TOAST_CONFIGURATION` and not directly multiple `const` ?
* In the cc-example-component, we tried to use a more compact JSDoc style (one-liner) for types in for constructor
* Interesting use of a region for private methods ;-)
* The cc-example-component says: "// use a private method to handle the event and prefix it with `_on`."
  * you used `handleSomething`
* Because `_handleMouseEnter` is not always use for mouseEnter, I would rename it to `_pause` and rename `_handleMouseLeave` to `_resume`
* I know it's personal, but I think this is more readable:
  * const tabIndex = this.timeout > 0 ? '0' : undefined;
  * const tabIndex = (this.timeout > 0) ? '0' : undefined;
* In some situations like this, we can put these templates on a single line
* I would be in favor of flattening the 3 config properties directly on the toast object

```
${this.message ? html`
  <div class="message-wrapper">${this.message}</div>
` : ''}
```

* Should we handle special combination of with blur/focus on close button AND mouseout?
* I'd rather use a "private" CSS custom prop for this `animation-duration: ${this.timeout}ms`
* We could rely on the intent attribute instead of applying a class to style stuffs
  * I'm not sure if it's better or not
* There's a weird white border around the colored rectangle on the left
  * I think it comes from the `padding-right: calc(var(--cc-toast-padding) + var(--cc-toast-border-width));`
* The progress bar animation seems a bit complex
* We could also use the Element.animate method and maybe simplify this
  * I tried and it's kinda fun ;-) I'll show you
* Not a great fan of those selectors `.progress-bar > div {`
* I think we need to move the progress-bar in content so we can use full width and no calc

### cc-toaster

* `This component is responsible for handling the `cc-toast` lifecycle.`
  * I would write it differently, something like "display zone for toasts" but we need to discuss it ;-)
* `Add this component only once in your application, place it where you want, and you'll have the ability to display some notification messages (`cc-toast`).`
  * I would make it more explicit that this component listens to X events and displays toasts
* `### Controlling maximum number of toasts.`
  * We can remove the dot at the end of a title
* Nice idea of having a `toastDefaultConfig`
  * There seem to be a conbination of multiple layers of defaults, I think we can simplify this.
  * I'm not a great fan of using the default `const` from toast in the toaster, we have double defaults
* Why is there a public `show` method on the toasterElement?
  * Do we need this?
* "The `cc-toaster` listens to these events and react by removing the corresponding `cc-toast` element from the DOM."
  * react => reacts
* Why is there a way to dismiss toasts programatically?
  * When do we intend to use this?
* `This triggers a css animation.`
  * css => CSS
* I think we cannot use markdown after using JSDoc tags
* Don't forget to order properties (alpha)
* Same as cc-toast, please use a compact JSDoc definition (one-line) in the constructor
  * For the toastDefaultConfig, we may need several lines
* The @private is not necessary (I think...)
* I like the `_attachEventHandler`, it would be interesting to move it in lib at some point and even a reactive controller once we have Lit 2
* You don't need to use the dot `.foo` assignment style for strings (or maybe you do if they can be null :-( )
* For booleans, please use `?myBoolean`
* Nice usage of `@media (prefers-reduced-motion) {`
  * I'm not sure you need to use `.toaster cc-toast`, just `cc-toast` should be enough
* `class CcToasterComponent` => `class CcToaster`
* I'm not a great fan of adding private attributes on toast objects `toast._dismissing`
* I'm a bit confused but the `dismiss` vs `_dismiss`
  * We could go for a private low lever `add/remove` that take care of the animation
* We should look at this in the future https://lit.dev/playground/#sample=examples/motion-list
* I tried to use element.animate to remove toasts, it feels cleaner but it's not that simple for toast insertions
  * should we use both CSS animation + JS animation ?
* I think the slide animation only animates the current toast
  * How hard would it be to also hide the other ones

pdesoyres-cc:
Thank you everyone for the feedback. I've taken them into account:

**cc-toast**
* rename the component class
* remove the timer and use animate API
* use `<button>` for closing toast
* remove custom CSS properties
* rename some methods, reorder properties, inline JSDoc
* use intent attribute instead of class
* simplify the progress bar placement (use grid instead of flex)
* fix the weird blank border when zooming on chrome
* remove the `DEFAULT_TOAST_CONFIGURATION` constant.
* add an `alt` attribute to the icon.
* use same border radius for all corners
* use `em` size unit for border radius and progress bar height.
* set tabindex only when timeout > 0

**cc-toaster**
* rename the component class
* rework the doc
* remove public methods
* add a new way to dismiss a toast by adding a `dismiss` function to the `cc-toast:show` event. (thanks to [Justin Fagnani](https://www.youtube.com/watch?v=x9YDQUJx2uw))
* replace the `_dismissing` private `Toast` property with an internal `_dismissingSet` set
* fix the `maxToasts` limit handling

I have published a new [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-toast/index.html?path=/story/%F0%9F%9B%A0-toast-cc-toaster--default-story).

src/toast/cc-toast.js
```diff
@@ -0,0 +1,348 @@
...
+        ${this.closeable ? html`
+          <div class="close-button-wrapper">
+            <button
+                class="close-button"
+                @click=${this._onCloseButtonClick}
+                @focus=${this._pause}
+                @blur=${this._resume}
+                title="${i18n('cc-toast.close')}"
+            >
+              <img src="${closeSvg}" alt="${this._getIconAlt()}"/>
```
florian-sanders-cc, line 196: the `alt` of the image is used as the label (accessible name) of the button.
It should describe the action (close) and not the type of toast.

```html
   <img src="${closeSvg}" alt="${i18n('cc-toast.close')}"/>
```

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,447 @@
...
+          display: flex;
+          flex-direction: column;
+          gap: 1em;
+          pointer-events: none;
+        }
+        
+        /*region POSITION*/
+        .toaster.left {
+          align-items: start;
+        }
```
florian-sanders-cc, line 311: line breaks

florian-sanders-cc (commented):
I added a few comments but most of my comments are about insignificant details :smile: 

I didn't get the chance to say it during the sync review but this is impressive and it's great :clap: !

src/toast/cc-toast.js
```diff
@@ -0,0 +1,348 @@
...
+    /** @type {string|null} The message to display. */
+    this.message = null;
+
+    /** @type {boolean} Whether to show the progress bar. */
+    this.showProgress = false;
+
+    /** @type {number} The amount of time (in millis) before the `cc-toast:dismiss` event is fired. The timer starts as soon as the component is attached to the DOM. */
+    this.timeout = 5000;
+
+    /** @type {string|null} The title to display. */
```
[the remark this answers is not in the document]
pdesoyres-cc: Good catch. I'll stand for `header`

roberttran-cc:
Everything looks good to me! And congrats on the `.animate()` use, looks neat! :tada:

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+
+      if (activeToasts.length >= this.maxToasts) {
+        this._dismiss(activeToasts[activeToasts.length - 1].key);
+      }
+    }
+
+    const completedToast = this._completeToast(toast);
+
+    // do nothing if a toast with the same key already exists.
+    if (this._toasts.find((toast) => toast.key === completedToast.key)) {
```
hsablonniere, line 175: Can you explicitly compare to null:

```
if (this._toasts.find((toast) => toast.key === completedToast.key) != null) {
```

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+   * @param {Toast} toast the toast to show
+   * @returns {Toast|undefined} the completed toast. Returns undefined if no toast was shown.
+   */
+  _show (toast) {
+    if (this.maxToasts === 0) {
+      return;
+    }
+
+    // dismiss last toast if max is reached
+    if (this.maxToasts) {
```
hsablonniere, line 164: Can you do an explicit check with null or `> 0`?

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+      const activeToasts = this._toasts.filter((toast) => this._isDismissing(toast.key) !== true);
+
+      if (activeToasts.length >= this.maxToasts) {
+        this._dismiss(activeToasts[activeToasts.length - 1].key);
+      }
+    }
+
+    const completedToast = this._completeToast(toast);
```
hsablonniere, line 172: I would rename `completed` toast to something else.

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+    this._toasts = [completedToast, ...this._toasts];
+
+    return completedToast;
+  }
+
+  _dismiss (key) {
+    dispatchCustomEvent(this, 'cc-toast:dismiss', { key, timeoutExpired: false });
+  }
+
+  _startHide (key) {
```
hsablonniere, line 188: I'm not sure but it seems like we use `hide` and `dismiss` for the same thing. Maybe we don't...

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  /**
+   * @param {AnimationEvent} e
+   * @private
+   */
+  // TODO: in the future (when we have Lit2) we may want to implement that with https://github.com/lit/lit/tree/main/packages/labs/motion
+  _onHideAnimationEnd (e) {
+    const toast = this._getToastByKey(e.target.key);
+    if (toast) {
```
hsablonniere, line 203: Please add a != null

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+      this._toasts = this._toasts.filter((t) => t !== toast);
+      this._dismissingSet.delete(toast.key);
```
hsablonniere, line 205: This is weird because the `this._dismissingSet.delete(toast.key);` requires a `this.requestUpdate();` but the presence of `this._toasts = ` makes it useless...

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+    }
+  }
+
+  _isDismissing (key) {
+    return this._dismissingSet.has(key) === true;
```
hsablonniere, line 210: Here the ` === true` is useless as it already returns a boolean ;-)

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+
+    this._detachToastShowEventListener = attachEventListener('cc-toast:show', (event) => {
+      const toast = this._show(event.detail);
+      if (toast) {
+        event.dismiss = () => {
+          this._dismiss(toast.key);
+        };
+      }
+      else {
+        event.dismiss = () => {};
```
hsablonniere, line 226: I would prefer this to be called `dismissToast`

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+          flex-direction: column-reverse;
+        }
+        /*endregion*/
+
+        /*region ANIMATION*/
+        @media (prefers-reduced-motion) {
+          cc-toast {
+            animation-duration: 0s !important;
+          }
+        }
```
hsablonniere, line 310: We could mention this in the docs.

stories/toast/cc-toaster.stories.js
```diff
@@ -0,0 +1,276 @@
...
+        return html`<cc-button
+            danger
+            image=${intentsButtonProp[intent]?.icon}
+            outlined
+            data-intent="${intent}"
+            @cc-button:click=${onClick}>
+          ${intent}
+        </cc-button>`;
+      }
+    }
```
hsablonniere, line 205: It feels like we could inline this...

stories/toast/cc-toaster.stories.js
```diff
@@ -0,0 +1,276 @@
...
+                               @cc-input-number:input=${onMaxToastChanged}
+              ></cc-input-number>
+            </div>
+            <div>
+              <span>Position</span>
+              <cc-toggle value=${position} .choices=${positions} @cc-toggle:input=${onPositionChanged}></cc-toggle>
+            </div>
+            <div>
+              <span>Kind of animation to be played when the toast appears</span>
+              <cc-toggle value=${animation} .choices=${animations} @cc-toggle:input=${onAnimationChanged}></cc-toggle>
```
hsablonniere, line 235: The `<cc-toggle>` has a `legend` attribute.

hsablonniere (commented):
We're almost there, I propose that we debrief this synchronously :wink: 

## cc-toast

* The docs is really really nice :heart:
* I like the usage of `background-color: var(--toast-color);` and then modification through `:host([intent="XXX"])`
* Nice `getItems()` helper in the story

## cc-toaster

* Again, the docs is really really nice :heart:
* nice usage of the event.dismiss(), I would rename it dismissToast to be more explicit
* The system we currently have to insert arbitrary HTML/DOM is limited, I'd rather avoid using lit inside our stories and use pure DOM but I don't want to ask you to rework everything

## others

* nice touch the `attachEventListener` in a lib

src/toast/cc-toast.js
```diff
@@ -0,0 +1,348 @@
...
+        /*endregion*/
+
+        /* region PROGRESS */
+        .progress-bar {
+          background-color: transparent;
+          bottom: 0;
+          grid-column-start: 2;
+          height: 0.3em;;
+          position: absolute;
+          width: 100%;
```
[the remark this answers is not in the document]
pdesoyres-cc: Works fine for

* linux: chromium, brave, firefox
* windows: chrome, firefox, edge

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+   * @param {Toast} toast
+   * @returns {Toast}
+   * @private
+   */
+  _completeToast (toast) {
+    return {
+      ...toast,
+      key: toast.key ?? Math.random().toString(36).slice(2),
+      config: {
+        ...this.toastDefaultConfig,
```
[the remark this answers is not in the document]
pdesoyres-cc: yes it works:

`console.log({...null})` return an empty object.

src/toast/cc-toaster.js
pdesoyres-cc: I wanted to include `NaN` in the condition.

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+
+  /* region PRIVATE METHODS */
+  /**
+   * Based on the given toast, forges a new toast where the default config is applied, and where the key is generated if not present.
+   *
+   * @param {Toast} toast
+   * @returns {Toast}
+   * @private
+   */
+  _completeToast (toast) {
```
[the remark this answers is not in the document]
pdesoyres-cc: I don't like it neither :laughing: . Let's go for `initToast`

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,432 @@
...
+    return completedToast;
+  }
+
+  _dismiss (key) {
+    dispatchCustomEvent(this, 'cc-toast:dismiss', { key, timeoutExpired: false });
+  }
+
+  _startHide (key) {
+    const toast = this._getToastByKey(key);
+    if (toast) {
```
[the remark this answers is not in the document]
pdesoyres-cc: as `this._getToastByKey(key)` returns a `Toast` or `undefined`, I will compare to `undefined` explicitly.

src/toast/types.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  options?: ToastOptions;
+}
+
+export interface ToastOptions {
+  timeout?: number;
+  closeable?: boolean;
+  showProgress?: boolean;
+}
+
+export interface Toast extends Notification{
```
hsablonniere, line 20: Missing a space after `Notification`

stories/tcp-redirections/cc-tcp-redirection-form.smart.md
```diff
@@ -5,6 +5,13 @@ title: '💡 Smart'
 
 # 💡 Smart `<cc-tcp-redirection-form>`
 
+<cc-smart-container context='{
+  "ownerId": "orga_3547a882-d464-4c34-8168-add4b3e0c135",
+  "appId": "app_b75977aa-563f-40fd-a592-224a5f6afbd6"
+}'>
+  <cc-tcp-redirection-form></cc-tcp-redirection-form>
+</cc-smart-container>
```
hsablonniere, line 13: I think you missed this. This file should not be modified in this PR.

src/toast/cc-toast.js
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * A component dedicated to display a short message and dismiss it after a short period of time.
+ *
+ * ## Details
+ *
+ * This component is not to be used directly but should be used through the `cc-toaster` component.
+ *
+ * * The component displays a message. Optionally, this message can be topped by a heading.
+ * * The component dispatches a `cc-toast:dismiss` event after a certain amount of time. The timer starts when the component is attached to the DOM.
+ * * Optionally, it can be closed manually with a close button.
```
hsablonniere, line 27: I think we should be more explicit on the fact that clicking on the close button will also trigger a `cc-toast:dismiss` event.

src/toast/cc-toast.js
```diff
@@ -0,0 +1,339 @@
...
+ * * Optionally, it can display a progress bar showing to the user that the message will be dismissed in a short amount of time.
+ * * When user moves the mouse over the component, the timer is paused. It will be resumed as soon as the user moves the pointer out.
+ * * As soon as a timeout is set, the toast becomes focusable. Gaining the focus will pause the timer. Losing the focus will resume the timer.
+ *
+ * The `message` property can be a simple string, but you can also pass a DOM `Node`.
```
hsablonniere, line 32: Can we add some kind of warning about this so users understand that they must be careful not to pass any DOM (code injection...)

src/toast/cc-toast.js
```diff
@@ -0,0 +1,339 @@
...
+    if (this.intent === 'warning') {
+      return i18n('cc-toast.icon-alt.warning');
+    }
+    if (this.intent === 'danger') {
+      return i18n('cc-toast.icon-alt.danger');
+    }
+  }
+
+  _onCloseButtonClick () {
+    this._dismiss(false);
```
hsablonniere, line 119: You can remove `false` now

hsablonniere: FYI: using the `change signature` would have helped you not to forget this.

src/toast/cc-toast.js
```diff
@@ -0,0 +1,348 @@
...
+        /*endregion*/
+
+        /*region CONTENT*/
+        .content {
+          background-color: var(--color-bg-default);
+          color: var(--toast-color);
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
+          height:100%;
```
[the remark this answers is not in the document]
hsablonniere: This is still missing @pdesoyres-cc

src/translations/translations.en.js
```diff
@@ -660,6 +660,13 @@ export const translations = {
     return `${formattedValue} ${request} (${formatPercent(lang, percent)})`;
   },
   //#endregion
+  //#region cc-toast
+  'cc-toast.close': `Close`,
```
hsablonniere, line 664: @florian-sanders-cc do we need to improve this with some kind of `close this toas/notification`?

src/toast/cc-toaster.js
```diff
@@ -0,0 +1,402 @@
...
+    `;
+  }
+
+  /**
+   * @param {Toast} toast
+   */
+  _renderToast (toast) {
+    const dismissing = this._isDismissing(toast.key);
+    const onAnimationEnd = dismissing ? () => this._finishDismissing(toast.key) : undefined;
+    const onDismiss = dismissing ? undefined : () => this._startDismissing(toast.key);
```
hsablonniere, line 229: It feels like too much protection. Will this ever happen?

hsablonniere (commented):
## Review

* Nice touch for the ADR 14 type (could have been in its own commit but it's OK)
* Nice of you to complete a few type annotations here and there
* Can you add a small document in the `docs/guides` to briefly describe "How to use the notification/toaster system"?
* I think we can simplify something a bit more now that the `Toast` object is something that is only known and used inside the toaster.

pdesoyres-cc:
Facing a bug with safari (https://bugs.webkit.org/show_bug.cgi?id=242513) we need to find a workaround.

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,142 @@
...
+
+We designed a notification system based on the toast pattern. This document explains how to integrate it in your application.
+
+We provide two web components:
+
+1. The `cc-toast` component that represents a single notification.
+2. The `cc-toaster` component that helps you display them on your application.
+
+First of all, you need to place the `cc-toaster` somewhere in your application. A common practice is to have a single
+`cc-toaster` for your whole application and to place it at the end of the `body` tag.
```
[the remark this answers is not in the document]
hsablonniere: This is true for the whole document.

hsablonniere:
@pdesoyres-cc As we discussed on 2022-07-13, this is almost ready :pray:

TODO:

* [ ] Update the docs as we discussed

src/toast/cc-toast.js
```diff
@@ -0,0 +1,336 @@
...
+            ${this.heading ? html`<div class="heading">${this.heading}</div>` : ''}
+            ${this.message ? html`<div>${this.message}</div>` : ''}
+          </div>
+
+          ${this.closeable ? html`
+            <button class="close-button"
+                    @click=${this._onCloseButtonClick}
+                    @focus=${this._pause}
+                    @blur=${this._resume}
+                    title="${i18n('cc-toast.close')}">
```
florian-sanders-cc, line 185: Let's avoid relying on the `title` attribute alone to generate the accessible name of an element.
Specs say that this should work everywhere but most people say it's a bad idea to rely on it anyway because in reality, its support (as an accessible name) is dodgy.

Rely on the button content (the image in that case) to provide the accessible name:
`${i18n('cc-toast.close')}`

You can keep the `title` as it provides a tooltip for other users. It is redundant but it's tolerated when dealing with interactive elements that only contain images :wink: 

```js
<button class="close-button"
  @click=${this._onCloseButtonClick}
  @focus=${this._pause}
  @blur=${this._resume}
  title="${i18n('cc-toast.close')}">
    <img src="${closeSvg}" alt="${i18n('cc-toast.close')}">
</button>
```

florian-sanders-cc: (sorry for not pointing this out earlier, last time I reviewed this part, it was a `cc-button` and it uses an `aria-label` so it was ok)

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,129 @@
...
+cc-toaster {
+  position: fixed;
+  z-index: 1000;
+  margin: 1em;
+  top: 0;
+  right: 0;
+}
+```
+
+When doing that you need to tell the `cc-toaster` where you placed it so that he knows how to align toasts.
```
hsablonniere, line 55: Shouldn't we also mention animations here?

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,129 @@
...
+
+We designed a notification system based on the toast pattern.
+
+First, this document will explain how to use it.
+Then, it will show how we designed a fully decoupled way to integrate it into our _smart components_(*).
+
+## The toaster
+
+We provide two web components:
+1. The `cc-toast` component that represents a single notification.
```
hsablonniere, line 14: Can you put an empty line between paragraph and list?

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,129 @@
...
+
+```html
+<cc-toaster toast-default-options='{ "closeable": true, "timeout": 3000 }'></cc-toaster>
+```
+
+## Integration in smart components
+
+You must wonder what is a smart component as we never described it before (we planned to provide a complete documentation about it).
+All you need to know right now is that smart components are classic web components, but they have something special:
+They communicate with the outside. They are the glue between pure UI web components and the clever API.
```
hsablonniere, line 103:
```suggestion
All you need to know right now is that smart components are classic web components, but they have something special: they communicate with the outside.
They are the glue between pure UI web components and the Clever API.
```

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,129 @@
...
+And at some point, they need to trigger some notifications when something goes wrong while contacting our API.
```
hsablonniere, line 104: It's also when something goes right.

docs/guides/how-to-use-notification-system.md
```diff
@@ -0,0 +1,129 @@
...
+```javascript
+const toast = {
+  message: 'Something went wrong while contacting clever API.',
+  intent: 'danger',
+};
+
+node.dispatchEvent(new CustomEvent('cc:notify', {detail: toast, bubbles: true, composed: true }));
+```
+
+At this point, nothing really happen.
```
hsablonniere, line 120:
```suggestion
At this point, nothing really happens.
```

hsablonniere (approved):

Issue #419, opened 2022-04-13
cc-ssh-keys-thirdparty-list: init component


roberttran-cc:
# Context

A component that displays the list of all third-party SSH keys: initially supporting GitHub platform.

Each SSH key displays the following informations:
* the name,
* the public key value,
* the key fingerprint,
* a static icon.

Each SSH key possesses the following action:
* import.

# Data

* Read: https://api.clever-cloud.com/v2/github/keys

# Input

| Property | Type | Description |
| --- | --- | --- |
| `keys` | `Key[]` | list of keys |
| `skeleton` | `boolean` | UI skeleton state |
| `waiting` | `boolean` | UI waiting state |
| `error` | `null\|loading\|saving` | UI error states |

NOTE: error happening when trying to import a key might be handled with the console toast notification system.

# Output

| Event | Parameters | Description |
| --- | --- | --- |
| `import` | `{Key}` | when importing a key from the list |

# Stories

* Skeleton.
* Empty.
* Loaded.
* Waiting.
* Error (null).
* Error (loading).
* Error (importing).

# Wording and i18n

* Block title.
* Button labels.
* Help text.
* Empty state label.
* Error messages.

Issue #418, opened 2022-04-13
cc-ssh-key-list: init component


roberttran-cc:
# Context

A component that displays a list of SSH keys associated with your account and allows to add new ones:
* first displays a form to associate a new SSH key with your account;
* then displays the list of personal keys currently associated with your account;
* finally displays the list of keys available from third parties (GitHub for example) that you can associate with your account.

Each SSH key displays the following informations:
* the name,
* the key fingerprint,
* a static icon.

Each personal SSH key possesses the following action:
* delete.

Each third parties SSH key possesses the following action:
* import.

The user can manually add a key via a creation form.

# Datas

* Read (personal keys): https://api.clever-cloud.com/v2/self/keys
* Read (third parties keys): https://api.clever-cloud.com/v2/github/keys
* Create/import/delete: https://api.clever-cloud.com/v2/self/keys/my-new-key

# Inputs

| Property | Type | Description |
| --- | --- | --- |
| `keys` | `SSHKey[]` | list of personal keys |
| `keysThirdParties` | `SSHKey[]` | list of third parties keys |
| `newKey` | `SSHKey` | key being added |
| `adding` | `Boolean` | UI state when adding key |
| `deleting` | `String` | UI state when deleting a personal key (value is the key name) |
| `importing` | `String` | UI state when importing a third parties key (value is the key name) |
| `listing` | `Boolean` | UI state when listing personal keys |
| `listingThirdParties` | `Boolean` | UI state when listing third parties keys |
| `unlinked` | `Boolean` | UI state hen no third party account is linked to the main account |
| `error` | `String` | UI error states `null\|'loading'\|'loading-third-parties'\|'adding'\|'deleting'\|'importing'` |

```ts
interface SSHKey {
  name: string,
  key?: string,
  fingerprint?: string,
}
```

NOTE: some errors might be handled with the console toast notification system.

# Outputs

| Event | Parameters | Description |
| --- | --- | --- |
| `create` | `{SSHKey}` | when adding a new key |
| `delete` | `{SSHKey}` | when removing an existing personal key |
| `import` | `{SSHKey}` | when importing a third parties key |

# Stories

* Default - a single key.
* Multiple keys.
* Keys with long name.
* Third parties accounts unlinked.
* Skeleton.
* Waiting when adding personal key.
* Waiting when deleting personal key.
* Waiting when importing third parties key.
* Error when loading personal keys.
* Error when loading third parties keys.
* Error when new key name is empty.
[...]

PR #475, opened 2022-05-27
cc-ssh-key-list: init component

roberttran-cc:
Here are the last updates concerning the `cc-ssh-key-list` component:
- preview: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story
- issue: https://github.com/CleverCloud/clever-components/issues/418

# Last improvements

Mainly the UI:
- merged with #419: now includes the third parties section;
- various minor changes.

No need to check the code yet.

# To review: UI

Not happy with delete/import button positioning.

Not happy with the whole key block in fact. :sweat_smile:

# To review: component name

Since SSH keys are no more split in two distinct components, should we update the component name?

Check next comment to vote. :+1: :-1:

# To review: wording

Wording is hard. :sweat_smile:

About the third parties section, the console only manages for now GitHub account linkage. I try to avoid mentioning the name GitHub in the UI (and leaving "third parties") but it will feel weird when integrated in the console. That's why I left an occurrence in the left menu info.

And a little bit of the same problem when mentioning Clever Cloud: should we use "your (main) account" instead of "Clever Cloud account"?

All feedback are welcome!

# Misc

The component does not provide feedback if a key delete/import fails: we assume the context (the console) will handle the feedback (toast mechanism).


roberttran-cc:
Vote :+1: or :-1: to change the component name from `cc-ssh-key-list` to `cc-ssh-key-form` or not!

florian-sanders-cc:
Hey !

I think the UI looks clean and it feels easy to use (using `cc-block-section` was such a good call :+1: ). 

# Accessibility & UX:
## Error management inside the form

Right now, the form validation stops as soon as there is one error, meaning we display one error at a time to users. This is exactly what browsers do when using native controls (required, type email, etc.) but I really feel this is not the best for users.
I feel like we should indicate all errors. For instance, if one leaves both the fields empty and tries to submit, we should add error messages to both of these fields.
Of course in your case, it might seem a little bit dumb, because "who would try to submit a form without filling any of the fields ?!" but we may very well have longer forms (sign up process for instance), and applying the same rule for all our forms would be better I think (even if it means handling unrealistic cases like in this case) ?
If we agree, this could be added to our a11y guidelines (although this is more UX than a11y per se)

## Button labels

This is something we need to discuss together because I have the same "issue" but I think we could improve delete and import button labels by specifying the name of the item related to the button. For instance "Delete the Work PC key". This way people using screen readers can go from one button to another without having to check the context. Its a small benefit, especially in screens where there are not that many items but I feel like it's cheap enough that we should do it (and if we all agree, I would write this as an a11y guideline actually cause this is a common topic).

# Button positioning

I think the current positioning is acceptable, although I agree it might feel that there is something off. I think this is because we decided not to show any "border" for our "cards" and the whole UI is just floating without any distinction between elements other than spacing and headings. (this is just my opinion though, and I still think it could stay like that).

# Wording
Some suggestions that are far from perfect but could help you:

- "You need to associate an SSH key to your account to deploy via Git: associate a key with the dedicated form." => "You need to associate an SSH key to your account to deploy via Git. Use this form to do so.
- I feel like the "Create an SSH key" part is not that useful but I really don't know how to deal with it. Maybe we could sum up the rest of the process and link to documentation ? Like "Then you need to add your key to your SSH agent and copy the public key into this form. All of these steps are details in our SSH key documentation page" (Or maybe just don't provide any of the commands, sum up all steps and link to the documentation).
- (your keys section) "This is the SSH keys associated with your account." => "These are the SSH keys..."
- List the fingerprints of your keys" => "You can list the fingerprints..." (since its a very optional step, I wouldn't use imperative here). Maybe we could explain the use of doing that, like "If you want to know whether the key exists on your local machine, you can... and compare it with the fingerprints listed here" ?
- (Third parties keys) "This is the SSH keys" => "These are the SSH keys from your GitHub account. You can import them to associate them to your Clever Cloud account." (not a huge fan of my sentence though, but I think splitting sentences is better here)

pdesoyres-cc:
## Buttons

* Didn't we say that we should use the circle button mode when we do not have enough room for a standard button with a label. Here, do we have a lack of space that led us to choose a circle button?
* Moreover, a circle button next to a box (the fingerprint box) is a bit strange for me.

## Layout

If you want some ideas on the general layout, here is one. Do what you want with it :)

cc-ssh-key-list_btn-position-proposal

The same layout with labeled buttons which slightly erases the "floating" effect of the list items.

cc-ssh-key-list_btn-position-proposal-label-button

## component name

I think that we should stay with this naming: 
* `cc` prefix
* `ssh-key`: what we are talking about. In singular.
* `list`: This component is about managing a list of items.

By the way, I should rename the `cc-email` component into `cc-email-list`.

## Wording

Why don't you want to make `Clever Cloud` and `GitHub` labels appear?

roberttran-cc:
New iteration with this [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story).

# Last improvements

UI:
- global key item layout: thanks a lot for your suggestions, they were invaluable, the result is so much better!
- right-aligned documentation link in anticipation of future `<cc-block>` improvements,
- `inline-block` CLI commands.

Behavior:
- both creation form errors triggered at the same time.

Wording.

# Answers to previous questions

> Here, do we have a lack of space that led us to choose a circle button?

Nope, it is my bad: I followed `<cc-orga-member-card>` design but it wasn't adapted.

> Why don't you want to make `Clever Cloud` and `GitHub` labels appear?

For `Clever Cloud`, it is for the white/grey mark positioning. And for `GitHub`, it is in anticipation of other third party accounts. But as @hsablonniere stated, there's no need to pay too much attention to it for now.

> What did we decided for the loading errors, are we going to try a notice when the component is ready?

Definitively!

# To do

The public key input in the creation form has to be a `<textarea>` but via the `<cc-input-text>` component, which will need a modification beforehand.

florian-sanders-cc:
:wave: this looks really cool ! I love the new styling for SSH keys :heart_eyes: + the new import icon with text :+1: 

# Just a few notes about small details 

## `data loaded (third parties unlinked)` story

- "There are no GitHub account linked to your Clever Cloud account." => "There is no GitHub..." since we can only link one GitHub account at max
- Profile link => I would use a relative path so that it does not bring us to prod when we check it on next for instance (but also for instance if someone else runs our console on their environment, they would not want to be brought back to our prod I guess).

## `error (when adding new personal key)` story

- Is this story still relevant since we are going to use `cc-toast` to display these error messages ? (same for the related Simulation and same for errors on key import)

florian-sanders-cc:
sorry I forgot to look the the mobile views :sweat_smile: 

What do you think about aligning buttons to the right when viewport is small ?
image

I am not sure it is a good idea so this is really up for debate. I just felt it a little bit weird to see these buttons aligned on the left but it's not that big of deal.

pdesoyres-cc:
I really like it like that now. Great job @roberttran-cc!

On the point raised by @florian-sanders-cc about the alignment of the delete button when viewport is small, I remember we had this same discussion with the cc-orga-member-* component. Do we need to think about that when settling our design principles?

roberttran-cc:
Another round of [review](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story)!

### What changed

- (design) add key button now primary
- (design) add key button right-aligned in mobile view @florian-sanders-cc
- (design) unlinked account sentence updated @florian-sanders-cc
- (code) component properties major refactor

### What to review

Design-wise, not much except the modifications listed above.

Code-wise, I guess you can start looking at the component code (without the smart part).

### Misc

> "error (when adding new personal key) story" and related not relevant
@florian-sanders-cc

Agree! Will remove them when wiring the toast component.

> "button alignment on small viewport" should be a design principle?
@pdesoyres-cc

I agree there's probably something to dig here but it might be more complex than expected. For example, in the `cc-orga-member-list` component, buttons are right-aligned at 750px but left-aligned at 550px... :thinking:

hsablonniere:

* I'm struggling with the "the smart definition sets the strict miminum and the UI component relies on it's own defaults"
  * I really like how what you proposed requires less code but I feel like it's less explicit and requires thinking about how smart and UI will work together to produce the final output.
* We have 2 booleans for errors at the form level
  * I would prefer an `error` property on each form field
    * For now the type of this error would only be `required` but others could appear afterwards
* I like the union type of strings for the SshKey state instead of having a union type of interfaces at the root level
  * It's simpler and more efficient in this situation
* I need to dig the "ready and unlinked" state, I completly forgot about this situation
* I like the flattened `state`, `name` and `fingerprint`
* I'm skeptical about the "fetch the whole list again from the server"
  * I think we should rely on local state and "patch" the list
  * Especially since this screen won't have an impact on other users
  * It's a good way to approach optimistic UI
  * We have an SPA, it's a bit sad to propose a UX that is closer to a form POST + reload
* I think your usage of target still has some race condition problems
* I don't think we need to use immer that much :wink: 
  * Most of the time, the way we designed the new states is meant to give the smart definition the ownership of what's inside the UI component properties (state and data).
* It seems like you're refreshing the whole list on an update, what do you display while you do this?

pdesoyres-cc:
## State modeling

* I like the SSHKey lists state modeling
* I'm wondering if `KeyListsStateReadyAndUnlinked` may have its state property set to 'ready-and-unlinked' (so that the component will rely on the `state` property rather than on the `unlinked` boolean property)

## Smart component

* I think you should always apply the rule:
  * if I need to modify the component state, I must do it using the `CcEventTarget` indirection.
* Sometimes, you don't need immerJs because you want to override the full state. (for instance line 37, line 46)

## Component

* Maybe you should avoid using `this.requestUpdate();`. I can see a `TODO` just above, so maybe you already planned to remove it.

roberttran-cc:
Back and ready for review! :skull:
Thanks for your last feedback by the way!

Since much has changed (yes, state and smart, I'm looking at you :eyes:), a global review is the way to go IMHO.

Note: due to a bug in the current Storybook smart implementation, GitHub keys are sadly inaccessible. Ping me for a workaround (mocking the API resolve in the code :see_no_evil:).

About the GitHub list, after a check in our data, there are more users with no linked account than with their GH account associated (45% vs 55%, not much of a difference but a difference nevertheless).
That is why the component default ready state is with no GH account linked (although it feels a little weird in the code).

About refreshing list locally (without fetching both lists again), there's not much more we can do with the information currently at our disposal.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -0,0 +1,438 @@
...
+        </div>
+      </div>
+    `;
+  }
+
+  /**
+   * @param {"personal"|"github"|"skeleton"} type
+   * @param {SshKey[]} keys
+   */
+  _renderKeyList (type, keys = [{}]) {
```
pdesoyres-cc, line 241: I'm not very comfortable with this single entry that is used for skeleton type.
I would prefer to be more explicit:

```javascript
_renderKeyList (type, keys = []) {
  const skeleton = (type === 'skeleton');

  if (skeleton) {
    keys = [
      {
        state: 'idle',
        name: fakeString(32),
        fingerprint: fakeString(32),
      },
    ];
  }
  // ...
}
```

src/components/cc-ssh-key-list/cc-ssh-key-list.types.d.ts
```diff
@@ -0,0 +1,51 @@
...
+// when an error has occurred
+interface KeyDataStateError {
+  state: "error";
+}
+//#endregion
+
+//#region common
+interface FormField {
+  value: string;
+  error?: "required";
```
pdesoyres-cc, line 42: The form field is "required". But the error is "empty"

src/translations/translations.en.js
```diff
@@ -566,6 +566,39 @@ export const translations = {
...
+  'cc-ssh-key-list.github.empty': `There are no SSH keys available for import from your Github account.`,
+  'cc-ssh-key-list.github.import': `Import`,
+  'cc-ssh-key-list.github.info': `These are the SSH keys from your GitHub account. You can import them to associate them to your Clever Cloud account.`,
+  'cc-ssh-key-list.github.title': `GitHub keys`,
+  'cc-ssh-key-list.github.unlinked': () => sanitize`There is no GitHub account linked to your Clever Cloud account. Check your <a href="./information">profile</a> to activate the linkage.`,
+  'cc-ssh-key-list.personal.delete': `Delete`,
+  'cc-ssh-key-list.personal.empty': `There are no SSH keys associated with your account.`,
+  'cc-ssh-key-list.personal.info': () => sanitize`<p>These are the SSH keys associated with your account.</p><p>If you want to check whether a key is already associated or not, you can list the fingerprints of your local keys with the following command:</p><code>ssh-add -l -E sha256</code>`,
+  'cc-ssh-key-list.personal.title': `Your keys`,
+  'cc-ssh-key-list.success.add': ({ name }) => {
```
pdesoyres-cc, line 591: Could we have that (and the next 2 ones) on one line?

src/translations/translations.fr.js
```diff
@@ -579,6 +579,39 @@ export const translations = {
...
+  'cc-ssh-key-list.github.empty': `Il n'y a aucune clé SSH disponible à l'import depuis votre compte GitHub.`,
+  'cc-ssh-key-list.github.import': `Importer`,
+  'cc-ssh-key-list.github.info': `Voici les clés provenant de votre compte GitHub. Vous pouvez les importer pour les associer à votre compte Clever Cloud.`,
+  'cc-ssh-key-list.github.title': `Clés GitHub`,
+  'cc-ssh-key-list.github.unlinked': () => sanitize`Il n'y a pas de compte GitHub lié à votre compte Clever Cloud. Rendez-vous dans votre <a href="./information">profil</a> pour activer le lien.`,
+  'cc-ssh-key-list.personal.delete': `Supprimer`,
+  'cc-ssh-key-list.personal.empty': `Il n'y a aucune clé SSH associée à votre compte.`,
+  'cc-ssh-key-list.personal.info': () => sanitize`<p>Voici la liste des clés SSH associées à votre compte.</p><p>Si vous souhaitez vérifier qu'une clé est déjà associée ou non, vous pouvez lister les empreintes de vos clés locales avec la commande suivante :</p><code>ssh-add -l -E sha256</code>`,
+  'cc-ssh-key-list.personal.title': `Vos clés`,
+  'cc-ssh-key-list.success.add': ({ name }) => {
```
pdesoyres-cc, line 604: Could we have that on one line?

src/components/cc-ssh-key-list/cc-ssh-key-list.smart.js
```diff
@@ -0,0 +1,145 @@
...
+    onEvent('cc-ssh-key-list:create', ({ name, publicKey }) => {
+      updateComponent('createSshKeyForm', (createSshKeyForm) => {
+        createSshKeyForm.state = 'creating';
+      });
+
+      addKey({ apiConfig, key: { name, key: publicKey } })
+        .then(() => {
+          notifySuccess(component, i18n('cc-ssh-key-list.success.add', { name }));
+          updateComponent('createSshKeyForm', CcSshKeyList.CREATE_FORM_INIT_STATE);
+          refreshList();
```
pdesoyres-cc, line 56: There must be a reason that leads us to refresh the whole list instead of adding the key directly to the model. Can you add a comment to explain why?

pdesoyres-cc (changes requested):
This code is so easy to read. GG @roberttran-cc :clap:

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -0,0 +1,438 @@
...
+export class CcSshKeyList extends LitElement {
+
+  static get properties () {
+    return {
+      createSshKeyForm: { type: Object, attribute: 'create-ssh-key-form' },
+      keyData: { type: Object, attribute: 'key-data' },
+    };
+  }
+
+  static get CREATE_FORM_INIT_STATE () {
```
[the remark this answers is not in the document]
hsablonniere: Yeah, I forgot to tell you @pdesoyres-cc, that's what we decided on the orga members too.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
hsablonniere: We were also a bit "is it really OK?" but I also don't like when we modify the reference of a function param. We can find something better together ;-)

src/components/cc-ssh-key-list/cc-ssh-key-list.types.d.ts
hsablonniere: Yeah, we need to decided together between empty and required for required text fields.

src/components/cc-ssh-key-list/cc-ssh-key-list.smart.js
hsablonniere: Yes, there's a good reason :wink: :+1: for a comment

hsablonniere (changes requested):

florian-sanders-cc (changes requested):
:wave: This component is so great !!
I compared to what we currently have and really, it's painful to go back to the old one :joy:.
Gg :clap: 

The comments I left about CSS are suggestions.
I may have missed usecases for properties but I tried to check what seemed to be unnecessary but it's always tricky with CSS as you may easily miss the usecase (responsive, long content, browser specific issue...).
Feel free to really ignore any of these :wink:.

A few comments in general:

* looks like we allow blank spaces to be submitted in the input fields, we should probably trim (I believe someone already pointed this out during our peer coding session so sorry for the redundancy)
* there are some spacing issues with paragraphs inside the block-section infos:
image
image
image
We could cancel the paragraph margins, cancel only the first paragraph margins, place the github key info inside a `<p>`, it all depends on the spacing you want :+1:

src/components/cc-ssh-key-list/cc-ssh-key-list.types.d.ts
roberttran-cc: I went for `required` everywhere until we make a decision between `required` and `empty`!

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -0,0 +1,438 @@
...
+          flex-direction: column;
+          gap: 1em;
+        }
+        
+        .create-form__public-key {
+          --cc-input-font-family: var(--cc-ff-monospace);
+        }
+
+        .create-form__footer {
+          align-items: baseline;
```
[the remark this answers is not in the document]
roberttran-cc: Definitively a dead rule! Used when an error message was displayed in a `<cc-error>`.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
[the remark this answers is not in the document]
roberttran-cc: FYI, this syntax add an entry in the Storybook "Docs" tab, in the properties/attributes/events tables:
image

roberttran-cc:
Thanks again for all for your top tier feedback! :heart: 
You can check related updates in the separated commit here: caf196e397780daaf7b00e5228da8ece3712f965.

A note on the `new LostFocusController(...)` syntax that triggers an `eslint` error: it seems legit as the triggered rule (`no-new`) is part of the `standard` eslint config [we use](https://github.com/CleverCloud/clever-components/blob/master/.eslintrc.cjs#L3) (config details [here](https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json#L139)).

src/translations/translations.fr.js
```diff
@@ -592,6 +592,33 @@ export const translations = {
...
+  'cc-ssh-key-list.error.required.name': `Saisissez un nom pour votre clé SSH`,
+  'cc-ssh-key-list.error.required.public-key': `Saisissez la valeur de votre clé publique`,
+  'cc-ssh-key-list.github.empty': `Il n'y a aucune clé SSH disponible à l'import depuis votre compte GitHub.`,
+  'cc-ssh-key-list.github.import': `Importer`,
+  'cc-ssh-key-list.github.info': () => sanitize`<p>Voici les clés provenant de votre compte GitHub. Vous pouvez les importer pour les associer à votre compte Clever Cloud.</p>`,
+  'cc-ssh-key-list.github.title': `Clés GitHub`,
+  'cc-ssh-key-list.github.unlinked': () => sanitize`Il n'y a pas de compte GitHub lié à votre compte Clever Cloud. Vous pouvez lier vos comptes depuis votre <a href="./information">profil</a>`,
+  'cc-ssh-key-list.personal.delete': `Supprimer`,
+  'cc-ssh-key-list.personal.empty': `Il n'y a aucune clé SSH associée à votre compte.`,
+  'cc-ssh-key-list.personal.info': () => sanitize`<p>Voici la liste des clés SSH associées à votre compte.</p><p>Si vous souhaitez vérifier qu'une clé est déjà associée, vous pouvez lister les empreintes de vos clés locales avec la commande suivante :</p><code>ssh-add -l -E sha256</code>`,
```
pdesoyres-cc, line 615: You could also add a `&nbsp;` here too before the colon.

pdesoyres-cc (changes requested):
Great Job @roberttran-cc

roberttran-cc:
Another round of feedback (including some from our support team)! :tada: :taco: 

For those who want to check the last updates, go for [this commit](https://github.com/CleverCloud/clever-components/pull/475/commits/094fc4c8ba478e51c0ab218a56948a817a808193).\
For the others, start with the first feedback [commit](https://github.com/CleverCloud/clever-components/commit/caf196e397780daaf7b00e5228da8ece3712f965).\
Otherwise, checking out the whole branch is still the way to go.

In this round, the main thing to review is: when a new key is submitted, the component checks if the public key is possibly a private key by searching the ` private ` string (with trailing spaces). There is a related story you can check in the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--error-with-when-public-key-is-private).

Note: the component also benefits from #585 after rebasing on master.

src/translations/translations.fr.js
```diff
@@ -603,6 +603,7 @@ export const translations = {
   'cc-ssh-key-list.error.delete': ({ name }) => `Une erreur est survenue pendant la suppression de votre clé personnelle "${name}".`,
   'cc-ssh-key-list.error.import': ({ name }) => `Une erreur est survenue pendant l'import de votre clé personnelle "${name}".`,
   'cc-ssh-key-list.error.loading': `Une erreur est survenue pendant le chargement de vos clés.`,
+  'cc-ssh-key-list.error.private-key': () => sanitize`Format incorrect&nbsp;: avez-vous saisi votre clé privée au lieu de votre clé publique&nbsp;?`,
```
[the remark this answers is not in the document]
roberttran-cc: Actually, it's totally the place and the moment to display this information... but it kinda show a lot of text in the UI. :see_no_evil: 

I think I will leave it like that for now and wait for users feedback :innocent: : if they complain, we can be confident that we should add some more info whereas if we add the info now and we have no feedback, we'll never know if it is useful or not (I don't know if that makes sense :sweat_smile:).

src/translations/translations.fr.js
```diff
@@ -592,6 +592,34 @@ export const translations = {
   //#region cc-select
   'cc-select.required': `obligatoire`,
   //#endregion
+  //#region cc-ssh-key-list
+  'cc-ssh-key-list.add.btn': `Ajouter la clé`,
+  'cc-ssh-key-list.add.info': () => sanitize`<p>Vous devez associer une clé SSH à votre compte si vous désirez déployer via Git. Utilisez ce formulaire à cet effet.</p><p>Vous pouvez créer une clé SSH avec la commande suivante&nbsp;:</p><code>ssh-keygen -t ed25519 -C "my-email@example.com"</code><p>La clé publique générée est sauvegardée dans le fichier "*.pub".</p>`,
+  'cc-ssh-key-list.add.name': `Nom`,
+  'cc-ssh-key-list.add.public-key': `Clé publique`,
+  'cc-ssh-key-list.add.title': `Ajouter une nouvelle clé`,
+  'cc-ssh-key-list.doc.info': () => sanitize`Pour plus d'aide, vous pouvez consulter notre <a href="https://www.clever-cloud.com/doc/admin-console/ssh-keys/">documentation</a> (en anglais).`,
```
florian-sanders-cc, line 601: very small details:
It would be better if the "en Anglais" mention was part of the link text. This way, screen reader users don't have to check the context of the link.
Anglais should start with a cap because it's a noun here and not an adjective ("les utilisateurs anglais" vs "en Anglais, on dit ça").

florian-sanders-cc (changes requested):
A very tiny little detail but apart from this, all good (very good even) for me :sunglasses:

florian-sanders-cc (approved):
LGTM ! Gg :clap: :tada: :sunglasses:

pdesoyres-cc (approved):
Well done :clap:  LGTM :tada:

src/components/cc-ssh-key-list/cc-ssh-key-list.smart.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+async function addKey ({ apiConfig, key }) {
+  const name = encodeURIComponent(key.name);
+  const publicKey = key.key;
+  return addSshKey({ key: name }, JSON.stringify(publicKey))
+    .then(sendToApi({ apiConfig }));
+}
+
+const importKey = addKey;
```
hsablonniere, line 141: A bit unusual but why not ;-)

hsablonniere:
@roberttran-cc LGTM, impressive work. This is really the solid foundation we want all the new components to follow :clap: :heart_eyes:

florian-sanders-cc (approved):
Gg :clap: :sunglasses:

PR #479, opened 2022-06-02
cc-button: update waiting loader animation in circle state

### LINKED ISSUE #449: cc-button: improve loader style when it contains only an image
When `cc-button` contains only an image, the loader style is not optimal.

image

The loader could circle around the border instead.

roberttran-cc:
Here are some design proposal for the `cc-button` waiting mode in circle state :
- [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button-circle-loader/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-button--circle-waiting)
- fixes #449 

I tried to present multiple combinations in the story, varying on:
- mode (simple, primary, success, etc.)
- icon size
- ring radius
- opacity (global, icon only, none)
- icon transparent or not

Note: feel free to zoom with the related action on the top toolbar.

Just vote for your overall favorite option! (and do not look at the code yet :smile:)

### DIFF
--- a/src/atoms/cc-button.js
+++ b/src/atoms/cc-button.js
@@ -52,7 +52,7 @@ export class CcButton extends LitElement {
       primary: { type: Boolean },
       skeleton: { type: Boolean },
       success: { type: Boolean },
-      waiting: { type: Boolean },
+      waiting: { type: Boolean, reflect: true },
       warning: { type: Boolean },
       _cancelMode: { type: Boolean, attribute: false },
     };
@@ -210,8 +210,13 @@ export class CcButton extends LitElement {
         ${delay != null ? html`
           <progress class="delay ${classMap({ active: this._cancelMode })}" style="--delay: ${delay}s"></progress>
         ` : ''}
-        ${waiting ? html`
+        ${waiting && !modes.circle ? html`
           <progress class="waiting"></progress>
+        ` : ''}
+        ${waiting && modes.circle ? html`
+          <svg class="circle-loader" viewBox="25 25 50 50" stroke-width="4" aria-hidden="true">
+            <circle fill="none" cx="50" cy="50" r="15" />
+          </svg>
         ` : ''}
       </button>
     `;
@@ -460,6 +465,54 @@ export class CcButton extends LitElement {
           margin-left: calc(50% - calc(var(--width) / 2));
           position: absolute;
           width: var(--width);
+        }
+
+        /* circle waiting mode - keyframes */
+        @keyframes rotate {
+          100% {
+            transform: rotate(360deg);
+          }
+        }
+
+        @keyframes stretch {
+          0% {
+            stroke-dasharray: 1, 200;
+            stroke-dashoffset: 0;
+          }
+          50% {
+            stroke-dasharray: 90, 200;
+            stroke-dashoffset: -35px;
+          }
+          100% {
+            stroke-dashoffset: -124px;
+          }
+        }
+
+        /* circle waiting mode - partial opacity */
+        :host([waiting]) button.circle {
+          opacity: 1;
+        }
+
+        :host([waiting]) button.circle .text-wrapper img {
+          opacity: 0.25;
+        }
+
+        /* circle waiting mode - core animation */
+        .circle-loader {
+          --bcw-speed: 2s;
+          animation: rotate var(--bcw-speed) linear infinite;
+          inset: 0;
+          position: absolute;
+          transform-origin: center;
+          vertical-align: middle;
+        }
+
+        .circle-loader circle {
+          animation: stretch calc(var(--bcw-speed) * 0.75) ease-in-out infinite;
+          stroke: currentColor;
+          stroke-dasharray: 1, 200;
+          stroke-dashoffset: 0;
+          stroke-linecap: round;
         }
 
         /* We can do this because we set a visible focus state */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@ title: Changelog
 * `<cc-doc-card>`: rename title property to heading to fix a conflict.
 * `<cc-doc-list>`: use heading when initializing the cards instead of title.
 
-### Fixes 
+### Components 
 * `<cc-article-list>`: fix error mode not triggering on XML parsing failure (smart). 
 * `parseRssFeed()`: trim XML string before parse to avoid whitespaces error.
+* `<cc-button>`: update waiting loader animation in circle state.
 
 ## 7.12.0 (2022-05-20)
 
--- a/stories/atoms/cc-button.stories.js
+++ b/stories/atoms/cc-button.stories.js
@@ -158,6 +158,13 @@ As you can see here, \`circle\` can only be used if there is an \`image\` and in
   ],
 });
 
+export const waitingAndCircle = makeStory(conf, {
+  items: [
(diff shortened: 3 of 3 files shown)


florian-sanders-cc:
Just dropping this here even though I haven't made my mind up yet but wow this preview looks really nice.
Thanks for making it so easy to compare :+1: 

I don't know which is my favorite yet but I really like most of them :sunglasses:

pdesoyres-cc:
Thank you @roberttran-cc for this preview with so many alternatives. 

* First choice: Inner loader / Icon only opacity
* Second choice: Outer loader / Icon only opacity

florian-sanders-cc:
I finally took time to decide and my favorite is:
Inner loader / icon only opacity. (this may not match exactly what we re doing with text buttons since opacity is applied globally but I don't think its an issue).

Others also seem really good to me except Inner loader / global Opacity as you can imagine since the loader is difficult to see.

pdesoyres-cc:
I keep my first vote: icon only opacity

florian-sanders-cc:
:wave: same, I still prefer the first option (icon only opacity). I think the "loader" is more visible with this option but the disabled state is less obvious.
I think it's better to have a strong visible loader and a very discrete disabled state in this case.

Still, I think both options are valid choices, it's just a small preference that does not really make any difference for users I think :+1:

hsablonniere:
I'm OK with both I guess :smile: 

I think I didn't like the first one because the loader was stacked on top of the icon but I like the fact that the loader is full white compared to the second version.

Anyway, let's wait for @Galimede's feedback but first option (icon only opacity) is OK with me :+1:

Galimede:
While having some hesitation I do think I prefer the first one too 😛

roberttran-cc:
The people have spoken: "icon only opacity" shall it be!

[Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button-circle-loader/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-button--circle) has been updated, this PR is ready for code review! :eyes:

src/atoms/cc-button.js
```diff
@@ -462,6 +468,52 @@ export class CcButton extends LitElement {
...
+          }
+          100% {
+            stroke-dashoffset: -124px;
+          }
+        }
+
+        /* circle waiting mode - partial opacity */
+        button.circle[waiting] {
+          opacity: 1;
+        }
```
florian-sanders-cc, line 495: missing a line break between your selector blocks

src/atoms/cc-button.js
```diff
@@ -210,9 +211,14 @@ export class CcButton extends LitElement {
         ${delay != null ? html`
           <progress class="delay ${classMap({ active: this._cancelMode })}" style="--delay: ${delay}s"></progress>
         ` : ''}
-        ${waiting ? html`
+        ${waiting && !modes.circle ? html`
           <progress class="waiting"></progress>
         ` : ''}
+        ${waiting && modes.circle ? html`
+          <svg class="circle-loader" viewBox="25 25 50 50" stroke-width="4">
```
florian-sanders-cc, line 218: Let's consider this `<svg>` is decorative because we will rely on something else (dunno what yet, see #486) to convey the "loading" status for screen reader users. 

This means that you should add an `aria-hidden="true"` attribute to the `<svg>` element:
`<svg ... aria-hidden="true">`

(I'll create a new "rule" in our a11y document to sum up how we handle `<svg>` elements, sorry for not doing it before :wink:)

pdesoyres-cc:
## UI

* I really like this mode with icon opacity.
* It works when zooming
* It works when changing the default font-size
* :thinking: In the story, should we show the icon-only & waiting with all the modes (as for `Waiting`, `Waiting and outlined`)?

## Code

* I find it a bit strange to use a `waiting` attribute on the `<button>` tag. I don't think this follows the HTML standards. Maybe we could use a class instead?
* I like the SVG and animation trick!

hsablonniere:
About @pdesoyres-cc feedbacks:

> thinking In the story, should we show the icon-only & waiting with all the modes (as for Waiting, Waiting and outlined)?

I was also wondering how we would maybe reorganize the stories but I don't any interesting suggestion.

> I find it a bit strange to use a waiting attribute on the <button> tag. I don't think this follows the HTML standards. Maybe we could use a class instead?

What do you mean? We're already using this attribute here and there without problems.
Right now we never really used CSS classes as component inputs.

pdesoyres-cc:
> I find it a bit strange to use a waiting attribute on the `<button>` tag. I don't think this follows the HTML standards. Maybe we could use a class instead

I mean that we are adding a non-standard attribute (`waiting`) to a standard HTML tag (`<button>`). Yes, it works, but I am wondering if we should do that. Maybe some HTML analyzer will fail when encountering that. So I was suggesting using a class instead of a non-standard attribute.

hsablonniere:
> I mean that we are adding a non-standard attribute (waiting) to a standard HTML tag (<button>). Yes, it works, but I am wondering if we should do that. Maybe some HTML analyzer will fail when encountering that. So I was suggesting using a class instead of a non-standard attribute.

Ahhhhhhh. I 200% misunderstood your comment, sorry. I agree with you :+1: 

@roberttran-cc can you replace this attribute with a class?

florian-sanders-cc:
Do we need a class at all ?
We could use this kind of selector:
```css
:host([waiting]) button.circle { }
```

src/atoms/cc-button.js
roberttran-cc: Nice catch! Fixed with last update.

src/atoms/cc-button.js
```diff
@@ -462,6 +468,52 @@ export class CcButton extends LitElement {
...
+        button.circle[waiting] {
+          opacity: 1;
+        }
+        button.circle[waiting] .text-wrapper img {
+          opacity: 0.25;
+        }
+
+        /* circle waiting mode - core animation */
+        .circle-loader {
+          --bcw-speed: 2s;
```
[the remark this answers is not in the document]
roberttran-cc: It stands for "Button Circle Waiting"... :see_no_evil: 
I dropped a kinda-random prefix to "couple" the custom property (which should be private) with the component but I'm not really sure if it's worth it since it's not really understandable.

roberttran-cc:
Thanks for your feedbacks! :heart_eyes: 
Quick preview [here](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button-circle-loader/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-button--waiting-and-circle).

What changed:
- `aria-hidden` added ;
- `waiting` implementation updated ;
- story split.

For the `cc-button` stories, they may need a bigger cleanup/refactor.

florian-sanders-cc:
LGTM, well done @roberttran-cc this is really nice :clap:

src/atoms/cc-button.js
```diff
@@ -462,6 +467,53 @@ export class CcButton extends LitElement {
...
+          animation: rotate var(--bcw-speed) linear infinite;
+          inset: 0;
+          position: absolute;
+          transform-origin: center;
+          vertical-align: middle;
+        }
```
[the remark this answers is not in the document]
roberttran-cc: Nice catch once again! Fixed with last update.

hsablonniere:
We just need the CHANGELOG update in the same commit.

pdesoyres-cc (approved):
Good to go!

PR #481, opened 2022-06-15
cc-badge: init component

### LINKED ISSUE #462: cc-badge: init component
## Context

We need a component that emphasizes small chunks of text content within a UI.
This component would rely on modes, just like `cc-button`, with an "info" (primary light bg color / primary border when outlined).
This component may contain an icon.
image

### Icon management

Do we define an icon that matches the mode so that one only has to enable or disable icon mode to toggle the icon visibility ? 

Or do we let people insert the icon they want just like `cc-button`.

I think in most cases, the default icon is what we want but maybe there are cases where it would be useful to override it ?

### Attributes/properties

| `mode`    | `Mode` ('neutral', 'info', 'success', 'danger', 'warning') | `neutral`        | `false`    | Sets the style of the badge (bg color and text color and border color when outlined      |
| `outlined`    | `Boolean` | `false`     | `false`    | Sets a border and removes the plain bg color + the text color      |
| `icon`    | `String` / `null` | `false`     | `false`    | If set, enables icon mode and sets the `src` of the inner native `` element.      |

### Slots

Default slot for text content

## Stories & states

* Modes
* Outlined
* Icon

florian-sanders-cc:
:wave: Hey there ! 

Fixes #462 

The [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-badge--dimmed) has been updated following your feedback (thanks again :heart:)

Padding and font-size have been reduced just a little bit. The size is very close to what what we use in `<cc-orga-member-card>` and `<cc-email-list>` and it matches what was already used in `<cc-header-orga>`.

# To review:

## UI

* [`<cc-badge>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-badge--dimmed),
* [`<cc-header-orga>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-overview-cc-header-orga--default-story) (org status and hotline use `<cc-badge>`),
* [`<cc-tile-instances>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/🛠-overview-cc-tile-instances--default-story)  (bubble uses `<cc-badge>`),
* [`<cc-tcp-redirection-form>` "context admin" story](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-tcp-redirections-cc-tcp-redirection-form--data-loaded-with-context-admin) (notif bubble uses `<cc-badge>`),
* [`<cc-header-app>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-overview-cc-header-app--default-story) (footer bg is neutral now),
* [`<cc-header-addon>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-addon-cc-header-addon--default-story) (footer bg is neutral now),

## Code

Every commit :sweat_smile: 

# TODO

- [x] Lint fix
- [x] Rebase

### DIFF
--- a/src/atoms/cc-badge.js
+++ b/src/atoms/cc-badge.js
@@ -1,0 +1,145 @@
+import { css, html, LitElement } from 'lit-element';
+import { defaultThemeStyles } from '../styles/default-theme.js';
+
+/**
+ * @typedef {import('./types.js').BadgeIntent} BadgeIntent
+ * @typedef {import('./types.js').BadgeWeight} BadgeWeight
+ */
+
+/**
+ * A component to highlight a small chunk of text.
+ *
+ * @cssdisplay inline-flex
+ */
+export class CcBadge extends LitElement {
+  static get properties () {
+    return {
+      circle: { type: Boolean, reflect: true },
+      iconAlt: { type: String, attribute: 'icon-alt' },
+      iconSrc: { type: String, attribute: 'icon-src' },
+      intent: { type: String, reflect: true },
+      weight: { type: String, reflect: true },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {Boolean} Sets the badge to a bubble style. Should only be used to display 1 or 2 digits figures. */
+    this.circle = false;
+
+    /** @type {String|null} Sets the `alt` attribute value on the `<img>` tag. Only use if the image conveys additional info compared to surrounding text. */
+    this.iconAlt = null;
+
+    /** @type {String|null} Sets the icon displayed on the left of the text inside the badge. */
+    this.iconSrc = null;
+
+    /** @type {BadgeIntent} Sets the accent color used for the badge. */
+    this.intent = 'neutral';
+
+    /** @type {BadgeWeight} Sets the style of the badge depending on how much one wants it to stand out. */
+    this.weight = 'dimmed';
+  }
+
+  render () {
+    return html`
+      ${this.iconSrc != null ? html`
+        <img src=${this.iconSrc} alt=${this.iconAlt ?? ''}>
+      ` : ''}
+      <span>
+        <slot></slot>
+      </span>
+    `;
+  }
+
+  static get styles () {
+    return [
+      defaultThemeStyles,
+      // language=CSS
+      css`
+        :host {
+          align-items: center;
+          border-radius: 1em;
+          display: inline-flex;
+          font-size: 0.8em;
+          gap: 0.3em;
+          padding: 0.2em 0.8em;
+        }
+
+        :host([circle]) {
+          border-radius: 50%;
+          font-size: 1em;
+          height: 1.5em;
+          justify-content: center;
+          min-height: unset;
+          padding: 0;
+          width: 1.5em;
+        }
+
+        :host([weight="dimmed"]) {
+          background-color: var(--accent-color, #ccc);
+          color: var(--color-text-default);
+        }
+
+        :host([weight="strong"]) {
+          background-color: var(--accent-color, #777);
+          color: var(--color-text-inverted, #fff);
+        }
+
+        :host([weight="outlined"]) {
+          background-color: var(--color-bg-default, #fff);
+          /* roughly 1px. We want the border to scale with the font size so that outlined
+          badges still stand out as they should when font-size is increased. */
+          box-shadow: inset 0 0 0 0.06em var(--accent-color, #777);
+          color: var(--accent-color, #777);
+        }
+
+        :host([intent="info"]) {
+          --accent-color: var(--color-bg-primary);
+        }
+
+        :host([intent="info"][weight="dimmed"]) {
+          --accent-color: var(--color-bg-primary-light);
+        }
+
+        :host([intent="success"]) {
+          --accent-color: var(--color-bg-success);
+        }
+
+        :host([intent="success"][weight="dimmed"]) {
+          --accent-color: var(--color-bg-success-light);
+        }
+
+        :host([intent="danger"]) {
+          --accent-color: var(--color-bg-danger);
+        }
+
+        :host([intent="danger"][weight="dimmed"]) {
+          --accent-color: var(--color-bg-danger-light);
+        }
+
+        :host([intent="warning"]) {
+          --accent-color: var(--color-bg-warning);
+        }
+
+        :host([intent="warning"][weight="dimmed"]) {
+          --accent-color: var(--color-bg-warning-light);
+        }
+
+        :host([intent="neutral"]) {
+          --accent-color: var(--color-bg-strong);
+        }
+
+        :host([intent="neutral"][weight="dimmed"]) {
+          --accent-color: var(--color-bg-neutral-alt);
+        }
+
+        img {
+          height: 1em;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-badge', CcBadge);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ title: Changelog
 * `<cc-tcp-redirection-form>`: Use `<cc-badge>` to display redirection count.
 * `<cc-header-app>`: Change footer background to neutral.
 * `<cc-header-addon>`: Change footer background to neutral.
+* `<cc-header-orga>`: Use `<cc-badge>` to display org status and hotline number. 
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ title: Changelog
 * Color design tokens: add darker shades for light colors.
 * New component:
   * `<cc-badge>`
+* `<cc-tcp-redirection-form>`: Use `<cc-badge>` to display redirection count.
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ title: Changelog
 * New component:
   * `<cc-badge>`
 * `<cc-tcp-redirection-form>`: Use `<cc-badge>` to display redirection count.
+* `<cc-header-app>`: Change footer background to neutral.
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ title: Changelog
   * `<cc-badge>`
 * `<cc-tcp-redirection-form>`: Use `<cc-badge>` to display redirection count.
 * `<cc-header-app>`: Change footer background to neutral.
+* `<cc-header-addon>`: Change footer background to neutral.
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ title: Changelog
 * `<cc-article-list>`: fix error mode not triggering on XML parsing failure (smart). 
 * `parseRssFeed()`: trim XML string before parse to avoid whitespaces error.
 * `<cc-button>`: update waiting loader animation in circle state.
+* Color design tokens: add darker shades for light colors.
+...
 
 ## 7.12.0 (2022-05-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ title: Changelog
 * `parseRssFeed()`: trim XML string before parse to avoid whitespaces error.
 * `<cc-button>`: update waiting loader animation in circle state.
 * Color design tokens: add darker shades for light colors.
+* New component:
+  * `<cc-badge>`
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/src/addon/cc-header-addon.js
+++ b/src/addon/cc-header-addon.js
@@ -133,6 +133,7 @@ export class CcHeaderAddon extends LitElement {
           border: 1px solid #bcc2d1;
           border-radius: 0.25rem;
           display: block;
+          overflow: hidden;
         }
 
         .main {
@@ -181,7 +182,7 @@ export class CcHeaderAddon extends LitElement {
           --cc-gap: 0.5rem;
           --cc-align-items: center;
           align-items: center;
-          background-color: var(--color-bg-primary-light);
+          background-color: var(--color-bg-neutral);
           box-shadow: inset 0 6px 6px -6px #a4b1c9;
           box-sizing: border-box;
           color: #2e2e2e;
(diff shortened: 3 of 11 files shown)


Galimede:
Hiya 👋,

This looks pretty cool. 👍 😍 
 However, I have some doubts about some colors with the dimmed mode especially the primary and the green one on a white background. 
I have tried to check the preview on different monitors and I had a hard time focusing and reading the text on the primary color and it is the same thing for the green one while I have less trouble reading/focusing on it. 
I then have tried to put a plain black background and they seem to be much more easier to read. While putting the background in black I found that the "this is info" badge and the "this is neutral" badge seems to have the same color ? Are they the same color or it is just a matter of contrast ? 🤔 
This makes me think that we might need strong colors at least for some cases we would need to determine indeed.

> circle, do we want this to be handled by the cc-badge or not ? 

You mean the notification circle in the corner right? If so in my opinion, yes the cc-badge should handle the "notification" circle. At least it seems to be the case on the design systems/component libraries that I have checked. 
Of what I have seen they also seem to provide these features to their badge:
* Have a default "notification" circle when there is no number in it
* Some have a square corner with the possibility to make the border of the badge more rounded (circled)
* Some even provide a badge with just an icon

I don't know if we want to make these features available but that might be useful in the future?


> When we rework the tile components, we could use the cc-badge for these ?

If we handle the "notification"/circle in the cc-badge do think we could indeed 😃

florian-sanders-cc:
Thanks @Galimede !!

> I have some doubts about some colors with the dimmed mode 

I definitely agree with you. I went with the primary light that we use in some other components (header-addon in overviews, the bottom part I think). It is way too light to be used over a white background I think ?
What do you think about these ?
image

> You mean the notification circle in the corner right?
Yes exactly, I'll check other design systems. I think all of what you mention can be done with the current example (see custom style / circle story) but it could be improved so it is interesting to see what other design systems do :+1:

Galimede:
> It is way too light to be used over a white background I think ?

Yeah I do think so too

> What do you think about these ?

Seems to be much better like this :+1:

florian-sanders-cc:
:wave: @pdesoyres-cc thanks for you feedback ! 

I'll update the component based on your review ;)

> :point_up_2: interactive: true/false (false by default). When interactive, a :hover style is applied and the mouse cursor switches to pointer when hovered.

Well, I don't know ? :sweat_smile: 

I thought it through and I am not sure this should be handled by the `cc-badge` component.

In case of an interactive badge (which could be a button, a checkbox, a radio input or a link depending on the context), one would have to determine a hover/focus style, an active/selected style (checkbox / radio input), a disabled style (if relevant).

It feels weird to provide the hover style for an element that won't be interactive by itself.

You would have to use the `cc-badge` inside a link (or `cc-link` if we create one) and depending on the context, build a style that works.
These styles could be handled using the CSS custom props. 

If we need an interactive badge, we may need to build another component (that is, if we want to reuse it in different parts of the our console) that would wrap the `cc-badge` with the right semantics for our use case (but I don't think we have any use case at the moment).

> :heavy_multiplication_x: removable: true/false (false by default). When removable, a cross icon-only button is displayed on the right and the click dispatches a cc-badge:remove event.

I think we can do this yes :+1:.

roberttran-cc:
Mostly agree with everything that have been said! :smile: 

- Strong color version is a must-have in my opinion.
- Last version of the dimmed colors is much better.
- Was not bothered with the icon size.
- Thumbs up for the badge and the closable props (not really with the interactive one, agree the wrapper approach though) but...
- ...Maybe with should ship the minimal working badge and see how it integrates with our products, giving us more insights for future component upgrades.

hsablonniere:
> There is a small alignment issue on FF for linux Users due to the default font. This is also the case for other components.

Yes, this will be solved by the default theme I guess.

* Most of the time we name the first story default but maybe it's an exception.
* Is there a default weight?
  * If not I'd suggest `dimmed` and we need to explicit it in the docs.
* I prefer the colors you proposed in your reply to @Galimede 
  * I had a hard time differentiating success and warning in the first proposition
* Can we order intents like this in the stories: neutral, info, success, warning, danger
* Why do we have a custom styles story?
  * Do we recommend to use this? When?
* Do we really need these CSS custom properties?
  * I looked at the code quickly and we may be able to remove the span and do everything on the `:host` which would be let users override stuffs directly without custom props.
  * We'll discuss this in a sync review
* Why is there a innerHTML prop in the docs?
* About the strong weight
  * I would say let's do it and use it directly in the tile components
  * The problem is, the tile components don't use the real theme colors
    * Solution A: these tile components will be reworked so don't need weight:strong right now
    * Solution B: we do the weight:strong, we use the cc-badge in the tile components with weight:strong and custom colors
    * I prefer solution B but I wanted to emphasize Solution A as I often prefer the YAGNI principle
* I agree with Pierre, icons are very small
  * We'll discuss this in a sync review
* @pdesoyres-cc I don't understand the motivation for interactive and removable features
  * My first reaction is: do we need this in any of our components? (YAGNI)
  * If we don't, let's postpone this to a later time

This last point tend to be confirmed this @roberttran-cc :arrow_down: 

> ...Maybe with should ship the minimal working badge and see how it integrates with our products, giving us more insights for future component upgrades.

florian-sanders-cc:
:wave: Thanks for your replies !

@hsablonniere some quick answers before updating the code:

- CSS Custom props, I figured why not, because it did not cost any extra dev but I'll remove them, especially since we are going to remove the span and class system ;)
- Strong weight and instant tiles: I was thinking solution A for tile instances but I thought it could still be used here (tcp redir, admin): 
image
Let's go for solution B :+1: 

Code things: 
- text is inserted through a slot (hence the innerHTML inside stories), which is not necessary since its just plain text.
- I'll remove the span :+1: 
- Cleaning everything this morning, hopefully it will be better :+1:

pdesoyres-cc:
About interactive mode, I agree with you @florian-sanders-cc, it should be doable from outside the component itself.

About removable mode, I agree with you @hsablonniere: let's follow the YAGNI principle.

Sorry for disturbing the review with that.

roberttran-cc:
@pdesoyres-cc you mispelled "disturbing the review" with "giving a broader scope of the component with some potential evolutions while giving insights on unforeseen use cases". ;)

Please keep on doing it, this is invaluable feedback!

hsablonniere:
> Sorry for disturbing the review with that.

@pdesoyres-cc Don't be sorry and please do continue to propose ideas :pray: :+1:

florian-sanders-cc:
> Sorry for disturbing the review with that.

@pdesoyres-cc I don't know if it's my comments / answers that made you feel like it was disturbing the review, if so, very sorry for that :hugs: 
On the contrary, :+1: to what @roberttran-cc said, nothing more to add, these were great suggestions !

florian-sanders-cc:
:wave: Thanks for all your reviews ! ;) 

The reason I went with very light colors at the beginning is that some of these already were being used in some of our components:

- primary-light is used by `cc-header-app` / `cc-header-addon`,
- warning-light is used by `cc-warning-payment`.

I wanted to see it we could work with what existed before changing these but I think we all agree it does not work here.

We have 2 options:

- change primary-light and warning light colors with the ones we all agreed on in this review (making them a little bit darker),
- introduce new shades "dimmed", in the design system so we can use these inside `¢c-badge` and avoid impacting `cc-payment-warning`, `cc-header-addon` and `cc-header-app`.

To see how it would impact these components if we change light colors, you can compare with links below:

- [Payment warning prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-notices-cc-warning-payment--default-story) vs [Payment warning darker](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/light-colors-update/index.html?path=/story/%F0%9F%9B%A0-notices-cc-warning-payment--default-story)
- [Header app/addon prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-overview-cc-header-app--default-story) vs [Header app/addon darker](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/light-colors-update/index.html?path=/story/%F0%9F%9B%A0-overview-cc-header-app--default-story)

We could also try to swap the header app/addon bg for a neutral:
image

pdesoyres-cc:
I would keep the very light color we have now and add a darker version. We could adopt a naming like `primary-light` vs `primary-lighter`.

hsablonniere:
* Looking at your previews, I think we need both. The "darker" colors used for badges don't really work for such big areas like the notice.
* I like the proposition of a neutral background for the 2 header components.

src/atoms/cc-badge.js
```diff
@@ -0,0 +1,150 @@
...
+          align-items: center;
+          /* only used as fallback in case theme is not set */
+          background-color: #777;
+          border-radius: 1em;
+          box-sizing: border-box;
+          color: var(--color-text-default);
+          display: inline-flex;
+          font-size: 0.8em;
+          gap: 0.3em;
+          min-height: 1.7em;
```
florian-sanders-cc, line 72: This is something I really don't like to do but I wanted to make sure badges have the same size when outlined (border) and not outlined (no border).

I am open to suggestions if you find a better solution.

florian-sanders-cc:
:wave: Hey there ! 

The [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-badge--dimmed) has been updated following your feedback (thanks again :heart:)

Padding and font-size have been reduced just a little bit. The size is very close to what what we use in `<cc-orga-member-card>` and `<cc-email-list>` and it matches what was already used in `<cc-header-orga>`.

# To review:

## UI

* [`<cc-badge>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-badge--dimmed),
* [`<cc-header-orga>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-overview-cc-header-orga--default-story) (org status and hotline use `<cc-badge>`),
* [`<cc-tile-instances>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/🛠-overview-cc-tile-instances--default-story)  (bubble uses `<cc-badge>`),
* [`<cc-tcp-redirection-form>` "context admin" story](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-tcp-redirections-cc-tcp-redirection-form--data-loaded-with-context-admin) (notif bubble uses `<cc-badge>`),
* [`<cc-header-app>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-overview-cc-header-app--default-story) (footer bg is neutral now),
* [`<cc-header-addon>` stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge/index.html?path=/story/%F0%9F%9B%A0-addon-cc-header-addon--default-story) (footer bg is neutral now),

## Code

Every commit :sweat_smile:

pdesoyres-cc:
:+1: GG

## UI

* `cc-badge`: I still find the icons a bit small. But maybe it depends on the icon used as I find them ok on the `cc-header-orga`.
* `cc-tcp-redirection-form`: I find it a bit "too much in your face". Maybe we could use the dimmed weight instead? Let's see what the others think...

## Code

* `cc-badge`: very nice CSS-based implementation.
* `cc-badge`: wrong style order in `border: solid 0.1em var(--cc-badge-accent-color);`. border-size should be before border-style.
* `cc-tile-instances`: You're lucky that `--bubble-d` is the same as the default `cc-badge`'s height in circle mode (1.5em). Even if it is the same value today, I think that we should force the `cc-badge` in `cc-tile-instances.js` to have the `--bubble-d` size (width and height).

src/atoms/cc-badge.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        :host([weight="strong"]) {
+          background-color: var(--cc-badge-accent-color);
+          color: var(--color-text-inverted);
+        }
+
+        :host([weight="outlined"]) {
+          background-color: var(--color-bg-default);
+          border: solid 0.1em var(--cc-badge-accent-color);
```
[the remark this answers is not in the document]
florian-sanders-cc: To be honest it depends, in most cases I don't want the border to scale with the font size but in this case, the border is what makes the "outlined" badges stand out so I thought it was a good idea to make them scale so that if you increase font size, they would still stand out from the rest of the content.

You can compare below, with font size increase a lot (@400% on FF):
- 1px: image
- 0.1em: 
image

src/atoms/cc-badge.js
```diff
@@ -0,0 +1,145 @@
...
+    /** @type {BadgeIntent} Sets the accent color used for the badge. */
+    this.intent = 'neutral';
+
+    /** @type {BadgeWeight} Sets the style of the badge depending on how much one wants it to stand out. */
+    this.weight = 'dimmed';
+  }
+
+  render () {
+    return html`
+      ${this.iconSrc != null ? html`
```
florian-sanders-cc, line 46: considering we use a slot, should we also slot the image ? :roll_eyes: 

I guess right now it matches the `cc-button` behavior (slot for text, prop for image) so I'm okay with it :+1:

src/atoms/cc-badge.js
```diff
@@ -0,0 +1,145 @@
...
+        <img src=${this.iconSrc} alt=${this.iconAlt ?? ''}>
+      ` : ''}
+      <span>
+        <slot></slot>
+      </span>
```
florian-sanders-cc, line 51: using slot instead of prop opens up for more HTML errors. Only text should be inserted or inline elements (span, etc.) should ever be inserted here.

I think this is okay and we can mitigate this with an HTML validation tool anyway :+1:

hsablonniere:
@florian-sanders-cc I'll let you get confirmations from others about TCP redirs etc but IMO it's ready to be merged (after @Galimede changes and after @roberttran-cc PR on cc-button waiting).

roberttran-cc:
GG on the overall work: this component is really polished and clean, I like it very much! :+1: 
And the usages in other components look really good (and TFW deleting custom style :sunglasses:).

As for the `<tcp-redirection-form>`, I'm ok with the current badge!

For me, it's good to go!

PR #482, opened 2022-06-15
cc-input-, cc-select, cc-toggle: support inline layout (label next to form field)

### LINKED ISSUE #451: cc-input-*, cc-select, cc-toggle: add support for inline layout (label and form control on the same line)
We need to add a prop to make it possible to display our form controls and labels on the same line.

This would rely on setting `inline-flex` display on all these components with `flex-direction: column;` by default and switching to `flex-direction: row;` when the host has the inline attribute.

florian-sanders-cc:
Fixes #451.

:wave: Hey there !

I updated the [Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/form-fields/inline-label/index.html).

Most of the changes are related to `<cc-toggle>` and `<cc-invoice-list>` :+1: 

# Things to review:

## UI 

- "inline" stories (+ inline required and inline with error / help messages) for these components:
	- `<cc-input-number>`,
	- `<cc-input-text>`,
	- `<cc-select>`,
	- `<cc-toggle>` (only one inline story since `cc-toggle` does not support required and help messages (and I'm not sure it should but it's another subject).
- `<cc-invoice-list>` stories (both desktop and mobile), the "year" filter is now "inlined".

## Code

Every commit.

# TODO:

- [x] Lint fix
- [x] Rebase after #462 merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ title: Changelog
 * `<cc-header-orga>`: Use `<cc-badge>` to display org status and hotline number. 
 * `<cc-input-text>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-input-number>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
+* `<cc-select>`: add `inline` prop to place the label on the left of the `<select>` element. Add new `inline` story to show this behavior.
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ title: Changelog
 * `<cc-input-number>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-select>`: add `inline` prop to place the label on the left of the `<select>` element. Add new `inline` story to show this behavior.
 * `<cc-toggle>`: add `inline` prop to place the label on the left of the group of radio input elements. Add new `inline` story to show this behavior.
+* `<cc-invoice-list>`: inline year filters (`<cc-toggle>` for desktop and `<cc-select>` for mobile).
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ title: Changelog
 * `<cc-input-text>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-input-number>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-select>`: add `inline` prop to place the label on the left of the `<select>` element. Add new `inline` story to show this behavior.
+* `<cc-toggle>`: add `inline` prop to place the label on the left of the group of radio input elements. Add new `inline` story to show this behavior.
 ...
 
 ## 7.12.0 (2022-05-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ title: Changelog
 * `<cc-header-addon>`: Change footer background to neutral.
 * `<cc-header-orga>`: Use `<cc-badge>` to display org status and hotline number. 
 * `<cc-input-text>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
+* `<cc-input-number>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 ...
 
 ## 7.12.0 (2022-05-20)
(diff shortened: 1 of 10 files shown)


Galimede:
Hiya :wave:,

here's my review

## UI 

That's great gj :heart_eyes:!

> Not very responsive friendly if you set a "long" label. I included a warning about this in the inline prop documentation text.

That's what I was wondering. Maybe we could check our forms and find the average length of our labels and see what it would look like for our use cases to see what we are up to before taking a decision about further tweaks to the responsive, WDYT?

I was also wondering if we should put a simple input that is not inlined it the inline story to see the difference between the two?

Also, is it just me or am I the only one who find that the label with the form control seems to be a little bit offset? I have check on the devtools and it is OK but I don't know I have the impression that the label is slightly above the input. I might have this impression because of the "taller" letters (h, l ,b..) though.


## Code 

LGTM on my side. Maybe we could just add a line break between the two sentences on the type description before the "Only use this..." as the line is very long.
Maybe like this: 
```
/** @type {boolean} Sets the `<label>` on the left of the `<input>` element.
  * Only use this if your form contains 1 or 2 fields and your labels are short.
  */
  this.inline = false;
```
WDYT?

pdesoyres-cc:
Hey, this is a pretty good job!

## UI

* I find that the placement of the `REQUIRED` makes the whole thing a bit too tight. Could we try to inline it too?
  
inline-required

## Code

This pure CSS implementation is very elegant. I have nothing to say.

florian-sanders-cc:
I am pausing the reviews for the moment and resetting the PR back to draft because I think Pierre and Mathieu's feedback are enough to iterate a better version for the next review phase.

florian-sanders-cc:
Following latest feedback, I updated the preview and code. (@Galimede I followed your suggestions and tried to make the label text inlining less weird...).

The main issue is for "required" fields. I did not want to inline the "required" text (@pdesoyres-cc example) because it makes it even less responsive friendly but in the end it might be the best solution.

My last try at not inlining the required text meant reducing its font size (still above the minimum of 12px but very close).

- What do you think about the preview ?
- Is the "required" text too small ?
- Do you prefer the full inline version (see @pdesoyres-cc suggestion below) ?

hsablonniere:
## Code review

Nothing to say. The CSS implementation is really nice. :+1

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ title: Changelog
 * `<cc-input-text>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-input-number>`: add `inline` prop to place the label on the left of the `<input>` element. Add new `inline` story to show this behavior.
 * `<cc-select>`: add `inline` prop to place the label on the left of the `<select>` element. Add new `inline` story to show this behavior.
+* `<cc-toggle>`: add `inline` prop to place the label on the left of the group of radio input elements. Add new `inline` story to show this behavior.
```
florian-sanders-cc, line 12: Ignore the whole commit right now. I'd rather discuss this in sync as the `<fieldset>` / `<legend>` styling is quite tricky :roll_eyes:

src/atoms/cc-toggle.js
```diff
@@ -47,6 +47,7 @@ export class CcToggle extends LitElement {
       choices: { type: Array },
       disabled: { type: Boolean },
       hideText: { type: Boolean, attribute: 'hide-text' },
+      inline: { type: Boolean, reflect: true },
```
florian-sanders-cc, line 50: Ignore the whole commit right now. I'd rather discuss this in sync as the `<fieldset>` / `<legend>` styling is quite tricky :roll_eyes:

florian-sanders-cc:
:wave: Hey there !

I updated the [Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/form-fields/inline-label/index.html).

# Things to review:

# UI 

- "inline" stories (+ inline required and inline with error / help messages) for these components:
	- `<cc-input-number>`,
	- `<cc-input-text>`,
	- `<cc-select>`,
	- `<cc-toggle>` (only one inline story since `cc-toggle` does not support required and help messages (and I'm not sure it should but it's another subject).
- `<cc-invoice-list>` stories (both desktop and mobile), the "year" filter is now "inlined".

# Code

Every commit except the one related to `<cc-toggle>` (I'd rather discuss this one in sync ;))

src/atoms/cc-input-text.js
```diff
@@ -359,12 +386,30 @@ export class CcInputText extends LitElement {
...
+          gap: 0;
+          grid-area: label;
+          justify-content: center;
+          line-height: normal;
+          padding: 0;
+        }
+        
+        :host([inline][multi]) label {
+          /* Allows the label text to be aligned with the first line of the input. */
+          height: 2em;
```
florian-sanders-cc, line 400: this is not great but it's the only way I found to do this:
image
image

roberttran-cc:
Thanks for the answers to both of my questions!
- totally agree with waiting insights on long labels,
- okay for the font issue.

Otherwise, everything looks good to me!

hsablonniere:
@florian-sanders-cc I'll do a final review tomorrow but I guess it's ready to be rebased and merge after the cc-badge PR.

pdesoyres-cc (approved):
Good to go!

roberttran-cc:
And everything looks good to me as well! :muscle:

PR #484, opened 2022-06-21
Integrate cc-toast into smart components

### LINKED ISSUE #478: Use cc-toast in smart components
Since we've got a `cc-toast` component (see #374), smart components can now display toasts when there are errors coming from the server. As a consequence, we can use the web component `error` property just for loading errors.

We can also use `cc-toast` to display a success message when an operation has succeeded.

Below is the list of involved components:

- [x] cc-grafana-info
- [x] cc-tcp-redirection-form
- [x] cc-env-var-form

Some components don't have their underlying smart component. The move to `cc-toast` usage is to be made in the console code directly :

* cc-addon-admin
* cc-env-var (sometimes in console, the component is used directly instead of using the smart one)
* 

Some components are used in the website. The move must be made on the website directly:

* `cc-pricing`: do we want to use toast when the user clicks on the 'Add' button?

## How to display toast?

you just need to dispatch a custom event:

```javascript
const toast = {
  message: 'This is a notification message',
  intent: 'success',
}
window.dispatchEvent(new CustomEvent('cc-toast:show', { detail: toast, bubbles: true, composed: true }));
```

pdesoyres-cc:
fixes #478

### DIFF
--- a/src/lib/notifications.js
+++ b/src/lib/notifications.js
@@ -1,0 +1,42 @@
+import { dispatchCustomEvent } from './events.js';
+
+/**
+ * @typedef {import('../toast/types.js').Notification} Notification
+ */
+
+/**
+ * @param {EventTarget} target
+ * @param {Notification} notification
+ * @return {CustomEvent} the <code>cc:notify</code> event that has been dispatched.
+ */
+export function notify (target, notification) {
+  return dispatchCustomEvent(target, 'cc:notify', notification);
+}
+
+/**
+ * @param {EventTarget} target
+ * @param {string} message
+ * @param {string?} title
+ * @return {CustomEvent} the <code>cc:notify</code> event that has been dispatched.
+ */
+export function notifyError (target, message, title) {
+  return notify(target, {
+    message,
+    title,
+    intent: 'danger',
+  });
+}
+
+/**
+ * @param {EventTarget} target
+ * @param {string} message
+ * @param {string?} title
+ * @return {CustomEvent} the <code>cc:notify</code> event that has been dispatched.
+ */
+export function notifySuccess (target, message, title) {
+  return notify(target, {
+    message,
+    title,
+    intent: 'success',
+  });
+}
--- a/src/tcp-redirections/cc-tcp-redirection-form.js
+++ b/src/tcp-redirections/cc-tcp-redirection-form.js
@@ -78,7 +78,6 @@ export class CcTcpRedirectionForm extends LitElement {
               ?skeleton=${skeleton}
               ?waiting=${redirection.waiting}
               ?private=${redirection.private}
-              ?error=${redirection.error}
             ></cc-tcp-redirection>
           `)}
         ` : ''}
--- a/stories/env-var/cc-env-var-form.stories.js
+++ b/stories/env-var/cc-env-var-form.stories.js
@@ -1,6 +1,8 @@
 import '../../src/env-var/cc-env-var-form.js';
-// Load smart definition so we can use it in the Markdown docs
+// Load smart definition, so we can use it in the Markdown docs
 import '../../src/env-var/cc-env-var-form.smart-config-provider.js';
+import '../../src/env-var/cc-env-var-form.smart-env-var-addon.js';
+import '../../src/env-var/cc-env-var-form.smart-exposed-config.js';
 import { makeStory } from '../lib/make-story.js';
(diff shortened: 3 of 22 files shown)


src/lib/notifications.js
```diff
@@ -0,0 +1,68 @@
...
+  return promise
+    .then((resp) => {
+      notifySuccess({ message: i18n(`${k}.success`, d) });
+      return resp;
+    })
+    .catch((error) => {
+      notifyError({ message: i18n(`${key}.error`, d) });
+      throw error;
+    });
+}
```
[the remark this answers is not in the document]
hsablonniere: I'm not sure it's better but we could do this:

```
promise
  .then((resp) => notifySuccess({ message: i18n(`${k}.success`, d) }))
  .catch((error) => notifyError({ message: i18n(`${key}.error`, d) }));

return promise;
```

I'm not sure about it...

src/lib/notifications.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @typedef {import('../toast/types.js').Toast} Toast
+ */
+
+/**
+ * @param {Toast} detail
+ * @param {Window|Document|HtmlElement} target
+ */
+export function notify (detail, target = window) {
```
hsablonniere, line 12: I'm wondering if this whole file should and methods should be named `notify`. What about `emitToast` or `dispatchToast`?

src/tcp-redirections/cc-tcp-redirection-form.js
```diff
@@ -72,12 +72,11 @@ export class CcTcpRedirectionForm extends LitElement {
...
-              .sourcePort=${redirection.sourcePort}
-              ?skeleton=${skeleton}
-              ?waiting=${redirection.waiting}
-              ?private=${redirection.private}
-              ?error=${redirection.error}
+                namespace=${redirection.namespace}
+                .sourcePort=${redirection.sourcePort}
+                ?skeleton=${skeleton}
+                ?waiting=${redirection.waiting}
+                ?private=${redirection.private}
```
hsablonniere, line 79: Be careful with the formatter (we don't have :p)

stories/env-var/cc-env-var-form.stories.js
```diff
@@ -1,6 +1,8 @@
 import '../../src/env-var/cc-env-var-form.js';
-// Load smart definition so we can use it in the Markdown docs
+// Load smart definition, so we can use it in the Markdown docs
 import '../../src/env-var/cc-env-var-form.smart-config-provider.js';
+import '../../src/env-var/cc-env-var-form.smart-env-var-addon.js';
+import '../../src/env-var/cc-env-var-form.smart-exposed-config.js';
```
hsablonniere, line 5: Good catch!

stories/saas/cc-grafana-info.stories.js
```diff
@@ -1,4 +1,5 @@
 import '../../src/saas/cc-grafana-info.js';
```
hsablonniere, line 1: You can remove this one.

stories/toast/cc-toaster.stories.js
```diff
@@ -104,11 +104,25 @@ export const defaultStory = makeStory(conf, {
     let i = 0;
     function toast (node, { intent }) {
       i++;
+
+      // prevent other toaster to show the toast.
+      const mainToaster = document.querySelector('cc-toaster.main');
+      let resetMainToaster = () => {};
+      if (mainToaster) {
```
hsablonniere, line 111: `!= null`

hsablonniere (changes requested):
## Code review

I propose to discuss the details F2F

* Nice to add the toaster to all stories
* I really like how it simplifies the code
* You should try the i18n check command ;-)
* Nice to see that update env vars aren't in a last promise anymore :+1
* I'm wondering about the `withNotifications`
  * I think I'd rather have the calls to the toast helper inside the smart component and not the helpers
  * the way you did as a `withNotifications` wrapper has positive aspects
* I'm wondering if we should dispacth from the component and let them bubble
* I think we can simplify the grafana `error` a bit more (according to what we recently discussed)
* It seems like you shortened the translations for TCP redirection errors, is it because they can be too long for a screenreader?

hsablonniere (approved):

hsablonniere:
@pdesoyres-cc This PR looks good to me, we need to rebase it on the other one once the docs are OK.

PR #488, opened 2022-06-30
cc-article-list: fix error trigger when xml parsing fails (smart)

### LINKED ISSUE #487: cc-article-list: error mode is not triggered when xml parsing fails (smart)

Galimede:
Fixes #487


hsablonniere (changes requested):
* once you've changed the `return null` to `throw Error` can you create new unit tests to check that bad XML will throw an error?
* can you split the trim in another commit and add a dedicated to check that adding spaces still works?

hsablonniere:
LGTM :+1:

PR #512, opened 2022-07-21
Rework file structure

### LINKED ISSUE #501: Rework the project structure
## Context

All components are organized in folders that have some kind of meaning.
It can be semantic: `atoms` or related to the domain: `env-vars`.

## Problems

* Finding the right name for the folder is often cumbersome
* This folder must be manually listed in the inputs config for rollup
* The components are published in these folders on npm
  * Which means users need to know where to find them
  * It also means if we want to reorganize and move a component to another folder, it's a breaking change

NOTE: The folder is not used by Storybook for the left menu/tree (even if it's the same). This is done through CSF config.

### 1) Component files in `src/components`

Example: we move `src/atoms/cc-input-text.js` to `src/components/cc-input-text.js`.

* Where do we put the `types.d.ts`?
  * I guess this means every component refers to one giant common `types.d.ts` file.
* I guess this means the smart definition would be mixed up with all these files

### 2) Component files in `src/components/cc-my-component`

Example: we move `src/atoms/cc-input-text.js` to `src/components/cc-input-text/cc-input-text.js`.

* Where do we put the `types.d.ts`?
  * We could put all types only used by the component in a `types.d.ts` in the same folder as the component
  * And keep using the common `types.d.ts` for types used by multiple components
* This would allow us to colocate a smart definition with a component file
  * Example: `src/homepage/cc-article-list.smart.js` to `src/components/cc-article-list/cc-article-list.smart.js`.
  * Do we have smart definition that target multiple components?
    * I don't think so
* Do we want to colocate the story file in the component folder?
  * Example: `stories/atoms/cc-input-text.stories.js` to `src/components/cc-input-text/cc-input-text.stories.js`.
* Do we want to colocate the markdown file used for smart definitions in the component folder?
  * Example: `stories/homepage/cc-article-list.smart.md` to `src/components/cc-article-list/cc-article-list.smart.md`.
* Right now, we don't have automated tests for components but we may colocate the test file in this folder too

## Impacts

In both cases, I propose that we publish to npm with all components directly available in the `root` folder.

```js
import `@clevercloud/components/dist/cc-input-text.js`;
```

:warning: This is a breaking change.

hsablonniere:
Fixes #501


hsablonniere:
This is kinda nice now :wink: 

```
src
├── assets
│   ├── *.svg
│   └── world-110m.geo.js
├── components
│   ├── cc-*
│   │   ├── cc-*.d.ts
│   │   ├── cc-*.js
│   │   ├── cc-*.smart.js
│   │   ├── cc-*.smart.md
│   │   └── cc-*.stories.js
│   └── common.d.ts
├── lib
│   └── *.js
├── mixins
│   └── with-resize-observer
│       ├── with-resize-observer.js
│       ├── with-resize-observer.md
│       └── with-resize-observer.stories.js
├── stories
│   ├── all-form-controls.js
│   ├── assets
│   │   └── *.svg
│   ├── fixtures
│   │   └── *.js
│   └── lib
│       └── *.js
├── styles
│   ├── default-theme.css
│   ├── undefined-components.css
│   └── *.js
├── templates
│   ├── cc-addon-encryption-at-rest-option
│   │   └── cc-addon-encryption-at-rest-option.js
│   └── cc-link
│       ├── cc-link.js
│       ├── cc-link.md
│       └── cc-link.stories.js
└── translations
    ├── translations.en.js
    └── translations.fr.js
```

hsablonniere:
:ballot_box_with_check: The CDN build worked almost without any change

The npm build required a few work and as you remember, we had open questions about how to flatten or not the non component files. I went for a "everything that should be used by users is at the root, the rest is in sub folders like in the source".

The only exceptions are `src/styles/default-theme.css` and `src/styles/undefined-components.css` which goes into `dist/styles/*.css`.

The result of the npm build would be like that:

```
dist
├── assets
│   ├── *.svg
│   └── world-110m.geo.js
├── cc-*.js
├── cc-*.smart.js
├── i18n.js
├── lib
│   └── *.js
├── mixins
│   └── with-resize-observer
│       └── with-resize-observer.js
├── smart-manager.js
├── styles
│   ├── default-theme.css
│   ├── undefined-components.css
│   └── *.js
├── templates
│   ├── cc-addon-encryption-at-rest-option
│   │   └── cc-addon-encryption-at-rest-option.js
│   └── cc-link
│       └── cc-link.js
├── translations.en.js
├── translations.fr.js
└── visualizer-stats.html
```

WDYT?

pdesoyres-cc:
I totally agree with your choices about the npm build. :+1:

florian-sanders-cc:
> "everything that should be used by users is at the root, the rest is in sub folders like in the source

Definitely agree with this logic.
I'm ok with the overall structure :+1:

hsablonniere:
I would say this PR is done, now we need to do some testing:

* [ ] test storybook local
* [ ] test storybook prod
* [ ] test local npm usage in the console
* [ ] test sourcemap support via npm usage

## About the npm build

Justin wrote a recap here:

https://justinfagnani.com/2019/11/01/how-to-publish-web-components-to-npm/

I agree with most of what he says:

* I'm still defending the fact that we need to minify the HTML/CSS before publishing on npm
  * Users should not know we're using lit and should not have to cater their build for this
* I would be OK not to minify before pushing to npm...

## About the index.js

Seems like lots of people publish an `index.js` with everything in it. There's no way to treeshake that kind of big import because f the side effects so I'm really wondering what's the point.

:arrow_right: I created a small plugin to generate the `index.js` anyway

* It should help a few CDN
* It should help bundlephobia

Here's a few examples of other libs and how they expose an index.js

* https://www.fast.design/docs/components/getting-started
  * https://bundlephobia.com/package/@microsoft/fast-components@2.30.6
* https://graphitedesignsystem.com/guidelines/components-usage/#usage-with-bundlers
  * https://bundlephobia.com/package/@graphiteds/core@1.9.2
* https://github.com/carbon-design-system/carbon-web-components#how-to-install
  * https://bundlephobia.com/package/carbon-components@10.56.0
* https://crayons.freshworks.com/introduction/#getting-started
  * https://bundlephobia.com/package/@freshworks/crayons@3.5.0

rollup/rollup-cdn.config.js
```diff
@@ -45,8 +44,6 @@ export default {
   plugins: [
     // output directory cleanup
     clearPlugin({ outputDir }),
-    // transform .json files into ES6 modules. (Used by statuses library). TODO: we should get rid of this library
-    json(),
```
pdesoyres-cc, line 49: Maybe I missed something here. I thought we needed that because we use the `statuses` dependency. We do not use it anymore?

hsablonniere: Oh, you mean this error? :rofl: 

It's not needed in one of our two builds but it's the other one. It should be in the CDN build but not in the npm build.

Capture d’écran du 2022-08-26 10-32-39

hsablonniere: It's actually what the commit says :-( Thanks again for catching this. I fixed id.

florian-sanders-cc:
LGTM :+1:  

Pretty sure I did not even say it because it's always kind of assumed that your work is great but this is impressive and inspiring ! 
I would probably have done it all manually and would have spent days crying :laughing: 

If I face this kind of stuff later, I'll definitely try it your way, this is so great :clap:

hsablonniere:
> LGTM +1
> 
> Pretty sure I did not even say it because it's always kind of assumed that your work is great but this is impressive and inspiring ! I would probably have done it all manually and would have spent days crying laughing
> 
> If I face this kind of stuff later, I'll definitely try it your way, this is so great clap

Ah ah thanks. The automation allowed us to change our minds a few times, to rebase easily on other's work and to have more confidence in the result ;-)

PR #519, opened 2022-07-25
cdn-ui: show style endpoint if relevant

### LINKED ISSUE #508: CDN UI: display info related to the theme stylesheet
# Context 

Following #403 / #472, the common CSS for our components (the theme) is now provided through a CSS file instead of a JS file.

This means people using our components need to use this CSS file:
* by relying on a `<link ref="stylesheet">` tag,
* or a CSS `@import`.

This issue can only be handled after we have updated the CDN back end so that it can serve the CSS file corresponding to a specific version of the component. (see issue [smart CDN #2](https://github.com/CleverCloud/clever-components-cdn/issues/2)).

# Goals

The CDN UI needs to display the end point serving this CSS file (the specific endpoint would depend on the components version).

# TODO

- [ ] show the relevant sections from the cdn-ui index file (they are ready but hidden at the moment).
- [ ] only show if manifest version === 2
- [ ] update the ["How to use the components theme?"](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%93%8C-docs-how-to-use-the-components-theme--page) doc with a Smart CDN section.

florian-sanders-cc:
Fixes #508 

I went for the quick and dirty way.

I feel like the UI is evolving and the code is becoming quite verbose.
Using a declarative pattern, maybe with a very lightweight framework like [petite vue](https://github.com/vuejs/petite-vue) could be great here, even though this UI is WIP ?

### DIFF
--- a/cdn-ui/index.html
+++ b/cdn-ui/index.html
@@ -109,11 +109,11 @@
   </div>
   <dl id="styles-area">
     <dt>Styles URL:</dt>
-    <dd><a id="cdn-styles-url" target="_blank" href="https://components.clever-cloud.com/styles.css">https://components.clever-cloud.com/styles.css</a></dd>
+    <dd><a id="cdn-styles-url" target="_blank">?</a></dd>
   </dl>
   <dl>
     <dt>Script URL:</dt>
-    <dd><a id="cdn-script-url" target="_blank" href="https://components.clever-cloud.com/load.js">https://components.clever-cloud.com/load.js</a></dd>
+    <dd><a id="cdn-script-url" target="_blank">?</a></dd>
   </dl>
 
 </section>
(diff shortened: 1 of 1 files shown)


cdn-ui/index.html
```diff
@@ -121,6 +128,7 @@ <h1>Clever Components smart CDN (WIP)</h1>
 import semver from 'https://jspm.dev/es-semver@1';
 
 const FIRST_CDN_VERSION = '5.3.1';
+let manifest_version = 0;
```
florian-sanders-cc, line 131: should I avoid the global variable and pass this from function to function instead ?

pdesoyres-cc: I don't like this design very much but here it's ok ;)

pdesoyres-cc:
If we come to re-implement this UI, I would opt for a web component for the UI and a smart component for the business logic.

pdesoyres-cc (approved):

roberttran-cc (approved):
Looks good to me!

I agree with making the rendering stack evolve (maybe in the future): it could be done while rethinking this page and what features we could add.

cdn-ui/index.html
hsablonniere: Same, it's OK for now ;-)

PR #522, opened 2022-07-28
cc-select: use value property instead of selected attribute

### LINKED ISSUE #514: cc-select: use the value property instead of the selected attribute

florian-sanders-cc:
Fixes #514 

# Issues:

## The selected attribute issue

The `cc-select` component was relying on the `selected` attribute instead of the value property.

```js
<option
  value=${option.value}
  ?selected=${option.value === this.value}
>
  ${option.label}
</option>
```

This led to inconsistent renders. 
The `value` of the component and the visually selected option could end up completely disconnected when trying to set the `value` of the component from outside (changing the value programmatically with no user action).

Also, the `selected` attribute is supposed to only be used to set the initial / default value of the `<select>` element.
It is not supposed to reflect the `<select>` value as explained in the [MDN Docs about the select element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#attr-selected).
It behaves just like the `checked` attribute, which I've always found confusing.

## The value property issue

Using the `value` property is not as simple as it looks with the `<select>` element.

The issue happens during the first render, if you want to set a default value.
The `value` property must match the `value` of one of the `<option>` elements inside the `<select>` element.
These `<option>` elements are generated from the `options` prop.
What seems to happen (thanks @Galimede for finding this out) is visible in the screenshot below:
image

1. update - Properties have been set with new values, it triggers a render.
2. rendering start - The component properties are set as they should, render updates the DOM based on these values.
3. during the rendering - It first tries to set the value of the `<select>` element, but no `<option>` has been added yet. The `<select>` value does not match any `<option>` value.
4. during the rendering - It then generates the `<option>` elements.
5. updated - render is over. The native `<select>` value does not match the component `value`. It seems it falls back to the first item from the list (which is the default behavior when no `value` is specified). 

If you want to check the debug yourself, you can use the [cc-select/debug-value-issue branch](https://github.com/CleverCloud/clever-components/tree/cc-select/debug-value-issue).
Use the "Required" story of the `cc-select` component.

# Solutions

We decided to add an internal `_value` property.
This property is updated **after any render** that was triggered by an update to the `options` or the `value` (`firstUpdated` lifecycle hook).
This property is synced with the native `<select>` value.

This way, we make sure that <option> elements have been rendered before updating the `<select>` value.

# Questions:

- do we rely on lit and an internal property like "_value" or on an imperative update of the `<select>` value ?
- when no value is provided to the `<cc-select>`, should we fallback to the first option (lit behavior ?) or display an empty `<select>` (native behavior) ? 
- when the value provided is inconsistent (does not match any option), what should we do ? Empty `<select>` vs first option select

### DIFF
--- a/src/components/cc-select/cc-select.js
+++ b/src/components/cc-select/cc-select.js
@@ -10,6 +10,11 @@ import { i18n } from '../../lib/i18n.js';
 /**
  * A component wrapping native select element and its label.
  *
+ * Caution:
+ *
+ *   * Setting an empty / undefined value displays a `<select>` element with no selected value unless a placeholder is set.
+ *   * Setting a value that does not match any option displays a `<select>` element with no selected value.
+ *
  * @cssdisplay inline-block
  *
  * @event {CustomEvent<string>} cc-select:input - Fires the `value` whenever the `value` changes.
@@ -55,17 +60,17 @@ export class CcSelect extends LitElement {
     /** @type {Option[]|[]} Sets the list of options displayed inside the select element. */
     this.options = [];
 
-    /** @type {string|null} Sets a default option with empty value. */
+    /** @type {string|null} Sets a disabled option with empty value. This option will always be the first of the list. It can be selected by default by setting `value=''`. */
     this.placeholder = null;
 
     /** @type {boolean} Sets the "required" text inside the label */
     this.required = false;
 
-    /** @type {string|null} Sets the selected value of the element. */
+    /** @type {string|null} Sets the selected value of the element. This prop should always be set. It should always match one of the option values. */
     this.value = null;
 
     // use this unique id for isolation (Safari seems to have a bug)
-    /** @type {string} used by the aria-describedby attribute on the `<select>` element and the `id` attribute on the error slot container */
+    /** @type {string} used by the `aria-describedby` attribute on the `<select>` element and the `id` attribute on the error slot container */
     this._uniqueErrorId = Math.random().toString(36).slice(2);
 
     // use this unique id for isolation (Safari seems to have a bug)
@@ -73,8 +78,19 @@ export class CcSelect extends LitElement {
     this._uniqueHelpId = Math.random().toString(36).slice(2);
 
     // use this unique name for isolation (Safari seems to have a bug)
-    /** @type {string} used by the for/id relation between `<label>` and `<select>` */
+    /** @type {string} used by the `for`/`id` relation between `<label>` and `<select>` */
     this._uniqueInputId = Math.random().toString(36).slice(2);
+  }
+
+  updated (changedProperties) {
+    /*
+     * The `<select>` value must match the value of an `<option>` element.
+     * We need to make sure the value of the `<select>` element in only updated after
+     * `<option>` elements have been rendered.
+    */
+    if (changedProperties.has('value') || changedProperties.has('options')) {
+      this.shadowRoot.querySelector('select').value = this.value;
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/components/cc-select/cc-select.stories.js
```diff
@@ -163,7 +174,7 @@ export const disabled = makeStory(conf, {
   items: [
     {
       label: 'Favourite artist',
-      placeholder: '-- Select an artist --',
+      value: 'LENNON',
```
florian-sanders-cc, line 177: it did not make much sense to have a placeholder saying "Select an artist" in a disabled select element.

roberttran-cc (approved):
LGTM! :muscle:

src/components/cc-select/cc-select.js
```diff
@@ -10,6 +10,10 @@ import { i18n } from '../../lib/i18n.js';
 /**
  * A component wrapping native select element and its label.
  *
+ * Caution:
+ *   * Setting an empty / undefined value displays a `<select>` element with no selected value unless a placeholder is set.
+ *   * Setting a value that does not match any option displays a `<select>` element with no selected value.
```
hsablonniere, line 15: I'm nit picking here but can you write it like this:

```
 * Caution:
 *
 * * Setting an empty / undefined value displays a `<select>` element with no selected value unless a placeholder is set.
 * * Setting a value that does not match any option displays a `<select>` element with no selected value.
 ```

File: src/components/cc-select/cc-select.js [hunk 3]
```diff
@@ -55,28 +59,38 @@ export class CcSelect extends LitElement {
...
     /** @type {string} used by the `aria-describedby` attribute on the `<select>` element and the `id` attribute on the help text container */
     this._uniqueHelpId = Math.random().toString(36).slice(2);
 
     // use this unique name for isolation (Safari seems to have a bug)
-    /** @type {string} used by the for/id relation between `<label>` and `<select>` */
+    /** @type {string} used by the `for`/`id` relation between `<label>` and `<select>` */
     this._uniqueInputId = Math.random().toString(36).slice(2);
   }
 
+  updated (_changedProperties) {
```
hsablonniere, line 84: `changedProperties` is not "private", we can skip the underscore prefix

src/components/cc-select/cc-select.js
```diff
@@ -55,28 +59,38 @@ export class CcSelect extends LitElement {
...
+    /*
+     * The `<select>` value must match the value of an `<option>` element.
+     * We need to make sure the value of the `<select>` element in only updated after
+     * `<option>` elements have been rendered.
+    */
+    this.shadowRoot.querySelector('select').value = this.value;
+    super.updated(_changedProperties);
```
hsablonniere, line 91: Call to super is not required according to the docs ;-)

src/components/cc-select/cc-select.js
```diff
@@ -103,14 +117,14 @@ export class CcSelect extends LitElement {
...
-            <option value="" disabled selected>${this.placeholder}</option>
+            <option value="" disabled>${this.placeholder}</option>
           ` : ''}
           ${this.options.map((option) => html`
             <option
               value=${option.value}
-              ?selected=${option.value === this.value}
             >
               ${option.label}
             </option>
```
hsablonniere, line 130: Can we put this on just one line?

hsablonniere (changes requested):
@florian-sanders-cc A few details here and there but it's great and ready to be merged after the fixes. :clap:

PR #526, opened 2022-08-02
task: improve i18n check display

### LINKED ISSUE #91: Improve i18n task display ;-)
Inspiration https://github.com/murar8/react-intl-locale-manager

roberttran-cc:
Fixes #91.

- Remove log of each single file scanned.
- Display amount of file scanned, key missing and key unused.
- Display success message when no key are missing or being unused.
- Regroup languages for each key missing or unused.
- Add some colors and emojis.
- Display task execution duration.

### DIFF
--- a/tasks/check-i18n.js
+++ b/tasks/check-i18n.js
@@ -1,7 +1,9 @@
 import { promises as fs } from 'fs';
 import util from 'util';
+import chalk from 'chalk';
 import rawGlob from 'glob';
 import { extractFromCode } from 'i18n-extract';
+import table from 'text-table';
 import { translations as en } from '../src/translations/translations.en.js';
 import { translations as fr } from '../src/translations/translations.fr.js';
 
@@ -20,7 +22,11 @@ async function getUsedKeys (sourceFilepaths) {
 
 async function run () {
 
+  const startTime = process.hrtime();
+
   let errors = false;
+  const allMissingKeys = new Map();
+  const allUnusedKeys = new Map();
 
   const sourceFilepaths = await glob('./src/*/*.js', {
     ignore: [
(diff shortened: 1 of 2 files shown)


tasks/check-i18n.js
```diff
@@ -32,25 +38,44 @@ async function run () {
...
+    const formattedMissingKeys = [];
+    allMissingKeys.forEach((values, key) => {
+      formattedMissingKeys.push(['> ' + key, [...values].join(',')]);
+    });
+    console.log(table(formattedMissingKeys));
+  }
+
   // UNUSED KEYS
   const allUsedKeys = Object.values(usedKeysByFile)
     .reduce((a, b) => [...a, ...b]);
```
florian-sanders-cc, line 81: Hey I know this is not in the scope of the quick win cause this code was already here but I feel like this `reduce` is confusing, it's actually just a `flat()` no ?

florian-sanders-cc (approved):
This looks great, gg ! 
I find some colors difficult to read but it's a terminal based tool so let's not dive into this subject :joy: 

I left one comment about legacy code (I mean, code that you havn't modified) but it can be merged as it is I think :+1:

pdesoyres-cc (approved):
GG. I have nothing to say. This is good to go.

hsablonniere:
I love it, it's a very nice improvement. I also love the emojis and the nice touch for execution duration.

* Could we add :warning: emojis for missing/unused keys?
* About the colors, I like theme but I would say it's not that common to use specific hex colors. Most CLIs rely on the palette so it works with the user's theme. We could use `bgWhite`, `bgRed`, `bgGreen` and `bgYellow`. It's a suggestion, we'll discuss it :wink:

roberttran-cc:
> About the colors, I like theme but I would say it's not that common to use specific hex colors. Most CLIs rely on the palette so it works with the user's theme.

Totally agree, we could go for it. The screenshot below shows hex values on the left and palette colors on the right: WDYT?

And both shows emojis for missing and unused keys.

image

(note: click to view at full size!)

pdesoyres-cc:
Maybe I would go for console palette color as it is more standard and is not very far from the custom hex color you used.

hsablonniere:
@roberttran-cc Looking at the screenshot, I think I would also go for the palette. Maybe with some bold.

florian-sanders-cc:
I'd go with the palette colors too :+1:

roberttran-cc:
Colors updated to palette and bold added for all texts! :heavy_check_mark:

hsablonniere:
Capture d’écran du 2022-08-25 16-30-36

Maybe it's just me but I think we should force black for yellow (at least).

hsablonniere:
Nice!!! :tada:

PR #528, opened 2022-08-04
cc-toggle: update host display css property

### LINKED ISSUE #180: cc-toggle: reconsider default display flex (to inline-flex maybe)
* `<cc-toggle>` default display is flex
* This means it's a block and takes the full width
* The inner `.toggle-group` takes the full width *AND* has a white background

The result is this:

Capture d’écran_2020-12-11_14-50-34

We should reconsider default display to `inline-flex`

roberttran-cc:
Fixes #180.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ title: Changelog
 
 * `<cc-map>`: make dot markers not focusable.
 * `<cc-zone-input>`: make server markers not focusable.
-
-...
+* `<cc-toggle>`: update component host default `display` CSS property (BREAKING CHANGE).
 
 ## 9.0.0 (2022-07-19)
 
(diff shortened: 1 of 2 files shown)


hsablonniere:
LGTM :+1

hsablonniere:
@roberttran-cc Nice ;-)

* You forgot to update the JSDoc `@cssdisplay flex` to `@cssdisplay inline-flex`.
* I don't get why I added `flex-direction: column` :thinking: thanks for removing it

hsablonniere:
Also, FYI, the original bug was partly fixed in v9.0.0. Before this version the focus ring and the white background was applied to the whole line. With the modifications of the theme system, those were only applied to the cc-toggle but it was still on its own like.

With this PR, it's now properly inline :+1:

PR #529, opened 2022-08-08
tools: Add stylelint

### LINKED ISSUE #122: Add stylelint
It was brought to my attention that stylelint handles CSS in template literals.

This means our custom ESLint rule for to sort CSS declarations in JavaScript LitElement component files is useless and not the right approach :sob:

Galimede:
Fixes #122

## Context

The base context was that stylelint is apparently able to handle CSS in template literals which we use in our web components.
It would replace the sorting plugin of eslint that we have right now.

## Remarks and questions

Right now, I have only added the base dependency, added a basic configuration and played a bit with it.
Here are a few remarks and questions that I have right now:
* The base issue was of 2020, when adding stylelint, it doesn't go through template literals anymore by default. You have to install `@stylelint/postcss-css-in-js` to make it work. 
* I have also installed the `stylelint-config-standard` dependency which is a default configuration for stylelint. I guess we will be able to get rid of it when we will be sure of what rules we want to use.
* I have tried to see how stylelint handles the sort of CSS declarations but it seems that it doesn't do it by default which led me to install `stylelint-order`. It works great if you use the `stylelint --fix` (pretty much like the `eslint --fix` that we have). The only downside is that it doesn't tell you precisely where it should be if you wanted to do it manually but that's not a big downside since the --fix option of stylelint does it for you.
* stylelint does check a lot of things and for now and they are probably things that we might not want to check in there.
* I have provided a `stylelint` and `stylelint:fix` in the package.json if you want to try it.
* The config file is in JSON, do we keep it in JSON or do we want to change it to JS or whatever? 

## What's next?
* Play a bit more with it
* Discuss with everyone about rules that we want to check and have
* ...


Galimede:
Hello guys, 

After playing a bit more and some tests here are my conclusions and questions: 

I have tried three configurations: Standard, Recommended, Airbnb (sort of), and a custom one. 
You will find a recap of my findings below. 

## Standard

Standard does trigger a lot of warnings and errors. It extends recommended but enforces its own convention.
It's the one that I chose to extend when adding the custom rules for our config.

Here are some redundant errors below when using stylelint on our components:
```
*  Unexpected empty line => no-empty-first-line                                                                                  
*  Expected empty line before declaration => declaration-empty-line-before  
*  Expected modern color-function notation => color-function-notation        
*  Expected "0.4" to be "40%" => alpha-value-notation            
*  Expected selector ".title" to come before selector ":host([mode="small"]) .title" => no-descending-specificity      
*  Expected empty line before declaration => declaration-empty-line-before  
*  Expected "#eeeeee" to be "#eee" => color-hex-length                
*  Unexpected whitespace at end of line => no-eol-whitespace              
*  Unexpected missing end-of-source newline => no-missing-end-of-source-newline
* Unexpected duplicate selector "img", first used at line 300 => no-duplicate-selectors          
* Expected a leading zero => number-leading-zero            
* Expected modern color-function notation => color-function-notation        
* Expected ".25" to be "25%" => alpha-value-notation            
* Unexpected vendor-prefix "-moz-user-select" => property-no-vendor-prefix      
* Expected empty line before comment => comment-empty-line-before      
* Expected "#000000" to be "#000" => color-hex-length                
* Expected class selector to be kebab-case =>   selector-class-pattern                            
* Unexpected longhand value '0.5em 0 0 0' instead of '0.5em 0 0' => shorthand-property-no-redundant-values
* Expected indentation of 12 spaces => indentation                          
* Expected whitespace after "/*" => comment-whitespace-inside
* Expected double quotes => string-quotes
* Expected "0" to be "0deg" => hue-degree-notation                  
* Expected line length to be no more than 120 characters => max-line-length                
* Expected single space after "," in a single-line function => function-comma-space-after     
```

## Recommended

There are rules that are only made to avoid CSS errors but won't check anything else.
It could be used as a base for making our own custom config instead of using standard for this purpose.

Here are some redundant errors below when using stylelint on our components:

```
*  Expected selector ".title" to come before selector ":host([mode="small"]) .title" => no-descending-specificity      
*  Unexpected duplicate selector "img", first used at line 300 => no-duplicate-selectors          

```

## Airbnb

Airbnb config hasn't been updated since 2018 therefore I wasn't able to test it as it is.
However on its GitHub page you can find most of the rules they use and an explanation of why it was important for them here: [[https://github.com/airbnb/css](https://github.com/airbnb/css)](https://github.com/airbnb/css%5D(https://github.com/airbnb/css))

## Custom

As I mentioned yesterday, I have found a website where you can answer a set of questions to choose the rules that you want to create a custom config ([[https://maximgatilin.github.io/stylelint-config/](https://maximgatilin.github.io/stylelint-config/)](https://maximgatilin.github.io/stylelint-config/%5D(https://maximgatilin.github.io/stylelint-config/)))

As I said earlier I have tried it while extending the standard config but we might want to use the recommended one to only enforce those that we want.

Here is the preview of what the config file would look like:

```
{
    "extends": "stylelint-config-standard",
    "rules": {
        "indentation": 2,
        "string-quotes": "single",
        "no-duplicate-selectors": true,
        "color-hex-case": "lower",
        "color-hex-length": "short",
        "selector-no-id": true,
        "selector-combinator-space-after": "always",
        "selector-attribute-quotes": "always",
        "selector-attribute-brackets-space-inside": "never",
        "declaration-block-trailing-semicolon": "always",
        "declaration-colon-space-before": "never",
        "declaration-colon-space-after": "always",
        "value-no-vendor-prefix": true,
        "number-leading-zero": "always",
        "font-weight-notation": "named-where-possible",
        "font-family-name-quotes": "always-unless-keyword",
        "comment-whitespace-inside": "always",
        "comment-empty-line-before": "always",
        "rule-empty-line-before": "always-multi-line",
        "selector-pseudo-element-colon-notation": "double",
        "selector-pseudo-class-parentheses-space-inside": "never",
        "selector-no-universal": true,
        "media-feature-range-operator-space-before": "always",
        "media-feature-range-operator-space-after": "always",
        "media-feature-parentheses-space-inside": "never",
        "media-feature-name-no-vendor-prefix": true,
        "media-feature-colon-space-before": "never",
        "media-feature-colon-space-after": "never"
    }
}
```
I think it's better to use standard with our custom rules if we want to enforce some styles and standard. 
Otherwise, we could use recommended with our custom rules but it will not enforce any convention/standard.
I do prefer the first option imo.

Still I have some questions about errors that are often present and not fixed automatically:

* Do we want to enforce a case for our selector like kebab-case for example? 
* Do we want to fulfill no-descending-specificity? (this rule means that some selectors should appear before other one like an img selector should appear before a .skeleton img one)
*I have seen that we have some universal and id selectors so I have chosen to deactivate the rules related. But, do we want to keep these selectors as is or change them?

_Note: This config is not definitive and we will be able to change it if we are doing things that triggers errors and that we want to keep doing._ 

## Orders

As for the properties order I have two options to offer you:
* Alphabetical order
* Order based on idiomatic CSS => The idea behind this is that in a selector you first place positioning properties then the display and box model and finally the rest by alphabetical order. (more info here: [https://github.com/necolas/idiomatic-css#declaration-order)](https://github.com/necolas/idiomatic-css#declaration-order) (note: we would only use the order of the properties not the rest)

I let you comment and choose the one you would prefer.

roberttran-cc:
> I think it's better to use standard with our custom rules if we want to enforce some styles and standard.

Agree!

> Do we want to enforce a case for our selector like kebab-case for example?

Hmm... I think we should.

> Do we want to fulfill no-descending-specificity?

I think we should too, but it might trigger some style fixing, not quite sure.

> As for the properties order I have two options to offer you...

I'd go for idiomatic order but honestly, alphabetic is fine too (i got used to it I guess :smile:).

florian-sanders-cc:
:wave: thanks for this great work @Galimede :+1: 

> Do we want to enforce a case for our selector like kebab-case for example?

I'm not sure.
Does this mean BEM style like `my-selector__whatever` or `my-selector--modifier` would trigger errors ? We don't really need BEM in our case but I do like to use "_" to separate the parent name with the child name in a selector and the "--" for modifiers.
It's more of a question than an answer, if we choose to go with kebab case only, I'm ok with it.

> Do we want to fulfill no-descending-specificity?

I think I really like this rule and I definitely agree with stylelint approach & explanation.
Still, this feels like it's the kind of rule that is great when you use it from the start. Enforcing it now might be too much work ? Dunno...

> As for the properties order I have two options to offer you:

Idiomatic is what I prefer but really, I'm fine with alphabetical too :smile:

Galimede:
> Does this mean BEM style like my-selector__whatever or my-selector--modifier would trigger errors ?
@florian-sanders-cc yes it would and sadly this is not an auto fixable error. As I said earlier we're not forced to use this rule but I think it will depends on if we want to enforce a naming style or not.

> Still, this feels like it's the kind of rule that is great when you use it from the start. Enforcing it now might be too much work ? Dunno...
Same here not auto fixable sadly. I think it would be a bit of work but nothing too complicated and it would be a one time change. 
If we want to use it I can fix the error when we will be set on the rules :wink:

hsablonniere:
It's OK with me to use standard and configure it.

> Do we want to enforce a case for our selector like kebab-case for example?

I don't have strong opinions on which one to chose but I would say yes.
As @florian-sanders-cc said, we don't really need BEM naming convention but still, we could use dashes and double dashes in some situations.

> Do we want to fulfill no-descending-specificity? (this rule means that some selectors should appear before other one like an img selector should appear before a .skeleton img one)

I'm not quite sure to understand this one. I need to dig.

> I have seen that we have some universal and id selectors so I have chosen to deactivate the rules related. But, do we want to keep these selectors as is or change them?

I'm not quite sure to understand this one. I need to dig.

> As for the properties order I have two options to offer you:

Let's go for idiomatic (@roberttran-cc won).

hsablonniere:
About the code/config:

* Isn't it possible to specify the custom syntax in the config file instead of the CLI param?
* I guess we only need `stylelint`, `stylelint-config-idiomatic-order` and `@stylelint/postcss-css-in-js`
  * Can you confirm the other new dependencies are WIP?
* This PR only changes the CSS of a component
  * Once we agree on the config, can we put all the automated (and manual) changes on current CSS in a different commit? Something like "refactor: apply stylelint config".
* We need to remove the custom eslint rule
  * update `.eslintrc.cjs` and remove `sort-lit-element-css-declarations`
  * remove the `/* eslint-disable sort-lit-element-css-declarations */` comment in `leaflet.js`
* We need a new ADR :pray: (by the way, ADR 11 was about the custom ESLint rule)
* Delete the `eslint-rules/sort-lit-element-css-declarations.js` file
* Remove the dependency of this custom rule `postcss`, `postcss-sorting`

pdesoyres-cc (changes requested):
Nice work! :clap: 

## customSyntax

The `customSyntax` should be specified in the config file instead of the command line:
```
{
	"customSyntax": "@stylelint/postcss-css-in-js"
}
```

## non-auto fixable rules

There are a lot of rules that are not fixable automatically. Do you plan to fix them all in this PR?

## Some interrogations about some rules:

### no-empty-first-line

The way we write CSS in js, we always have this first empty line. With this rule,

```
static get styles () {
    return [
      skeletonStyles,
      // language=CSS
      css`
        :host {
          display: block;
        }
      `,
    ];
```
needs to be written
```
static get styles () {
    return [
      skeletonStyles,
      // language=CSS
      css`:host {
          display: block;
        }
      `,
    ];
```
And I find this less readable.

It seems like you didn't have the problem on your side :thinking: 

### no-eol-whitespace and no-missing-end-of-source-newline
Same as above, these rules break the readability because

```
static get styles () {
    return [
      skeletonStyles,
      // language=CSS
      css`
        :host {
          display: block;
        }
      `,
    ];
```
needs to be written
```
static get styles () {
    return [
      skeletonStyles,
      // language=CSS
      css`
        :host {
          display: block;
        }
`,
    ];
```

hsablonniere (changes requested):
Well done @Galimede, we're almost there :hugs: 

* Nice the empty line between custom props
* Oh, it changes the rgb syntax to space separation?
  * support seems good so OK
  * https://caniuse.com/mdn-css_types_color_rgba_space_separated_parameters
* some vendor prefix are removed (appearance, user-select)
    * https://caniuse.com/mdn-css_properties_user-select
    * https://caniuse.com/css-appearance
    * safari still requires some prefix
* RIP pseudo aligned "grid-template-areas" :rofl:
* the formatting/wrapping of comments is weird
* the spacing around comment lines is weird
  * cc-env-var-create
  * cc-env-var-editor-simple
* the alignement of the last backtick is weird

pdesoyres-cc (approved):
Great! :clap:

hsablonniere (approved):

PR #530, opened 2022-08-09
cc-addon-admin: fix skeleton

### LINKED ISSUE #496: cc-addon-admin: skeleton story seems to be broken
The [`<cc-addon-admin>` skeleton story](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-addon-cc-addon-admin--skeleton) is supposed to show skeleton text inputs but it seems to be broken.

Haven't had time to check exactly why.

Galimede:
Fixes #496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ title: Changelog
 * `<cc-map>`: make dot markers not focusable.
 * `<cc-zone-input>`: make server markers not focusable.
 * `<cc-toggle>`: update component host default `display` CSS property (BREAKING CHANGE).
+* `<cc-addon-admin>`: fix skeleton mode
 
 ## 9.0.0 (2022-07-19)
 
--- a/src/addon/cc-addon-admin.js
+++ b/src/addon/cc-addon-admin.js
@@ -62,12 +62,6 @@ export class CcAddonAdmin extends LitElement {
     this._tags = [];
   }
 
-  set addon (addon) {
-    this._skeleton = (addon == null);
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
-...
+* `<cc-input-text>`: fix skeleton mode
```
florian-sanders-cc, line 9: should be `<cc-addon-admin>` :wink:

pdesoyres-cc (commented):
What you've done resolves the problem. But I find that playing with the component lifecycle for such small computing is overkill.
The code may be simpler if you:
* compute the `skeleton`, `name`, and `tags` information inside the `render()` method. (using local variables)
* remove the `_skeleton`, `_name`, and `_tags` properties
* remove the `update(changedProperties)` overriding.

hsablonniere:
@pdesoyres-cc How would you handle the internal state of the _name and _tags? I think we should merge this and maybe rework it once we make progress on forms and state.

pdesoyres-cc:
@hsablonniere you're right. I missed something during my review. What I said applies for `_skeleton`, but not for `_tags` and `_name` properties.

hsablonniere:
@Galimede We decided to move the `_skeleton` in the render for you :wink:

hsablonniere:
We're merging this as is. Thanks again @Galimede and thanks @pdesoyres-cc for the feedbacks.

PR #531, opened 2022-08-10
cc-input-text with tags: set default font back to `--cc-ff-monospace`

### LINKED ISSUE #510: cc-input-text (tags): change font-family back to monospace
# Context

Following #453, `cc-input-text` no longer uses the monospace font for tags.

# Goals

* We want the font to be monospace by default when using tags (**only when using tags**, it should still inherit in other cases).
* It should still be customizable by using the CSS custom prop.

florian-sanders-cc:
Fixes #510

### DIFF
--- a/src/atoms/cc-input-text.js
+++ b/src/atoms/cc-input-text.js
@@ -50,7 +50,7 @@ function arrayEquals (a, b) {
  * @event {CustomEvent} cc-input-text:requestimplicitsubmit - Fires when enter key is pressed in simple mode, in tags mode or when ctrl+enter is pressed in multi mode.
  * @event {CustomEvent<string[]>} cc-input-text:tags - Fires an array of tags whenever the `value` changes (separated by spaces).
  *
- * @cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `inherit`).
+ * @cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `inherit` or `--cc-ff-monospace` when using the tags mode).
  *
  * @slot error - The error message to be displayed below the `<input>` element or below the help text. Please use a `<p>` tag.
  * @slot help - The help message to be displayed right below the `<input>` element. Please use a `<p>` tag.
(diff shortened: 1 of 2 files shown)


roberttran-cc (approved):
LGTM! :+1:

pdesoyres-cc (approved):

hsablonniere:
I update the JSDoc to specify that the default changed:

```
@cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `--cc-ff-monospace` from the them).
```

hsablonniere:
@florian-sanders-cc I want to confirm with you wether this is a breaking change before merging.

florian-sanders-cc:
> @florian-sanders-cc I want to confirm with you wether this is a breaking change before merging.

Did not think about that. I think it is indeed a breaking change :sweat_smile:.
I'll reword the commit and add it to the changelog :+1:

florian-sanders-cc:
> I update the JSDoc to specify that the default changed:
> 
> ```
> @cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `--cc-ff-monospace` from the them).
> ```

thanks :wink:

src/atoms/cc-input-text.js
```diff
@@ -50,7 +50,7 @@ function arrayEquals (a, b) {
  * @event {CustomEvent} cc-input-text:requestimplicitsubmit - Fires when enter key is pressed in simple mode, in tags mode or when ctrl+enter is pressed in multi mode.
  * @event {CustomEvent<string[]>} cc-input-text:tags - Fires an array of tags whenever the `value` changes (separated by spaces).
  *
- * @cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `inherit`).
+ * @cssprop {FontFamily} --cc-input-font-family - The font-family for the input content (defaults: `--cc-ff-monospace` from the them).
```
florian-sanders-cc, line 53: @hsablonniere ahh it's a bit more complicated than this actually.
It only defaults to `--cc-ff-monospace` if you use tags. If not, it inherits.
I'll update the desc if that's ok with you ?

hsablonniere: Nice :+1:

PR #535, opened 2022-08-12
Tooling: automated a11y checks

### LINKED ISSUE #480: A11Y and HTML validation tools
# Goals

We need tools to:
* check that the HTML we produce is robust (tags are closed, attributes are closed and not duplicated, etc.),
* check things related to a11y that can be checked automatically (alt attributes, aria usage, etc.).

# Solutions

After asking on the Lit slack, we think we could rely on 3 layers:

- linting and checking in the editor (easy since it relies on the [a11y lit eslint plugin](https://open-wc.org/docs/linting/eslint-plugin-lit-a11y/overview/)),
- automated and exhaustive testing, to be used before committing for instance ? (would rely on [Open WC Chai Axe](https://open-wc.org/docs/testing/chai-a11y-axe/), as explained earlier),
- specific and more thorough testing for complex components, to be used before committing just to make sure nothing ever breaks in there without us noticing it ? (would rely on [Modern Web Test Runner - a11y snapshot command](https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot)).
This last one looks quite specific and heavy because it generates an accessible tree so that you can check it. That's why we would only use it for a few components.

# What's missing / Up for debate

From what I've tested, all these tools mainly focus on a11y and some basic HTML checks (tags must be closed, etc.).
We're still missing some pieces for HTML checking:

- Content models: for instance, buttons are phrasing contents. They only accept phrasing content as descendants. Using a `<div>` inside it should raise an error but it does not in these tools. They only check stuff directly related to a11y. We used to consider this an accessibility issue but this is debatable (they are considering removing html validation checking from WCAG + I fail to see the issues it could actually cause).
- [Need more testing, I'll edit later... :sweat_smile: ]

### LINKED ISSUE #558: cc-addon-admin: provide visible labels
Input fields in the `<cc-addon-admin>` have no labels.
image

We provided visually hidden labels for now but we actually need to add visible labels to these input fields.
These labels may be redundant with the headings on the left but there should always be a visible label **next to** the input field. The headings are too far from the input to be considered as visible labels.

florian-sanders-cc:
Fixes #480 
Fixes #558 

Alright, this PR is not very easy to review, sorry for this.

# What does it do ?

- add a new eslint plugin to check for a11y issues
- add temp script that generates the test files (you should ignore this commit)
- add new test files and test helpers to run a11y tests, update the package.json scripts used to run the tests
- fix all accessibility issues detected
- mitigate resize observer issues (thanks @pdesoyres-cc. This commit might be dropped if we move to lit's observer before we merge this one).
- fix missing storiy enhanced names
- fix JS error for `cc-tile-status-codes` in error state 
- change `cc-html-frame` `title` prop to `iframe-title` prop. (breaking change)

# What and how to review

## Code 

I would advise to first review commit by commit (I've added comments to provide more context, especially for translations because they are related to hidden texts that you won't see unless you check the code).
You can skip most of the new test files since they are generated from the same template. I added comments of 2 test files that you can review (one with ignore rules, one with stories to exclude).

I especially need feedback on the test helpers, I have added comments to explain the context. 

## UI

Finally, if you are alive and willing to do so, it would be nice to check that nothing is broken visually in the stories of the components impacted by a11y fixes. I suggest we split the workload (I'll recheck everything myself and you only check yours, as suggested below)

@roberttran-cc 

- [x] cc-addon-admin
- [x] cc-button
- [x] cc-env-var-create

@Galimede 

- [x] cc-env-var-editor-expert
- [x] cc-env-var-editor-json
- [x] cc-env-var-input

@hsablonniere 

- [x] cc-header-addon
- [x] cc-html-frame
- [x] cc-input-number
- [x] cc-input-text
- [x] cc-invoice

@pdesoyres-cc 

- [x] cc-pricing-estimation
- [x] cc-pricing-product-consumption
- [x] cc-pricing-table
- [x] cc-zone-input

# TODO

- [x] REBASE to retrieve deps bump from master once #557 is merged.
- [x] refactor web dev server and web test runner shared config / plugins once #557 is merged.
- [x] regenerate all test files.
- [x] Add a commit to remove the tmp script and its related script entries in the `package.json`


pdesoyres-cc:
## Investigation on `ResizeObserver loop limit exceeded` error

To reproduce the error, use the command `wtr 'test/a11y/unique.test.js' --node-resolve`

---

### Attempt 1

According to https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

> This error means that ResizeObserver was not able to deliver all observations within a single animation frame. It is benign (your site will not break). – [Aleksandar Totic](https://github.com/WICG/ResizeObserver/commits?author=atotic) Apr 15 at 3:14

I tried to modify resize-observer.js with the code below:
```
const ro = new window.ResizeObserver((entries) => {
	  window.requestAnimationFrame(() => {
	    // NOTE: We could use entries[0].borderBoxSize.inlineSize but not supported in Chrome, Safari or polyfill
	    const { width } = this.getBoundingClientRect();
	    this._onResize({ width });
	  });
});
````
By wrapping code in `window.requestAnimationFrame` we limit the executions to a single frame.

This doesn't break the with-resize-observer story.

To check if nothing is broken, I added a second story that simulates a more realistic usage. It asks for the user to really resize its window. And, good news, nothing is broken.

Code can be found on branch `tooling/automated-a11y-checks-ro-fix-1`

### Attempt 2

Another approach is to ignore the error when running tests:

I don't really like this workaround but it is another approach that affects only tests but not the code itself.

Code can be found on branch `tooling/automated-a11y-checks-ro-fix-2`

**Comment by florian-sanders-cc** [comment]
Hey there,

Sorry for this but actually it does not seem like the fix 1 work.

We can discuss it later :wink:

File: src/components/cc-input-number/cc-input-number.js [hunk 1]
```diff
@@ -186,7 +191,7 @@ export class CcInputNumber extends LitElement {
       <div class="meta-input">
         ${controls ? html`
           <button class="btn" @click=${this._onDecrement} ?disabled=${this.disabled || this.readonly || minDisabled}>
-            <img class="btn-img" src=${decrementSvg} alt="">
+            <img class="btn-img" src=${decrementSvg} alt="${i18n('cc-input-number.decrease')}">
```
florian-sanders-cc, line 194: translations:
`decrease`
`décrémenter` (not sure this is a good idea, maybe I should have gone with `augmenter` and `réduire`)

src/mixins/with-resize-observer/with-resize-observer.js
```diff
@@ -32,9 +32,11 @@ export function withResizeObserver (ParentClass) {
...
       }
       const ro = new window.ResizeObserver(() => {
-        // NOTE: We could use entries[0].borderBoxSize.inlineSize but not supported in Chrome, Safari or polyfill
-        const { width } = this.getBoundingClientRect();
-        this._onResize({ width });
+        window.requestAnimationFrame(() => {
+          // NOTE: We could use entries[0].borderBoxSize.inlineSize but not supported in Chrome, Safari or polyfill
+          const { width } = this.getBoundingClientRect();
+          this._onResize({ width });
+        });
```
florian-sanders-cc, line 39: thanks @pdesoyres-cc apparently it does not break anything and it does fix a lot of `resize observer loop limit exceeded issue` occurrences so why not ?

Anyway, this commit might be dropped if we implement the lit resize observer before the merge.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ title: Changelog
 * `<cc-pricing-table>`: add text content to `<cc-button>` elements (toggle State and add buttons) so that they can be identified by assistive technologies.
 * `<cc-zone-input>`: add `tabindex="0"` to the list container so that people can focus and scroll the list without having the change their selection.
 * `resizeObserver`: add `window.requestAnimationFrame` in the resize observer callback to mitigate the `resize observer loop limit exceeded issue`.
+* stories: fix miscellaneous stories not being passed to `enhancedStoryNames`.
```
florian-sanders-cc: At one point, my testing helpers relied on `story.storyName`. The `storyName` is added by the `enhancedStoryNames` function if i'm right.

Testing would fail because some of our stories were not going through the function so I figured I'd fix them even though we don't rely on these for story testings. 

It might be something that we could check with eslint later on ?

CHANGELOG.md
```diff
@@ -43,7 +43,8 @@ title: Changelog
 * `<cc-pricing-table>`: add text content to `<cc-button>` elements (toggle State and add buttons) so that they can be identified by assistive technologies.
 * `<cc-zone-input>`: add `tabindex="0"` to the list container so that people can focus and scroll the list without having the change their selection.
 * `resizeObserver`: add `window.requestAnimationFrame` in the resize observer callback to mitigate the `resize observer loop limit exceeded issue`.
-* stories: fix miscellaneous stories not being passed to `enhancedStoryNames`.
+* stories: fix miscalleneous stories not being passed to `enhancedStoryNames`. 
+* `<cc-tile-status-codes>`: fix JavaScript errors in case initial loading fails (error state).
```
florian-sanders-cc, line 48: you can check the error in prod, just open the `error` story of the component:
image

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ title: Changelog
 * `<cc-input-text>`: set default font back to `--cc-ff-monospace` when the input contains tags (BREAKING CHANGE).
 * Introduce a new project file structure (BREAKING CHANGE).
 * all components: change `rem` units to `em` (BREAKING CHANGE).
+* `<cc-html-frame>`: change `title` prop to `iframeTitle` (`iframe-title` when used as an attribute) to avoid conflicts with the native `title` attribute (BREAKING CHANGE).
```
florian-sanders-cc, line 15: @hsablonniere this is a breaking change but I figured it was ok to do it because it's only used in our `cc-invoice` component.
Do you know of any use outside this, for instance the admin console or the docs website ?

src/components/cc-addon-admin/cc-addon-admin.js
```diff
@@ -108,10 +108,11 @@ export class CcAddonAdmin extends LitElement {
         ${!loadingError ? html`
 
           <cc-block-section>
-            <div slot="title">${i18n('cc-addon-admin.addon-name')}</div>
+            <div slot="title">${i18n('cc-addon-admin.heading.name')}</div>
```
florian-sanders-cc, line 111: translations:
`Name`
`Nom`

src/components/cc-addon-admin/cc-addon-admin.js
```diff
@@ -108,10 +108,11 @@ export class CcAddonAdmin extends LitElement {
...
             <div slot="info"></div>
             <div class="one-line-form">
               <cc-input-text
+                label="${i18n('cc-addon-admin.input.name')}"
```
florian-sanders-cc, line 115: all translations of `cc-addon-admin` are visible in the story (no hidden label) so you can skip this and only check visually.

src/components/cc-addon-admin/cc-addon-admin.js
```diff
@@ -180,6 +182,10 @@ export class CcAddonAdmin extends LitElement {
           flex: 1 1 10em;
           margin-right: 0.5em;
         }
+        
+        .one-line-form cc-button {
+          margin-top: var(--cc-margin-top-btn-horizontal-form);
+        }
```
florian-sanders-cc, line 188: now that we've added labels, we need this to make sure everything is aligned

src/components/cc-block/cc-block.js
```diff
@@ -71,12 +71,14 @@ export class CcBlock extends LitElement {
...
+    /* eslint-disable lit-a11y/click-events-have-key-events */
     return html`
 
       ${this.ribbon != null && this.ribbon !== '' ? html`
         <div class="info-ribbon">${this.ribbon}</div>
       ` : ''}
-
+      
       ${!this.noHead ? html`
         <div class="head" @click=${this._clickToggle}>
```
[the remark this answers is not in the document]
hsablonniere: I thought it was OK because we had a button.

package.json
```diff
@@ -24,8 +24,8 @@
     "components:check-i18n": "node tasks/check-i18n.js",
     "components:check-lit": "lit-analyzer src",
     "components:docs": "cem analyze --litelement --outdir dist",
-    "test": "wtr 'test/**/*.test.*' --node-resolve",
-    "test:watch": "wtr 'test/**/*.test.*' --node-resolve --watch",
```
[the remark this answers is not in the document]
hsablonniere: Nice :+1:

src/.eslintrc.cjs
```diff
@@ -7,4 +7,7 @@ module.exports = {
       'peerDependencies': false,
     }],
   },
+  'env': {
+    'mocha': true,
+  },
```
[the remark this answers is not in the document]
hsablonniere: Yes but now we can use describe in non test files. I think we need to add a env comment in each test file.

src/components/cc-input-text/cc-input-text.stories.js
```diff
@@ -221,14 +226,6 @@ export const tagsWithClipboard = makeStory(conf, {
...
 });
 
-export const tagsWithLabel = makeStory(conf, {
-  // language=CSS
-  css: `cc-input-text {
-    width: 300px;
-  }`,
-  items: tagsItems.map((p) => ({ ...p, label: 'Tags here' })),
-});
-
```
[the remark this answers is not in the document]
hsablonniere: You mean without?

CHANGELOG.md
```diff
@@ -19,7 +19,6 @@ title: Changelog
 * `<cc-zone-input>`: make server markers not focusable.
 * `<cc-addon-admin>`: fix skeleton mode
 * `<cc-env-var-form>`: fix toggling to JSON mode while in skeleton state.
-* `<cc-badge>`: add skeleton mode
```
hsablonniere, line 22: It seems like you removed something from the CHANGELOG

test/helpers/accessibility.js
```diff
@@ -0,0 +1,70 @@
+import { expect, fixture, elementUpdated } from '@open-wc/testing';
```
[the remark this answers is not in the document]
hsablonniere: The name of the file and function seems OK to me.

test/helpers/accessibility.js
```diff
@@ -0,0 +1,70 @@
...
+      });
+
+      it(`Mobile: width = ${viewports.mobile.width} height = ${viewports.mobile.height}`, async function () {
+        await setViewport(viewports.mobile);
+        const element = await fixture(story({}, storyConf));
+
+        await elementUpdated(element);
+
+        await expect(element).to.be.accessible({ ignoredRules });
+      });
```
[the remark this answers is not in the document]
hsablonniere: This is OK like this :+1:

test/helpers/stories.js
```diff
@@ -0,0 +1,22 @@
+/**
```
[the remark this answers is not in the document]
hsablonniere: We could rename the file to `get-stories.js`.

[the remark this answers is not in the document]
hsablonniere: Can you try this instead of the reduce:

```
/**
 * Transform the result of an imported module from a story file into an array of story functions that can be used to render every story.
 * The imported module is an object containing every named export (story functions mostly) as well as the default export (metadata about the stories).
 *
 * @param {Object} importedModule the result of a story file import.
 * @param {Array} excludedStories the names of the stories to exclude. These names correspond to the names exported from the story file (for instance `defaultStory`).
 * @returns {Array} the story functions, minus the simulations and the excluded stories. These functions can be used to render each story.
 */
export const getStories = (importedModule, excludedStories = []) => {

  const allExcludedStories = [
    ...(importedModule.default?.excludeStories ?? []),
    ...excludedStories,
  ];

  const filteredStories = Object
    .entries(importedModule)
    .filter(([name, exportItem]) => {
      const isNotExcludedStory = !allExcludedStories.includes(name);
      const isNotSimulation = !name.match(/simulation/i);
      const isFunction = typeof exportItem === 'function';
      return isNotExcludedStory && isNotSimulation && isFunction;
    })
    .map(([key, value]) => value);

  return filteredStories;
};
  ```

File: web-test-runner.config.js [hunk 17]
```diff
@@ -0,0 +1,65 @@
...
+    '**/*.md': 'js',
+    '**/*.json': 'js',
+    '.**/*.json': 'js',
+  },
+  testFramework: {
+    config: {
+      ui: 'bdd',
+      timeout: '5000',
+    },
+  },
```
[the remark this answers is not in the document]
hsablonniere: * Do we need the mime types in the wtr config?
  * I guess it allows us to import stories and some of them do import MD/JSON files, right?

web-test-runner.config.js
```diff
@@ -0,0 +1,65 @@
...
+  testRunnerHtml: (testFramework) => `
+    <html>
+      <body>
+        <link rel="stylesheet" href="src/styles/default-theme.css" >
+        <script type="module" src="${testFramework}"></script>
+      </body>
+    </html>
+  `,
```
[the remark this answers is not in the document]
hsablonniere: * Shouldn't we put a doctype?
* Shouldn't we move script & link tags in a `<head>`?

web-test-runner.config.js
```diff
@@ -0,0 +1,55 @@
...
+  plugins: [
+    cemAnalyzerPlugin,
```
hsablonniere, line 36: * Why do we need the CEM analyzer plugin in the test runner?
  * I guess it allows us to import stories and some of them do import MD/JSON files, right?

web-test-runner.config.js
```diff
@@ -0,0 +1,55 @@
...
+    }),
+    commonjs({
+      // the commonjs plugin is slow, list the required packages explicitly:
+      include: commonJsIdentifiers([
+        'statuses',
+        // used by clever-client
+        'oauth-1.0a',
+        'component-emitter',
+      ]),
+    }),
```
hsablonniere, line 53: Can we group some config between WTR and WDS?

test/helpers/accessibility.js
```diff
@@ -0,0 +1,70 @@
...
+  };
+});
+
+/**
+ * Loop through stories to run accessibility tests on their content both on desktop and mobile
+ *
+ * @param {Array} stories the story functions to execute and run a11y test
+ * @param {Array} ignoredRules names of the rules to ignore
+ */
+export const testAccessibility = (stories, ignoredRules) => stories.forEach(function (story) {
```
hsablonniere, line 48: Can we add a body to the function so we don't "arrow" directly a forEach?

CHANGELOG.md
```diff
@@ -38,6 +38,10 @@ title: Changelog
 * Upgrade from `@custom-elements-manifest/analyzer@0.4.1` to `@custom-elements-manifest/analyzer@0.6.4`.
 * Fix Custom Element Manifest generation web dev server plugin: disable caching
 * Add new eslint plugin `eslint-plugin-lit-a11y` to prevent accessibility issues.
+* Add new dependency `@open-wc/testing` to allow Web Test Runner to run tests for accessibility issues.
+* Add new dependency `@web/test-runner-commands` so that Web Test Runner can test both on desktop and mobile.
+* Add new helpers to extract stories and run accessibility test on them.
+* Add new test files in most of the component folders. These test files only contain accessibility tests for the moment.
```
hsablonniere, line 71: This commit contains both manual updates and automated updates (with the script). Next time, I would prefer these in two commits so the automated modifications are isolated. It's also easier if you need to update them during the review.

CHANGELOG.md
hsablonniere: * :clap: Well done for the batch of enhanceStoryNames.
* Yes, the `storyName` is optional and added by the `enhancedStoryNames`.
* I'd rather have a different story format that does not require this patch (or even a WDS plugin).

CHANGELOG.md
hsablonniere: :clap: Nice catch!

CHANGELOG.md
hsablonniere: :clap: Nice catch. I'm OK with this. No usage apart from invoice to the best of my knowledge.

src/components/cc-env-var-create/cc-env-var-create.js
```diff
@@ -99,6 +99,8 @@ export class CcEnvVarCreate extends LitElement {
       <cc-flex-gap>
 
         <cc-input-text
+          label=${i18n('cc-env-var-create.name.placeholder')}
```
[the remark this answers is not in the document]
hsablonniere: * :warning: You used the name translation instead of the value one
* I tried to put explicit/visible labels and no placeholder, it's not that bad actually, what do others think?

src/components/cc-header-addon/cc-header-addon.js
```diff
@@ -86,8 +86,8 @@ export class CcHeaderAddon extends LitElement {
           <div class="details">
             <div class="name"><span class="${classMap({ skeleton })}">${addon.name}</span></div>
             <cc-flex-gap>
-              <cc-input-text readonly clipboard value="${ifDefined(addon.id)}" ?skeleton=${skeleton}></cc-input-text>
-              <cc-input-text readonly clipboard value="${ifDefined(addon.realId)}" ?skeleton=${skeleton}></cc-input-text>
+              <cc-input-text label=${i18n('cc-header-addon.id-label')} hidden-label readonly clipboard value="${ifDefined(addon.id)}" ?skeleton=${skeleton}></cc-input-text>
```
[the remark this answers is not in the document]
hsablonniere: Idendentifiant => Identifiant

src/components/cc-input-text/cc-input-text.stories.js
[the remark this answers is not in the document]
hsablonniere: I like this :+1:

src/translations/translations.fr.js
```diff
@@ -378,7 +385,7 @@ export const translations = {
   'cc-header-orga.hotline': `Numéro d'urgence :`,
   //#endregion
   //#region cc-heptapod-info
-  'cc-heptapod-info.description': () => sanitize`Cette instance Heptapod héberge des dépôts Mercurial. Plus d'informations sur <a href="https://about.heptapod.host" rel="noreferrer noopener">https://about.heptapod.host</a>.`,
+  'cc-heptapod-info.description': () => sanitize`Cette instance Heptapod héberge des dépôts Mercurial. Plus d'informations sur <a href="https://about.heptapod.host">https://about.heptapod.host</a>.`,
```
hsablonniere, line 388: Good catch!

hsablonniere (changes requested):
Please keep the tmp commit in the PR and add another commit which removes the script at the end of the PR. This way, we end up with no tmp script on master but it's here in the history if needed.

Can you explain the problem with the onClick on cc-block title bar?

* I thought SR users would use the button and mouse users would have a bonus
* I guess even with this, we need to tell via ARIA that the block is hidden/collapsed

src/.eslintrc.cjs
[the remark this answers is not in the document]
florian-sanders-cc: :dart: oupsy !
:+1: for an env comment in each test file then !

web-test-runner.config.js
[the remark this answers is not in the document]
florian-sanders-cc: Well, thanks for this, I forgot to remove `.md` and `.**/*.json` :wink:

web-test-runner.config.js
florian-sanders-cc: `**/*.json` mimeType + `cemAnalyzerPlugin` are useful because `makeStory` imports the CEM. 

But it's true that the part of `makeStory` that requires CEM is only useful for storybook and not for tests.

Do you think we should optimize this ? (I do see how, but I did not want to modify the `makeStory` function unless I was told to do so, considering how many changes we're already introducing :smile: )

web-test-runner.config.js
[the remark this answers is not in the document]
florian-sanders-cc: yes we should :scream_cat:

florian-sanders-cc:
> Can you explain the problem with the onClick on cc-block title bar?
>
>    - I thought SR users would use the button and mouse users would have a bonus
>    - I guess even with this, we need to tell via ARIA that the block is hidden/collapsed

@hsablonniere it is acceptable like this, and this is why I did not make any change, only added a comment to disable the eslint rule for this specific case.
We do convey the status of the area (closed / open) by changing the button name so in this case `aria-expanded` is not mandatory and could even be confusing.
I still think we can do better and rely on a more standard pattern (I commented about this in #225 so that it's tracked).

src/components/cc-env-var-create/cc-env-var-create.js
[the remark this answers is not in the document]
pdesoyres-cc: I like it with the label, but do we really need a placeholder then?

CHANGELOG.md
```diff
@@ -43,7 +43,8 @@ title: Changelog
 * `<cc-pricing-table>`: add text content to `<cc-button>` elements (toggle State and add buttons) so that they can be identified by assistive technologies.
 * `<cc-zone-input>`: add `tabindex="0"` to the list container so that people can focus and scroll the list without having the change their selection.
 * `resizeObserver`: add `window.requestAnimationFrame` in the resize observer callback to mitigate the `resize observer loop limit exceeded issue`.
-* stories: fix miscellaneous stories not being passed to `enhancedStoryNames`.
+* stories: fix miscalleneous stories not being passed to `enhancedStoryNames`.
```
pdesoyres-cc, line 48: typo in `miscalleneous`

pdesoyres-cc (changes requested):
I don't have much to say except that this is an impressive piece of work! GG :clap:

src/components/cc-pricing-table/cc-pricing-table.js
```diff
@@ -243,7 +243,9 @@ export class CcPricingTable extends withResizeObserver(LitElement) {
               hide-text
               circle
               @cc-button:click=${() => this._onToggleState(plan)}
-            ></cc-button>
+            >
+                ${i18n('cc-pricing-table.feature.toggle')}
```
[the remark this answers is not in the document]
roberttran-cc: Maybe `Basculer l'affichage`? :thinking:

src/components/cc-tile-status-codes/cc-tile-status-codes.js
```diff
@@ -80,75 +80,77 @@ export class CcTileStatusCodes extends LitElement {
...
-                }
-              });
-              this.chart.update();
-            },
-            onHover: (e) => {
-              this._ctx.style.cursor = 'pointer';
-            },
-            onLeave: (e) => {
-              this._ctx.style.cursor = null;
+    if (!this.error) {
```
roberttran-cc, line 83: Is there a case for the following syntax?
```js
firstUpdated () {
  if (this.error) {
    return;
  }
  // [initial function content]
}
```
Spares an nested block and gives a more legible diff... :hugs:

CHANGELOG.md
```diff
@@ -12,14 +12,41 @@ title: Changelog
...
+* `accessibility Styles`: add new `accessibilityStyles` containing a `visually-hidden` class to hide content visually but not from assistive technologies.
+* `<cc-input-number> stories`: 
+  * update stories to always show visible labels.
+  * remove `tagsWithLabel` story since there is always a visible label now. 
+* `<cc-input-number>`: 
+  * add `alt` attribute values for controls (+ / -) so that they can be identified by assistive technologies. 
+  * add `hiddenLabel` prop to allow the label to be visually hidden in some cases.
+* `<cc-input-text>`: add `hiddenLabel` prop to allow the label to be visually hidden in some cases.
+* `<cc-addon-admin>`: add visually hidden label for `addon name` and `addon tags` input fields so that these fields can be identified by assistive technologies.
+* `<cc-env-var-create>`: add visually hidden label for all input fields so that these fields can be identified by assistive technologies.
```
roberttran-cc, line 33: Breaking change (as the `:host` display changed)? :thinking:

src/components/cc-env-var-create/cc-env-var-create.js
```diff
@@ -20,7 +20,7 @@ import { linkStyles } from '../../templates/cc-link/cc-link.js';
  * * The validation of the variable name format is handled with [@clevercloud/client](https://github.com/CleverCloud/clever-client.js/blob/master/esm/utils/env-vars.js)
  * * The validation of existing names is handled with the `variablesNames` property which is a list of already existing names.
  *
- * @cssdisplay block
+ * @cssdisplay flex
```
roberttran-cc, line 23: I think the layout is broken for `650px` and `450px` viewport thresholds (inputs and button do not collapse anymore).

roberttran-cc (changes requested):
This PR is :exploding_head: ! GG !
I also reviewed the components assigned to me and `cc-env-var-create` appears twice in the first message of this PR... :thinking:

src/components/cc-env-var-create/cc-env-var-create.js
florian-sanders-cc: ah yes, this is part of a quick test following Hubert's feedback, shouldn't have been pushed like this.
Good catch, thanks :hugs:

CHANGELOG.md
florian-sanders-cc: I fixed it but in the wrong commit :P I'll move the fix to the right commit when squashing the feedback commits, thanks !

src/components/cc-env-var-editor-simple/cc-env-var-editor-simple.js
```diff
@@ -139,8 +139,12 @@ export class CcEnvVarEditorSimple extends LitElement {
         display: none;
       }
 
+      /* Negative margin + padding to make the background-color span through the full width of the cc-block, despite the cc-block padding */
       cc-env-var-create {
-        margin-bottom: 1em;
+        background-color: var(--cc-color-bg-neutral);
+        margin-bottom: 0.5em;
+        margin-inline: -1em;
+        padding: 1em;
```
florian-sanders-cc, line 147: The `overflow:hidden` of the `cc-expand` in the `cc-env-var-form` makes it tricky to obtain a full width background color for `cc-env-var-create`.
image

src/components/cc-tile-status-codes/cc-tile-status-codes.js
```diff
@@ -80,71 +80,73 @@ export class CcTileStatusCodes extends LitElement {
...
-        type: 'doughnut',
-        options: {
-          responsive: true,
-          maintainAspectRatio: false,
-          plugins: {
-            datalabels: {
-              display: false,
+    if (this.error) {
+      return;
+    }
```
florian-sanders-cc, line 85: @roberttran-cc feedback, using a guard clause is cleaner here :+1:

test/helpers/get-stories.js
```diff
@@ -0,0 +1,27 @@
...
+    .filter(([name, exportItem]) => {
+      const isNotExcludedStory = !allExcludedStories.includes(name);
+      const isNotSimulation = !name.match(/simulation/i);
+      const isFunction = typeof exportItem === 'function';
+      return isNotExcludedStory && isNotSimulation && isFunction;
+    })
+    .map(([key, value]) => value);
+
+  return filteredStories;
+};
```
florian-sanders-cc, line 27: @hsablonniere copy pasted from your suggestion. I really like this, thanks !

test/helpers/accessibility.js
```diff
@@ -1 +1,72 @@
...
+          const element = await fixture(story({}, storyConf));
+
+          await elementUpdated(element);
+
+          await expect(element).to.be.accessible({ ignoredRules });
+        });
+      });
+    });
+  });
+};
```
florian-sanders-cc, line 72: only added a body { } to the function instead of using forEach directly after the arrow.

src/.eslintrc.cjs
```diff
@@ -7,7 +7,4 @@ module.exports = {
       'peerDependencies': false,
     }],
   },
-  'env': {
-    'mocha': true,
-  },
```
florian-sanders-cc: @hsablonniere I used `/* eslint-env node, mocha */` comment as you suggested, placed after the import statements in every test file :+1:

pdesoyres-cc (commented):
I think we have a regression on the `cc-zone-input` component. The zone is no more changeable with the keyboard. On the list at the right, the up/down arrow keys apply on the scroll bar instead of moving the selected item.
(I tried on firefox)

florian-sanders-cc:
> I think we have a regression on the cc-zone-input component. The zone is no more changeable with the keyboard. On the list at > the right, the up/down arrow keys apply on the scroll bar instead of moving the selected item.
> (I tried on firefox)

You have two successive tab stops:

* first you reach the list container. The fact you can focus it means you can scroll through it without having to change the selected value because we don't hijack the arrows.
* then you tab once more and reach the selected option. From this, we hijack the arrows and pressing up / down no longers scrolls but it selects a value.

There's a rule saying if something is scrollable, then it should be focusable.
In this specific case, i'm not sure it actually benefits users so I'm going to rollback this change: remove the `tabindex="0"` on the list container and disable this rule for this specific component. That's what I did in the first place but it's really debatable :roll_eyes:

Thanks @pdesoyres-cc :wink:

Galimede (commented):
Hiya, here is my short review for the a11y tooling. 

Before anything else, great job @florian-sanders-cc, that's great and it's cool to have this automated! :clap: :heart:


# UI 

## cc-env-var-editor-expert

* Imo it's weird to have `Variable Editor` in english and `Edition des variables` for the invisible label.
  * I would either put `Variables edition` in english
  * Or change `Edition des variables` in french by `Editeur de variables`

## cc-env-var-editor-json

* Same thing as above for the invisible label naming.

## cc-env-var-input

* The name of the variable seems to be a little off but I don't know if it's related to the Arch font bug.

## Helpers

* Could we add some more context on the comment of the `Accessibility` helper to inform that it will be used by generated test cases?
=> I was wondering where it was used before remembering that it will be used by generated test cases. 
* I don't know if it's possible nor if it's already there but it would be cool to be able to test a subset of the test cases nor generate only some. That would be bonus though :stuck_out_tongue: 
=> e.g: Imagine that I work on the `cc-admin` component. I might want at some point to run the accesibility tests but only for this component and not the others. 
Or maybe if I was working on a feature that has two components I would like to test/generate them both. 
* As generating and deleting accesibility test files commands are sort of tasks shouldn't we group them as `tasks:...` in case we have others "tasks" in the future in the `package.json`?

florian-sanders-cc:
Ideas from @Galimede (as discussed in sync):

* we could add an example file for test files like we do for components and smart components.
* we could document how to test a single file (using the focus mode of wtr or running wtr on a single test file).

roberttran-cc (approved):
I checked the following commits and everything looks good to me!
- Feedback General - Refacto config wds / wtr
- Feedback General - cc-env-var-create visible labels
- Feedback Bob

pdesoyres-cc (approved):
Good to go. Nice job :clap:

PR #544, opened 2022-08-30
cc-env-var-form: fix toggling to JSON mode while in skeleton state

### LINKED ISSUE #543: cc-env-var-form: switching to JSON mode while in skeleton state throws an error
To reproduce:
- open the Clever Components Storybook
- open one of the skeleton stories for the `cc-env-var-form` component
- click on the `JSON` option in the upper-right toggle

Expected behavior: component should display the `JSON` mode.

Observed behavior: component does not update the current mode. If you check the console, you'll see an error is thrown.

This bug was introduced with #497.

roberttran-cc:
Fix #543.

Initial code was using array spreading on a variable (when the user toggles to JSON mode). But this variable can be `null` when the component is in skeleton state.

The fix is checking the variable value and spreading only if it is not null.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ title: Changelog
 
 ## Unreleased (????-??-??)
 
+### Components
+
+* `<cc-env-var-form>`: fix toggling to JSON mode while in skeleton state.
+
 ### ⚠️ BREAKING CHANGES
 
 * `<cc-toggle>`: update component host default `display` CSS property (BREAKING CHANGE).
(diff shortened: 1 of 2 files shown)


pdesoyres-cc (approved):
LGTM

Galimede:
I guess we could also use the nullish coalescing operator `??` but it's fine the way it is.

LGTM 👍

florian-sanders-cc:
LGTM ! :+1:

hsablonniere (approved):

hsablonniere:
@roberttran-cc Well done!!

PR #546, opened 2022-08-30
Fix the plugin used to inject auth for smart component

### LINKED ISSUE #545: Update Plugin to inject auth for smart component
Following the project file structure update, the `injectAuthForSmartComponentsPlugin` does not work anymore.

Need to fix:

* the regex that matches the smart story path,
* the `updateRootContext` import path.

florian-sanders-cc:
Fixes #545


hsablonniere (approved):
Good catch, thanks :+1:

roberttran-cc (approved):
Thanks for catching this! :muscle:

PR #547, opened 2022-08-31
cc-badge: add skeleton mode / impl 1

### LINKED ISSUE #521: Add skeleton mode on cc-badge
The `cc-badge` component is lacking a `skeleton` mode.

What is to be done?

* add a `skeleton` boolean property
* use the `skeleton` global style when `waiting` is `true`
* use the same colors as for `<cc-button>`

pdesoyres-cc:
Fixes #521

## UI

On the UI part, I used the same colors as `<cc-button>`.

Preview is available here: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge-skeleton-impl-1/index.html

:thinking: Maybe, we would like to put these gray colors (background, border) into the theme.

## Implementation

I tried two implementations as I need your thought about what I encountered. (See another PR for second implementation #548) 

In this implementation, I tried to follow the same philosophy already present: there is no `<div>` tag encapsulating the whole component. Instead, it uses directly the `:host` CSS selector to apply styles for the different modes (weight, intent, ...).

In order to add the skeleton mode to our components, we use `skeleton.js` which exports a `skeletonStyles` constant that can be added to the `static get styles()` array of our component.

But here, I don't want this full `skeletonStyles` provided by `skeleton.js`. I just want a part of it and apply some pieces to a `:host([skeleton])` CSS selector.

That is what I did in this implementation: I split the `skeletonStyles` provided by `skeleton.js` into multiple parts.

And then, I used those parts like that:
```javascript

import { innerSkeletonStyles, skeletonPulseStyles } from '../../styles/skeleton.js';

// ...

  static get styles () {
    return [
      skeletonPulseStyles,
      // language=CSS
      css`
        :host([skeleton]) {
          background-color: #bbb !important;
          ${innerSkeletonStyles};
        }
      `,
    ];
  }
```

### DIFF
--- a/src/components/cc-badge/cc-badge.js
+++ b/src/components/cc-badge/cc-badge.js
@@ -1,4 +1,5 @@
 import { css, html, LitElement } from 'lit';
+import { innerSkeletonStyles, skeletonPulseStyles } from '../../styles/skeleton.js';
 
 /**
  * @typedef {import('./cc-badge.types.js').BadgeIntent} BadgeIntent
@@ -17,6 +18,7 @@ export class CcBadge extends LitElement {
       iconAlt: { type: String, attribute: 'icon-alt' },
       iconSrc: { type: String, attribute: 'icon-src' },
       intent: { type: String, reflect: true },
+      skeleton: { type: Boolean, reflect: true },
       weight: { type: String, reflect: true },
     };
   }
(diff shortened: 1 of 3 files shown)


pdesoyres-cc:
`lit-analyzer` checks for CSS validity in 
```
css`...`
```

There is no possibility to ignore a rule for a part of the code using comments.

I had to use `unsafeCSS` to make it work. I don't like it at all :disappointed:

PR #548, opened 2022-08-31
cc-badge: add skeleton mode / impl 2

### LINKED ISSUE #521: Add skeleton mode on cc-badge

pdesoyres-cc:
Fixes #521 

## UI

Visually there is no difference but I added the preview so that you can compare the DOM which differs from the first implementation: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-badge-skeleton-impl-2/index.html

## Implementation

This implementation breaks the initial philosophy which is based on the `:host` CSS selector to apply styles. Instead, it adds a root `<div>` tag which receives all the CSS classes needed to apply style for the different modes (intent, weight, ...)

I add the modify all the CSS selectors to be based on classes instead of `host:`.

Adding the skeleton mode is straightforward as it uses the `skeleton.js` in the same way as usual.

### DIFF
--- a/src/components/cc-badge/cc-badge.js
+++ b/src/components/cc-badge/cc-badge.js
@@ -1,4 +1,6 @@
 import { css, html, LitElement } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
 
 /**
  * @typedef {import('./cc-badge.types.js').BadgeIntent} BadgeIntent
@@ -8,16 +10,17 @@ import { css, html, LitElement } from 'lit';
 /**
  * A component to highlight a small chunk of text.
  *
- * @cssdisplay inline-flex
+ * @cssdisplay inline-block
  */
 export class CcBadge extends LitElement {
   static get properties () {
     return {
-      circle: { type: Boolean, reflect: true },
+      circle: { type: Boolean },
       iconAlt: { type: String, attribute: 'icon-alt' },
       iconSrc: { type: String, attribute: 'icon-src' },
-      intent: { type: String, reflect: true },
-      weight: { type: String, reflect: true },
+      intent: { type: String },
+      skeleton: { type: Boolean },
+      weight: { type: String },
     };
   }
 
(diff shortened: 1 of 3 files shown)


hsablonniere (changes requested):
I prefer this one ;-)

* We could remove some `reflect` in the properties as we don't really need them anymore
* Because you just replaced the `:host` with a `div` the component no longer has a default display. We need to add back a `:host { display: inline-flex }` or `:host { display: inline-block }`, I'm not sure
* I'm not sure this should be a div but maybe a span...

src/components/cc-badge/cc-badge.js
```diff
@@ -36,35 +39,67 @@ export class CcBadge extends LitElement {
...
+      </div>
     `;
   }
 
   static get styles () {
     return [
+      skeletonStyles,
       // language=CSS
       css`
-        :host {
```
florian-sanders-cc, line 58: Missing
```css
:host {
  display: inline-block;
};
```

florian-sanders-cc (changes requested):
Nice job as usual :smile: 

I think this is a breaking change: we cannot change the `font-size` the `height` etc. directly on `<cc-badge>` anymore.

This was especially useful with circle badges in `<cc-block>` (see [cc-tcp-redirection-form (context admin](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-tcp-redirections-cc-tcp-redirection-form--data-loaded-with-context-admin).

We could set the font-size in the `:host` selector instead of the `div`. I don't think we need to be able the height / width yet, if we do, we'd have to introduce custom properties I guess.

florian-sanders-cc:
> I think this is a breaking change: we cannot change the font-size the height etc. directly on <cc-badge> anymore.

Answering myself :smile::
We might be ok actually because we have moved to `em`. 

I'm not a huge fan but it works out at the moment because the badge font-size is `1em` when in `circle` mode and it's based on the `font-size` of its parent (the host) so if we set `cc-badge { font-size: 0.8em }` it works.

src/components/cc-badge/cc-badge.js
```diff
@@ -36,35 +39,67 @@ export class CcBadge extends LitElement {
...
+    /** @type {boolean} Whether the component should be displayed as skeleton. */
+    this.skeleton = false;
+
     /** @type {BadgeWeight} Sets the style of the badge depending on how much one wants it to stand out. */
     this.weight = 'dimmed';
   }
 
   render () {
+    const modes = {
+      dimmed: this.weight == null || this.weight === 'dimmed',
```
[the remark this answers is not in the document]
florian-sanders-cc: Thanks !

I'm just wondering where our limit is on this and how I should handle this type of cases in the future.
What I mean is that when we do this, it means we support `null` as a value for `weight` but it's not documented anywhere because `weight` is supposed to only accept strings with specific values.
Also, we handle `null` but we don't handle an empty string for instance.

I'm ok with this though, It does not cause any issue so why not, this was more of a question than a request for change :wink:

src/components/cc-badge/cc-badge.js
```diff
@@ -36,35 +39,69 @@ export class CcBadge extends LitElement {
...
     return html`
-      ${this.iconSrc != null ? html`
-        <img src=${this.iconSrc} alt=${this.iconAlt ?? ''}>
-      ` : ''}
-      <span>
+      <span class="${classMap(modes)}">
+        ${this.iconSrc != null ? html`
+          <img src=${this.iconSrc} alt=${this.iconAlt ?? ''}>
+        ` : ''}
         <slot></slot>
```
florian-sanders-cc, line 68: you need this <span> around the <slot> because we are using inline-flex.
If we slot something like my super <strong>slotted text</strong> then there will be a flex gap between the text node and the <strong>.

It might also cause other issues which I don't remember (I used to do this a lot before @hsablonniere reviews smile

florian-sanders-cc (changes requested):

hsablonniere:
@pdesoyres-cc This looks good an ready to me
@florian-sanders-cc I'm not sure I 100% understood the font-size stuffs, can you explain them to me? (maybe in a video call)

florian-sanders-cc (approved):
LGTM ! GG :sunglasses:

hsablonniere:
Suggestion for the `span.` specificity hack:

```
/* skeleton is more important */
        .skeleton {
          background-color: #bbb !important;
          color: transparent !important;
        }

        .skeleton.outlined {
          box-shadow: inset 0 0 0 0.06em #777 !important;
        }

        .skeleton img {
          visibility: hidden !important;
        }
        
        .circle {
          border-radius: 50%;
          font-size: 1em;
          height: 1.5em;
          justify-content: center;
          min-height: unset;
          padding: 0;
          width: 1.5em;
        }
        ```

florian-sanders-cc (changes requested):
Sorry, forgot to mention that we're missing changelog update :wink:

hsablonniere (approved):

florian-sanders-cc (approved):

PR #555, opened 2022-09-01
Rework types and cem plugin

### LINKED ISSUE #354: types.d.ts replace all `,` with `;`

### LINKED ISSUE #466: cem: refactor, tests and improve type definitions
## Context

Almost a year ago we introduced a plugin for cem analyser. 
The goal of the plugin was to be able to regroup types into one file for each folder of components and generate the related documentation. 
It allows us to: 
* not repeat interfaces
* not make typo and copy/paste the interfaces for components who has type in commons
* get some hints by our IDEs

While the plugin does its job there are still a few flaws:
* Some cases are not handled because we weren't thinking about them or having the need to use them at the time of initial dev. (e.g: if you use Array<type> the plugin will currently not find the subtypes associated if there are some while type[] would work and has been thought of)
* Although the code is commented and you can know easily what does what making new features for the plugin available is not ideal right now.

## What we want to do now

To make it more readable, simpler and make it easy to implement new "features" (use cases that are not covered by default) we do want to refactor the plugin and make some tests to make it "safer" and easy to maintain.
To be more compliant we will also use that time to go through all the current types/interfaces and make it more "typescript compliant".

## What's coming next?

If we do find cases that are not implemented we will just have to make some tests and implement it :wink:

Galimede:
fixes #354 (and types structure in new file organization)
fixes #466

## Context

A year ago we made a CEM plugin that generated the documentation of the types/interfaces of our components within `types.d.ts` files.
While it was great it had some flaws and bugs and was not that testable and tested.
Now that we are more on the team, we eventually tried some things that weren't thought of at the time of the first plugin iteration.
That's why we decided to make another version of the plugin that is more robust and that we can easily test and add new features to.
Moreover, with the new file organization we replaced all `,` with `,` in the interfaces as its standard TS and reorganized types.
Also, a basic script to check unused imports in the `@typedef ... ` imports has been made.

## So, what's the difference?

You might not see a huge difference using the plugin but there has been some changes behind the scenes like:

* There's now a log that tell you which component documentation is been generated. The purpose of adding a log is that if there's an error you will know which component is faulty and it will be easier to debug.
* If the file path is wrong, it will trigger an error but the rest of the components documentation will still be generated.
* If something is wrong within your `@typedef ...` imports the plugin will tell you that something is wrong.
* Code has been splitted into a lib and tested to ensure better readability and debugging.
* Comments on the plugin should help you understanding what's going on and especially when we're gathering something from an AST.
...

## What's next?

* With that new version it should be easier to implement new things that we might want. We'll need to discuss this later.
* The script that check `@typedef` imports right now is very basic and we could make some improvements and sort of features on it on a later date.


pdesoyres-cc (changes requested):
As usual, excellent job! :muscle: 

### Cycles are not detected when discovering subTypes

If you change the `SubInterface` like below, you'll get a stack overflow when running tests:
```typescript
interface SubInterface {
  parent: TheInterface;
}
```

### Unit Tests

The unit test file could be renamed `support-typedef-jsdoc-utils.test.js` as it is related to `support-typedef-jsdoc-utils.js`.

You could add some unit tests for scenarios where something goes wrong. For instance:

* You test 'should get the constructor node'. You could also test 'should return null when no constructor'
* You test 'should retrieve the types present in the constructor.'. You could also test 'should return an empty array when the constructor is empty'
* ...

Also, you can add some tests when Interfaces or types are malformed, when a typedef points to a missing file or missing type...

### Logs

We could dump more logs to console
* when aborting (for instance, when no constructor is found)
* during process
* at the end of the process (success)

Use the appropriate log level (info/warn/error)

Should we use chalk to make logs nicer?

### Misc

Should we consistently use `Foo[]` vs `Array<Foo>`?

package.json
```diff
@@ -22,8 +22,11 @@
     "components:build-cdn:versions-list": "node tasks/update-versions-list.js",
     "components:publish-cdn": "node tasks/publish-cdn.js",
     "components:check-i18n": "node tasks/check-i18n.js",
+    "components:check-type-imports": "node tasks/check-type-imports.js",
     "components:check-lit": "lit-analyzer src",
     "components:docs": "cem analyze --litelement --outdir dist",
+    "test-cem": "mocha cem/test/*.js",
+    "test-cem:watch": "mocha --watch cem/test/*.js",
```
roberttran-cc, line 29: Should we use `test:cem` and `test:cem:watch` to match the other tasks? WDYT? :thinking:

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+describe('findPath()', function () {
+
+  it('should retrieve the @typedef path from the first one in the test file.', function () {
+    const rootDir = process.cwd();
+    const moduleDir = 'cem/test/fixtures';
+    const importsNode = classNode.jsDoc[0].tags;
+    const path = findTypePath(importsNode[0], rootDir, moduleDir);
+    expect(path).to.equal(`${rootDir}/${moduleDir}/cc-test-component.types.d.ts`);
+  });
```
roberttran-cc, line 81: Could we use the `beforeEach` helper to factorize some code? (in this test and other ones)

roberttran-cc (commented):
Impressive work! I still have the main implementation to check in details but meanwhile, some minor suggestions. :smiley:

cem/support-typedef-jsdoc.js
```diff
@@ -1,186 +1,119 @@
...
+    imports.forEach((typesSet, filename) => {
+      let sourceCode;
+      let sourceAst;
+      const inMemoryType = fileCache.get(filename);
+      if (inMemoryType == null) {
+        try {
+          sourceCode = readFileSync(filename).toString();
         }
+        catch (e) {
+          console.log(e);
```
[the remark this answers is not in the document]
hsablonniere: Yes :+1:

cem/support-typedef-jsdoc.js
```diff
@@ -1,186 +1,119 @@
...
+            }
+            moduleTypeCache.get(typePath).add(type);
+          }
+        });
 
       // Now that we have the types, and the path of where the types are located
       // We can convert the imports to md types
-      const convertedImports = convertImports(ts, imports);
+      const convertedImports = convertImports(ts, moduleTypeCache);
       const displayText = (convertedImports) ? '### Type Definitions\n\n' + convertedImports : '';
```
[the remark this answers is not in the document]
hsablonniere: If we go for backticks, we will need `dedent` to have proper indentation.

cem/support-typedef-jsdoc.js
```diff
@@ -1,186 +1,119 @@
...
- * @returns {string}
- */
-function convertInterface (ts, node, code, interfaceName, filename) {
-  const st = node?.statements.find((st) => st?.name?.getText() === interfaceName);
-  if (st == null) {
-    return '';
-  }
+  const typesStore = new Map();
+  const fileCache = new Map();
+  const moduleTypeCache = new Map();
```
hsablonniere, line 14: Could you add 3 comments, one for each to explain what is their purpose?

cem/support-typedef-jsdoc-utils.js
```diff
@@ -0,0 +1,194 @@
...
+          subtypes.push(...handleTypeDeclaration(td.type, node, ts));
+          break;
+        }
+        default: {
+          subtypes.push(...handleInterface(td, node, ts));
+        }
+      }
+    });
+
+  return Array.from(new Set(subtypes));
```
hsablonniere, line 120: You could remove the `const fetch` and group the `forEach` after the `filter`.

cem/test/fixtures/cc-test-component.js
```diff
@@ -0,0 +1,48 @@
+import { LitElement } from 'lit-element';
+
+/**
+ * @typedef {import('./cc-test-component.types.js').Foo} Foo
+ * @typedef {import('./cc-test-component.types.js').Bar} Bar
+ * @typedef {import('./cc-test-component.types.js').TheInterface} TheInterface
+ * @typedef {import('./cc-test-component.types.js').TheType} TheType
+ * @typedef {import('./cc-test-component.types.js').TupleFoo} TupleFoo
+ * @typedef {import('./cc-test-component.types.js').TupleBar} TupleBar
+ * @tyedef {import('./cc-test-component.types.js').TupleBar} TupleBar
```
hsablonniere, line 10: If it's supposed to be ignored, I would use it with something else than `TupleBar` so we can be sure it's ignored.

cem/test/common.types.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Common {
+
+}
```
hsablonniere, line 3: This is also a fixture, we should put it in the fixtures folder and add another folder in it for the test component.

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+import { expect } from 'chai';
+import ts from 'typescript';
+import {
+  convertInterface,
+  findCustomType, findSubtypes, findTypePath, getConstructorNode, getTypesFromConstructor,
+} from '../support-typedef-jsdoc-utils.js';
+
+const filename = 'cem/test/fixtures/cc-test-component.js';
+const sourceCode = fs.readFileSync(filename, { encoding: 'utf-8' });
+// console.log(sourceCode);
```
hsablonniere, line 11: We can remove this I guess.

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+const sourceAst = ts.createSourceFile(filename, sourceCode, ts.ScriptTarget.ES2015, true);
+const classNode = sourceAst.statements.find((node) => node.kind === ts.SyntaxKind.ClassDeclaration);
+
+describe('getConstructorNode()', function () {
+  it('should get the constructor node.', function () {
+    const constructorNode = getConstructorNode(classNode, ts);
+    expect(constructorNode.kind).to.equal(ts.SyntaxKind.Constructor);
+  });
```
hsablonniere, line 19: We need another test with a file that doesn't have a constructor to check that it's ignored properly.

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+});
+
+describe('getTypesFromConstructor()', function () {
+  const constructorNode = getConstructorNode(classNode, ts);
+
+  it('should retrieve the types present in the constructor.', function () {
+    const types = getTypesFromConstructor(constructorNode, ts);
+    expect(types).to.have.members(['Foo', 'Bar', 'TheInterface', 'TheType', 'TupleFoo', 'TupleBar']);
+  });
```
hsablonniere, line 28: Could we add another test with another file that only use one or two types from the file.

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+  it('should return `Type` for a `Array<Type>` type.', function () {
+    testCustomType('Array<Foo>', 'Foo');
+  });
+
+  it('should return `Type` for a non primitive type.', function () {
+    testCustomType('Foo', 'Foo');
+  });
+
+  // TODO: I have seen one of the member of the team uses this pattern. We might include it afterwards.
+  it('should...', function () {
```
hsablonniere, line 68: You can name this `should ignore custom types in generics` for now.

cem/test/support-typedef-jsdoc-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+  it('should return `Type` for a `Type[]` type.', function () {
+    testCustomType('Foo[]', 'Foo');
+  });
+
+  it('should return `null` for a `primitive[]` type.', function () {
+    testCustomType('string[]', null);
+  });
+
+  it('should return `null` for a `Array<primitive>` type.', function () {
```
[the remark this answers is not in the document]
hsablonniere: No it means with `Array<number>` we don't need to retrieve any custom type information for the documentation.

tasks/check-type-imports.js
```diff
@@ -0,0 +1,96 @@
...
+    console.log(chalk.bgHex('#098846')(` 🎉 No type imports in the components are unused! `));
+    return;
+  }
+
+  unusedImports.forEach((unusedImport) => {
+    console.log('🔍', unusedImport.component, '-', unusedImport.filepath, '\n');
+    unusedImport.errors.forEach((error) => console.log(chalk.red(`❌ Something might be wrong for @typedef: `), error));
+    console.log(chalk.hex('#c15807')(`\n ⚠️  Unused import(s) are: `));
+    unusedImport.unusedImports.forEach((importation) => console.log(`\t - ${importation}\n`));
+  });
```
hsablonniere, line 93: Errors should trigger a `process.exit(1)`.

src/components/cc-tcp-redirection/cc-tcp-redirection.js
```diff
@@ -11,7 +11,6 @@ import { waitingStyles } from '../../styles/waiting.js';
 const noRedirectionSvg = new URL('../../assets/redirection-off.svg', import.meta.url).href;
 const redirectionSvg = new URL('../../assets/redirection-on.svg', import.meta.url).href;
 
-
```
hsablonniere, line 18: Can you rollback this.

tasks/check-type-imports.js
```diff
@@ -0,0 +1,96 @@
...
+import { getConstructorNode, getTypesFromConstructor } from '../cem/support-typedef-jsdoc-utils.js';
+
+function glob (patternList, options) {
+  if (typeof patternList === 'string') {
+    return monoglob.sync(patternList, options);
+  }
+  return patternList.flatMap((pattern) => {
+    return monoglob.sync(pattern, options);
+  });
+}
```
hsablonniere, line 16: It seems like you're not using the muti version anymore, can we simplify this?

hsablonniere (changes requested):
@Galimede This is an impressive work, congrats :clap:

## Commits

I would rename the commits like this:

* refactor: rework types files following new project file structure
* cem: update CEM support-typedef-jsdoc plugin

(you can add a TODO in the PR and do this at the end)

## Types file

My mind went back and forth but in the end, I think we should remove all useless `.types.d.ts` files.

## Imports for private types

I'm skeptical about the "we don't explicitly import the types used in private properties".
I know it would add some work but I think we should add them again and adapt the plugin and the task before merging.

## Unused types in `.d.ts`

I may have missed a few but I saw a few types in `.d.ts` files that were not used by components:

* cc-addon-admin
  * the component import Addon from the common, yet there's a Addon in the component types file
* cc-addon-elasticsearch-options
  * the component only imports and uses ElasticOptions, yet the component types have a Option, Flavor and 
* cc-addon-jenkins-options
  * component types has types but they aren't used
* cc-addon-mongodb-options
  * component types has types but they aren't used
* cc-addon-option-form
  * component types has types but they aren't used
* cc-addon-postgresql-options
  * component types has types but they aren't used
* cc-env-var*
  * We have duplication between parser options, could we group them?
* cc-tcp-redirection
  * RedirectionNamespace and Redirection aren't imported but used in events
* cc-overview
  * the Mode is not the right one
  
## Missing dev deps in package.json

* Your task for imports uses typescript yet it's not explicitly in the package.json
  * It works because we have TS as a transitive dep but we must add it (please use the same version as the one used by CEM analyzer)
* You're using mocha, it should be in the package.json

## CI

* We should add the test-cem in the prepack task
* We should add the the check unused imports too

## Triggering errors

Your plugin and task should use `console.error` and stop the process with a `process.exit(1)`.

## Chack

I think we should do like we did on the i18n and rely on the palette instead of custom colors.

## Misc

* :+1: for @pdesoyres-cc feedbacks

## Future features (we need to create issues for those)

* We could need a task to check for missing imports
  * maybe running typescript would give us this information
* Add support for types in events
  * this may change with the event refactor so let's wait for now
* Add support for imports of custom types in type definitions
* As Pierre said, we should detect import cycles

cem/support-typedef-jsdoc.js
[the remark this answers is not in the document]
Galimede: Although it's quite lightweight, I do prefer not adding another dependency just for this case.
Imo we better leave it as it is.

package.json
Galimede: Seems fair :wink:

cem/support-typedef-jsdoc.js
```diff
@@ -1,186 +1,119 @@
...
-            const typeToTs = path.format({ name: typeName, ext: '.d.ts' });
-            const typePath = path.resolve(rootDir, moduleDir, typeDir, typeToTs);
+      typeDefNode?.tags
+        ?.forEach((tag) => {
 
-            const typeDefDisplay = tag.name.getText();
-            const type = types.find((type) => type === typeDefDisplay);
+          // Detect if an import is not correct and warn the plugin user.
+          if (tag.kind !== ts.SyntaxKind.JSDocTypedefTag) {
+            console.log('Something in your @typedef imports is wrong.');
```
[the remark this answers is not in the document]
Galimede: You can know what component is faulty because the plugin tells you before that phase which component documentation is being generated.
Nonetheless we should be able to print the faulty line as we do in the script that checks type imports. :wink:

src/components/cc-tile-scalability/cc-tile-scalability.types.d.ts
```diff
@@ -1,52 +1,7 @@
-export interface Consumption {
-  yesterday: number,
-  last30Days: number,
+interface Scalability {
+  minFlavor: Flavor;
+  maxFlavor: Flavor;
```
florian-sanders-cc, line 3: `Flavor` does not exist in this file. Should we import it from common or move all this to common ?

src/components/cc-toaster/cc-toaster.js
```diff
@@ -6,9 +6,7 @@ import '../cc-toast/cc-toast.js';
 /**
  * @typedef {import('./cc-toaster.types.js').ToastPosition} ToastPosition
  * @typedef {import('./cc-toaster.types.js').ToastAnimation} ToastAnimation
- * @typedef {import('./cc-toaster.types.js').Notification} Notification
```
florian-sanders-cc, line 9: this type is used in the JS Doc for the `show` method. Should probably keep the import to let the IDE do its magic.

src/components/cc-toaster/cc-toaster.js
```diff
@@ -6,9 +6,7 @@ import '../cc-toast/cc-toast.js';
...
  * @typedef {import('./cc-toaster.types.js').ToastOptions} ToastOptions
- * @typedef {import('./cc-toaster.types.js').Toast} Toast
```
florian-sanders-cc, line 11: this type is used in the JS Doc for the `createToast` / `renderToast` methods + for an internal property. Should probably keep the import to let the IDE do its magic.

florian-sanders-cc (changes requested):
:clap: Gg @Galimede this is impressive and it's great to dive into this subject (although it's not the easiest to dive into :exploding_head: ). 

Most of my comments are related to manual changes in the type files. Some of them may be irrelevant because I remember you saying we could remove some type imports without breaking anything but I'm not sure about this. You can ignore the ones you know are safe :wink:

tasks/check-type-imports.js
Galimede: I don't understand what you mean by multi ? :thinking:

Galimede:
Hello :wave:,

I've taken into consideration most of your feedbacks in the relevant commits from f9e9041to the latest.
You will also find the new feature that concerns the extends and import features in the last commit.

## Things that changed
* Got rid of the type files that are unused. Don't hesitate to tell me if I have missed some.
* Updated the task 
  * Takes into account private properties in the constructor.
  * Triggers a process.exit() on success or error for the CI
* Updated needed deps and added the task to the prepack
* Plugin now supports imports and extends :tada: 
* Updated plugins comments
* Added more tests

However they are still some things overall I'm not really sure about:
* Triggering process.exit() on the cem plugin is not a good idea imo, you should still be able to generate the rest of the documentation even in one of them fail, you will see it anyway in the logs.
  * We might want to better log them and make the errors more visible on the plugin in the future.
* I've had to put a:  `// eslint-disable-next-line no-unused-expressions` at some points in the tests because I was using the expect function with `.to.be.null` which is well unused for eslint. Do you think that we should keep this line or deactivate this rule for the cem tests?
* @hsablonniere I still don't understand what you mean by multi version in the task :thinking: 

## Still left to do
* change commit order
* Take into account your new feedbacks if you have some :wink:

pdesoyres-cc (changes requested):
GG :clap:

roberttran-cc (commented):
Nice work! Globally LGTM.

Only 2 feedback which may or may not lead to modification. :smiley:

florian-sanders-cc (approved):

cem/support-typedef-jsdoc-utils.js
```diff
@@ -0,0 +1,319 @@
+import { readFileSync } from 'fs';
+import path from 'path';
+
+export function getConstructorNode (classNode, ts) {
```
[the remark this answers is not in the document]
hsablonniere: @Galimede Can answer but it feels like the caller shouldn't really know how the TypeScript compiler will be used by the called function.

hsablonniere (approved):
> Triggering process.exit() on the cem plugin is not a good idea imo, you should still be able to generate the rest of the documentation even in one of them fail, you will see it anyway in the logs. 

Fine with me :+1: 

> I've had to put a: // eslint-disable-next-line no-unused-expressions at some points in the tests because I was using the expect function with .to.be.null which is well unused for eslint. Do you think that we should keep this line or deactivate this rule for the cem tests?

We can either try an eslint plugin as @pdesoyres-cc suggested or a chai plugin. I used this one a while ago: https://www.chaijs.com/plugins/dirty-chai/ It seems a bit old and may not work with our chai-esm.

Impressive work :clap: :+1: 
Well done!!

PR #567, opened 2022-09-14
Feat - Hide env value by default

jerethom:
Hide environment variable values by default.


hsablonniere:
Hi @jerethom thanks for your contribution but I'm sorry we cannot merge it.

We don't want all the env vars to be hidden as secrets. We considered only marking as secret some properties based on their name but haven't really made progress on whether it's a good idea. We're also considering asking for your password again to display the env vars but again, this needs to be discussed and we're focused on other topics right now.

PR #569, opened 2022-09-15
TO_DELETE cc-email: state & smart (tests Hubert)

hsablonniere:


pdesoyres-cc:
I really like the `updateComponent` helper. I don't feel like there is too much magic.

PR #570, opened 2022-09-16
icon system: init

### LINKED ISSUE #525: Icon system: init
# New icon system

The "why" of a new and improved icon system is discussed here #436 while this issue will host the "how", meaning everything related to implementation.

The several parts composing the new system are:

* an icon library;
* a component in charge of rendering an icon SVG;
* a component in charge of resolving a SVG content from an ID;
* scripts for wrapping SVG files in ES modules and checking if components import icons that are unused;
* development stack updates (Rollup and Web Dev Server) to deliver ES module icons on the fly.

## Choosing an icon library

We opted for [Remix Icon](https://remixicon.com/).

Note that the new icon system will be able to display legacy icons currently used accross our components.

### Context

This component will be in charge of rendering an SVG, allowing transformations trough properties and CSS custom properties.

### To be discussed

The list of inputs is not exhaustive.

Accessibility implementation needs validation from @florian-sanders-cc.

#### Attributes/properties

| Property | Type       | Default value | Mandatory | Description                                               |
| -------- | ---------- | ------------- | --------- | --------------------------------------------------------- |
| `size`   | `{string}` | `md`          | `false`   | icon size among `xs\|sm\|md\|lg\|xl`                      |
| `icon`   | `{string}` | `null`        | `false`   | icon ID from which the resolver will return a SVG content |

Note: the `size` property implies the component will always be a square.

#### Slots

| Name   | Description                            |
| ------ | -------------------------------------- |
| `null` | if valid SVG content, will be rendered |

#### CSS custom property

| Name      | Type           | Description              |
| --------- | -------------- | ------------------------ |
| `--size`  | `CSS <length>` | updates height and width |
| `--color` | `CSS <color>`  | updates SVG color        |

### Type Definitions

```typescript
interface IconWrapper {
  id: string;
  content: string;
}
```

### Data/APIs

TBD: in case no resolver is defined, the component may directly retrieves the requested icon from our CDN.

### Stories & states

Default (slotted icon).
Empty.
Waiting.
Icon property (aka with resolver).
Size property.

### Accessibility

Following [Lion icon component](https://lion-web.netlify.app/components/icon/use-cases/#accessibility):

- add `aria-label` and `aria-hidden` attributes that are exclusive;
- add a `role` attribute which defaults to `img`.

### Context

An ES module that allows to register icons (SVG wrapped in ES module) and return them, this through public methods and properties.

### Inputs

| Properties     | Type                   | Description         |
| -------------- | ---------------------- | ------------------- |
| `iconRegistry` | `{Map<string,object>}` | collection of icons |

| Methods      | Arguments         | Return          | Description                          |
| ------------ | ----------------- | --------------- | ------------------------------------ |
| `addIcon`    | `{IconWrapper}`   | `null`          | register a single icon               |
| `addIcons`   | `{IconWrapper[]}` | `null`          | register multiple icons              |
| `getIcon`    | `{string}`        | `{IconWrapper}` | get an icon from its ID              |
| `removeIcon` | `{string}`        | `null`          | remove a registered icon from its ID |

## Scripts

A script that parses all SVG files we want to wrap in ES module and write the output in one or multiple files. It should handle multiple sets of icons, like Remix vs legacy (TBD via a namespace or multiple maps).

A script that checks for duplicates icon IDs (maybe used in previous script).

A script that parses all components and warns when some of them import icon that are not used (cf. CEM analyzer plugin `list-images`).

## Development stack updates

A Rollup plugin that executes the previous script in development mode with live reload.

The plugin should work with Web Dev Server via `rollupAdapter()` from `@web/dev-server-rollup`.

## Misc

Add some tests for the component and the resolver.

TBD icon component error management, meaning when no proper SVG can be rendered.

## Documentation

Update `cc-example-component.js`.

Ensure Remix Icon copyright requirement are properly met.

Create a user guide to explain how to use the new icon system.

Create an ADR to explain our decision process (will be part of a different issue though).

## Migration

Finally, integrating the new icon system in components.

### Prior inventory

List all used icons with:

- size;
- color;
- caveat.

Suggest if a remix icon can replace a clever icon. Be careful of remarkable icons well-known by users (maybe part of a two-times migration).

### Clever Components updates

Some components will probably need substantial changes:

- `<cc-button hide-text image='path/to/svg.svg'>...</cc-button>`
- `<cc-toggle hide-text choices='[{"label":"...","value":"...","image":"..."}]' value="..."></cc-toggle>`

roberttran-cc:
Fixes #525.


pdesoyres-cc:
### HTML identifier

:+1: I'm ok with the current implementation. I don't think we need an additional prefix.
:+1: I'm ok with the colon character as separator.
:+1: I'm ok with the use of snake case for the icon identifier.

### ES modules

:+1: I'm ok with the current implementation. I don't think we need an additional prefix.
:+1: I'm ok with the use of param case.

### Plugin behaviour

:+1: I'm ok with the actual implementation. No need to have the category in the file name.

> Another implementation is having an unique file which exposes all sets (through a Map).

:bulb: We could also implement that by adding a `category` property to the icon.
We could also apply that to our own icons (maybe by using subdirectories too)

:thinking: It could help us to write a story where users can filter icons by category.

### Rendering

I think we have to distinguish two situations:

* error from the developer
  * The icon is not registered
  * => display a specific red icon. and log an error message to the console
* error at runtime
  * The icon could not be fetched (when using remote provider)
  * => display a transparent square (with the right size)

### code review

This code review may be a little bit anticipated so this is to be taken with "pincettes".

#### plugin

* `plugin-icon-clevercloud.js` and `plugin-icon-remix.js` are very similar. I think we could factorize them into one single impl
* `plugin-icon-remix.js` could apply the same SVG optimization plugins as the `plugin-icon-clevercloud.js` because we don't know if `remixicon` dependency will always provide optimized icons. (this could also help in the previous point)
* `rename rollup-plugin-icon.js` into `rollup-plugin-icons-assets.js` (same naming strategy as for `rollup-plugin-styles-assets.js`)
* add more logs
  * log number of generated icons
  * log process duration
  * use chalk to have pretty logs

#### resolver

* add JSDoc

#### component

* rename `iconId` => `id` or `name` (as we already are in the context of a `<cc-icon>` tag)
* rename `remoteProvider` => `remote`
* add type.d.ts for the size with all the possible values.
* should we allow developers to specify a specific size? (`15px`, `0.8em`, ...)
* I feel not very comfortable using `unsafeSVG` in a remote icon context. I'm not a specialist at all, but do we open a security breach if we do that?
* We should think about how we want to handle the case where the remote icon could not be fetched.
* Should we have a 'loading' state displayed until the remote icon is fetched?

* I'm wondering when developers would like to mix the usage of remote and non-remote in the same project.
As a developer, I want to choose a way to integrate icons in my app, but I'm not sure, I would like to mix two different technics.

Let's say, I want to use the module technic, then I would write:
```
// add all my icons once when my page loads
addIcons(cleverCompleteSet);

//...
// and then use them in my HTML
<cc-icon icon-id="clever:icon"></cc-icon>
```

Now, let's say, I want to use the remote technic, then I would write:
```
// I can use icons directly
<cc-icon icon-id="clever:icon" remote-provider></cc-icon>
```

In the above, I need to always specify `remote-provider` which I find a bit annoying.
I would prefer to say that I always want to use the remote provider once and for all, and then use the `<cc-icon>` without having to repeat the `remote-provider` attribute.

roberttran-cc:
Hello there, this PR is ready for review! :brain: 

## How to review

Pull checkout is the way: I think it's simpler to start from scratch.

## What has changed

A lot of things! :sweat_smile: 

### The component & Rollup plugin

Behavior has been reduced to focus on integration in the Clever Components, so no more resolver implementation (for now).

Meaning that we pass the icon directly to the `<cc-icon>` component, wrapped in an ES module (this point hasn't changed) and _voilà_.

You'll find this implementation in the first commit, alongside the Rollup plugin generating the ES module icons: this part has been refactored in a single file, should be easier to scan.

### Integration attempt on an existing component

Crash test dummy was `<cc-ssh-key-list>`. :eyes: 

The most impactful thing was usage in the `<cc-button>`.

First try was putting the new icon component in the button slot: that worked but it required to add some style in the parent component to tweak the layout.

So I went to test an update on `<cc-button>` where you pass the icon ES module to a new `icon` property and the button component handles the layout. You're left with only the icon size to specify in the parent component and you're done. Trade-off is that it adds some more complexity to the `<cc-button>` code.

You can see it in action in the second and third commits, all feedback are welcome!

## What is left to do

Test including `cc-icon` in `cc-button` via a named slot. Then decide what option we're taking on this topic.

Fix Rollup plugin launched in parallel to tests (and not consecutively).

Test with SSR.

Optionally live reload when SVG sources changes on dev mode.

Finally migrate all other Clever Components.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -8,17 +9,18 @@ import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { createRef, ref } from 'lit/directives/ref.js';
 import { repeat } from 'lit/directives/repeat.js';
+import {
+  IconRemixAddCircleFill as IconAdd,
+  IconRemixDeleteBin_5Fill as IconBin,
+  IconRemixKeyFill as IconKey,
+} from '../../assets/cc-remix.icons.js';
```
hsablonniere, line 16: Do you think we'll always alias them so it's easier to understand in the template like this?

I think I like that but we should update the example to reflect this.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -306,7 +308,13 @@ export class CcSshKeyList extends LitElement {
...
           return html`
             <div class="key ${classMap(classes)}">
-              <cc-img class="key__icon" src=${!skeleton ? keySvg : ''} ?skeleton=${skeleton}></cc-img>
+              <div class="key__icon">
+              ${
+                skeleton
+                ? html`<div class="skeleton"></div>`
+                : html`<cc-icon .icon="${IconKey}" size="lg"></cc-icon>`
+              }
+              </div>
```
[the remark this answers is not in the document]
hsablonniere: At first, I was like @pdesoyres-cc.

Then I was "in this situation, why don't we always display the icon since it's not an icon where we don't know what to display?

Then I was "we'll still need skeleton in some situations".

My conclusion is:

* maybe we should always use the icon (no skeleton) in this very example
* maybe we should also add a skeleton mode on cc-icon

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -441,6 +453,18 @@ export class CcSshKeyList extends LitElement {
...
+          --cc-icon-size: 18px;
+        }
+
+        .key__button--github {
+          --cc-icon-size: 20px;
+        }
+
+        .key__button--skeleton {
+          --cc-icon-size: 20px;
+        }
```
hsablonniere, line 467: I think this should be in `em`.

hsablonniere (changes requested):
Hey Bob, so nice to see this feature almost ready :heart_eyes: congrats :clap:

I wrote a but feedbacks.

* What did we decided for the existing icons? When do we migrate them to the new system?
  * Now in this PR once we agree on the basics
  * :ballot_box_with_check: Later in another PR
* I think some of the Clever icons are based on stoke and not fill and I think the custom color won't work
  * How do we handle this?
  * Are all remix icons based on fill?
  
> Test including cc-icon in cc-button via a named slot. Then decide what option we're taking on this topic.

I tried a slot for cc-icon and using the SVG string directly but not the named slot for cc-button. I think we should try it, at least to describe the downsides in the ADR (yes we need and ADR :p).

Using a slot for cc-icon would mean this:

```js
import { IconFoo } from '../some-icons.js';
import { unsafeSVG } from 'lit';

function render () {
  return html`
    <div>
      <cc-icon>${unsafeSVG(IconFoo)}</cc-icon>
      <span>Hello</span>
    </div>
  `;
}

All components would have to use the `unsafeSVG` directive, this is ugly and it prevents us from changing the implementation under the hood.

I'll help you with the ADR to bring other details and downsides of some ideas we had or some we have for the future :wink:

File: src/components/cc-ssh-key-list/cc-ssh-key-list.js (hunk 2 above)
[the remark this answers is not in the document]
Comment (roberttran-cc) [reply src/components/cc-ssh-key-list/cc-ssh-key-list.js:317]: Let's create a skeleton state for `cc-icon` and see how it goes!

File: src/components/cc-ssh-key-list/cc-ssh-key-list.js (hunk 1 above)
Comment (roberttran-cc) [reply src/components/cc-ssh-key-list/cc-ssh-key-list.js:16]: Self-memo: update how we import icons and alias usage in `cc-example-component.js`

File: src/components/cc-icon/cc-icon.js [hunk 4]
```diff
@@ -0,0 +1,109 @@
...
+import { unsafeSVG } from 'lit/directives/unsafe-svg.js';
+
+/**
+ * @typedef {import('./cc-icon.types.js').IconModel} IconModel
+ * @typedef {import('./cc-icon.types.js').IconSize} IconSize
+ */
+
+/**
+ * A component rendering an SVG icon.
+ *
```
florian-sanders-cc, line 11: I think we could introduce an `Accessibility guidelines` section because `<svg>` a11y is a tricky subject and we want to offer some guidance on how to use the `accessible-name` prop.

image


```suggestion
 * **Accessibility guidelines:**
 *
 * Use the `accessible-name` prop only if your icon provides information that is not already given in its surrounding text.
 *
 * If this prop has a value:
 *
 *  * sets a `role="img"` and `aria-label="the value of this prop"` attributes on the `<svg>` element.
 *  * creates a `<title>` element inside the `<svg>`.
 *
 * This allows assistive technologies to recognize the element as an image and convey its information.
 *
 * If this prop has no value, sets an `aria-hidden="true"` attribute on the `<svg>` element so that it can be ignored by assistive technologies.
```

You'll see below that I also suggest we add this info in the `accessible-name` prop doc.It might seem redundant but I think both the main doc and the prop doc should mention this because if you don't know about accessibility, you might not even check the `accessible-name` prop :roll_eyes:.

Feel free to rephrase any of these of course :wink:

src/components/cc-icon/cc-icon.js
```diff
@@ -0,0 +1,109 @@
...
+      accessibleName: { type: String, attribute: 'accessible-name' },
+      icon: { type: Object },
+      size: { type: String, reflect: true },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {string|null} Forces the values of the `aria-hidden`, `aria-label` and `role` attributes on the `svg` element. Also add or removes a `<title>` element in the SVG. */
```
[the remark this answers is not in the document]
florian-sanders-cc: I think we could make an exception to how we typically write our prop docs here to provide more guidance and a warning about this prop usage:

image
```suggestion
    /** @type {string|null} Only use this prop if your icon provides information that is not already given in its surrounding text.
     *
     * If this prop has a value:
     *
     *  * sets a `role="img"` and `aria-label="the value of this prop"` attributes on the `<svg>` element.
     *  * creates a `<title>` element inside the `<svg>`.
     *
     * This allows assistive technologies to recognize the element as an image and convey its information.
     *
     * If this prop has no value, sets an `aria-hidden="true"` attribute on the `<svg>` element so that it can be ignored by assistive technologies.
     */
```

florian-sanders-cc (changes requested):
Gg @roberttran-cc, this was not an easy subject but the code is easy to read and well documented :clap: 

I can't wait to use this inside all components.

This is going to help SO MUCH for dark mode :heart: and design consistency in general.

I left a few comments with suggestions about a11y documentation mainly :wink:

src/components/cc-icon/cc-icon.js
```diff
@@ -70,8 +77,11 @@ export class CcIcon extends LitElement {
 
   static get styles () {
     return [
+      skeletonStyles,
       css`
         :host {
+          --cc-skeleton-color: #bbb;
```
hsablonniere, line 83: We're trying to apply this pattern and maybe it should be documented somewhere :+1: 

1. private CSS variable and only used in a component => no `cc-` prefix
2. public CSS variable specific to the component and which can be configured by the user => `cc-foobar-something` prefix `foobar` being the component name
3. public CSS variable global to multiple components => `cc-something` prefix, those mostly come from the theme

We still don't have a variable and proper system for skeleton :-(

I think we should either document the skeleton as a (2) variable or go private (1) and even inline `#bbb` in the only place it's used for now while we wait for a better skeleton system.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -373,6 +369,8 @@ export class CcSshKeyList extends LitElement {
         /* region global */
 
         :host {
+          --cc-skeleton-color: #bbb;
```
hsablonniere, line 372: Same remark for this CSS variable as above.

hsablonniere (changes requested):
Super job @roberttran-cc, really looking forward to land this!!

* Nice test suite, simple, efficient
* Thanks for updating the example
* I love the new icon galleries, so nice
* Nice touch of adding the icon plugin for the Storybook static build

pdesoyres-cc (approved):
Well done!

hsablonniere (approved):

Galimede (approved):

florian-sanders-cc (approved):

roberttran-cc:
All previous commits (initial one and feedback) have been squashed in three commits:
- icon system initialization
- `cc-button` new `icon` property
- implementation in `cc-ssh-key-list`

A few reminders though, that may trigger some last minute updates:
- there is no fail-safe with the new `icon` property in `cc-button` : `image` property is still present and can be used alongside, which will break the display (icon and image will be superposed)
- the "key" Remix icon used for the `cc-ssh-key-list` implementation is quite different than the current one (check the new one [here](https://clever-components-preview.cellar-c2.services.clever-cloud.com/icon-system/init/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story))
- SRR has been successfully tested, though I did not manage to make the VM module scenario work, only the global scope one (check [here](https://lit.dev/docs/ssr/server-usage/) to look what it is all about)
- ADR will be treated in a separate issue
- components migration will be treated in a separate issue as well

pdesoyres-cc:
> there is no fail-safe with the new icon property in cc-button : image property is still present and can be used alongside, which will break the display (icon and image will be superposed)

I would say that we should deprecate now the image property and we need another issue for removing it.

> the "key" Remix icon used for the cc-ssh-key-list implementation is quite different than the current one

I'm ok with this one too. It looks like a key!

> SRR has been successfully tested, though I did not manage to make the VM module scenario work, only the global scope one

Do you mean that you did not have the cc-icon working, or you did not make any of our components work with the VM module? If the latter, I would say that we can make an issue for investigation.

hsablonniere (changes requested):
> there is no fail-safe with the new icon property in cc-button : image property is still present and can be used alongside, which will break the display (icon and image will be superposed)

Like Pierre, I think we should create an issue to deprecate this, or at least discuss the deprecation.

> the "key" Remix icon used for the cc-ssh-key-list implementation is quite different than the current one (check the new one [here](https://clever-components-preview.cellar-c2.services.clever-cloud.com/icon-system/init/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story))

I prefer the other one but I like the let's go full remix if possible and we'll reconsider once we use more icons.

> SRR has been successfully tested, though I did not manage to make the VM module scenario work, only the global scope one (check [here](https://lit.dev/docs/ssr/server-usage/) to look what it is all about)

Nice!

roberttran-cc:
> Do you mean that you did not have the cc-icon working, or you did not make any of our components work with the VM module? If the latter, I would say that we can make an issue for investigation.

The latter! But as a matter of fact, I only tested `cc-icon`: I'll test with other components soon. And a little reminder that the SSR package is still experimental and have some **[limitations](https://lit.dev/docs/ssr/overview/#library-status)**.

> Like Pierre, I think we should create an issue to deprecate this, or at least discuss the deprecation.

Agree! I'll submit an issue.

:arrow_right: Btw, unless contraindicated, I'll add a mention to Remix Icon (although not mandatory, see their [license](https://github.com/Remix-Design/remixicon#license)) in our [readme](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%8F%A0-home-readme--page) ("License" section).

.storybook/main.js
```diff
@@ -3,6 +3,11 @@ const json = require('@rollup/plugin-json');
 const { importMetaAssets } = require('@web/rollup-plugin-import-meta-assets');
 const { storybookRollupPlugin } = require('../src/stories/lib/markdown.cjs');
 
+let createIconAssetsPlugin;
+import('../rollup/rollup-plugin-icon.js').then(module => {
+  createIconAssetsPlugin = module.createIconAssetsPlugin;
+});
```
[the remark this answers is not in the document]
roberttran-cc: `rollup/rollup-plugin-icon.js` is an ES module file whereas `.storybook/main.js` is a CommonJS one: as ESM is asynchronous and CJS synchronous, I think it's the only way to combine both of them (without renaming files or change `package.json`'s type).

But I'm really not sure and open to any suggestion! :hugs:

hsablonniere:
> arrow_right Btw, unless contraindicated, I'll add a mention to Remix Icon (although not mandatory, see their [license](https://github.com/Remix-Design/remixicon#license)) in our [readme](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%8F%A0-home-readme--page) ("License" section).

Good catch!

.storybook/main.js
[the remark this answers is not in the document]
hsablonniere: Grrr, I hate this nonsense (not your fault lol).

Technically the code looks like the `createIconAssetsPlugin` could be undefined when the plugin is run but I guess this will never happen. I guess this is why I create a `.cjs` module for markdown.

Let's keep it like this :+1:

hsablonniere (approved):

PR #572, opened 2022-09-21
refactor: rollback the smart-manager to the old low level API and move the observable API to a different module

hsablonniere:
Hey folks,

We're going to deprecate (and then remove) the current `defineComponent` function from the smart manager which is based on Observables and RxJS.

In this PR:

* add some comments (just the first step)
* rollback the `defineComponent` function to the previous low level API with 3 callbacks `onConnect`, `onContextUpdate` and `onDisconnect`
* reimplement the current `defineComponent` function (based on just one `onConnect` callback and a `context$` observable) with the "new" low level API in a different file `smart-manager-observables.js`
* cleanup the test suite
  * BTW the test suite was kinda still testing the low level API with a dirty hack, it's better now

Next steps

* Define a new high level API (not based on observables) with a bit more magic
* Write some more comments, JSDoc and documentation

### DIFF
--- a/src/lib/smart-manager.js
+++ b/src/lib/smart-manager.js
@@ -1,6 +1,19 @@
 import { Subject } from './observables.js';
 import { objectEquals } from './utils.js';
 
+// In the global space of this module (for any module importing it), we maintain:
+// * a rootContext object
+// * a smartContainers Set with all containers currently in the page
+// * a smartComponentDefinitions Set with all definitions currently in the page
+
+// On each container, we maintain:
+// * a Map on container[COMPONENTS] with all components keyed by definition object
+// * an object with the current context on container[CURRENT_CONTEXT]
+
+// On each component, we maintain:
+// * a map on component[META] with pointers to context$, disconnectionController keyed by definition object
+// * an object with the last context on component[LAST_CONTEXT]
+
 /**
  * @typedef SmartComponentDefinition
  * @property {String} selector
--- a/src/lib/smart-manager.js
+++ b/src/lib/smart-manager.js
@@ -1,4 +1,3 @@
-import { Subject } from './observables.js';
 import { objectEquals } from './utils.js';
 
 // In the global space of this module (for any module importing it), we maintain:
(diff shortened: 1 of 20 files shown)


src/lib/smart-manager.js
```diff
@@ -148,17 +148,8 @@ export function updateRootContext (context) {
...
-  const disconnectionController = new AbortController();
-
-  if (component[META] == null) {
-    component[META] = new Map();
-  }
-  component[META].set(definition, { context$, disconnectionController });
   component[LAST_CONTEXT] = {};
-
-  definition.onConnect(container, component, context$, disconnectionController.signal);
+  definition.onConnect?.(container, component);
```
florian-sanders-cc, line 152: I've never used the optional chaining for this kind of case so I might have missed the point but it feels like  our 3 lifecycle hooks (`onConnect`, etc., not sure how to call these) are all mandatory for our low level smart manager to work properly.

If so, I would avoid using the optional chaining and let it fail if we forget to provide one of the hooks in our smart manager observables / magic implementations.
This way it fails with an explicit error instead of silently not working ?

package.json
```diff
@@ -28,7 +28,7 @@
     "test:watch": "wtr 'test/**/*.test.*' --node-resolve --watch",
     "lint": "eslint --rulesdir eslint-rules rollup/*.js cem src tasks test wds",
     "lint:fix": "eslint --rulesdir eslint-rules --fix rollup/*.js cem src tasks test wds",
-    "prepack": "npm run test && npm run lint && npm run components:check-lit && npm run components:check-i18n && npm run components:build && npm run components:docs",
+    "prepack": "npm run test && npm run lint && npm run components:check-lit && npm run components:check-i18n",
```
florian-sanders-cc: this one feels weird but I think you said you needed to do something about this yesterday. Just leaving a comment so you don't forget.

florian-sanders-cc (changes requested):

package.json
hsablonniere: I should not have committed this. Thanks.

src/lib/smart-manager.js
```diff
@@ -1,6 +1,19 @@
 import { Subject } from './observables.js';
 import { objectEquals } from './utils.js';
 
+// In the global space of this module (for any module importing it), we maintain:
+// * a rootContext object
+// * a smartContainers Set with all containers currenlty in the page
```
pdesoyres-cc, line 5: typo on currently

src/lib/smart-manager.js
```diff
@@ -1,6 +1,19 @@
...
+// * a smartComponentDefinitions Set with all definitions currently in the page
+
+// On each container, we maintain:
+// * a Map on container[COMPONENTS] with all components keyed by definition object
+// * an object with the current context on container[CURRENT_CONTEXT]
+
+// On each component, we maintain:
+// * a map on component[META] with pointers to context$, disconnectionController keyed by definition object
```
pdesoyres-cc, line 13: this one needs to be removed for the new implementation (on the next commit)

pdesoyres-cc (changes requested):

src/lib/smart-manager-observables.js
```diff
@@ -0,0 +1,42 @@
...
+
+      const context$ = new Subject();
+      const disconnectionController = new AbortController();
+
+      if (component[META] == null) {
+        component[META] = new Map();
+      }
+      component[META].set(definition, { context$, disconnectionController });
+
+      definition.onConnect(container, component, context$, disconnectionController);
```
pdesoyres-cc, line 32: oups. here you should pass `disconnectionController.signal` instead of `disconnectionController`

hsablonniere:
@florian-sanders-cc @pdesoyres-cc Thanks for your feedbacks. I fixed your remarks but I already did it in the respective commits, sorry, I should have pushed additional fixup commits.

pdesoyres-cc (approved):

PR #573, opened 2022-09-22
Add a new high level / magic API to define a smart component

hsablonniere:
* This is still WIP
* Feedbacks are welcomed
* You can rebase your components over this

### DIFF
--- a/src/lib/define-smart-component.js
+++ b/src/lib/define-smart-component.js
@@ -1,0 +1,104 @@
+import { produce } from './immer.js';
+import { defineSmartComponentCore } from './smart-manager.js';
+
+const META = Symbol('META');
+
+/**
+ * @typedef {Element} SmartContainer
+ * @param {object} context
+ */
+
+/**
+ * @typedef {object|((property: object) => void)} CallbackOrObject
+ */
+
+/**
+ * @param {object} definition
+ * @param {string} definition.selector
+ * @param {object} definition.params
+ * @param {({
+ *   container?: SmartContainer,
+ *   component?: Element,
+ *   context?: object,
+ *   onEvent?: (type: string, listener: (detail: object) => void) => void,
+ *   updateComponent?: (
+ *     type: string,
+ *     callbackOrObject: CallbackOrObject,
+ *   ) => void,
+ *   signal?: AbortSignal,
+ * }) => void} definition.onContextUpdate
+ * @param {AbortSignal?} signal
+ */
+export function defineSmartComponent (definition, signal) {
+
+  defineSmartComponentCore({
+    selector: definition.selector,
+    params: definition.params,
+    onConnect (container, component) {
+      if (component[META] == null) {
+        component[META] = new Map();
+      }
+      // Prepare a metadata object specific to this component and this definition
+      component[META].set(definition, {});
+    },
+    onContextUpdate (container, component, context) {
+
+      // Don't trigger the high level onContextUpdate if one of the params is null
+      const someContextParamsAreNull = Object
+        .keys(definition.params)
+        .some((name) => context[name] == null);
+      if (someContextParamsAreNull) {
+        return;
+      }
+
+      // Get the metadata object specific to this component and this definition
+      const meta = component[META].get(definition);
+
+      // Abort controller linked to the signal sent to the last onContextUpdate (if any)
+      meta.abortController?.abort();
+      // Prepare a new controller triggered when the context is updated or the component disconnected
+      meta.abortController = new AbortController();
+
+      const signal = meta.abortController.signal;
+
+      // Prepare a helper function to attach event listeners on the component
+      // and make sure they're removed if the signal is aborted
+      function onEvent (type, listener) {
+        component.addEventListener(type, (e) => {
+          listener(e.detail);
+        }, { signal });
+      }
+
+      // Prepare a helper function to update a component (via immer if necessary)
+      // We use an even target as an indirection so we can
+      // * ask for updates via un event
+      // * listen for the event to apply the update
+      // * remove the listener and don't apply the update if the signal is aborted
+      const target = new EventTarget();
+
+      target.addEventListener('update-component', (event) => {
+        const { property, callbackOrObject } = event.data;
+        if (typeof callbackOrObject === 'function') {
+          if (component[property] != null) {
+            component[property] = produce(component[property], callbackOrObject);
+          }
+        }
+        else {
+          component[property] = callbackOrObject;
+        }
+      }, { signal });
+
+      function updateComponent (property, callbackOrObject) {
+        const event = new Event('update-component');
+        event.data = { property, callbackOrObject };
+        target.dispatchEvent(event);
+      }
+
+      definition.onContextUpdate({ container, component, context, onEvent, updateComponent, signal });
+    },
+    onDisconnect (container, component) {
+      component[META].get(definition).abortController.abort();
+      component[META].delete(definition);
+    },
+  });
+}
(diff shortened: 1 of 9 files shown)


src/lib/smart-manager-foobar.js
```diff
@@ -30,34 +30,41 @@ export function defineSmartComponent (definition, signal) {
...
       function onEvent (type, listener) {
         component.addEventListener(type, (e) => {
           listener(e.detail);
         }, { signal });
       }
 
-      // ---
+      // Prepare a helper function to update a component (via immer if necessary)
+      // We use an even target as an indirection so we can
+      // * as for updates via un event
```
pdesoyres-cc, line 59: typo: `as` => `ask`

web-dev-server.config.js
```diff
@@ -21,7 +21,8 @@ const hmrI18nPlugin = {
     if (TRANSLATION_FILE_REGEX.test(context.path)) {
       // language=JavaScript
       context.body += `
-        ;;
+        ;
+        ;
```
pdesoyres-cc, line 25: what is that for?

pdesoyres-cc (changes requested):

src/lib/smart-manager-foobar.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @typedef SmartComponentDefinitionMagic
+ * @property {String} selector
+ * @property {Object} params
+ * @property {Function} onContextUpdate
+ */
+
+/**
+ * @param {SmartComponentDefinition} definition
```
pdesoyres-cc, line 14: should be `SmartComponentDefinitionMagic`

src/lib/smart-manager-foobar.js
```diff
@@ -0,0 +1,86 @@
+import { produce } from './immer.js';
+import { defineComponent } from './smart-manager.js';
+
+const META = Symbol('META');
+
+/**
+ * @typedef SmartComponentDefinitionMagic
+ * @property {String} selector
+ * @property {Object} params
+ * @property {Function} onContextUpdate
```
pdesoyres-cc, line 10: We should add the @callback definition

pdesoyres-cc (changes requested):

src/lib/smart-manager-foobar.js
```diff
@@ -0,0 +1,86 @@
...
+
+      // ---
+      const target = new EventTarget();
+
+      target.addEventListener('update-component', (event) => {
+        const { property, callbackOrObject } = event.data;
+        if (typeof callbackOrObject === 'function') {
+          component[property] = produce(component[property], callbackOrObject);
+        }
+        else if (typeof callbackOrObject === 'object') {
```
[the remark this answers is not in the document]
hsablonniere: Good catch.

src/lib/smart-manager-foobar.js
hsablonniere: Yes, I need to improve the whole JSDoc.

web-dev-server.config.js
hsablonniere: I rolled it back. I can explain later what the existing code does.

src/lib/define-smart-component.js
```diff
@@ -0,0 +1,98 @@
...
+      // We use an even target as an indirection so we can
+      // * ask for updates via un event
+      // * listen for the event to apply the update
+      // * remove the listener and don't apply the update if the signal is aborted
+      const target = new EventTarget();
+
+      target.addEventListener('update-component', (event) => {
+        const { property, callbackOrObject } = event.data;
+        if (typeof callbackOrObject === 'function') {
+          component[property] = produce(component[property], callbackOrObject);
```
pdesoyres-cc, line 78: `produce(null, callback)` produces a cryptic error.
Should we check if `component[property]` is `null` before using `produce` function ?

src/lib/define-smart-component.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @typedef {Element} SmartContainer
+ * @param {object} context
+ */
+
+/**
+ * @param {object} definition
+ * @param {string} definition.selector
+ * @param {object} definition.params
+ * @param {({
```
pdesoyres-cc, line 15: When using:

```javascript
defineSmartComponent({
  onContextUpdate ({ updateComponent }) {
  },
  selector: 'selector',
  params: {},
}, null);
```

WebStorm is complaining because the `onContextUpdate` is not destructuring all the properties.
Making the properties optional in JSDoc makes WebStorm happy:

```javascript
 * @param {({
 *   container?: SmartContainer,
 *   component?: Element,
 *   context?: object,
 *   onEvent?: (type: string, listener: (detail: object) => void) => void,
 *   updateComponent?: (
 *     type: string,
 *     callbackOrObject: ((property: object) => void),
 *   ) => void,
 *   signal?: AbortSignal,
 * }) => void} definition.onContextUpdate

```

src/lib/define-smart-component.js
```diff
@@ -0,0 +1,98 @@
...
+ *   component: Element,
+ *   context: object,
+ *   onEvent: (type: string, listener: (detail: object) => void) => void,
+ *   updateComponent: (
+ *     type: string,
+ *     callbackOrObject: ((property: object) => void),
+ *   ) => void,
+ *   signal: AbortSignal,
+ * }) => void} definition.onContextUpdate
+ * @param {AbortSignal} [signal]
```
pdesoyres-cc, line 26: I didn't know this syntax to make a param optional but I think I prefer the `?` syntax:
` @param {AbortSignal?} signal`

pdesoyres-cc (changes requested):

src/lib/define-smart-component.js
```diff
@@ -0,0 +1,98 @@
...
+ * @param {string} definition.selector
+ * @param {object} definition.params
+ * @param {({
+ *   container: SmartContainer,
+ *   component: Element,
+ *   context: object,
+ *   onEvent: (type: string, listener: (detail: object) => void) => void,
+ *   updateComponent: (
+ *     type: string,
+ *     callbackOrObject: ((property: object) => void),
```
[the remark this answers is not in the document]
hsablonniere: Yes, I tried many combinations, including an algebric type so the name of the param could be different whether it's a callback or an object but it didn't work so I kinda gave up :disappointed:

I tried your proposition but it didn't bring the `property: object` in the IDE. I also don't think it's valid JSDoc to have a `@type` below a `@typedef` like this.

I tried this and it seems to "work", even if it's just minor declarative help for now:

```js
/**
 * @typedef {object|((property: object) => void)} CallbackOrObject
 */
```

src/lib/define-smart-component.js
hsablonniere: Good idea, I'll add that.

src/lib/define-smart-component.js
hsablonniere: I applied this but we may mark some of them as required in the future (once we have better typechecking).

src/lib/define-smart-component.js
hsablonniere: Yes, I discovered that the `?` syntax is not really documented for JSDoc but we used it here and there.
I'll update it here and in the other places I used the `[optional]` syntax.

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#handbook-content

hsablonniere:
@pdesoyres-cc I applied your feedbacks in the latest `!fixup` commit

pdesoyres-cc (approved):

PR #576, opened 2022-09-22
Add a FocusLostController

### LINKED ISSUE #575: Investigate focus lost management
When an element with focus is removed from the DOM, the focus is lost.

For instance, a user using the keyboard to 'click' on the delete button in a list of items where each item can be deleted. After the item is removed from the DOM, the focus is lost. As a user, I would prefer the focus to be set to the closest delete button (if any).

We want to investigate a solution to ease the developer's effort in managing such a situation.

pdesoyres-cc:
Fixes #575 

## implementation

* This Reactive controller helps detect a focus loss situation when the focused element is removed from the DOM.
* You provide a query selector which will select the elements for which you're tracking removal.
* You provide a callback that will be fired when one of these elements is removed from the DOM. The callback is fired only if the deleted element had focus (or one of its children had focus)
* The callback will suggest a new element to focus on. It suggests the nearest element matching the same query selector.

## review

Preview is [here](https://clever-components-preview.cellar-c2.services.clever-cloud.com/focus-lost-ctrl/index.html?path=/story/%F0%9F%95%B9%EF%B8%8F-controllers-focuslostcontroller--default-story)

* check the code
* check the story
* check the storybook doc tab

### DIFF
--- a/test/controller/lost-focus-controller.test.js
+++ b/test/controller/lost-focus-controller.test.js
@@ -1,0 +1,190 @@
+import { expect } from '@bundled-es-modules/chai';
+import { fixture, defineCE, nextFrame } from '@open-wc/testing';
+import * as hanbi from 'hanbi';
+import { html, LitElement } from 'lit';
+import { repeat } from 'lit/directives/repeat.js';
+import { LostFocusController } from '../../src/controllers/lost-focus-controller.js';
+import { dispatchCustomEvent } from '../../src/lib/events.js';
+
+describe('lost-focus-controller', () => {
+
+  const ce = defineCE(
+    class extends LitElement {
+      static get properties () {
+        return {
+          items: { type: Array },
+        };
+      }
+
+      constructor () {
+        super();
+        this.items = ['1', '2', '3'];
+
+        // eslint-disable-next-line no-new
+        new LostFocusController(this, '.item', (event) => {
+          dispatchCustomEvent(this, 'my-element:lostFocus', event);
+        });
+      }
+
+      removeItem (it) {
+        this.items = this.items.filter((item) => item !== it);
+      }
+
+      clear () {
+        this.items = [];
+      }
+
+      focusHeader () {
+        this.shadowRoot.querySelector('.header').focus();
+      }
+
+      focusItem (it) {
+        this.getItemElement(it)?.focus();
+      }
+
+      focusItemButton (it) {
+        this.getItemElement(it)?.querySelector('button').focus();
+      }
+
+      getItemElement (it) {
+        return this.shadowRoot.querySelector(`[data-item="${it}"]`);
+      }
+
+      render () {
+        return html`
+          <div class="header"></div>
+          ${repeat(this.items, (item) => item, (item) => {
+            return html`
+              <div tabindex="0" class="item" data-item="${item}"><button>${item}</button></div>
+            `;
+          })}
+        `;
+      }
+    },
+  );
+
+  /**
+   * @return {Promise<{element: Element, spy: Stub}>}
+   */
+  const createElement = async () => {
+    const element = await fixture(`<${ce}></${ce}>`);
+
+    const spy = hanbi.spy();
+    element.addEventListener('my-element:lostFocus', (e) => spy.handler(e.detail));
+
+    return { element, spy };
+  };
+
+  it('should notify focus lost when an item is focused', async () => {
+    const { element, spy } = await createElement();
+
+    element.focusItem('2');
+    element.removeItem('2');
+
+    await nextFrame();
+
+    expect(spy.called).to.equal(true);
+  });
+
+  it('should notify focus lost when a child of item is focused', async () => {
+    const { element, spy } = await createElement();
+
+    element.focusItemButton('2');
+    element.removeItem('2');
+
+    await nextFrame();
+
+    expect(spy.called).to.equal(true);
+  });
+
+  it('should not notify focus lost when an item is not focused', async () => {
+    const { element, spy } = await createElement();
+
+    element.focusItem('1');
+    element.removeItem('2');
+
+    await nextFrame();
+
+    expect(spy.called).to.equal(false);
+  });
+
+  it('should not notify focus lost when no items is not focused', async () => {
+    const { element, spy } = await createElement();
+
+    element.focusHeader();
+    element.removeItem('2');
+
+    await nextFrame();
+
+    expect(spy.called).to.equal(false);
+  });
+
+  it('should not notify focus lost when an item is not focused', async () => {
+    const { element, spy } = await createElement();
+
+    element.focusHeader();
+    element.removeItem('2');
+
+    await nextFrame();
+
+    expect(spy.called).to.equal(false);
+  });
+
+  it('should identify the deleted and focused item properly', async () => {
+    const { element, spy } = await createElement();
+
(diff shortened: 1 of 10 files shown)


florian-sanders-cc:
First off, it's great and I really love it :heart: , well done :clap: :sunglasses:.

I had not checked how reactive controllers worked but the API is so great, the code is fairly easy to understand I think and it's a smart way to deal with this subject :+1:. 

I have a few concerns, some of which are minor and have easy solutions:

- The name of the controller should be a lot less general. It does not deal with all focus loss situations but specifically with focus loss following the removal of a list item. Maybe we could rename it to `itemRemovalFocusLossController` but this is long and ugly ? I’ll try to come up with something better 😂.
- We need to handle cases where users delete the last of the list items. We could add another param like `fallbackElementToFocus` (sorry for the name) to be focused if the list is empty ? This case is fairly common, it will happen when you delete all your secondary emails, all your ssh keys, and all members corresponding the the active filter (unless we reset the filters everytime we remove a member but I’m not sure I would want this).
image
image
- This last point also means the current controller is tricky for `cc-orga-member-list` because we are watching the list of elements but in reality the list displayed to users is the `filteredMemberList`. If we watch the filtered list in the controller then it won’t work because elements can be deleted and added everytime we change the filters 😅. (we can discuss it on Discord, I'm not sure my point is clear here, sorry)
- Focus is only moved upon removal of the item if the user was focusing the item being removed. It’s a smart design choice but it's only acceptable for a11y because we also use toasts to warn users about the result of their async action. We also have to be careful with smart choices in SPA as sometimes, what we anticipate is not what users expect. Sometimes an imperfect pattern is better than a really smart one (for instance, form control on blur vs form control on submit). It may also be tricky once we introduce modal window confirmations before deleting a member. More about this below but in the end I'm okay with it :wink: 

# Only moving the focus if the user was focusing the deleted element

Moving the focus to another element when an element is removed usually answers two objectives:

- making sure the focus is not "lost" and prevent inconsistent focus management by the browser heuristics (may be placed back on the `body` if it fails). This is useful for both screen reader users and keyboard users.
- making sure screen reader users are made aware of the result of their action and UI change. If we don't move the focus when an element is removed, nothing happens for the reader. The browser will only try to determine where to move the focus if the user tries to move the virtual cursor (using screen reader shortcuts or tab).

The way we choose to handle focus management in case of element removal is smart because we're trying to only move the user if they are still focusing the element that has been removed. It fulfills the first objective but it fails the second one. We are not consistently telling screen readers what happened.

Thankfully, in most cases for now, we use toasts for async calls that trigger removal of stuff so this is ok.
I do think we should be very careful though and maybe document that using this "smart" behavior is only ok if you provide a consistent way to warn users about the removal of an item (mostly toasts I guess).

I had two concerns about this choice but in the end I think we’re good anyway:

- I haven’t seen this done anywhere else (but it does not mean that it has not been done anywhere else, and it does not mean we should not do it). It just means the behavior may not be as predictable as usual for screen reader users but this is subjective, and in the end, I really think it’s ok. The API call is usually fast, most people won’t move and people who do move won’t complain about us not moving them back somewhere else ?
- What do we do when we introduce modal windows to confirm the removal ? Will this still work ? We can leave this subject for later I think but it’s interesting:
    - will we wait with the modal open and close it when API call is done ?
    - or will be close it and place the focus back on the waiting card, and then place the focus back to the previous item when the call is done ? 😅 This one is simpler technically with this focus controller but considering how fast API calls are (in most cases), it might not be the best UX wise :wink:

florian-sanders-cc:
Ah ! I forgot, it's a bit sad that reactive controllers don't provide a way to see what properties triggered the update because we're running all this everytime the component is updated (whether it's just the form or the list being updated).
I did not see anything within the docs to prevent this and the code is light so it's ok ?

src/controllers/focus-lost-controller.js
```diff
@@ -0,0 +1,71 @@
...
+    const previousElements = this.elements;
+    // get the elements we are interested in
+    this.elements = [...this.host.shadowRoot.querySelectorAll(this.selector)];
+
+    // compare elements to the previous one and check if some elements have been removed
+    const removedElements = previousElements.filter((e) => !this.elements.includes(e));
+    if (removedElements.length > 0) {
+      // among these removed elements, find the one containing the focused element
+      // elementWithFocus may not be the focused element itself, it can be the element matching the query selector containing the focused element.
+      const elementWithFocus = removedElements.find((e) => e === this.activeElement || isParentOf(e, this.activeElement));
```
florian-sanders-cc, line 58: I don't understand how this works:
we are using the `hostUpdated` hook.
This means we are running this after the new rendering phase is complete (before the `updated` lifecycle, but still, after the `render` phase).
At this point, the element we clicked has been removed and is no longer part of the DOM. The focus should already be lost (meaning, the `activeElement` can be determined by the browser, which is unreliable).

I don't understand how an element that is no longer part of the DOM could still be the `activeElement` ?
(I do see how it could be the parent of the activeElement though, depending on what the browser does when the focus is lost)

src/controllers/focus-lost-controller.stories.js
```diff
@@ -0,0 +1,97 @@
...
+      elements: { type: Array },
+    };
+  }
+
+  constructor () {
+    super();
+    this.elements = [];
+
+    // eslint-disable-next-line no-new
+    new FocusLostController(this, 'div', ({ suggestedElement }) => {
```
florian-sanders-cc, line 20: I'm wondering if we should warn against using very generic selectors like this within the controller documentation ?
It's ok for the story here but `querySelectorAll('div')` or `p` etc. would also query all nested `div` children in the light DOM with these tags no ?
But then again, who would query a `div` in real components ?

src/controllers/focus-lost-controller.js
florian-sanders-cc: Ahh I got it sorry, we get the `activeElement` during the `update` phase, gg :smile:

src/controllers/focus-lost-controller.stories.js
```diff
@@ -0,0 +1,97 @@
...
+      // language=CSS
+      css`
+        div {
+          margin-bottom: 0.2em;
+        }
+      `,
+    ];
+  }
+}
+window.customElements.define('cc-list', MyList);
```
hsablonniere, line 59: I would name the tag `my-list`, this way it's clear it's an example, it's named like the class and won't be clashing with a future list component.

hsablonniere (changes requested):
Hey @pdesoyres-cc, again, really nice work, congrats :clap: 

I suggested a few name changes.
I was wondering, is it possible and not too difficult to also test the reactive controller?

test/controller/lost-focus-controller.test.js
```diff
@@ -0,0 +1,190 @@
...
+    },
+  );
+
+  /**
+   * @return {Promise<{element: Element, callback: Function}>}
+   */
+  const createElement = async () => {
+    const element = await fixture(`<${ce}></${ce}>`);
+
+    const callback = sinon.fake();
```
hsablonniere, line 72: We already have a simpler/smaller sinon alternative (hanbi), used in the smart manager tests.
Can you remove sinon and use hanbi please :wink:

The docs are simple but also because the API is simple.
https://github.com/43081j/hanbi

hsablonniere (changes requested):
Almost there :wink:

hsablonniere (approved):

PR #579, opened 2022-10-03
feat(cc-button): use aria-disabled attribute instead of disabled

### LINKED ISSUE #578: cc-button loses focus after disabling it (chrome only)
We identified a scenario that doesn't behave the same on firefox and safari VS on chrome.

Here is the scenario:

* focus a `<button>` element
* disable it
* => on firefox and safari: the focus is kept on the button
* => on chrome the focus is lost

## how to fix that

In fact, it depends on which behavior we choose to have:

1. do we want the Chrome scenario: lose the focus when disabling
2. do we want the Firefox scenario: keep the focus when disabling

### loose focus when disabling

Here we need to reset focus to `<body>` when we detect that the `disabled` property switches to `true`.

### keep focus when disabling

Here we can use `aria-disabled` attribute instead of `disabled` so that we are still accessible. In this case, the `<button>` will still be focusable (which is what we want). But it will also be clickable, so we need to prevent action on the click event.

The blog post [here](https://css-tricks.com/making-disabled-buttons-more-inclusive/) shows that this will make buttons more inclusive, so this is maybe a good idea to implement it like that.

## To be discussed

We have two alternatives that we need to discuss:

* we really want to force a coherent cross-browser behavior. In this case, we need to implement one of the two scenarios explained above.
* we want to let developers decide which behavior they want on each instance of a `<cc-button>`. In this case, we need to add a property like `use-aria-disabled-when-I-am-disabled`.

pdesoyres-cc:
Fixes #578

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -123,6 +123,11 @@ export class CcButton extends LitElement {
 
     e.stopPropagation();
 
+    // we need to check that because we use aria-disabled which doesn't prevent the onclick event to be fired.
+    if (this.disabled) {
+      return;
+    }
+
     // delay=0 is needed in some situations where you want the button to have the same width
     // as buttons with delay > 0 but without any delay
     if (this.delay == null || this.delay === 0 || this.link) {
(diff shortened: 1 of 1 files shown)


src/components/cc-button/cc-button.js
```diff
@@ -123,6 +123,10 @@ export class CcButton extends LitElement {
 
     e.stopPropagation();
 
+    if (this.disabled) {
```
hsablonniere, line 126: Can we add a comment here where we explain that because we use `aria-disabled` we need to take care of this.

hsablonniere (changes requested):
This is perfect. I just think we can add a small comment explaining aria-disabled does not really disable :stuck_out_tongue:

florian-sanders-cc (approved):
All good ! :clap:

hsablonniere (approved):

PR #580, opened 2022-10-03
form controls: remove unique IDs and rely on shadow dom isolation

### LINKED ISSUE #571: Remove unique ID techniques
As mentioned in #401 we may be able to remove the unique ID techniques we use inside our form atom components.

- [x] check the [Webkit ticket](https://bug.webkit.org/show_bug.cgi?id=199568)
- [x] check that it has indeed been fixed with a safari (use `name` attribute with `cc-toggle` + check `aria-describedby`)
- [x] update `cc-input-number`, `cc-input-text`, `cc-toggle`, `cc-select`!

florian-sanders-cc:
Fixes #571

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ title: Changelog
 * `<cc-badge>`: add skeleton mode
 * `<cc-input-text>`: remove the unique id generation technique and rely on Shadow DOM isolation instead.
 * `<cc-input-number>`: remove the unique id generation technique and rely on Shadow DOM isolation instead.
+* `<cc-select>`: remove the unique id generation technique and rely on Shadow DOM isolation instead.
 * New component:
   * `<cc-action-dispatcher>`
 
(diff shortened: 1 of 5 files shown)


hsablonniere (approved):
Nice :+1:
I was wondering if we should add a name on toggle but this would be a breaking change and maybe we should do this in another commit.

florian-sanders-cc:
> I was wondering if we should add a name on toggle but this would be a breaking change and maybe we should do this in another issue.

Yes, I thought we could deal with names when we switch to element internals and form data ? Or do you want to do it earlier ?
I believe we're missing a "name" attribute and prop on the `cc-select` element.

I'll create the issue anyway and we'll see when we deal with it :wink:

pdesoyres-cc (approved):
Perfect :+1:

hsablonniere:
> Yes, I thought we could deal with names when we switch to element internals and form data ? Or do you want to do it earlier ?

Yes, we'll do this when doing element internals

PR #590, opened 2022-10-12
feat(cc-button): add an accessibleName prop

### LINKED ISSUE #588: cc-button: add accessible-name prop
# Context

We want a way to set an accessible name on the `cc-button` that would be different from its text content.
Currently, the `aria-label` and `title` attributes are used if the component only display an image.

There are many cases where we want to provide more context than the visible text.
For instance, for a button that would delete an app, we would want to tell screen reader users what app they are about to delete:
"Delete ${appName}"

We would also like the `title` to reflect this because it might be useful for screen magnifier users who would not be able to see both the button and its context at the same time.

At the moment, the `cc-button` only populate these attributes when it displays an image and no text. 

**Note:** 
This will also allow us to solve a very specific case for `cc-orga-member` where we slot different texts into a single button that is used for different purposes and we switch the visibility of these texts. 
The goal of this is to prevent layout shifts. 
Currently, we end-up with weird `title` values because the `title` value is based on the text content:

# How

We want to introduce a prop that would populate both the `aria-label` and the `title` attribute with the provided value.
For instance:
```html
  <cc-button accessible-name="edit John Doe's role">edit</cc-button>
```
would result in this:
```html
  <button aria-label="edit John Doe's role" title="edit John Doe's role">edit</button>
```

florian-sanders-cc:
Fixes #588

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -43,6 +43,7 @@ export class CcButton extends LitElement {
 
   static get properties () {
     return {
+      accessibleName: { type: String, attribute: 'accessible-name' },
       circle: { type: Boolean },
       danger: { type: Boolean },
       delay: { type: Number },
@@ -62,6 +63,9 @@ export class CcButton extends LitElement {
 
   constructor () {
     super();
+
+    /** @type {string|null} Forces the values of the `aria-label` and `title` attributes on the `button` element. CAUTION: The accessible name should always start with the visible text if there is one. For instance "add to estimation - NodeJS XS" */
+    this.accessibleName = null;
 
     /** @type {boolean} Sets button UI to a circle form when in `hide-text` and `image` mode. */
     this.circle = false;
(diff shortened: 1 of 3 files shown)


pdesoyres-cc (changes requested):
I have not much to say except that the changelog entry is missing.

roberttran-cc (approved):
Neat addition! :pray:

src/components/cc-button/cc-button.js
```diff
@@ -115,6 +119,40 @@ export class CcButton extends LitElement {
...
+  _getAriaLabel () {
+    if (this.accessibleName != null) {
+      return this.accessibleName?.trim() ?? '';
+    }
+
+    if (this.hideText && this.image != null) {
+      return this.textContent?.trim() ?? '';
+    }
+
+    return undefined;
```
hsablonniere, line 136: :rofl: :cry: When I think we cannot use null because of lit-analyzer...

src/components/cc-button/cc-button.js
```diff
@@ -115,6 +119,40 @@ export class CcButton extends LitElement {
...
   }
 
+  /**
+   * The `aria-label` attribute should only be set if the `accessibleName` prop is set or if the button only shows an image with no text.
+   *
+   * @returns {string|undefined} the value of the `aria-label` attribute or `undefined` if the attribute should not be set.
+   */
+  _getAriaLabel () {
+    if (this.accessibleName != null) {
+      return this.accessibleName?.trim() ?? '';
```
hsablonniere, line 129: If accessibleName is not null, should we really use `?.trim()` instead of `.trim()`?

src/components/cc-button/cc-button.js
```diff
@@ -115,6 +119,40 @@ export class CcButton extends LitElement {
...
+   *
+   * @returns {string|undefined} the value of the `title` attribute or `undefined` if the attribute should not be set.
+   */
+  _getTitle () {
+    if (this.accessibleName != null) {
+      return this.accessibleName?.trim() ?? '';
+    }
+
+    if (this.hideText && this.image != null) {
+      return this.textContent?.trim() ?? '';
```
hsablonniere, line 150: I don't think `this.textContent` can be null but let's keep this `?.` to be sure for now.

hsablonniere (changes requested):

src/components/cc-button/cc-button.js
florian-sanders-cc: Eh indeed, it should be an empty string if there's no text content, and there shouldn't be cases with no text content cause it's bad for a11y anyway :smile: 

I'll remove it :+1:

roberttran-cc (approved):

PR #591, opened 2022-10-12
fix(cc-button): fix grid text-wrapper that causes a Safari bug

Galimede:
## Why 

A few days ago we discovered a bug on Safari within our `cc-header-app` component. After some investigation with @florian-sanders-cc we found that the bug was caused by the `grid-template-columns` grid part that we use on our `cc-button` which is part of the technique we use to prevent layout shifting with the wording of a button when doing a cancel action.

## How

We managed to fix this problem by changing the value from `1fr` to `auto` yet this issue only appeared within the `cc-header-app` component but not when we tried to reproduce the bug on a simple container with some butons. While it solves the bug it will need further investigation to know why.


roberttran-cc (approved):
Tested before and after and looks like the fix works! :+1:

florian-sanders-cc (approved):

florian-sanders-cc:
LGTM :+1: !

pdesoyres-cc (approved):

PR #593, opened 2022-10-12
feat(cc-stretch): init new component

### LINKED ISSUE #592: cc-max-content: new component
# Context

There are several cases where we want some content to replace other content without creating any layout shifts.
For instance, when you click the "info" / "close" button within `cc-tile-status-code`, the box keeps the same size even when you are switching its content.

This is because it's size is actually based on the biggest content of the two.

The technique used is described in one of the many blog posts (:trollface:) on on @hsablonniere's blog: ["Prevent layout shifts with CSS grid stacks"](https://www.hsablonniere.com/prevent-layout-shifts-with-css-grid-stacks--qcj5jo/).
image
image

We want to build a layout component that implements this technique because we use it in several cases.

# How

Implement a component that uses the technique.

The component would be slotted with as many containers as you want.
You add a CSS class on every item you slot.
You provide the component with the name of the CSS class you want to set visible.
All containers would be set to `visibility: hidden` within the component except the one(s) containing the provided class value.

For instance:
```html
<cc-max-content active-class="member-3">
  <p class="member-1">Member 1</p>
  <p class="member-2">Member 2</p>
  <p class="member-3">Member 3</p>
  <p class="member-4">Member 4</p>
</cc-max-content>
```
All members would be hidden except `.member-3`.

# Questions

We need to find a name for this component.
@pdesoyres-cc suggested two names that could work (gg :clap:):

* `cc-max-bounded-box`
* `cc-max-stretching-box`

Based on his suggestions, I also considered `cc-stretch-to-max-content` but it's probably too long ? :joy:

florian-sanders-cc:
# Context

Fixes #592 

# Preview

[cc-stretch](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-stretch/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-stretch--default-story)

# TODO:

- [x] update changelog and all references to the component name.
- [x] rebase + update changelog

### DIFF
--- a/src/components/cc-stretch/cc-stretch.js
+++ b/src/components/cc-stretch/cc-stretch.js
@@ -1,0 +1,105 @@
+import { css, html, LitElement } from 'lit';
+
+/**
+ * A layout component used to switch visibility between several elements while preventing layout shifts.
+ *
+ * Use `visibleElementId` prop to provide the `id` value corresponding to the element you want to make visible.
+ * Other elements are hidden with `visibility: hidden`.
+ *
+ * More information about the issues this component solves and the technique it uses can be found in [Hubert Sablonnière's blog Post - "Prevent layout shifts with CSS grid stacks"](https://www.hsablonniere.com/prevent-layout-shifts-with-css-grid-stacks--qcj5jo/).
+ *
+ * **Caution**:
+ *
+ * This component relies on a class named `cc-stretch--visible` applied to the relevant slotted element. Make sure you never override this class.
+ *
+ * @cssdisplay inline-grid
+ *
+ * @slot - The containers you want to switch between. The container you want to be displayed must have an `id` attribute matching the value of the `visibleElementId` prop.
+ *
+ * @cssprop {AlignItems} --cc-stretch-align-items - Specify how the items should be distributed / positioned vertically within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items).
+ * @cssprop {JustifyItems} --cc-stretch-justify-items - Specify how the items should be distributed / positioned horizontally within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items).
+ */
+
+export class CcStretch extends LitElement {
+
+  static get properties () {
+    return {
+      disableStretching: { type: Boolean, attribute: 'disable-stretching', reflect: true },
+      visibleElementId: { type: String, attribute: 'visible-element-id', reflect: true },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {boolean} Disables stretching by setting hidden elements to `display: none`.
+     * May be useful if you need stretching only under specific conditions (the size of the parent component for instance).
+     */
+    this.disableStretching = false;
+
+    /** @type {string|null} Sets the slotted element(s) matching this `id` to `visibility: inherit`.
+     * This makes the element visible unless one of its parents is set to `visibility: hidden` (in case you nest several `cc-stretch` for instance).
+     */
+    this.visibleElementId = null;
+  }
+
+  /*
+  * Triggered when the `visibleElementId` value changes or the slot content changes.
+  * Hides the currently visible element if there is one.
+  * Makes the element matching `visibleElementId` visible.
+  */
+  _makeElementVisible () {
+    // Since several `cc-stretch` may be nested, and we are working with light DOM, it is very important to only target the direct children with the `:scope >` selector
+    const elementToHide = this.querySelector(':scope > .cc-stretch--visible');
+    const elementToMakeVisible = this.querySelector(`:scope > #${this.visibleElementId}`);
+
+    elementToHide?.classList.remove('cc-stretch--visible');
+    elementToMakeVisible?.classList.add('cc-stretch--visible');
+  }
+
+  updated (changedProperties) {
+    if (changedProperties.has('visibleElementId')) {
+      this._makeElementVisible();
+    }
+  }
+
+  render () {
+    return html`
+      <slot @slotchange=${this._makeElementVisible}></slot>
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          align-items: var(--cc-stretch-align-items, center);
+          display: inline-grid;
+          justify-items: var(--cc-stretch-justify-items, center);
+          width: max-content;
+        }
+
+        :host([disable-stretching]) slot::slotted(:not(.cc-stretch--visible)) {
+          display: none;
+        }
+
+        slot::slotted(*) {
+          grid-area: 1 / 1 / 2 / 2;
+          visibility: hidden;
+        }
+
+        /* 
+         * Very important: do not set to visible as this would break nested cc-stretch components.
+         * In a parent with "visibility: hidden", a child with "visibility: visible" is visible.
+         * With inherit, the visibility of the child depends on the visibility of the parents, which is what we want.
+         */
+        slot::slotted(.cc-stretch--visible) {
+          visibility: inherit;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-stretch', CcStretch);
--- a/src/components/cc-stretch/cc-stretch.stories.js
+++ b/src/components/cc-stretch/cc-stretch.stories.js
@@ -1,0 +1,181 @@
+import './cc-stretch.js';
+import '../cc-toggle/cc-toggle.js';
+import { html, render } from 'lit';
+import { makeStory, storyWait } from '../../stories/lib/make-story.js';
+import { enhanceStoriesNames } from '../../stories/lib/story-names.js';
+
+export default {
+  title: '🧬 Atoms/<cc-stretch>',
+  component: 'cc-stretch',
+};
+
+const conf = {
+  component: 'cc-stretch',
+  // language=CSS
+  css: `
+    h1 {
+      font-size: 1.2em;
+    }
+    
+    .showcase-controls {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 1em 2em;
+      padding: 1em;
+      background-color: var(--cc-color-bg-neutral);
+    }
+    
+    cc-stretch {
+      border: 0.2em solid red;
+    }
+    
+    cc-toggle {
+      margin-bottom: 1.5em;
+      margin-right: 1em;
+    }
+
+    code {
+      background-color: var(--cc-color-bg-neutral);
+      padding-inline: 0.3em;
+    }
+
+    p {
+      margin: 0;
+    }
+    
+    .showcase-container {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 1em;
+      margin-top: 1em;
+    }
+  `,
+};
+
+export const defaultStory = makeStory(conf, {
+  dom: (container) => {
+    const choices = [
+      { label: 'start', value: 'start' },
+      { label: 'center (Default)', value: 'center' },
+      { label: 'end', value: 'end' },
+    ];
+
+    const choicesItemIds = [
+      { label: 'item-1', value: 'item-1' },
+      { label: 'item-2', value: 'item-2' },
+      { label: 'item-3', value: 'item-3' },
+    ];
+
+    const choicesStretching = [
+      { label: 'enabled', value: 'enabled' },
+      { label: 'disabled', value: 'disabled' },
+    ];
+
+    let justifyCustomProp = 'center';
+    let alignCustomProp = 'center';
+    let visibleElementId = 'item-1';
+    let stretching = 'enabled';
+
+    const refreshUi = () => render(template({
+      visibleElementId,
+      justifyCustomProp,
+      alignCustomProp,
+      stretching,
+    }), container);
+
+    const onVisibleElementIdChange = ({ detail: newVisibleElementId }) => {
+      visibleElementId = newVisibleElementId;
+      refreshUi();
+    };
+
+    const onStretchingChange = ({ detail: newStretching }) => {
+      stretching = newStretching;
+      refreshUi();
+    };
+
+    const onJustifyCustomPropChange = ({ detail: newJustifyCustomProp }) => {
+      justifyCustomProp = newJustifyCustomProp;
+      refreshUi();
+    };
+
+    const onAlignCustomPropChange = ({ detail: newAlignCustomProp }) => {
+      alignCustomProp = newAlignCustomProp;
+      refreshUi();
+    };
+
+    function template ({ visibleElementId, justifyCustomProp, alignCustomProp, stretching }) {
+      return html`
+        <div class="showcase-controls">
+          <div>
+            <h1>Component Props</h1>
+            <cc-toggle 
+              legend="Visible element ID"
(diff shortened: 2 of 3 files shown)


src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @slot - The containers you want to switch between. The container you want to be displayed must have an `id` attribute matching the value of the `visibleElementId` prop.
+ */
+
+export class CcStretchToMaxContent extends LitElement {
+
+  static get properties () {
+    return {
+      centerContentHorizontally: { type: Boolean, attribute: 'center-content-horizontally', reflect: true },
+      centerContentVertically: { type: Boolean, attribute: 'center-content-vertically', reflect: true },
```
florian-sanders-cc, line 24: could also provide a `center-content` property in addition to these ? (just thought of that while implementing the two attributes in `cc-orga-member` context...)

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /*
+  * Triggered when the `visibleElementId` value changes or the slot content changes.
+  * Hides the currently visible element if there is one.
+  * Makes the element matching `visibleElementId` visible.
+  */
+  _makeElementVisible () {
+    // Since several `cc-stretch-to-max-content` may be nested, and we are working with light DOM, it is very important to only target the direct children with the `:scope >` selector
+    const elementToHide = this.querySelector(':scope > .cc-stretch-to-max-content--visible');
```
pdesoyres-cc, line 53: This class selector is awesome :star_struck:

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
```diff
@@ -0,0 +1,110 @@
...
+    const elementToMakeVisible = this.querySelector(`:scope > #${this.visibleElementId}`);
+
+    elementToHide?.classList.remove('cc-stretch-to-max-content--visible');
+    elementToMakeVisible?.classList.add('cc-stretch-to-max-content--visible');
+  }
+
+  update (changedProperties) {
```
pdesoyres-cc, line 60: I think that we should override `willUpdate` (before update) or `updated` (after update)... Not sure if overriding the `update` is a good practice :thinking:
What do you think @hsablonniere?

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
```diff
@@ -0,0 +1,110 @@
...
+  * Triggered when the `visibleElementId` value changes or the slot content changes.
+  * Hides the currently visible element if there is one.
+  * Makes the element matching `visibleElementId` visible.
+  */
+  _makeElementVisible () {
+    // Since several `cc-stretch-to-max-content` may be nested, and we are working with light DOM, it is very important to only target the direct children with the `:scope >` selector
+    const elementToHide = this.querySelector(':scope > .cc-stretch-to-max-content--visible');
+    const elementToMakeVisible = this.querySelector(`:scope > #${this.visibleElementId}`);
+
+    elementToHide?.classList.remove('cc-stretch-to-max-content--visible');
```
pdesoyres-cc, line 56: Maybe, we should add a comment in the class JSDoc telling developers that we are using this class underneath and warning that it should not clash with any existing class.

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+export class CcStretchToMaxContent extends LitElement {
+
+  static get properties () {
+    return {
+      centerContentHorizontally: { type: Boolean, attribute: 'center-content-horizontally', reflect: true },
+      centerContentVertically: { type: Boolean, attribute: 'center-content-vertically', reflect: true },
+      disableStretching: { type: Boolean, attribute: 'disable-stretching', reflect: true },
+      visibleElementId: { type: String, attribute: 'visible-element-id' },
```
pdesoyres-cc, line 26: why not reflect here? It may be interesting for developers to have it available as an attribute.

pdesoyres-cc (changes requested):
I really like this.

Instead of `centerContentHorizontally` and `centerContentVertically`, what do you think of that:

* horizontalAlign : 'start' | 'middle' | 'end'
* verticalAlign: 'start' | 'middle' | 'end'

florian-sanders-cc:
> Instead of centerContentHorizontally and centerContentVertically, what do you think of that:
>  - horizontalAlign : 'start' | 'middle' | 'end'
>  - verticalAlign: 'start' | 'middle' | 'end'

That's even better, thanks :heart: :+1:

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
florian-sanders-cc: I did not see any use case for it but now that you're mentioning it, I could see it being useful in some cases. I'll add the reflect :+1:

src/components/cc-stretch-to-max-content/cc-stretch-to-max-content.js
florian-sanders-cc: I'll go for the `updated` hook if that's ok with you :+1:

florian-sanders-cc:
Actually, I'm going to switch to CSS Custom properties to let people customize how the content should be positioned / distributed within the grid.
I think it makes more sense than recreating the CSS API through JS props.

src/components/cc-stretch/cc-stretch.js
```diff
@@ -0,0 +1,101 @@
...
+ * **Caution**:
+ *
+ * This component relies on a class named `cc-stretch--visible` applied to the relevant slotted element. Make sure you never override this class.
+ *
+ * @cssdisplay inline-grid
+ *
+ * @slot - The containers you want to switch between. The container you want to be displayed must have an `id` attribute matching the value of the `visibleElementId` prop.
+ *
+ * @cssprop {AlignItems} --cc-stretch-align-items - Specify how the items should be distributed / positioned vertically within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items).
+ * @cssprop {JustifyItems} --cc-stretch-justify-items - Specify how the items should be distributed / positioned horizontally within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items).
```
roberttran-cc, line 20: I may be biased but I expect the CSS custom properties default value to match the CSS property [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value), meaning:
- for `align-items`, the value `normal`
- for `justify-items`, the value... `legacy` ? :thinking: (as stated in [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items#formal_definition))

roberttran-cc (changes requested):

src/components/cc-stretch/cc-stretch.js
pdesoyres-cc: I was thinking the same as you. But my final thought was that we should choose the default values that fit best our use cases. I think it makes sense here to use `center` for both.

src/components/cc-stretch/cc-stretch.stories.js
```diff
@@ -0,0 +1,100 @@
+// Don't forget to import the component you're presenting!
```
pdesoyres-cc, line 1: this comment can be removed

src/components/cc-stretch/cc-stretch.js
roberttran-cc: Makes sense! :+1:

src/components/cc-stretch/cc-stretch.stories.js
```diff
@@ -0,0 +1,100 @@
...
+export const defaultStory = makeStory(conf, {
+  items: [
+    {
+      visibleElementId: 'item-1',
+      innerHTML: baseHtml,
+    },
+  ],
+});
+
+export const contentPositioning = makeStory(conf, {
```
pdesoyres-cc, line 45: Could be nice to have all the alignment combinations.

roberttran-cc (approved):

src/components/cc-stretch/cc-stretch.stories.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      visibleElementId: 'item-2',
+      innerHTML: baseHtml,
+      style: '--cc-stretch-align-items: end',
+    },
+  ],
+});
+
+export const disableStretching = makeStory(conf, {
```
pdesoyres-cc, line 70: Could be nice to have a simulation here so that we really see the advantage of using the `cc-strech` component.

pdesoyres-cc: By the way, I'm wondering which use case this prop covers.

pdesoyres-cc (changes requested):
GG :clap: 

Could you rename the PR so that it matches the new component name.

src/components/cc-stretch/cc-stretch.stories.js
florian-sanders-cc: It's very situational and that's why I'm not sure how to showcase it properly in stories / simulations.

In `cc-orga-member-card`, we want the badges to be stretched + centered in desktop:
image

But you don't really want that when your viewport is reduced:
image

Here we disable the stretching when the component width is below a specific value, but you might also want to disable the stretching based on other conditions (the element being displayed for instance ?).
At the beginning I used CSS to do this, but it felt like I was bypassing the component behavior just because I knew very well how it worked.
This meant I was adding weird CSS to `cc-orga-member-card`:
```css
:host([w-lt-740]) .status__role-mfa cc-stretch > span:not(.cc-stretch--visible) {
  display: none;
}
```

florian-sanders-cc: I'll go for a simulation showcasing stretching enabled to stretching disabled.

Tell me if you think this property is not relevant / too specific. Do you think we should handle it with CSS from the parent component when needed ?

pdesoyres-cc: I understand now the use case so this is perfect like that.

src/components/cc-stretch/cc-stretch.stories.js
pdesoyres-cc: I really like it like that. Well done :clap:

pdesoyres-cc (approved):
LGTM

Galimede (changes requested):
Good job, the story is pretty cool! :heart_eyes:  :clap: 
I just found a _typo_ and was wondering about the line of the comments but you will see it in my review below :wink:

roberttran-cc (approved):

src/components/cc-stretch/cc-stretch.js
```diff
@@ -0,0 +1,105 @@
+import { css, html, LitElement } from 'lit';
+
+/**
+ * A layout component used to switch visibility between several elements while preventing layout shifts.
+ *
+ * Use `visibleElementId` prop to provide the `id` value corresponding to the element you want to make visible.
+ * Other elements are hidden with `visibility: hidden`.
+ *
+ * More information about the issues this component solves and the technique it uses can be found in [Hubert Sablonnière's blog Post - "Prevent layout shifts with CSS grid stacks"](https://www.hsablonniere.com/prevent-layout-shifts-with-css-grid-stacks--qcj5jo/).
```
hsablonniere, line 9: :heart:

src/components/cc-stretch/cc-stretch.js
hsablonniere: About this, I would have asked to remove those and let the user define the properties directly but since we want different defaults, the custom props make sense. It's also better if we want to add another layer later.

src/components/cc-stretch/cc-stretch.js
```diff
@@ -0,0 +1,105 @@
...
+ * @cssprop {AlignItems} --cc-stretch-align-items - Specify how the items should be distributed / positioned vertically within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items).
+ * @cssprop {JustifyItems} --cc-stretch-justify-items - Specify how the items should be distributed / positioned horizontally within the grid (Default: `center`. Possible values: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items).
+ */
+
+export class CcStretch extends LitElement {
+
+  static get properties () {
+    return {
+      disableStretching: { type: Boolean, attribute: 'disable-stretching', reflect: true },
+      visibleElementId: { type: String, attribute: 'visible-element-id', reflect: true },
```
hsablonniere, line 28: Seems like we don't need this one to have `reflect`, WDYT?

src/components/cc-stretch/cc-stretch.js
```diff
@@ -0,0 +1,105 @@
...
+  constructor () {
+    super();
+
+    /** @type {boolean} Disables stretching by setting hidden elements to `display: none;`.
+     * May be useful if you need stretching only under specific conditions (the size of the parent component for instance).
+     */
+    this.disableStretching = false;
+
+    /** @type {string|null} Sets the slotted element(s) matching this `id` to `visibility: inherit;`.
+     * This makes the element visible unless one of its parents is set to `visibility: hidden;` (in case you nest several `cc-stretch` for instance).
```
hsablonniere, line 41: can we remove the `;` in the CSS examples?

src/components/cc-stretch/cc-stretch.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        slot::slotted(*) {
+          grid-area: 1 / 1 / 2 / 2;
+          visibility: hidden;
+        }
+
+        /* 
+         * Very important: do not set to visible as this would break nested cc-stretch components.
+         * In a parent with visibility: hidden;, a child with visibility: visible; is visible.
```
hsablonniere, line 94: * can we remove the `;` in the CSS example?
* can we put it in backticks as we do in other places?

hsablonniere (changes requested):
This is great!!! I really like the result and the "in your face" stories to explain what it does. Good job :clap:

src/components/cc-stretch/cc-stretch.js
florian-sanders-cc: Well, I didn't at first, then @pdesoyres-cc suggested reflecting it and I figured why not:
if developers want to style the box with a different background when a specific element is displayed ? (writing this makes me think it's far fetched actually but it does not hurt ? :joy:).
WDYT ?

src/components/cc-stretch/cc-stretch.js
florian-sanders-cc: backticks break my syntax highlighting because we're already inside css\`...\`.

I could escape the backticks in these comments but since they are only visible in the code and not inside md docs, I don't really like this idea ?

florian-sanders-cc: use double quotes instead

florian-sanders-cc:
@hsablonniere ready to be merged :+1:

hsablonniere (approved):

PR #598, opened 2022-10-18
chore(eslint): update `no-new` rule to "off"

### LINKED ISSUE #597: `new MyClass()` syntax triggering an `eslint` error
Because we use the [`standard`](https://github.com/CleverCloud/clever-components/blob/master/.eslintrc.cjs#L3) eslint config which contains the [`no-new`](https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json#L139) rule. More details on this rule [here](https://eslint.org/docs/latest/rules/no-new).

Troublesome when using [`LostFocusController`](https://github.com/CleverCloud/clever-components/blob/master/src/controllers/lost-focus-controller.md#usage).

roberttran-cc:
Fixes #597.

We choose to deactivate the `no-new` rule.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -48,6 +48,7 @@ module.exports = {
     'lit/no-useless-template-literals': 'error',
     'lit/no-value-attribute': 'error',
     'multiline-ternary': 'off',
+    'no-new': 'off',
     'operator-linebreak': ['error', 'before'],
     'padded-blocks': 'off',
     'quotes': ['error', 'single', { 'allowTemplateLiterals': true }],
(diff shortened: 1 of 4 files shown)


roberttran-cc:
I'm not sure about the impact of merging this PR before/after the conventional commits PR, so I'll leave it to @pdesoyres-cc to tell us what is the best option. :sweat_smile: :hugs:

pdesoyres-cc (approved):
Perfect.

We will merge the conventional commit PR just after the next release. Until then, we should keep on modifying the CHANGELOG manually.

PR #605, opened 2022-10-31
fix(cc-select): disable placeholder option only when field is required

### LINKED ISSUE #604: cc-select: only disable the placeholder option if the field is required
# Context

Currently, we disable the placeholder option. You may select it as the default option but users cannot select it themselves.
This means that once users select a value, they cannot go back to the placeholder (empty value / no value).

# Issue

This behavior matches our current usecases but it is not future proof.
It won't work with optional (not required) `select` fields.

# Solution

Only disable the placeholder option if the component is set to `required`.

florian-sanders-cc:
Fixes #604 

# How to test

In the [Default Story - Select component - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-select/fix-disabled-placeholder/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-select--default-story), you should be able to select a value from the first select, and then switch back to the placeholder. In other words, the placeholder is selectable.

In the [Required Story - Select component - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-select/fix-disabled-placeholder/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-select--required), if you select a value, then you should not be able to switch back to the placeholder value. In other words, the placeholder is not selectable / disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,9 @@ title: Changelog
   * add `alt` attribute values for controls (+ / -) so that they can be identified by assistive technologies.
   * add `hiddenLabel` prop to allow the label to be visually hidden in some cases.
   * add red border and redish focus ring when error slot is used
-* `<cc-select>`: remove the unique id generation technique and rely on Shadow DOM isolation instead.
+* `<cc-select>`: 
+  * remove the unique id generation technique and rely on Shadow DOM isolation instead.
+  * only disable the placeholder option if the component is in required mode.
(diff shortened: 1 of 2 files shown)


florian-sanders-cc:
Thanks @frinyvonnick for pointing this out :wink:

roberttran-cc (approved):

pdesoyres-cc (approved):

hsablonniere (approved):

PR #608, opened 2022-10-31
fix(cc-select): add red border / focus ring when error slot is used

### LINKED ISSUE #587: cc-select: add red border + redish focus ring when error slot is used
## Context

Right now the `<cc-input-number>` and `<cc-input-text>` get a red border and a redish focus ring when their error slot is used.

## TODO

We want this kind of display for `<cc-select>` when the `error` slot is used.

florian-sanders-cc:
Fixes #587 

The whole code is basically a copy / paste from what @pdesoyres-cc has done for `cc-input-` a few weeks ago :wink:.

[Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-select/fix-red-border-when-error/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-select--simulation)

# TODO

- [x] rebase after merging #605

### DIFF
--- a/src/components/cc-select/cc-select.js
+++ b/src/components/cc-select/cc-select.js
@@ -34,9 +34,7 @@ export class CcSelect extends LitElement {
       placeholder: { type: String },
       required: { type: Boolean },
       value: { type: String },
-      _uniqueErrorId: { type: String, state: true },
-      _uniqueHelpId: { type: String, state: true },
-      _uniqueInputId: { type: String, state: true },
+      _hasError: { type: Boolean, state: true },
     };
   }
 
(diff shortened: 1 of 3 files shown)


src/components/cc-select/cc-select.js
```diff
@@ -232,6 +237,14 @@ export class CcSelect extends LitElement {
           box-shadow: 0 0 0 .2em rgba(50, 115, 220, .25);
           outline: 0;
         }
+        
+        select.error {
+          border-color: var(--cc-color-border-danger) !important;
```
florian-sanders-cc, line 242: the `!important` is not necessary considering the selector specificity but since it's the same style / code as `cc-input-*` I chose to keep it.
WDYT ?

roberttran-cc: Agree, although we should try to use it as little as possible imo.

pdesoyres-cc (approved):

PR #622, opened 2022-11-09
feat(cc-notice): init new component

### LINKED ISSUE #343: cc-notice: init component
This would be an improved version of current `cc-error` that would handle several states : 
* error
* warning
* info
* success
Something like this:
https://getbootstrap.com/docs/5.1/components/alerts/

### LINKED ISSUE #634: color choices (core tokens): rename colors variants
## Context

For `cc-notice` we needed new color shades. Until now, we relied on `light`, `lighter` suffixes to name the shades but with the addition of new shades we can no longer rely on this system because we have more than two shades.

## Solution

We need to introduce a system that ranges from 0 to 100 to resolve that problem.

Galimede:
Fixes #343 
Fixes #634

preview here: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-notice/index.html?path=%2Fstory%2F%F0%9F%A7%AC-atoms-cc-notice--default-story

### DIFF
--- a/src/components/cc-notice/cc-notice.js
+++ b/src/components/cc-notice/cc-notice.js
@@ -1,0 +1,261 @@
+import { css, html, LitElement } from 'lit';
+import '../cc-img/cc-img.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
+import { i18n } from '../../lib/i18n.js';
+
+const closeSvg = new URL('../../assets/close-gray.svg', import.meta.url).href;
+const dangerSvg = new URL('../../assets/spam-2-fill.svg', import.meta.url).href;
+const infoSvg = new URL('../../assets/information-fill.svg', import.meta.url).href;
+const successSvg = new URL('../../assets/checkbox-circle-fill.svg', import.meta.url).href;
+const warningSvg = new URL('../../assets/alert-fill.svg', import.meta.url).href;
+
+/**
+ * @typedef {import('./cc-notice.types.js').NoticeIntent} NoticeIntent
+ */
+
+/**
+ * A component to display a block with a title and a message with different modes (info, success, danger, warning).
+ *
+ * You can also close it and override the message or the icon if needed.
+ *
+ * @cssdisplay block
+ *
+ * @slot icon - Icon slot to override the default one provided.
+ * @slot message - Message slot to override the message if you want more than just a short text.
+ *
+ * @event {CustomEvent} cc-notice:dismiss - Fires to inform that the notice should be dismissed.
+ */
+export class CcNotice extends LitElement {
+
+  static get properties () {
+    return {
+      closeable: { type: Boolean },
+      heading: { type: String },
+      intent: { type: String, reflect: true },
+      message: { type: String },
+      noIcon: { type: Boolean, attribute: 'no-icon' },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {boolean} Makes the notice closeable. */
+    this.closeable = false;
+
+    /** @type {string|null} The title of the notice. */
+    this.heading = null;
+
+    /** @type {NoticeIntent} The intent of the notice. */
+    this.intent = 'info';
+
+    /** @type {string} The content of the message of the notice. */
+    this.message = null;
+
+    /** @type {boolean} Remove icon if you don't want it. */
+    this.noIcon = false;
+  }
+
+  _getIcon () {
+    if (this.intent === 'danger') {
+      return dangerSvg;
+    }
+    if (this.intent === 'info') {
+      return infoSvg;
+    }
+    if (this.intent === 'success') {
+      return successSvg;
+    }
+    if (this.intent === 'warning') {
+      return warningSvg;
+    }
+  }
+
+  _getIconAlt () {
+    if (this.intent === 'danger') {
+      return i18n('cc-notice.icon-alt.danger');
+    }
+    if (this.intent === 'info') {
+      return i18n('cc-notice.icon-alt.info');
+    }
+    if (this.intent === 'success') {
+      return i18n('cc-notice.icon-alt.success');
+    }
+    if (this.intent === 'warning') {
+      return i18n('cc-notice.icon-alt.warning');
+    }
+  }
+
+  _onCloseButtonClick () {
+    dispatchCustomEvent(this, 'dismiss');
+  }
+
+  render () {
+
+    const layout = {
+      'no-icon': this.heading != null && this.noIcon,
+      'no-heading': this.heading == null && !this.noIcon,
+      'message-only': this.heading == null && this.noIcon,
+      closeable: this.closeable,
+    };
+
+    return html`
+      <div class="wrapper ${classMap(layout)}">
+        ${!this.noIcon ? html`
+          <slot name="icon">
+            <img src="${this._getIcon()}" alt="${this._getIconAlt()}" class="notice-icon">
+          </slot>
+        ` : ''}
+        ${this.heading != null ? html`
+          <div class="heading">
+            ${this.heading}
+          </div>
+        ` : ''}
+        <div class="message-container">
+          <slot name="message">
+            <p>${this.message}</p>
+          </slot>
+        </div>
+        ${this.closeable ? html`
+            <button class="close-button"
+              @click=${this._onCloseButtonClick}
+              title="${i18n('cc-toast.close')}">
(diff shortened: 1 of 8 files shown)


hsablonniere:
## Review

* :+1: I like the colors and icons and the design of the props.
* :thinking: I wonder if the title should also be a slot, maybe useful if we want to add some `em` or `code` but probably overkill and not needed for now. We can try this the day we feel it's necessary.
* I'm not about sure about the left/right spacing with `no-icon`.
* I'm not about sure about the spacing between the icon and the message with no heading.
* Should we adapt the color of the hover background for the closeable button?

:clap: Really nice work

* Can we add a commit in this PR to use this in the `cc-warning-payment`?
* Can we also try what some component would look like with a `cc-notice intent="warning"` instead of a cc-error?

roberttran-cc:
Overall design (colors, props, etc.) is really nice, GG! Looking forward to see them in our other components! :heart_eyes_cat: 

Some minor feedback:
- white-space between title and content could be slightly reduced
- white-space between icon and content in the "No Heading" story could be increased
- the close button interactive zone could use more padding
- it could be nice to have stories with large content, to check on the wrapping and leading

On a more global note (that may be discussed synchronously):
- the "Custom Message" story is nice but too close to the "Default" story and with much more white-space: consider reducing white-space in the slotted content? or maybe go all-in on slotting and use another component (for example `<cc-env-var-create>`)?
- the "No Heading" story alignment between the icon and the content seems off but... image/text alignment is [hard on the web](https://iamvdo.me/blog/css-avance-metriques-des-fontes-line-height-et-vertical-align) (and not sure if it still seems off with multi-line content)
- overall, I think we could use concrete text in the stories, to close the gap (CSS joke lmao) between the component in their stories and in their real world usage

pdesoyres-cc:
I like the overall design.

* Could we try with smaller margins?
* Could we have the close button more in the top right corner and with same top and right distance from the edge?
* Could we try less space between title and message?

src/components/cc-notice/cc-notice.stories.js
```diff
@@ -0,0 +1,107 @@
...
+});
+
+enhanceStoriesNames({
+  defaultStory,
+  withNoIcons,
+  withNoIconsAndNoHeading,
+  noHeading,
+  withCloseableNotice,
+  withCloseableAndNoIcons,
+  withOtherIcons,
```
florian-sanders-cc, line 106: Not really important but I think you are missing a few stories in there.

src/components/cc-notice/cc-notice.js
```diff
@@ -0,0 +1,262 @@
...
+ */
+export class CcNotice extends LitElement {
+
+  static get properties () {
+    return {
+      closeable: { type: String, reflect: true },
+      heading: { type: String, reflect: true },
+      intent: { type: String, reflect: true },
+      message: { type: String },
+      noIcon: { type: Boolean, attribute: 'no-icon', reflect: true },
```
florian-sanders-cc, line 36: I don't think we need to reflect most of the attributes now that we rely on a wrapper + class system.

It's still necessary for the intent I guess unless you want to make it rely on a class system like others.

src/components/cc-warning-payment/cc-warning-payment.js
```diff
@@ -2,8 +2,7 @@ import { ERROR_TYPES } from '@clevercloud/client/esm/utils/payment.js';
 import { css, html, LitElement } from 'lit';
 import { i18n } from '../../lib/i18n.js';
 import { linkStyles } from '../../templates/cc-link/cc-link.js';
-
-const warningSvg = new URL('../../assets/warning.svg', import.meta.url).href;
+import '../cc-notice/cc-notice.js';
 
 /**
  * @typedef {import('./cc-warning-payment.types.js').PaymentMethodError} PaymentMethodError
```
florian-sanders-cc, line 8: not related to this line but we need to update the doc about css display: 
```js
@cssdisplay block
```

src/components/cc-warning-payment/cc-warning-payment.js
```diff
@@ -35,21 +34,39 @@ export class CcWarningPayment extends LitElement {
   }
 
   render () {
+
+    const { title, error } = (this.mode === 'overview' || this.mode === 'billing') ? this._renderOrgaError(this.errors[0]) : '';
```
florian-sanders-cc, line 38: Not really related to your modifications but isn't `_renderOrgaError` essentially a method to return i18n strings that we usually name `_get...` ?

It does not seem to return anything to render so I wonder if the name is not misleading.

src/components/cc-warning-payment/cc-warning-payment.js
```diff
@@ -97,25 +123,19 @@ export class CcWarningPayment extends LitElement {
       // language=CSS
       linkStyles,
       css`
+
```
florian-sanders-cc, line 126: is this space now necessary for stylelint ? (I don't have any opinion about this, i'm fine with or without it :smile:)

florian-sanders-cc (changes requested):
:clap: Gg @Galimede and thanks a lot for handling the update of design token names.
It was annoying to do but very necessary :+1: 

I think this component looks great, I like the colors (:smirk:) and the spacing too :+1: 
The API seems easy to use which is all we want for an atom component !
Stories are great too.

I found very small details to adjust within the code and I just wonder if we could create a "long text + no heading", WDYT ?

src/components/cc-notice/cc-notice.js
```diff
@@ -0,0 +1,262 @@
...
+ * @slot icon - Icon slot to override the default one provided.
+ * @slot message - Message slot to override the message if you want more than just a short text.
+ *
+ * @event {CustomEvent} cc-notice:dismiss - Fires an event to inform that the notice should be dismissed.
+ */
+export class CcNotice extends LitElement {
+
+  static get properties () {
+    return {
+      closeable: { type: String },
```
hsablonniere, line 32: Isn't this a `Boolean` ?

src/components/cc-notice/cc-notice.stories.js
```diff
@@ -0,0 +1,120 @@
...
+      </div>
+      `,
+  })),
+});
+
+export const withOtherIcons = makeStory(conf, {
+  items: BASE_ITEMS.map((item) => ({
+    ...item,
+    innerHTML: `
+        <div slot="icon"><img src="http://placekitten.com/24/24" alt="placeholder kitten"/></div>
```
hsablonniere, line 53: Can you add a style `display:block` on the ``, it will remove the space below and align things correctly.

src/components/cc-warning-payment/cc-warning-payment.js
```diff
@@ -80,15 +97,24 @@ export class CcWarningPayment extends LitElement {
   /**
    * @param {PaymentMethodError} error
    */
-  _renderOrgaError ({ type }) {
+  _getOrgaError ({ type }) {
```
hsablonniere, line 100: Since it's not a subrender function, can we move it between the constructor and the main render?

hsablonniere (changes requested):
This is really nice :+1:

I added a few details but it's almost ready for prime time!!! :tada:

src/components/cc-notice/cc-notice.js
```diff
@@ -0,0 +1,263 @@
...
+          background-color: var(--cc-color-bg-primary-weaker);
+          border: 1px solid var(--cc-color-border-primary-weak);
+        }
+
+        :host([intent="danger"]) .wrapper {
+          background-color: var(--cc-color-bg-danger-weaker);
+          border: 1px solid var(--cc-color-border-danger-weak);
+        }
+
+
```
pdesoyres-cc, line 175: extra space here

src/components/cc-notice/cc-notice.js
```diff
@@ -0,0 +1,263 @@
...
+          box-shadow: 0 0 0 .15em rgba(50, 115, 220, .25);
+          outline: 0;
+        }
+
+        .close-button img {
+          display: block;
+          height: 1em;
+          width: 1em;
+        }
+
```
pdesoyres-cc, line 257: extra space here

pdesoyres-cc (changes requested):
Nice work!

I really like the UI. I find the margins and alignments really nice.

Don't forget to modify the 2 commit messages and add the `Fixes #xxx`: This will help the release-please bot to associate a link to the issue and the change log entry.

pdesoyres-cc (approved):

hsablonniere (approved):

PR #655, opened 2022-12-01
fix(cc-badge): override border in skeleton mode

### LINKED ISSUE #654: cc-badge: border color visible in skeleton (dimmed & strong)
# Context

With #630 we've decided to go back to using a `border` instead of an inner `box-shadow`.
We've added border in `dimmed` and `strong` modes as well so that all modes have the same size.

I forgot to override the border also in skeleton mode which means we see the border color.
image
image

# Solution

Set border color to the same color as the skeleton background.

florian-sanders-cc:
Fixes #654

2 reviews should be enough for this one :wink:

### DIFF
--- a/src/components/cc-badge/cc-badge.js
+++ b/src/components/cc-badge/cc-badge.js
@@ -96,6 +96,7 @@ export class CcBadge extends LitElement {
         /* skeleton is more important */
 
         .skeleton {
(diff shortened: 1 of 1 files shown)


hsablonniere (approved):

pdesoyres-cc (approved):
LGTM

PR #658, opened 2022-12-02
fix(cc-stretch): rollback width max-content rule

### LINKED ISSUE #657: cc-stretch: rollback width: max-content rule
# Context

In order to fix a safari issue (#647), we removed the `width: max-content` rule from the `cc-stretch` host.

This breaks the simulation in storybook and forces developers to set a `width: max-content` in many contexts.

It appears that setting a `width` on the inner `` element inside the `<cc-badge>` was enough to solve the issue so we can rollback this fix.

florian-sanders-cc:
Fixes #657 

# What do you need to review ?

In a  local branch, the new "differentCssLayouts" should show all `cc-stretch` with the same `width` and `height`.

### DIFF
--- a/src/components/cc-stretch/cc-stretch.js
+++ b/src/components/cc-stretch/cc-stretch.js
@@ -75,6 +75,7 @@ export class CcStretch extends LitElement {
       css`
         :host {
           display: inline-grid;
+          width: max-content;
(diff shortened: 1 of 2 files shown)


Galimede (approved):
LGTM ! :wink: :+1:

hsablonniere (approved):

PR #671, opened 2022-12-14
docs(cc-example-component-smart): add new Events fired section

### LINKED ISSUE #626: cc-example-component-smart.md: add a new "Event" section
# Context

We know that some of our smart components need to fire events in some cases:

- to warn the console about an important change (for instance, a user leaves an organisation and needs to be redirected).
- to see what features are used successfully, etc.

# Issue 

There is currently no dedicated section to document events emitted by the smart component in our template.

# Solution

Add a new "Event" section containing a table with info about events fired by the smart component.

florian-sanders-cc:
Fixes #626 

# What does this PR do?

* add a new "Events fired" section,
* remove the example telling that we can include the smart component in the `md` file (now that we use `demo-smart` system, it feels obsolete),
* use markdown syntax for most of the tables (first one cannot be done with markdown because it contains no headers?).

# What & how to review?

* code and markdown preview

### DIFF
--- a/docs/cc-example-component.smart.md
+++ b/docs/cc-example-component.smart.md
@@ -20,25 +20,38 @@ You can add some details in parens when the selector is more complex.
   <tr><td><strong>Requires auth</strong> <td>Yes/No
 </table>
 
+## 👋️ Events fired
+
+<!-- Here we describe the different events fired by the smart component. -->
+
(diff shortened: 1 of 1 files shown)


hsablonniere (approved):

roberttran-cc (approved):
GG! :muscle:

PR #684, opened 2023-01-02
fix(cc-icon): force svg size to resolve safari 15 issue

### LINKED ISSUE #681: Icon component not working on Safari 15.3
Component `<cc-icon>` displays no icon on Safari 15.3. Works fine on Safari 16.0 and other browsers.

florian-sanders-cc:
Fixes #681

# What does this PR do & how to review?

* 2 reviewers should be enough for this
- [x] check on every browser that `cc-icon` stories are still displayed normally.
- [x] If you have a Safari 15 available, check the the icons are displayed as they should.

# More info about the bug and this fix

The bug seems to correspond to what is described here:
https://bugs.webkit.org/show_bug.cgi?id=221234

The reduced test case can be found here: https://codepen.io/florian-sanders-cc/pen/OJwXxpW
If you check this on Safari 15, the `<svg>` element is not visible unless you uncomment the `flex-direction` rule or the `width` rule.

To fix this in the `cc-icon` I went for the explicit `width: 100%;` and `height: 100%` rules because I think it feels less magical than setting a different `flex-direction`.

# TODO

- [x] Should we report on the webkit tracker that this issue seems to be fixed with safari 16? 
edit: after checking with @roberttran-cc, the WebKit issue is not fixed with Safari 16 but our issue is. It might not be the same issue after all so let's not comment the WebKit issue.

### DIFF
--- a/src/components/cc-icon/cc-icon.js
+++ b/src/components/cc-icon/cc-icon.js
@@ -132,9 +132,11 @@ export class CcIcon extends LitElement {
         }
 
         svg {
+          width: 100%;
+          height: 100%;
           fill: var(--cc-icon-color, currentColor);
         }
-        
+
         .skeleton {
           width: var(--size, 1em);
           height: var(--size, 1em);


roberttran-cc (approved):
Nice fix!

Tested and approved on:
- Chrome 108.0.5359.124 (64-bit)
- Firefox 108.0.1 (64-bit)
- Safari 16.2

And I think it is a good idea to leave a note on the webkit tracker!

Galimede (approved):
gg 👏 

LGTM on Safari 16.2 on macOS latest 13.2 Ventura

hsablonniere (approved):

hsablonniere:
@florian-sanders-cc thanks for the investigations :+1:

PR #685, opened 2023-01-09
fix(focus outline): use contrasted focus outline

### LINKED ISSUE #281: Introduce --cc-focus-ring

florian-sanders-cc:
Fixes #281

# How to review

:tada: You may now review the code.
:monocle_face: You should also review most of the stories simply by tabbing through their content: the focus should always look like below
Focus outline example on a cc-input-text. A blue outline with some offset so that we can see the border of the input as well as the outline at the same time
:art: If you don't like the styling of the focus outline, feel free to say so in a comment !

# What we discussed

There are several ways to standardize our focus styling and I'm not sure which way to go.

The styling is made of two rules:
```css
outline: var(--color-blue-60) solid 2px;
outline-offset: 2px;
```

We'd like to apply these rules to every focusable element (minus some exceptions).

Currently, we use explicit selectors inside components like:
```css
.btn:not([aria-disabled='true']):focus {
   outline: var(--cc-focus-outline);
   outline-offset: var(--cc-focus-outline-offset);
}
```

The drawback is that it's easy to forget it, for instance when you implement a component and you use a simple button, a link, a checkbox, an input without using one of our custom components (like the close buttons in `cc-notice` / `cc-toast`).

We could rely on a general selector like:
```css
*:focus {
  outline: var(--cc-focus-outline);
  outline-offset: var(--cc-focus-outline-offset);
}
```

This is what I would do for a website with a stylesheet but when it comes to components, I'm not sure how to make this work because we don't have a global stylesheet (unless we rely on the `default-theme.css` but it's not really it's purpose so that would be weird).

Should every component import a `a11y-styling.js` that would contain this selector?
Do we stick to what we currently have (explicit selectors inside individual components, relying on CSS custom properties)?

## Decision

Go for explicit selectors inside individual components for now and discuss the idea of a common shared style module (see #690).

# TODO

Following a discussion with @hsablonniere It looks like I forgot to add fallbacks for focus styles in case the theme is not loaded.

- [x] add fallback

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -378,9 +378,9 @@ export class CcButton extends LitElement {
 
         /* STATES */
 
-        .btn:not([aria-disabled='true']):focus {
-          box-shadow: 0 0 0 0.2em rgb(50 115 220 / 25%);
-          outline: 0;
+        .btn:focus {
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: var(--cc-focus-outline-offset, 2px);
         }
 
         .btn:not([aria-disabled='true']):hover {
--- a/src/components/cc-header-orga/cc-header-orga.js
+++ b/src/components/cc-header-orga/cc-header-orga.js
@@ -155,7 +155,8 @@ export class CcHeaderOrga extends LitElement {
         }
 
         .hotline_number:focus {
-          box-shadow: 0 0 0 0.2em rgb(50 115 220 / 25%);
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: var(--cc-focus-outline-offset, 2px);
         }
 
         .hotline_number:hover {
--- a/src/components/cc-input-number/cc-input-number.js
+++ b/src/components/cc-input-number/cc-input-number.js
@@ -405,15 +405,17 @@ export class CcInputNumber extends LitElement {
 
         input:focus + .ring {
           border-color: #777;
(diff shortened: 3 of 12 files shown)


roberttran-cc:
For now, I'd go with explicit selectors: more straightforward and we'll collectively try to not forget about it. :stuck_out_tongue_closed_eyes: 

But I think your question puts back on the table the topic of having a global stylesheet, which usually serves two goals in my opinion:
- resetting some rules, to prevent some unwanted behavior,
- applying some minimal styles, to prevent some side effects from integration contexts of our components.

Ultimately allowing us to have a better control on how our components are displayed.

pdesoyres-cc:
I totally agree with @roberttran-cc comment.

hsablonniere:
Tricky question :stuck_out_tongue: 

* I specifically avoided creating a `shared-styles.js` that every component would import because it would mean that many component would import stuffs they don't need.
* Reading your comments and others' feedbacks, I guess we could find an in between solution

I think I agree with @roberttran-cc 

*  explicit selectors for now
* new issue to discuss a global styles module imported and use by all our components
  * we'll have to define something to make sure this does not include to many specific stuffs
  * we'll have to include it in the vendors module on the smart CDN

Galimede:
I vote (agree with) for @roberttran-cc :+1:

Galimede (approved):
LGTM, gg :clap:

pdesoyres-cc (approved):
GG

hsablonniere (approved):

PR #691, opened 2023-01-17
chore: update jenkins agent docker image

hsablonniere:


roberttran-cc (approved):

pdesoyres-cc (approved):
LGTM.

Maybe, in the future, we should consider installing volta instead of a specific version of NodeJs. This way, we rely on the package.json to get the right NodeJs version.

florian-sanders-cc (approved):

hsablonniere:
> Maybe, in the future, we should consider installing volta instead of a specific version of NodeJs. This way, we rely on the package.json to get the right NodeJs version.

Agreed ;-)

FYI: I was very close to move this to a GitHub action :rofl:

pdesoyres-cc:
Before moving to GitHub action we need to discuss about do we allow storing credentials with GitHub secrets

PR #696, opened 2023-02-01
fix(cc-ssh-key-list): proper form validation when adding a new key in some cases

### LINKED ISSUE #632: cc-ssh-key-list: the cc-input doesn't show the trimmed value after a validation failure
### Description

When a user wants to add an ssh key (when he clicks on the add button), we trim the values before sending them to the server.
But when the user input is just some spaces (`"    "` for instance), we trim the input, (and it becomes an empty string), and the validation doesn't pass because the field is mandatory.

:bug: But, the input still has the untrimmed value which is something we don't want.

cc-ssh-key-list-trim-bug

### Analysis

The problem here is that the `cc-ssh-key-list` component and the `cc-input` have their own different model. Here is what is going on:

| step | cc-ssh-key-list | cc-input |
|---|---|---|
| cc-ssh-key-list render <sup>(1)</sup>| "" | "" |
| user input | "" | "   " |
| user submits | "   " | "   " |
| trim | "" | "   " |
| validation fails | "" | "   " |
| cc-ssh-key-list render <sup>(2)</sup>| "" | "   " |

<sup>(1)</sup> In the first step, the <cc-input> tag is rendered like below:
```<cc-input value=""></cc-input>```

<sup>(2)</sup> In the last step, the <cc-input> tag is rendered like below:
```<cc-input value=""></cc-input>```

:bug: Lit doesn't find any difference so the `<cc-ssh-key-list>` is not re-rendered.

We have the same kind of problem when the user submits a private key, but it's a bit different.

cc-ssh-key-list-trim-bug-2

| step | cc-ssh-key-list | cc-input |
|---|---|---|
| `<cc-ssh-key-list>` render <sup>(1)</sup> | "" | "" |
| user input | "" | "   a private key   " |
| user submits | "   a private key   " | "   a private key   " |
| trim | "a private key" | "   a private key   " |
| validation fails | "a private key" | "   a private key   " |
| cc-ssh-key-list render <sup>(2)</sup> | "a private key" | "a private key" |
| user input spaces around again | "a private key" | "   a private key   " |
| user submits | "   a private key   " | "   a private key   " |
| trim | "a private key" | "   a private key   " |
| validation fails | "a private key" | "   a private key   " |
| cc-ssh-key-list render <sup>(3)</sup> | "a private key" | "   a private key   " |

<sup>(1)</sup> In the first step, the <cc-input> tag is rendered like below:
```<cc-input value=""></cc-input>```

<sup>(2)</sup> In the last step, the <cc-input> tag is rendered like below:
```<cc-input value="a private key"></cc-input>```

:tada: Lit sees a difference, so the `<cc-ssh-key-list>` is re-rendered.

<sup>(3)</sup> In the last step, the <cc-input> tag is rendered like below:
```<cc-input value="a private key"></cc-input>```

:bug: Lit doesn't find any difference so the `<cc-ssh-key-list>` is not re-rendered.

### Wrap up

We use a mix of controlled and uncontrolled and this leads to very strange behavior. We should not do that.

Instead, we should sync the `<cc-ssh-key-list>` model with the `<cc-input>` model by listening to the `cc-input:input` event.

The code below tries to show how we should do that:
```

render () {
  return html`
    <cc-input-text
            .value=${this.myFormModel.myInput.value}
            @cc-input-text:input=${this._onMyInputChange}
        >
  `
}

_onMyFormSubmit () {
  // we trim the value we have in our model (instead of getting the value from the <cc-input> element itself)
  const value = this.myFormModel.myInput.value.trim();
  
  // we do the validation
  const error = validate(value);

  if (error != null) {
    // we change our model with the trimmed value
    this.myFormModel = {
        ...this.myFormModel,
        myInput: {
          value: value,
          error: error,
        },
      };
  }
  else {
	// we do whatever we need to do when validation is ok
  }
}

_onMyInputChange ({ detail: value }) {
   // with that, our internal model is always in sync with the <cc-input> model
   this.myFormModel = {
      ...this.myFormModel,
      myInput: {
        ...this.myFormModel.myInput,
        value: value
      },
    };
  }
```

### LINKED ISSUE #633: cc-ssh-key-list: private key validation rule is not applied in some cases
An ssh key is considered private when we detect the " private " value in the input. But when the user enters this exact string, the validator doesn't detect the invalid format.

### Steps to reproduce

1. enter a name
2. enter an ssh key: " PRIVATE "
3. validate
4. :bug: no error is triggered and the `cc-ssh-key-list:create` event is fired

### Analysis

The value is trimmed before the validation but the validation detects " private " (with spaces around).

### PR DESCRIPTION
**Pull request by roberttran-cc**
Fixes #632, Fixes #633.

### What this PR fixes

In the component creation form, model and view are getting out of sync in some cases (check the very detailed issues for more details).

### How to review

Check the code : the fix is quite straightforward (alongside some minor refactor).

UI test 1 :
- in the creation form, enter a random key name ("The front-end team is awesome" for example" :fire:)
- enter white-spaces `    `  in the public key field
- what was happening (on submit) : the white-spaces are still visible in the input, the error message for an empty input is shown
- what is now expected (on submit) : the input is cleared, the error message for an empty input is still shown

UI test 2 :
- in the creation form, enter a random key name ("Vote for Bob" for example" :eyes:)
- enter the following string `    this is a private key    ` (with the trailing white-spaces)  in the public key field
- what was happening :
	- (on first submit) the input shows the string *without* the trailing white-spaces, the error message for a private key is shown
	- at this point, clear and enter the initial string again (with the trailing white-spaces) in the public key field
	- (on second submit) the input shows the string *with* the trailing white-spaces, the error message for a private key is still shown
- what is now expected (on each submit) : the input now shows the trimmed string (`this is a private key`), the error message for a private key is still shown

PS : don't forget you can use the `demo-smart/index.html` page.

### CODE REVIEW

**Comment by pdesoyres-cc** [comment]
I made a preview for this PR: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/fix-form-validation/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--empty-story

**Review by pdesoyres-cc — APPROVED** [review]
Well done!

File: src/components/cc-ssh-key-list/cc-ssh-key-list.js [hunk 1]
```diff
@@ -141,6 +149,33 @@ export class CcSshKeyList extends LitElement {
...
+  _onPublicKeyInput ({ detail: publicKey }) {
+    // ensuring model is synchronised with the view
+    this.createSshKeyForm = {
+      ...this.createSshKeyForm,
+      publicKey: {
+        ...this.createSshKeyForm.publicKey,
+        value: publicKey,
+      },
+    };
+  }
```
hsablonniere, line 177: I think I understand why we need to go from uncontrolled form inputs to controlled but I wonder if a `live` could be enough?

I also know we shoot ourselves in the foot with `live` a few times :disappointed:

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -141,6 +149,33 @@ export class CcSshKeyList extends LitElement {
     }
   }
 
+  _isPublicKeyPrivate (publicKey) {
+    const lowerCasePublicKey = publicKey.toLowerCase();
+    return lowerCasePublicKey.match(' private ') || lowerCasePublicKey === 'private';
+  }
```
hsablonniere, line 155: This is a bit funny because, the whole point was to detect ` private ` surrounded by spaces to identify a real private key but now because of the trim thing, we also detect just the keyword `private`.

It's not really a problem but maybe we could simplify this a bit.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -109,29 +109,37 @@ export class CcSshKeyList extends LitElement {
...
-        error: (publicKey?.length > 0) ? null : 'required',
+        error: isPublicKeyFilled ? null : 'required',
       },
     };
 
-    // testing if submitted key is possibly a private key
-    if (publicKey.toLowerCase().match(' private ')) {
+    if (isPublicKeyPrivate) {
       this.createSshKeyForm.publicKey.error = 'private-key';
     }
```
hsablonniere, line 132: Is it a problem to trigger Lit's reactive thing with an assignement like `this.createSshKeyForm = {` and then modify the object?

hsablonniere (changes requested):
I haven't tested the preview yet. I have some questions but this seems like a nice way to fix this nasty bug. Well done @roberttran-cc

florian-sanders-cc (approved):
All good for me :wink:, gg!

src/components/cc-ssh-key-list/cc-ssh-key-list.js
pdesoyres-cc: Maybe we should call this function with the untrimmed value?

src/components/cc-ssh-key-list/cc-ssh-key-list.js
pdesoyres-cc: You're right, theoretically, we shouldn't do that. But I'm not sure Lit has 'the time' to react before the object is modified.

src/components/cc-ssh-key-list/cc-ssh-key-list.js
roberttran-cc: I'll give it a try!

src/components/cc-ssh-key-list/cc-ssh-key-list.js
roberttran-cc: > Maybe we should call this function with the untrimmed value?

Good idea! I'll give it a try!

roberttran-cc:
Updates from last push:
- `live()` usage for model/view synchronization (as the workflow is quite simple, we shouldn't fall into weird related problems)
- matching against ` private ` string is done against untrimmed input value
- setting the model on submit is done in one step (like `cc-org-member-list`)

hsablonniere (approved):

PR #698, opened 2023-02-03
feat(cc-tile-metrics): init component

### LINKED ISSUE #240: cc-tile-metrics: init small metrics tile component for overview

Galimede:
Fixes #240 

Hello everyone, this is it, `cc-tile-metrics` is (almost) finished. o/
While we have our base for the component, and we are quite happy with its current state, there is still some questions to discuss.

## Questions

* What link do we provide to the component when it's loading as we don't have the link to grafana or the console yet?
* Do we want to keep the `0.0%` as the skeleton of the current percentage or do we want to provide a grey background-color?
* We discussed the title/tooltip which is currently in english: `about this chart...`
  * We do agree that we will get rid of the `...`
    * Please note that this is the same wording for every tile and that we will need to do that in the others.
  * We think that we should provide a verb 
    * We haven't found a one that we are satisfied with, so if you have any idea feel free to let us know.
  * Cedric proposed: `More details in...`
* From the API if we the app is down for more than 24 hours we will have X points at the value 0. 
  * Imo, we should put it in empty mode and change the message of empty to reflect that state.
* What should we do if the component as less than 24 points?
    * Option 1: leave it as it is it's not the component to handle that
    * Option 2: the component handles that
    * Option 3: leave it as it it is for the component and the smart puts it in error mode if it detects fewer or more points than needed 

## Feedbacks

Here are some of the feedback that we had for the component. 
I'll let you react on them if you have remarks and/or questions that we can discuss.

* It lacks a legend 
* It lacks tooltip
  * Please note that if we do a tooltip we might do it after having some users feedback and we will do it at some point after the component is released.
* There's should be a mention of if it's CPU load and more details on the average.
* There has been suggestions of changing the title of the tile from Metrics => Metrics over 24h/Metrics (24h)
* The info text in french might extend too much the tile blocks in height. (Check next vs prod)

:warning: Please note that this PR is blocked from merge until we have a new endpoint that doesn't trigger adblockers on the console


pdesoyres-cc:
Here are some feedbacks.

### Questions

> What link do we provide to the component when it's loading as we don't have the link to grafana or the console yet?
What about a link in skeleton mode?

> Do we want to keep the 0.0% as the skeleton of the current percentage or do we want to provide a grey background-color?
I like the 0.0% blinking graph.

> We discussed the title/tooltip which is currently in english: about this chart...
Maybe "Show more details about this chart" / "Voir plus d'info à propos de ce graphique".

> From the API if we the app is down for more than 24 hours we will have X points at the value 0.
I think that the API should say when the app is down. two choices:
* API returns X points with value `null` when app is not running
* API returns only points when app is running.

I would prefer the first solution which makes it easier for the frontend to know what to display.

In terms of display for a given point, I can propose:
* value = null (app is not running): we display a transparent bar (or no bar at all)
* value = 0 (app is running bu no CPU load): we display a gray bar (as it is now)
* value > 0: we display a gray bar with a blue watermark (as it is now)

This way, we make a good difference between `app is not running` versus `app has no CPU load`.

### Feedbacks

> It lacks a legend
I don't see what legend we could add.

> It lacks tooltip
I'm not sure we should provide this level of detail here. This tile is just to give an **idea** of the load, and users still have the ability to go to grafana to get full detail.

> There's should be a mention of if it's CPU load and more details on the average.
I agree and we can use the "more info" panel to provide this information.

> There has been suggestions of changing the title of the tile from Metrics => Metrics over 24h/Metrics (24h)
I would be ok for Metrics (24h)

src/components/cc-tile-metrics/cc-tile-metrics.smart.md
```diff
@@ -0,0 +1,121 @@
...
+    "OAUTH_CONSUMER_SECRET": "",
+  },
+  "ownerId": "",
+  "appId": ""
+}'>
+  <cc-tile-status-codes></cc-tile-status-codes>
+</cc-smart-container>
+```
+
+<h1>CPU BENCH</h1>
```
pdesoyres-cc, line 90: Don't forget to remove this

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+
+const ONE_HOUR = 60 * 60 * 1000;
+
+function fakeMetricData (numberOfPoints, usedValue, linearIncrease = false) {
+  return Array
+    .from({ length: numberOfPoints })
+    .map((_, index) => {
+      const randomShift = (usedValue !== 0) ? (Math.random() * 4) - 2 : 0;
+      const increase = (linearIncrease) ? index * 1.5 : 0;
+      console.log('value', usedValue + randomShift + increase);
```
pdesoyres-cc, line 41: to remove

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+export default {
+  title: '🛠 Overview/<cc-tile-metrics>',
+  component: 'cc-tile-metrics',
+};
+
+const conf = {
+  component: 'cc-tile-metrics',
+  // language=CSS
+  css: `cc-tile-metrics {
+    margin-bottom: 1rem;
```
pdesoyres-cc, line 15: don't we always use em unit?

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+      };
+    });
+}
+
+function now () {
+  return new Date().getTime();
+}
+
+function addTimestamp (array) {
+  const startTs = now();
```
pdesoyres-cc, line 53: you can use `Date.now()` instead

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+  items: baseItems.map((item) => ({ ...item, metrics: { state: 'loading' } })),
+});
+
+export const empty = makeStory(conf, {
+  items: baseItems.map((item) => ({ ...item, metrics: { state: 'empty' } })),
+});
+
+export const error = makeStory(conf, {
+  items: baseItems.map((item) => ({ ...item, metrics: { state: 'error' } })),
+});
```
pdesoyres-cc, line 266: We often have those stories just after the default story

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+        state: 'loaded',
+        value: {
+          cpuData: addTimestamp(fakeMetricData(24, 25)),
+          memData: addTimestamp(fakeMetricData(24, 16)),
+        },
+      },
+    })),
+});
+
+export const highValues = makeStory(conf, {
```
pdesoyres-cc, line 76: rename into dataLoadedWithHighValues so that the story name is well decorated in Storybook menu. (and apply to all story names)

src/components/cc-tile-metrics/cc-tile-metrics.smart.js
```diff
@@ -0,0 +1,92 @@
...
+      });
+
+    fetchGrafanaAppLink({ apiConfig, ownerId, appId, grafanaBaseLink, signal })
+      .then((grafanaAppLink) => {
+        updateComponent('grafanaLink', grafanaAppLink);
+      })
+      .catch((error) => {
+        console.error(error);
+        // TODO: find a way to remove console logic from this
+        const pathname = ownerId.startsWith('user_')
```
pdesoyres-cc, line 47: Maybe a redirection logic inside the console may help us not have to do that here.

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
```diff
@@ -0,0 +1,30 @@
+interface Metric {
+  // Timestamp in ms
+  timestamp: number,
+  // Value is a percentage (e.g: '14.02')
+  value: string,
```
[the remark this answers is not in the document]
hsablonniere: Agreed. The stories send a number and you even try to divide it by 100 at some point. The smart part should convert if the API sends a string.

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+
+const TOP_THRESHOLD = 80;
+const BOTTOM_THRESHOLD = 20;
+
+const NUMBER_OF_POINTS = 24;
+
+const SKELETON_REQUESTS = Array
+  .from(new Array(NUMBER_OF_POINTS))
+  .map((_, index) => {
+    const startTs = new Date().getTime();
```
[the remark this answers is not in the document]
hsablonniere: I think I prefer to have a proper skeleton set of values but you're right, in the end, it won't change anything.

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
```diff
@@ -0,0 +1,30 @@
...
+  state: 'error';
+}
+
+interface TileMetricsStateEmpty {
+  state: 'empty';
+}
+
+interface TileMetricsStateLoaded {
+  state: 'loaded';
+  value: MetricsData;
```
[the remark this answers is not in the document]
hsablonniere: I was skeptical but with the suggestion below, it would mean `component.metrics.data.cpu` and I like that.

src/components/cc-tile-metrics/cc-tile-metrics.smart.js
hsablonniere: Yes, I wonder how we can properly avoid this and a redirection would be a solution.

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+      </div>
+      <div class="tile_body ${classMap({ 'tile--hidden': panel !== 'chart' })}">
+        <div class="category">
+          <div>${i18n('cc-tile-metrics.cpu')}</div>
+          <div class="chart-wrapper">
+            <div class="chart-container ${classMap({ skeleton: this.metrics.state === 'loading' })}">
+              <canvas id="cpu_chart"></canvas>
+            </div>
+          </div>
+          <div class="current-percentage ${classMap({ skeleton: this.metrics.state === 'loading' })}" style=${`color: ${cpuColorType}`}>${i18n('cc-tile-metrics.percent', { percent: lastCpuValue / 100 })}</div>
```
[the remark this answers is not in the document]
hsablonniere: For values between 0% and 1% it could be better but not sure

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
+import { BarController, BarElement, CategoryScale, Chart, LinearScale, Title, Tooltip } from 'chart.js';
+import { css, html, LitElement } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { i18n } from '../../lib/i18n.js';
+import { tileStyles } from '../../styles/info-tiles.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
+import { ccLink, linkStyles } from '../../templates/cc-link/cc-link.js';
+import '../cc-stretch/cc-stretch.js';
```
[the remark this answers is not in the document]
hsablonniere: Yes, because of funky FF bugs with canvas.

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+
+  }
+
+  render () {
+
+    const lastCpuValue = (this.metrics.state === 'loaded') ? this.metrics.value.cpuData[this.metrics.value.cpuData.length - 1].value : 0;
+    const lastMemValue = (this.metrics.state === 'loaded') ? this.metrics.value.memData[this.metrics.value.memData.length - 1].value : 0;
+
+    const cpuColorType = (this.metrics.state === 'loaded') ? this._getColorLegend(lastCpuValue) : '#bbb';
+    const memColorType = (this.metrics.state === 'loaded') ? this._getColorLegend(lastMemValue) : '#bbb';
```
hsablonniere, line 213: Someone already mentioned it elsewhere but the `#bbb` should be in a const.

src/components/cc-tile-metrics/cc-tile-metrics.js
hsablonniere, line 241: I was wondering why:
```
style=${`color: ${cpuColorType}`}
```
and not:
```
style="color: ${cpuColorType}"
```
but I'm not sure which one is the best.

Same for the other chart.

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        .grafana-logo {
+          display: block;
+          width: 1em;
+          height: 1em;
+        }
+
+        .docs-toggle,
+        .docs-grafana-btn {
```
hsablonniere, line 292: I think `.docs-grafana-btn` is no longer used

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+          /* We need this because: https://github.com/chartjs/Chart.js/issues/4156 */
+          height: 100%;
+        }
+
+        /* endregion */
+
+        /* region tile-body */
+
+        .tile_body {
+          display: grid;
```
hsablonniere, line 325: `.tile_body` is already `display: grid;` in the tile styles

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  // updated and not update because we need this._chart before
+  updated (changedProperties) {
+
+    if (changedProperties.has('metrics')) {
+
+      this._cpuChart.data = (this.metrics.state === 'loaded')
+        ? this._getChartData(this.metrics.value.cpuData)
+        : this._getChartData(null);
```
hsablonniere, line 192: Not sure why but I would prefer `SKELETON_REQUESTS` to be passed to `_getChartData` instead of being implicit with null.

Same for the other graph.

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,369 @@
...
+            enabled: false,
+          },
+        },
+      },
+    });
+  }
+
+  _getChartData (inputData) {
+
+    const data = inputData ?? SKELETON_REQUESTS;
```
hsablonniere, line 109: Not sure why but I would prefer `SKELETON_REQUESTS` to be passed to `_getChartData` instead of being implicit with null.

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
hsablonniere: I think the whole `css` config is no longer useful in this.

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+function now () {
+  return new Date().getTime();
+}
+
+function addTimestamp (array) {
+  const startTs = now();
+  return array.map((item, index) => {
+    return {
+      ...item,
+      timestamp: startTs + index * ONE_HOUR,
```
hsablonniere, line 57: These fake data are generated in the future. I know we don't use the timestamp (for tooltips) yet but I would prefer if we generated something in the past to match the API behaviour.

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -0,0 +1,324 @@
...
+    };
+  });
+}
+
+export const defaultStory = makeStory(conf, {
+  items:
+    baseItems.map((item) => ({
```
hsablonniere, line 64: There are lots of `items:` with an empty like afterwards. It feels a bit weird.

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export type TileMetricsState = TileMetricsStateLoading | TileMetricsStateError | TileMetricsStateLoaded | TileMetricsStateEmpty;
```
hsablonniere, line 30: Do we need the `Tile` prefix in all of these?

src/lib/api-helpers.js
```diff
@@ -142,3 +142,12 @@ export async function fetchAllZones ({ signal }) {
   return getAllZones()
     .then(sendToApi({ signal, cacheDelay: ONE_DAY }));
 }
+
+// TODO: move this to clever client
+export function getAppMetrics (params) {
+  return Promise.resolve({
+    method: 'get',
+    url: `/v4/metrics/organisations/${params.id}/resources/${params.appId}/metrics?interval="P1D"&span="PT1H"&only=cpu&only=mem`,
```
hsablonniere, line 150: The clever-client will generate something with `queryParams` and as a user, you'll have to pass `interval`, `span` and `only` as params. Maybe we can do this right now.

src/templates/cc-link/cc-link.js
```diff
@@ -14,11 +14,11 @@ function isDifferentOrigin (rawUrl) {
 }
 
 // NOTE: we could just create raw DOM but here we benefit from lit-html safe/escaping system on "content"
-export const ccLink = (url, content, skeleton = false) => {
+export const ccLink = (url, content, skeleton = false, title) => {
```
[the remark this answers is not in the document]
hsablonniere: We really need to create this `cc-link` component.

hsablonniere (changes requested):
:clap: Wow, impressive work 

* I prefer the lighter red we had before but I guess it was not dark enough to use it for the text
* The docs panel feels too simple, I'll try a few ideas...
* Can we let the bars grow a bit more than always 2em height ?
* I think the whole smart doc page is outdated with some copy/paste

NOTE: too late but I'll try to use https://conventionalcomments.org/ for my next review

## Questions

> What link do we provide to the component when it's loading as we don't have the link to grafana or the console yet?

I think we should not display it while we wait

> Do we want to keep the 0.0% as the skeleton of the current percentage or do we want to provide a grey background-color?

Grey bg color, no text

> We discussed the title/tooltip which is currently in english: about this chart... 

"Show details about this chart"

> From the API if we the app is down for more than 24 hours we will have X points at the value 0. 

I agree.

> What should we do if the component as less than 24 points? 

Tricky...

* The component should not handle that
* The API should always return 24 points
* Maybe we could check this in the smart definition and trigger an error

## Feedbacks

> It lacks a legend

I agree, and even if the idea is to display something simple we could try a legend like for the tile-requests.

> It lacks tooltip 

Agree on the let's do this later

> There's should be a mention of if it's CPU load and more details on the average.

Can we fit this in the docs panel ?

> There has been suggestions of changing the title of the tile from Metrics => Metrics over 24h/Metrics (24h)

I would like to keep Metrics, maybe "Server metrics" to explicit that it's about the server (as in VM).
I would try the 24h thing in a bottom text legend like tile-requests.

> The info text in french might extend too much the tile blocks in height. (Check next vs prod)

Yes, this one is annoying.

src/components/cc-tile-metrics/cc-tile-metrics.js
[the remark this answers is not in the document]
Galimede: Imo we want to give some precision when we're between 0 and 1. And that's the case a lot of time.

src/templates/cc-link/cc-link.js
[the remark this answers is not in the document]
Galimede: We'll keep it as it is and we'll change it when we'll do the `cc-link` component.

src/components/cc-tile-metrics/cc-tile-metrics.js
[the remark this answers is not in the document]
Galimede: Let's keep timestamps

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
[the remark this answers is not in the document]
Galimede: We have value for our other components that have states so we'll keep `value`

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
```diff
@@ -0,0 +1,30 @@
+interface Metric {
+  // Timestamp in ms
+  timestamp: number,
+  // Value is a percentage (e.g: '14.02')
+  value: string,
+}
+
+interface MetricsData {
+  cpuData: Metric[];
+  memData: Metric[];
```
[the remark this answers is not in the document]
Galimede: As we'll keep `value` instead of data we'll keep `memData`

src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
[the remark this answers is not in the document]
Galimede: The type hasn't been changed yet but it was already converted to a float if it was a string from the smart :wink:

src/components/cc-block/cc-block.js
```diff
@@ -160,7 +160,7 @@ export class CcBlock extends LitElement {
 
         ::slotted([slot='title']) {
           flex: 1 1 0;
-          color: var(--cc-color-text-strong);
+          color: var(--cc-color-text-primary-strongest);
```
pdesoyres-cc, line 163: nitpick: Could we move this into another commit (along with modification on cc-button and cc-env-var-form and default-theme.css)

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -367,3 +564,18 @@ export class CcTileMetrics extends LitElement {
 }
 
 window.customElements.define('cc-tile-metrics', CcTileMetrics);
+
```
pdesoyres-cc, line 567: chore: Don't forget to remove this

src/lib/chart-add-points.js
```diff
@@ -0,0 +1,57 @@
...
+
+export function roundHour (ts) {
+  return ts - (ts % ONE_HOUR_MS);
+}
+
+export function addMissingPoints (points) {
+
+  const fPoints = points.map(([timestamp, usedPercent]) => ({ timestamp, usedPercent: usedPercent / 100 }));
+
+  if (points.length === 24) {
```
pdesoyres-cc, line 12: todo: `24` should be a parameter

src/lib/chart-add-points.js
```diff
@@ -0,0 +1,57 @@
...
+  if (lastTs !== end / 1000 && arr.length < 24) {
+    arr.push(...generateMissing(lastTs, end / 1000));
+  }
+
+  return arr;
+}
+
+function generateMissing (startTs, end) {
+  const missingHours = (end - startTs) / 3600;
+  console.log('missingHours', missingHours - 1);
```
pdesoyres-cc, line 51: todo: Remove this log

test-warp10/results_ram.json
```diff
@@ -0,0 +1,60 @@
+{
```
pdesoyres-cc, line 1: todo: Invalid JSON file

test-warp10/request-metrics.cjs
```diff
@@ -0,0 +1,182 @@
+const fs = require('fs/promises');
```
pdesoyres-cc, line 1: question: Is this file now obsolete?

test/add-missing-points.test.js
```diff
@@ -0,0 +1,74 @@
+// import { expect } from '@bundled-es-modules/chai';
```
pdesoyres-cc, line 1: question: Do we really want to ignore those tests?

florian-sanders-cc:
Gg this looks nice !

Just a few details.

## Info text spacing & alignment

I think the info part of the tile lacks a tiny bit of spacing and when the "Plus d'informations" text wraps, I think we could improve the alignment of the links


Suggestion, we could add `align-items: flex-end` and a `margin-top: 0.5em`


## Table caption 

I forgot to tell you (sorry) we need a `<caption>` tag to provide a name for the table.
For instance:
```
<table>
  <caption>Server metrics (+ the name of the app if you have it?)</caption>
  ...
</table>
```

##  A11Y tests

The test file is missing (sorry since I did not review the whole PR, I only saw it now).
If you encounter a11y issues during automated tests, ping me and we'll fix it together :wink:

hsablonniere (changes requested):
@Galimede Wow impressive work!!

I mostly added some nitpicks and a few todos for files that should be removed.

Here are my real change requests:

* changing the color of the CPU and RAM icons breaks the `cc-addon-features` component. I think we need two icons for both of them while we wait or a proper full refactor of the new icon system.
* I would add a small space above the legends, something like `0.5em` or `0.25em`
* I would reduce the space between the "more metrics" and the links to be more like a regular text space

pdesoyres-cc (changes requested):
GG Mathieu!

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -0,0 +1,562 @@
...
+    this.metricsLink = '';
+
+    /** @type {boolean} */
+    this._docsPanelVisible = false;
+
+    /** @type {Ref<Canvas>} */
+    this._cpuCtxRef = createRef();
+
+    /** @type {Ref<Canvas>} */
+    this._memCtxRef = createRef();
```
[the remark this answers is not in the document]
Galimede: This was a request from Hubert but when discussing with him we agreed on not changing it due to litanalyzer plus we use `ctx` in other components such as `tile-requests`

src/translations/translations.en.js
```diff
@@ -778,6 +778,28 @@ export const translations = {
...
+  'cc-tile-metrics.close-btn': `Display chart`,
+  'cc-tile-metrics.docs.more-metrics': `More metrics: `,
+  'cc-tile-metrics.docs.msg': () => sanitize`<p>Server metrics received in the last 24 hours.</p> 
+    <p>Each bar represents a time window of <strong>1 hour</strong>.</p>
+    <p>The percentage shown is the average over the last hour.</p>`,
+  'cc-tile-metrics.empty': `No metrics. Application is stopped.`,
+  'cc-tile-metrics.error': `Something went wrong while loading metrics.`,
+  'cc-tile-metrics.grafana': `Grafana`,
+  'cc-tile-metrics.legend.cpu': `CPU load average over 24h`,
+  'cc-tile-metrics.legend.mem': `Memory usage average over 24h`,
```
hsablonniere, line 795: * `CPU usage over 24h`
* `RAM usage over 24h`

src/translations/translations.fr.js
```diff
@@ -791,6 +791,28 @@ export const translations = {
...
+  'cc-tile-metrics.close-btn': `Afficher le graphique`,
+  'cc-tile-metrics.docs.more-metrics': `Plus de métriques : `,
+  'cc-tile-metrics.docs.msg': () => sanitize`<p>Métriques reçues durant les dernières 24 heures.</p>
+    <p>Chaque barre représente une fenêtre de temps de <strong>1 heure</strong>.</p> 
+    <p>Le pourcentage affiché représente une moyenne sur la dernière heure.</p>`,
+  'cc-tile-metrics.empty': `Pas de métriques. L'application est arrêtée.`,
+  'cc-tile-metrics.error': `Une erreur est survenue pendant le chargement des métriques.`,
+  'cc-tile-metrics.grafana': `Grafana`,
+  'cc-tile-metrics.legend.cpu': `Charge moyenne du load CPU sur 24h`,
+  'cc-tile-metrics.legend.mem': `Charge moyenne de la RAM sur 24h`,
```
hsablonniere, line 808: * `Utilisation CPU sur 24h`
* `Utilisation RAM sur 24h`

src/translations/translations.en.js
```diff
@@ -778,6 +778,28 @@ export const translations = {
   'cc-tile-instances.status.running': `Running`,
   'cc-tile-instances.title': `Instances`,
   //#endregion
+  //#region cc-tile-metrics
+  'cc-tile-metrics.a11y.table-header.cpu': `Average CPU load`,
+  'cc-tile-metrics.a11y.table-header.mem': `Average memory usage`,
```
hsablonniere, line 783: * `CPU usage over 24h`
* `RAM usage over 24h`

src/translations/translations.fr.js
```diff
@@ -791,6 +791,28 @@ export const translations = {
   'cc-tile-instances.status.running': `En ligne`,
   'cc-tile-instances.title': `Instances`,
   //#endregion
+  //#region cc-tile-metrics
+  'cc-tile-metrics.a11y.table-header.cpu': `Moyenne d'utilisation du CPU`,
+  'cc-tile-metrics.a11y.table-header.mem': `Moyenne d'utilisation de la mémoire`,
```
hsablonniere, line 796: * `Utilisation CPU sur 24h`
* `Utilisation RAM sur 24h`

hsablonniere (changes requested):

hsablonniere (approved):

florian-sanders-cc (approved):

PR #704, opened 2023-02-14
fix(cc-ssh-key-list): creation form mistakenly updated

### LINKED ISSUE #703: cc-ssh-key-list: creation form mistakenly updated in some cases
How to reproduce:
- type some random values in an empty creation form
- immediately delete or import an existing key
- what is expected: creation form is left as is
- what is happening: creation form is cleared

The form is updated to the state it was before the action, meaning any modifications occurring after the previous action is lost.

roberttran-cc:
Fixes #703

## How the fix works

Removing the `live()` directive and manually synchronizing the view and model does the job.

## How to review

You'll have to checkout the branch to test locally with real data as you need to perform the delete or import action. Check the issue #703 for details on how to reproduce.


hsablonniere (approved):
Have you tried using `live`? Just kidding, thanks for this fix and sorry for  suggesting it in the first place.

PR #723, opened 2023-03-09
fix(eslint): check for improper sanitize function usage

### LINKED ISSUE #718: Translation sanitize function improper usage
`sanitize` function in `src/lib/i18n-sanitize.js` should be used as a template literal and not as a regular function, as it can generate unwanted behavior.

There is one place where this is the case, for the `cc-doc-card.link` translation, both `fr` and `en`.

We should fix this and also add an `eslint` custom rule to detect this scenario.

roberttran-cc:
Fixes #718.

# What the PR does

Adds a custom `eslint` rules that detects and fixes improper usage of the `sanitize` function in translation files.

Most patterns can be detected, like the following:
```js
{
  'cc-example.pattern-1': ({ foo }) => sanitize(`<em>improper ${foo} usage number 1</em>`),
  'cc-example.pattern-2': ({ foo }) => {
    return sanitize(`<em>improper ${foo} usage number 2</em>`);
  },
  'cc-example.complex-1': ({ foo }) => {
    return foo
      ? sanitize(`<em>improper ${foo} complex usage number 1</em>`)
      : `bar`;
  },
  'cc-example.complex-2': ({ foo }) => {
    const foobar = sanitize(`<em>improper ${foo} complex usage number 2</em>`);
    return foo ? foobar : `bar`;
  },
}
```

Those patterns are automatically fixed (with the `lint:fix` npm task) to:
```js
{
  'cc-example.pattern-1': ({ foo }) => sanitize`<em >improper ${foo} usage number 1</em>`,
  'cc-example.pattern-2': ({ foo }) => {
    return sanitize`<em >improper ${foo} usage number 2</em>`;
  },
  'cc-example.complex-1': ({ foo }) => {
    return foo
      ? sanitize`<em >improper ${foo} complex usage number 1</em>`
      : `bar`;
  },
  'cc-example.complex-2': ({ foo }) => {
    const foobar = sanitize`<em >improper ${foo} complex usage number 2</em>`;
    return foo ? foobar : `bar`;
  },
}
```

Note that the parentheses that were removed.

# How to review

You can check the diff or checkout the branch and test locally as well.

- There may be more edge cases: feel free to break the rule!
- Do not forget the wording (on CLI outputs).

# Limitations

Actually, there are more complex patterns that are detected but cannot be automatically fixed:

```js
{
  'cc-example.unfixable-1': ({ foo }) => sanitize(foo ? `<em>improper ${foo} unfixable usage number 1</em>` : `<em>foobar</em>`),
  'cc-example.unfixable-2': ({ foo }) => {
    return sanitize(foo ? `<em>improper ${foo} unfixable usage number 1</em>` : `<em>foobar</em>`);
  },
}
```

:arrow_right: Assuming those cases won't happen often and that the rule still detects them, we consider them non-blocking.

### DIFF
--- a/eslint-rules/i18n-always-template-literal-sanitize.js
+++ b/eslint-rules/i18n-always-template-literal-sanitize.js
@@ -1,0 +1,67 @@
+/**
+ * Rule to enforce using `sanitize` function as a template literal and not as a regular function.
+ * Check the `docs/translations.example.js` file for more details.
+ *
+ * Note: this rule only applies in translation files.
+ *
+ * Limitations: the automatic fix script only affects `sanitize()` call with a single parameter,
+ * which must be a template literal (`TemplateLiteral` AST type).
+ */
+
+'use strict';
+
+const {
+  getClosestParentFromType,
+  isTranslationFile,
+} = require('./i18n-shared.js');
+
+function report (context, key, callExpressionNode) {
+  context.report({
+    node: callExpressionNode,
+    messageId: 'sanitizeAlwaysTemplateLiteral',
+    data: { key },
+    fix: (fixer) => {
+      if (callExpressionNode.arguments?.length !== 1) {
+        return;
+      }
+
+      const argument = callExpressionNode.arguments[0];
+      if (argument.type === 'TemplateLiteral') {
+        const contents = context.getSourceCode().text.substring(argument.start, argument.end);
+        return fixer.replaceText(callExpressionNode, `sanitize${contents}`);
+      }
+    },
+  });
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce template literal when using the sanitize function',
+      category: 'Translation files',
+    },
+    fixable: 'code',
+    messages: {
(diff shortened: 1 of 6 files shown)


eslint-rules/i18n-always-template-literal-sanitize.js
```diff
@@ -0,0 +1,57 @@
...
+  getClosestParentFromType,
+  isTranslationFile,
+} = require('./i18n-shared.js');
+
+function report (context, key, callExpressionNode) {
+  context.report({
+    node: callExpressionNode,
+    messageId: 'sanitizeAlwaysTemplateLiteral',
+    data: { key },
+    fix: (fixer) => {
```
pdesoyres-cc, line 13: doc: Could you add what this fix support and what it doesn't?

pdesoyres-cc (changes requested):
Well done @roberttran-cc!

I just have a few nitpicks.

Galimede (approved):
GG @roberttran-cc! :clap:

pdesoyres-cc (approved):

PR #725, opened 2023-03-14
feat(cc-zone): add infra if any to the getText() static method

### LINKED ISSUE #724: cc-zone: getText() static method should contain the infra tag if any.
The getText static method provides a nice identifier for a given zone. If a zone has an infra tag, this tag must also be part of the identifier.

pdesoyres-cc:
Fixes #724

### DIFF
--- a/src/components/cc-zone/cc-zone.js
+++ b/src/components/cc-zone/cc-zone.js
@@ -61,23 +61,32 @@ export class CcZone extends LitElement {
   // This is a bit irregular to do this but we need to reuse this text logic in a <select>.
   // Moving this to a separated module feels overkill right now.
   static getText (zone) {
-    const { title, subtitle } = CcZone._getTextParts(zone);
-    return [title, subtitle].filter((a) => a != null).join(', ');
+    const { title, subtitle, infra } = CcZone._getTextParts(zone);
+    const titleAndSubtitle = [title, subtitle].filter((a) => a != null).join(', ');
+    return (infra != null)
+      ? `${titleAndSubtitle} (${infra})`
+      : titleAndSubtitle;
   }
 
   static _getTextParts (zone) {
-    return (zone.tags.includes(PRIVATE_ZONE) && zone.displayName != null)
-      ? { title: zone.displayName }
-      : { title: zone.city, subtitle: i18n('cc-zone.country', { code: zone.countryCode, name: zone.country }) };
+    if (zone.tags.includes(PRIVATE_ZONE) && zone.displayName != null) {
+      return { title: zone.displayName };
+    }
+
+    const infraTag = zone.tags.find((t) => t.startsWith('infra:'));
+    const infraSlug = infraTag?.split(':')[1] ?? null;
+    return {
+      title: zone.city,
+      subtitle: i18n('cc-zone.country', { code: zone.countryCode, name: zone.country }),
+      infra: infraSlug,
+    };
   }
 
   render () {
 
     const skeleton = (this.zone == null);
     const zone = skeleton ? SKELETON_ZONE : this.zone;
-    const { title, subtitle } = CcZone._getTextParts(zone);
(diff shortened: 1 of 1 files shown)


src/components/cc-zone/cc-zone.js
```diff
@@ -87,8 +94,8 @@ export class CcZone extends LitElement {
             <span class="title ${classMap({ skeleton })}">${title}</span>
             <span class="subtitle ${classMap({ skeleton })}">${subtitle}</span>
           </div>
-          ${infraProviderSlug != null ? html`
-            <cc-img class="infra-logo" src=${ifDefined(getInfraProviderLogoUrl(infraProviderSlug))} text=${ifDefined(infraProviderSlug)}></cc-img>
+          ${infra != null ? html`
+            <cc-img class="infra-logo" src=${ifDefined(getInfraProviderLogoUrl(infra))} text=${ifDefined(infra)}></cc-img>
```
roberttran-cc, line 98: **🎯 suggestion:** I think we can safely remove the `ifDefined` directive here (both usage), as `infra` will always be defined.

roberttran-cc (approved):
Nice and clean PR, making the code cleaner step by step! :muscle: 

I commented with a question and a suggestion, but both non-blocking so I'll approve as well!

package.json
```diff
@@ -31,7 +31,7 @@
     "test:cem:watch": "mocha --watch cem/test/*.js",
     "lint": "eslint --rulesdir eslint-rules rollup/*.js cem src tasks test wds",
     "lint:fix": "eslint --rulesdir eslint-rules --fix rollup/*.js cem src tasks test wds",
-    "prepack": "npm run test && npm run lint && npm run components:check-lit && npm run typecheck && npm run components:check-i18n && npm run components:build && npm run components:check-type-imports && npm run components:docs && npm run stylelint",
+    "prepack": "npm run components:build",
```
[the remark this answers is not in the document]
pdesoyres-cc: Oups, this is an horrible mistake. I'll revert that!

src/components/cc-zone/cc-zone.js
pdesoyres-cc: `infra` can still be `null` if you look closely at the `CcZone._getInfraProviderSlug` function

roberttran-cc: My bad, I should've been more clear: it will always be defined considering the predicate line 97 `infra != null ? ...` outputting the HTML.

florian-sanders-cc (approved):
Gg !
Nothing more to add compared to @roberttran-cc 's review :wink:

src/components/cc-zone/cc-zone.js
pdesoyres-cc: You're absolutely right, I've just fixed that

Galimede (approved):
GG! :clap:

src/components/cc-zone/cc-zone.js
```diff
@@ -87,8 +94,8 @@ export class CcZone extends LitElement {
             <span class="title ${classMap({ skeleton })}">${title}</span>
             <span class="subtitle ${classMap({ skeleton })}">${subtitle}</span>
           </div>
-          ${infraProviderSlug != null ? html`
-            <cc-img class="infra-logo" src=${ifDefined(getInfraProviderLogoUrl(infraProviderSlug))} text=${ifDefined(infraProviderSlug)}></cc-img>
+          ${infra != null ? html`
+            <cc-img class="infra-logo" src=${getInfraProviderLogoUrl(infra)} text=${infra}></cc-img>
```
hsablonniere, line 98: praise: nice simplification :+1:

hsablonniere (changes requested):

hsablonniere (approved):
LGTM :+1: Well done Pierre :clap:

PR #727, opened 2023-03-15
feat(cc-overview): add cc-tile-metrics to the app-mode story

### LINKED ISSUE #722: cc-overview: replace consumption with tile-metrics in the app-mode story

Galimede:
Fixes #722


pdesoyres-cc (approved):

roberttran-cc (approved):

src/components/cc-overview/cc-overview.stories.js
```diff
@@ -28,6 +28,10 @@ const conf = {
     cc-overview {
       min-height: 750px;
     }
+    
+    cc-tile-metrics {
+      max-width: 100% !important;
+    }
```
hsablonniere, line 34: Curious to know why we need this

hsablonniere (changes requested):

src/components/cc-overview/cc-overview.stories.js
Galimede: We need that because we have a min-width on the `tile-metrics` stories and if we don't do that the tile doesn't scale in width and isn't responsive

PR #728, opened 2023-03-16
refactor: make use of the new cc-notice instead of cc-error

### LINKED ISSUE #645: refactor: make use of the new cc-notice in place of cc-error
## Context

With the addition of the new `cc-notice` component there are places where we could use it instead of our actual `cc-error` component.

## Components concerned

- [ ] cc-addon-admin (handled by #777)
- [x] cc-addon-backups
- [x] cc-addon-credentials
- [x] cc-addon-elasticsearch-options
- [x] cc-addon-encryption-at-rest-option
- [x] cc-addon-linked-apps
- [x] cc-addon-features
- [x] cc-addon-jenkins-options
- [x] cc-addon-mongodb-options
- [x] cc-addon-mysql-options
- [x] cc-addon-postgresql-options
- [x] cc-addon-redis-options
- [x] cc-addon-option-form (in story, slotted)
- [x] cc-addon-option (in story, slotted)
- [x] cc-article-list
- [x] cc-block (in story, slotted)
- [x] cc-doc-list
- [x] cc-elastic-search-info
- [x] cc-jenkins-info
- [x] cc-email-list
- [x] cc-env-var-create
- [x] cc-env-var-editor-expert
- [x] cc-env-var-editor-json
- [x] cc-env-var-linked-services
- [x] cc-env-var-form
- [x] cc-grafana-info
- [x] cc-header-addon
- [x] cc-header-app
- [x] cc-header-orga
- [x] cc-heptapod-info
- [x] cc-matomo-info
- [x] cc-pricing-estimation
- [x] cc-pricing-product
- [x] cc-pricing-product-consumption
- [x] cc-map
- [x] cc-tcp-redirection-form
- [x] cc-tile-consumption
- [x] cc-tile-deployments
- [x] cc-tile-instances
- [x] cc-tile-requests
- [x] cc-tile-status-codes
- [x] cc-invoice-list
- [x] cc-zone-input

## Problem

A few tests has already been made on some components and already some problems were found:
* components like `cc-header-*` are weird looking with the whole notice in the "box" on error
* it also seems to not be a valid option on components that have the cc-error on the front with the background of the component blurred on error

Galimede:
Fixes #645

Preview here: https://clever-components-preview.cellar-c2.services.clever-cloud.com/refactor-cc-notice/index.html

### DIFF
--- a/src/components/cc-addon-option-form/cc-addon-option-form.js
+++ b/src/components/cc-addon-option-form/cc-addon-option-form.js
@@ -100,6 +100,24 @@ export class CcAddonOptionForm extends LitElement {
         [name='description'] {
           line-height: 1.5;
         }
+
+        .option-warning {
+          display: grid;
+          gap: 0.5em;
+          grid-template-columns: min-content 1fr;
+          text-align: left;
+        }
+
+        .option-warning p {
+          margin: 0;
+        }
+
+        .icon-warning {
+          align-self: center;
+          color: var(--cc-color-text-warning);
+
+          --cc-icon-size: 1.25em;
+        }
       `,
     ];
   }
--- a/src/components/cc-addon-option-form/cc-addon-option-form.stories.js
+++ b/src/components/cc-addon-option-form/cc-addon-option-form.stories.js
@@ -1,17 +1,21 @@
 import './cc-addon-option-form.js';
-import '../cc-error/cc-error.js';
 import { html } from 'lit';
+import {
+  iconRemixAlertFill as iconAlert,
+} from '../../assets/cc-remix.icons.js';
 import { makeStory } from '../../stories/lib/make-story.js';
 import { enhanceStoriesNames } from '../../stories/lib/story-names.js';
 
 const htmlExample = html`
-<div class="option-details">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed libero risus,
-porttitor et turpis sed, mollis ultricies quam. Sed quis fermentum sem, sed dictum sapien.
-Donec rutrum ante vel dolor bibendum, eu pretium velit gravida</div>
-<cc-error class="option-warning">
-Nullam non nulla convallis, tincidunt nibh at, blandit eros. Cras arcu quam, faucibus eget neque id,
-scelerisque ornare neque
-</cc-error>
+  <div class="option-details">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed libero risus,
+    porttitor et turpis sed, mollis ultricies quam. Sed quis fermentum sem, sed dictum sapien.
+    Donec rutrum ante vel dolor bibendum, eu pretium velit gravida
+  </div>
+  <div class="option-warning">
+    <cc-icon .icon="${iconAlert}" accessible-name="Warning" class="icon-warning"></cc-icon>
+    <p>Nullam non nulla convallis, tincidunt nibh at, blandit eros. Cras arcu quam, faucibus eget neque id,
+      scelerisque ornare neque</p>
+  </div>
 `;
 
 export default {
--- a/src/components/cc-addon-option/cc-addon-option.stories.js
+++ b/src/components/cc-addon-option/cc-addon-option.stories.js
@@ -1,4 +1,3 @@
-import './cc-addon-option.js';
 import { iconRemixShieldKeyholeFill as iconEncryptionAtRest } from '../../assets/cc-remix.icons.js';
 import { makeStory } from '../../stories/lib/make-story.js';
 import { enhanceStoriesNames } from '../../stories/lib/story-names.js';
@@ -12,10 +11,10 @@ const htmlExample = `
 <div class="option-details">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed libero risus,
 porttitor et turpis sed, mollis ultricies quam. Sed quis fermentum sem, sed dictum sapien.
 Donec rutrum ante vel dolor bibendum, eu pretium velit gravida</div>
-<cc-error class="option-warning">
-Nullam non nulla convallis, tincidunt nibh at, blandit eros. Cras arcu quam, faucibus eget neque id,
-scelerisque ornare neque
-</cc-error>
+<div class="option-warning">
+    <p>Nullam non nulla convallis, tincidunt nibh at, blandit eros. Cras arcu quam, faucibus eget neque id,
+      scelerisque ornare neque</p>
+</div>
 `;
 
 const optionExample = {
--- a/src/components/cc-env-var-create/cc-env-var-create.js
+++ b/src/components/cc-env-var-create/cc-env-var-create.js
@@ -1,6 +1,6 @@
 import '../cc-button/cc-button.js';
 import '../cc-input-text/cc-input-text.js';
-import '../cc-error/cc-error.js';
+import '../cc-notice/cc-notice.js';
 import { validateName } from '@clevercloud/client/esm/utils/env-vars.js';
 import { css, html, LitElement } from 'lit';
 import { dispatchCustomEvent } from '../../lib/events.js';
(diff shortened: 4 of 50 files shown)


florian-sanders-cc (changes requested):
Gg @Galimede so many commits and changes :smile: .

I really like the way our components look like when they're in error (though I do hope people won't get to see this state very often).

Overall it feels more consistent and clean :clap:.

My comments are mostly nitpicks about two things:

- using properties to pass plain strings to the components,
- using slots instead of props to insert plain strings.

These do not cause any harm so I'm not even sure if it was relevant to point them out or not (it depends on whether we want to enforce this kind of usage or not, which I don't really know).

Here are a few comments that are up for debate:

- `cc-orga-member-card` / `cc-orga-member-list` I prefer the `danger` intent for these errors as it was before (but if more people prefer warning, let's stick with that),
- I think we can remove the `cc-env-var-form (error and readonly)` story since it's the same as `error` now.
- I think we could improve the translation in `cc-env-var-form` (error saving) => Change "Something went wrong while loading variables." with "Something went wrong while loading variables".
- I don't know where to discuss this but the `cc-addon-option` component implements classes supposed to be used within a slotted element (`.error-warning`). I don't know enough about this component yet to suggest a solution but at least we can document it I guess :+1:

src/components/cc-addon-option/cc-addon-option.stories.js
```diff
@@ -11,10 +15,11 @@ const htmlExample = `
 <div class="option-details">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed libero risus,
 porttitor et turpis sed, mollis ultricies quam. Sed quis fermentum sem, sed dictum sapien.
 Donec rutrum ante vel dolor bibendum, eu pretium velit gravida</div>
-<cc-error class="option-warning">
-Nullam non nulla convallis, tincidunt nibh at, blandit eros. Cras arcu quam, faucibus eget neque id,
-scelerisque ornare neque
-</cc-error>
+<div class="option-warning">
+    <cc-icon .icon="${iconAlert}" .accessibleName="Warning" class="icon-warning"></cc-icon>
```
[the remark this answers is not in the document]
pdesoyres-cc: I agree, we should not try to use cc-icon here. Show how to slot a simple paragraph is enough for this story.

src/components/cc-addon-option-form/cc-addon-option-form.js
```diff
@@ -99,6 +99,25 @@ export class CcAddonOptionForm extends LitElement {
...
+          grid-template-columns: min-content 1fr;
+          text-align: left;
+        }
+
+        .option-warning p {
+          margin: 0;
+        }
+
+        .icon-warning {
+          /* TODO: center or not center that is the question. */
```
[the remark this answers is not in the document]
pdesoyres-cc: If you need my opinion, I would say: let's go for centering it.

src/components/cc-env-var-editor-json/cc-env-var-editor-json.js
```diff
@@ -157,12 +157,12 @@ export class CcEnvVarEditorJson extends LitElement {
       ${this._errors.length > 0 ? html`
         <div class="error-list">
           ${this._errors.map(({ msg, isNotice }) => html`
```
pdesoyres-cc, line 159: The `isNotice` becomes confusing. We could replace it by the intent directly or `isWarning`.
By the way, the `this._errors` is of type `ParseError` which does not describe the `isNotice` property. Maybe its the right moment to fix that.
Also, the `line` property of the `ParseError` should be optional.

src/components/cc-tile-deployments/cc-tile-deployments.js
```diff
@@ -1,7 +1,11 @@
 import '../cc-datetime-relative/cc-datetime-relative.js';
-import '../cc-error/cc-error.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-notice/cc-notice.js';
```
pdesoyres-cc, line 3: fix: In fact you don't need cc-notice import

src/components/cc-tile-instances/cc-tile-instances.js
```diff
@@ -1,8 +1,12 @@
 import '../cc-expand/cc-expand.js';
+import '../cc-icon/cc-icon.js';
 import '../cc-loader/cc-loader.js';
-import '../cc-error/cc-error.js';
+import '../cc-notice/cc-notice.js';
```
pdesoyres-cc, line 4: fix: not needed

src/components/cc-tile-scalability/cc-tile-scalability.js
```diff
@@ -1,7 +1,11 @@
-import '../cc-error/cc-error.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-notice/cc-notice.js';
```
pdesoyres-cc, line 2: fix: not needed

pdesoyres-cc (changes requested):
Well done Mathieu. I had just some small comments.

src/components/cc-env-var-create/cc-env-var-create.js
```diff
@@ -126,17 +126,33 @@ export class CcEnvVarCreate extends LitElement {
       </div>
       
       ${(isNameInvalidStrict && this.mode === 'strict' && this._variableName !== '') ? html`
-        <cc-error>${i18n(`cc-env-var-create.errors.invalid-name`, { name: this._variableName })}</cc-error>
+        <cc-notice intent="warning">
+          <div slot="message">
+            ${i18n(`cc-env-var-create.errors.invalid-name`, { name: this._variableName })}
+          </div>
+        </cc-notice>
```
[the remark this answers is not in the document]
Galimede: The reason behind `slots` is that if I don't use them, the color of the link will be wiped out.
I haven't checked if there's a way to handle that and not use slots.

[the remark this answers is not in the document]
florian-sanders-cc: I had not thought about this :+1: 

**suggestion:** `cc-notice` could import the link styles, WDYT?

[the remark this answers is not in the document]
Galimede: This would do the trick for the links but not for the `code` sadly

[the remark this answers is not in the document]
florian-sanders-cc: Then I agree with you, let's stick with slots (thanks for the answers) :+1:

pdesoyres-cc (approved):
LGTM. Well done!

florian-sanders-cc (changes requested):
Gg @Galimede I bumped a few comments that you might have missed about properties used instead of attributes.

If you think we should not care about this type of nitpicks, resolve the threads or tell me, but I preferred bumping because I saw you fixed some of these but not all of them.

Other than these trivial nitpicks I only have an issue with `cc-orga-member-list`. I think the `cc-notice intent` in `error loading` should be `warning`. 

I think `danger` is fine for errors when you try to delete an admin or yourself.

florian-sanders-cc (approved):
Gg @Galimede, great work! 
All good for me :wink:

src/components/cc-addon-option-form/cc-addon-option-form.stories.js
```diff
@@ -12,7 +12,7 @@ const htmlExample = html`
     Donec rutrum ante vel dolor bibendum, eu pretium velit gravida
   </div>
   <div class="option-warning">
-    <cc-icon .icon="${iconAlert}" accessible-name="${'cc-tile-deployments.error.icon-a11y-name'}" class="icon-warning"></cc-icon>
+    <cc-icon .icon="${iconAlert}" accessible-name="Warning" class="icon-warning"></cc-icon>
```
hsablonniere, line 15: Why did we go for a translated message to a hard coded english one?

hsablonniere (changes requested):
Hey @Galimede, impressive and useful work, congrats :clap:

* Do we still need/user cc-error ?
  * Seems like we don't (apart from the cc-addon-admin but next PR takes care of this)
  * It's still imported but not used in cc-addon-*-options components, I think you missed this
* Would it be a good idea to use cc-notice info for "empty" states instead of our classic italic sentence?
  * I'm in favor of a discussion on this but in another issue and another PR, WDYT?
* We kept the cc-notice warning inside a cc-block for cc-elasticsearch-info
  * I mean, this is the only one where we kept the padding around instead of "merging" it like we do on the others
* The spacing around the cc-notice for cc-env-var-form (error loading) is not consistent with what we do for other components
* I know we decided left aligned for errors on tiles but it seems a bit odd
  * Not blocking, we'll revisit this later if necessary
* Is there a rule for message attribute vs slot?
  * Seems like the rule is "if you don't need any HTML, use the attribute", should we document it somewhere?
* I like the new early returns for error :+1:
* Some commits are named "use cc-notice..." but don't use cc-notice
* I like this one: `feat(cc-block): get rid of overlay error story` :+1:

src/components/cc-addon-option-form/cc-addon-option-form.stories.js
Galimede: Well, it was a story and I found it overkill to add a translation just for this. But I can if you want, just tell me.

Galimede:
> Do we still need/user cc-error ? 
* Seems like we don't indeed, I'll remove it on the `cc-addon-admin` before merging, WDYT?
* _It's still imported but not used in cc-addon-*-options components, I think you missed this_ => Thanks for the catch that's my bad.

> Would it be a good idea to use cc-notice info for "empty" states instead of our classic italic sentence? 

Yeah, I will create another issue for this but it seems like a good idea to add up some coherence. :wink: 

> We kept the cc-notice warning inside a cc-block for cc-elasticsearch-info 

If I remember well we decided this due to the info "banner".

> The spacing around the cc-notice for cc-env-var-form (error loading) is not consistent with what we do for other components

I will have a look. :wink: 

> I know we decided left aligned for errors on tiles but it seems a bit odd. Not blocking, we'll revisit this later if necessary.

Ok let's stick with that for the moment.

> Is there a rule for message attribute vs slot? 

I don't think so, I think we'll have to discuss this before documenting it.

> Some commits are named "use cc-notice..." but don't use cc-notice

I will try to find a better naming :wink:

src/components/cc-addon-option-form/cc-addon-option-form.stories.js
hsablonniere: NP

PR #736, opened 2023-04-05
tokens: add neutral border colors and use them when relevant

### LINKED ISSUE #733: design tokens (colors): introduce border colors and use it where relevant
# Context

For the pricing components, we need to be able to override the default border color.

We need to add new tokens to allow border color customization:

- [ ] `--cc-color-border-default` and use it in:
  - [ ] `cc-addon-option.js`
  - [ ] `cc-article-card.js`
  - [ ] `cc-article-list.js`
  - [ ] `cc-block.js`
  - [ ] `cc-doc-card.js`
  - [ ] `cc-doc-list.js`
  - [ ] `cc-env-var-form.js`
  - [ ] `cc-env-var-linked-services.js`
  - [ ] `cc-error.js`
  - [ ] `cc-header-addon.js`
  - [ ] `cc-header-app.js`
  - [ ] `cc-header-orga.js`
  - [ ] `cc-input-number.js`
  - [ ] `cc-input-text.js`
  - [ ] `cc-logsmap.js`
  - [ ] `cc-map.js`
  - [ ] `cc-select.js`
  - [ ] `cc-ssh-key-list.js`
  - [ ] `cc-zone-input.js`
- [ ] `--cc-color-border-hovered` (only to be used within the pricing component for the moment) => Should this be a component level token instead?
- [ ] `--cc-color-border-focused` (only to be used within the pricing component for the moment) => Should this be a component level token instead?

florian-sanders-cc:
Fixes #733

## TODO

- [x] decide on how many tokens we want
- [x] decide on colors used by each of the tokens we introduce
- [x] decide on token naming strategy
- [x] rebase

## How to review?

- check the commits (using GitHub should be enough),
- check all the components impacted (do the new border colors work for all of them?):

  - `cc-addon-option.js`,
  - `cc-article-card.js`,
  - `cc-article-list.js`,
  - `cc-block.js`,
  - `cc-block-section.js`,
  - `cc-doc-card.js`,
  - `cc-doc-list.js`,
  - `cc-env-var-form.js`,
  - `cc-env-var-linked-services.js`,
  - `cc-error.js`,
  - `cc-header-addon.js`,
  - `cc-header-app.js`,
  - `cc-header-orga.js`,
  - `cc-input-number.js`,
  - `cc-input-text.js`,
  - `cc-invoice-table.js`,
  - `cc-logsmap.js`,
  - `cc-map.js`,
  - `cc-select.js`,
  - `cc-ssh-key-list.js`,
  - `cc-tile-consumption.js`,
  - `cc-tile-scalability.js`,
  - `cc-zone-input.js`,
  - (all tiles).

<details>
<summary>Archived discussion</summary>

### Two tokens 
This one was a little bit trickier than anticipated in the issue.

We can distinguish at least two types of "neutral" borders:

- **strong borders**, usually for interactive controls (form controls). For these, we want users to clearly visualize the boundaries of the control because they need to click inside it. => **need at least `3:1` of contrast ratio**.
- **subtle borders**, usually to suggest boundaries that already are visible through the layout & spacing. For instance cards, table rows, etc. These are somewhat "decorative" (it's definitely debatable). They are not essential to the interface but they emphasize something that should already stand out (with good use of spacing and visual headings). => **no minimum contrast ratio** 

This means we probably need not one but two different tokens.

### Token naming

Working on this made me realize that most of the "neutral" borders we use are subtle (card borders, separators).

Strong borders are only required in a few of our components.

I suggest we change the naming of the `default` token since it's actually not going to be used that often:
`--cc-color-border-strong`.

Also, I wonder if we should mention "neutral" in their name like we do for background tokens?
`--cc-color-border-neutral-strong`
`--cc-color-border-neutral-weak`

### Do we need a 3rd token?

For subtle borders, we have different cases that may require different tokens.

Card & block borders, where we currently have fairly visible (although < 3:1 of contrast ratio) border:
the cc-block component

Separators, where we currently have less visible borders:
image

Check out these 3 previews to see our options:

- [one accessible token to rule them all](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tokens/add-border-colors-full-default/index.html?path=/story/%F0%9F%A7%AC-molecules-cc-block-section--default-story)
- [two tokens (strong for input borders, weak for everything else like card borders and separators)](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tokens/add-border-colors/index.html?path=/story/%F0%9F%A7%AC-molecules-cc-block-section--default-story)
- [three tokens (strong for input borders, weak for card borders, weaker for separators](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tokens/add-border-colors-three-shades/index.html?path=/story/%F0%9F%A7%AC-molecules-cc-block-section--default-story)

If we use a very subtle grey for both, or we go with 3 tokens:

- `--cc-color-border-neutral-strong` => for input, select, etc.
- `--cc-color-border-neutral` => for cards etc.
- `--cc-color-border-neutral-weak` => for row separators for instance.

or, if we consider `default` should mean it's accessible, we could go for:

- `--cc-color-border-neutral-default` => for input, select, etc.
- `--cc-color-border-neutral-weak` => for cards etc.
- `--cc-color-border-neutral-weaker` => for row separators for instance.

## Meeting about this subject

During the meeting, we discussed how important it was to ensure the token system remain as simple as possible by limiting how many tokens we introduce.
We all agreed that 3 tokens was the best option for "design" as it allows for more subtle design choices.
Some of us felt like 2 tokens were also enough for decent design and that it may not be worth it to add a third token.

In the end, I chose to go with 3 tokens because:

- before introducing tokens, it seems we were mostly relying on 3 shades of grey anyway, which means there was a need for these 3 shades,
- generating grey shades is fairly easy to do so later on, we may easily "hide" the complexity of overriding the border colors by letting developers only select the strongest grey and compute other shades from there (for grey shades, HSL system is ok),
- while we all agreed on the benefits of having 3 shades, it seemed we did not all agree that we could make it work with only two shades.

</details>

### DIFF
--- a/src/styles/default-theme.css
+++ b/src/styles/default-theme.css
@@ -1,6 +1,8 @@
 :root {
   --cc-ff-monospace: "SourceCodePro", "monaco", monospace;
-/* All custom properties are sorted alphabetically within each region. */
+  
+  /* All custom properties are sorted alphabetically within each region. */
+  
   /*region Color choices*/
   /* color choices - these should not be referenced inside components (use decisions below instead)
     tool used for grey shades: https://mdigi.tools/color-shades/#333333 (10 or 11 steps)
@@ -52,17 +54,13 @@
   /* Usage: Sensitive information that require attention.
   For instance: error messages */
   --cc-color-text-danger: var(--color-red-100);
-  /* Usage: regular text.
-  -- Not used at the moment - will be when we allow theme override. -- */
+  
+  /* Usage: regular text. */
   --cc-color-text-default: var(--color-grey-90);
 
   /* Usage: Opposite of default text color - use this with plain backgrounds like primary / success / danger, etc.
   For instance: text inside primary cc-button. */
   --cc-color-text-inverted: var(--color-white);
-
-  /* Usage: text less important than normal text.
-  For instance: sidenotes, the required mention inside forms. */
-  --cc-color-text-weak: var(--color-grey-80);
 
   /* Usage: info, main color used through the application.
   For instance: text color inside outlined primary cc-button. */
@@ -91,6 +89,10 @@
   /* Usage: cautionnary text.
   For instance: text saying an app is stopped. */
   --cc-color-text-warning: var(--color-orange-100);
+
+  /* Usage: text less important than normal text.
+  For instance: sidenotes, the required mention inside forms. */
+  --cc-color-text-weak: var(--color-grey-80);
   /*endregion*/
 
   /*region Color Decisions (background)*/
@@ -211,6 +213,12 @@
 
   /*region Color Decisions(border)*/
 
+  /* Usage: for danger border */
+  --cc-color-border-danger: var(--color-red-100);
+
+  /* Usage: for danger border */
+  --cc-color-border-danger-weak: var(--color-red-20);
+
   /* Usage: for primary border */
   --cc-color-border-primary-weak: var(--color-blue-20);
 
@@ -220,11 +228,6 @@
   /* Usage: for warning border */
   --cc-color-border-warning-weak: var(--color-orange-20);
 
-  /* Usage: for danger border */
-  --cc-color-border-danger: var(--color-red-100);
-
-  /* Usage: for danger border */
-  --cc-color-border-danger-weak: var(--color-red-20);
   /*endregion*/
 
   /*region Margin Decisions*/
--- a/src/styles/default-theme.css
+++ b/src/styles/default-theme.css
@@ -219,6 +219,10 @@
   /* Usage: for danger border */
   --cc-color-border-danger-weak: var(--color-red-20);
 
+  /* Usage: for contrasted borders.
+  * For instance: click targets like input fields */
+  --cc-color-border-neutral-strong: var(--color-grey-50);
+
   /* Usage: for primary border */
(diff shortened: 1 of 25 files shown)


pdesoyres-cc (commented):
Nice work!

I just have a consideration on the variable naming.
Maybe it's just me, but I would prefer:

* --cc-color-border-neutral-strong
* --cc-color-border-neutral
* --cc-color-border-neutral-weak

roberttran-cc (approved):
LGTM, nice addition to our tokens!
And I agree with @pdesoyres-cc's suggestion. :ok_hand:

hsablonniere (approved):
Same as @pdesoyres-cc for the names but apart from that, LGTM :+1:

src/styles/default-theme.css
```diff
@@ -211,6 +211,10 @@
 
   /*region Color Decisions(border)*/
 
+  /* Usage: for contrasted borders.
```
pdesoyres-cc, line 214: nitpick: Could we try to order the variables? Could be by thickness first. Or could be intent first.

pdesoyres-cc (changes requested):
LGTM apart from a nitpick about variables ordering in the default theme

Galimede (approved):

PR #738, opened 2023-04-06
fix: issue in Custom Element Manifest generation when detecting @typedef imports

### LINKED ISSUE #737: Issue in Custom Element Manifest generation when detecting @typedef imports
When detecting typedef imports, we iterate over all comments and collect all typedef tags (`@typedef`). But, in comments without any tags, the node's `tags` property is `undefined`. We should take care of this nullity before iterating over tags.

The error occurs when the file contains a comment without any tag.
```
/**
 * comment without tag
 */
```

pdesoyres-cc:
Fixes #737

### DIFF
--- a/cem/support-typedef-jsdoc.js
+++ b/cem/support-typedef-jsdoc.js
@@ -133,7 +133,7 @@ export default function supportTypedefJsdoc () {
       const types = getTypesFromConstructor(constructorNode, ts);
 
       // This finds the comment where the imports are located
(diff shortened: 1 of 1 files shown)


roberttran-cc (approved):
Nice catch!

hsablonniere (approved):
Nice catch indeed :+1:

florian-sanders-cc (approved):
Nice catch :sunglasses:

PR #750, opened 2023-05-22
Tokens: add new neutral border colors for component states (disabled, hovered, focused)

### LINKED ISSUE #511: design tokens: introduce a readonly neutral bg
# Context

Following the introduction of design tokens, the background of our readonly `<cc-input-text>` has become a little bit too dark:
before design tokens (v6.0.0
image
after design tokens (prod)
image

This color currently relies on `var(--cc-color-bg-neutral-hovered)` because:
* it felt overkill to create a design token for only one case (`<cc-input-*>` that are read only).
* we have settled on 10 grey shades and didn't want to introduce more.

This color might work for hover but it's a little bit too dark for readonly.

# Goal

* We need one or two more grey shades to cover all cases (we especially need more "light" shades),
* We want a `--cc-color-bg-neutral-readonly` that would rely on one of these new shades and that would be used in the `<cc-input-*>` (readonly states)

### LINKED ISSUE #749: tokens: missing border colors for element states (hovered, focused, disabled)
## Context

With #736, we introduced new color tokens for neutral borders.
I forgot to introduce color tokens for borders depending on the element's state : `hovered` | `focused` | `disabled`

These colors are mainly useful for `cc-input-*` & `cc-select`, as well as pricing components.

We need these colors for #716.

florian-sanders-cc:
Fixes #511 
Fixes #749 

## How to review

This PR is fairly simple.

- review every commit
- check `cc-input-*` & `cc-select` stories (new very subtle change of border on hover, focus)
- check one of the `all form controls` stories and compare to prod to see the fix for readonly bg color.

### DIFF
--- a/src/components/cc-input-number/cc-input-number.js
+++ b/src/components/cc-input-number/cc-input-number.js
@@ -419,7 +419,7 @@ export class CcInputNumber extends LitElement {
         }
 
         input:hover + .ring {
-          border-color: #777;
+          border-color: var(--cc-color-border-neutral-hovered, #777);
         }
 
         :host([disabled]) .ring {
--- a/src/components/cc-input-number/cc-input-number.js
+++ b/src/components/cc-input-number/cc-input-number.js
@@ -423,7 +423,7 @@ export class CcInputNumber extends LitElement {
         }
 
         :host([disabled]) .ring {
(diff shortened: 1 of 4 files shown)


roberttran-cc (approved):

pdesoyres-cc (approved):
LGTM

PR #755, opened 2023-05-25
chore: change icons assets generation policy

### LINKED ISSUE #754: Change icons assets generation policy
## What do we have?

Icons assets are made of two js files (`src/assets/cc-clever.icons.js` and `src/assets/cc-remix.icons.js`).

Those 2 files are generated
- at build time (components, CDN)
- every time storybook loads in dev mode
- every time unit tests are launched

The generation is CPU-consuming and can sometimes be quite long => it can become a bit boring in dev mode.

## What we could do?

Instead, we could generate the assets with an npm command and then commit the generated files.

Every time we change the icons, we would need to re-run the command manually. This is not really a problem since we do not create icons very often.

## What do we gain?

* storybook in dev mode is much faster
* unit tests are faster
* code is simpler (we don't use the assets generation process in many places)
* we have code completion in our IDEs when importing icons

pdesoyres-cc:
Fixes #754


package.json
```diff
@@ -112,7 +113,7 @@
     "remark-highlight.js": "^6.0.0",
     "remark-html": "^13.0.1",
     "remark-parse": "^9.0.0",
-    "remixicon": "^2.5.0",
+    "remixicon": "3.3.0",
```
florian-sanders-cc, line 116: nitpick: not sure if it's voluntary so I'm asking, do we want to be locked on the `3.3.0` or do we go for the same as before `^3.3.0`
I don't know much about the Remix project but if they are serious with semver and breaking changes, we could keep the `^`.

florian-sanders-cc (changes requested):
Gg ! :clap: 

Just a small suggestion about the remix version management but all good to me for the rest.

package.json
pdesoyres-cc: This is voluntary: Since the operation (transforming remix icons into .js) is now totally manual, there is no purpose in using the `^` notation.

florian-sanders-cc: Got it, thanks for the answer, this is perfect, LGTM then ! :smile:

florian-sanders-cc (approved):

roberttran-cc (approved):
LGTM! That will definitively smooth out developer experience. GG!

PR #716, opened 2023-03-03
cc-pricing-*: update components for the new website

florian-sanders-cc:
Fixes #692

## TODO after the review (Florian)

- [x] rebase and clean commits
- [x] adapt tokens following the merge of #736 
- [x] create a specific commit for `cc-notice` migration (and modify the intent to match what has been done in `cc-notice` migration PR)
- [x] check everything again with Safari :skull: 
- [x] check everything again with the preprod website :art: 
- [x] discuss last details with Clément
- [x] last rebase

<details>
<summary>archived content</summary>

## How to review (last round of feedback)

Review the following commits:

- `feedback [your name]`,
- all other following fix commits that are not prefixed with `feedback`:
  - `fix(cc-pricing-product): add missing stories (taken from `cc-pricing-table`)`
  - `fix(cc-pricing-*): remove `List` suffix from variables to use plural instead`
  - `fix(shoelace select): change hover styling & clean custom props`
  - `fix(cc-pricing-): change hover styling of quantity buttons`
  - `fix(cc-pricing-product-consumption): use hard coded values instead of `productId` for `productName``
  - `fix(cc-zone): split tags only when relevant`
  - `fix(cc-pricing-*): add missing hover styling`

Be sure to check stories and all from the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-pricing/update/index.html?path=/story/%F0%9F%9B%A0-pricing-cc-pricing-page--default-story).

## What does this PR do?

* update pricing component visual styles as well as their behavior:
  * `cc-pricing-page` is now a wrapper component supposed to be slotted with pricing components. This gives a lot more freedom on how to arrange the pricing components inside it (the estimation, header and products can be placed wherever you want within the component).
  * `cc-pricing-table` has been merged into `cc-pricing-product`.
  * `cc-pricing-product` only handles the table and no longer has head + icon slots. Heading + icon + desc introducing a product are handled by the website itself because everything is slotted within the pricing page.
  * `cc-pricing-estimation` has been entirely reworked so that it can be positioned as a sticky part (right / bottom depending on the available space).
    * This is the change that might trigger the most questions / issues for this review.
    * This is also the part that you should focus the most when playing with the component in demo-smart or pricing-page stories.
  * `cc-header` no longer displays the total price and it now features a `temporality` dropdown that updates all the pricing components when they are placed within the pricing-page component.
* new border design tokens to make the customization easier for the website
* This PR also migrates the components to
  *  the new smart system.
[...]

PR #774, opened 2023-06-15
chore: upgrade shoelace to 2.4.0

pdesoyres-cc:


florian-sanders-cc:
LGTM

This means the pricing components rework needs to be part of the next release but I don't think that's an issue :+1:

florian-sanders-cc (approved):

Galimede:
It seems to break the `cc-pricing-header` style, if you check between the preview: https://clever-components-preview.cellar-c2.services.clever-cloud.com/upgrade-shoelace/index.html?path=%2Fstory%2F%F0%9F%9B%A0-pricing-cc-pricing-header--default-story and the prod: https://www.clever-cloud.com/doc/clever-components/?path=%2Fstory%2F%F0%9F%9B%A0-pricing-cc-pricing-header--default-story

@florian-sanders-cc, can you confirm that it is not a problem with this version of Shoelace with the new styling of the components in the pricing PR?

pdesoyres-cc:
@Galimede in fact this component is totally broken:
* The selection doesn't work
* On the Zone input, the element disappears when hovered
* There is also a margin issue

florian-sanders-cc:
Yes shoelace introduced many breaking changes that I've solved in the pricing PR.

This is why both this PR and the pricing PR will need to be part of the same release.

pdesoyres-cc:
This PR will be covered by #716

PR #794, opened 2023-06-27
cc-pricing-*: remove `apiConfig` from smart context parameters

### LINKED ISSUE #798: cc-pricing-*: remove `apiConfig` from the list of smart context parameters
## Context

When we updated the pricing components to use the new smart system, we added `apiConfig` to the list of parameters so that developers may change the `apiHost`.

This was not possible with the previous versions of the `pricing` components.

Unfortunately, since our new smart system does not allow optional parameters yet (see #771), this forces developers to set an empty `apiConfig` object in the context of the smart container.

We need to rollback this change, meaning we need remove `apiConfig` from the list of parameters until the smart components support optional parameters.

florian-sanders-cc:
Fixes #798 

## What does this PR do?

- Removes all `apiConfig` parameters from the newly updated pricing smart components.

## How to review

- Review every commit.
@pdesoyres-cc I used `chore` as commit types because I did not want these to be part of the changelog (because we haven't released the changes I'm rolling back with this PR so it would be confusing to include these commits in the changelog).
=> Is there a better way to do this?

### DIFF
--- a/demo-smart/cc-pricing-page/index.html
+++ b/demo-smart/cc-pricing-page/index.html
@@ -101,7 +101,7 @@
 </head>
 <body>
 
-<cc-smart-container context='{ "apiConfig": {}, "zoneId": "par", "addonFeatures":["cpu","memory","max-db-size","disk-size","connection-limit","version","databases","has-logs","has-metrics"]}'>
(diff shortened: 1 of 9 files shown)


pdesoyres-cc (approved):
LGTM

roberttran-cc (approved):

Galimede (approved):
LGTM, GG! :muscle:

hsablonniere (approved):
:+1: All good for me, I also checked locally and it works!

PR #802, opened 2023-06-28
Web Dev Server: enable development mode for our dependencies

### LINKED ISSUE #801: Lit: enable dev mode to benefit from lit warning & errors
## Context

I noticed that Lit playground (https://lit.dev/playground/) throws errors and warnings when you do something Lit forbids or recommends not to do.

I think we should enable this when using storybook in dev mode.

This can be done by editing the Web Dev Server `nodeResolve` config option (see [Lit docs - Development and production builds](https://lit.dev/docs/tools/development/#development-and-production-builds)).

We could also discuss enabling it for tests (but I don't know if it's a good idea so we better discuss it before doing it).

florian-sanders-cc:
Fixes #801

## How to review?

- Check the commit,
- Pull the branch, run `npm run storybook:dev` and check some of the stories like `cc-html-frame`. Open your dev tools console and you should see warnings like this:

Lit warning about dev mode and mutating a reactive property inside the update hook

### DIFF
--- a/web-dev-server.config.js
+++ b/web-dev-server.config.js
@@ -109,7 +109,11 @@ const injectAuthForSmartComponentsPlugin = {
 
 export default {
   port: 6006,
-  nodeResolve: true,
+  // There is a PR to include this in the default config of Web Dev Server (https://github.com/modernweb-dev/web/pull/2109)
+  // Once it's merged, we may revert this part of the config to `nodeResolve: true`
+  nodeResolve: {
+    exportConditions: ['development'],
+  },
   // watch: true,
   mimeTypes: {
     '**/*.md': 'js',


web-dev-server.config.js
```diff
@@ -1,10 +1,12 @@
 import { hmrPlugin, presets } from '@open-wc/dev-server-hmr';
 import json from '@rollup/plugin-json';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { fromRollup, rollupAdapter } from '@web/dev-server-rollup';
 import { storybookWdsPlugin } from './src/stories/lib/markdown.cjs';
 import { cemAnalyzerPlugin } from './wds/cem-analyzer-plugin.js';
 import { commonjsPluginWithConfig, esbuildBundlePluginWithConfig } from './wds/wds-common.js';
 
+console.log('CONFIG HERE');
```
Galimede, line 9: nitpick: maybe we can remove this console.log

Galimede (approved):
Just a nitpick, otherwise GG, great initative! 💪

web-dev-server.config.js
florian-sanders-cc: oupsie yes, thanks @Galimede

pdesoyres-cc (approved):
Really nice.

I'm wondering if we should do that for tests too.

roberttran-cc (approved):

hsablonniere (approved):

thepassle:
This was added to WDS in https://github.com/modernweb-dev/web/pull/2375, fyi

florian-sanders-cc:
Thanks for the update @thepassle :+1:

PR #807, opened 2023-07-04
cc-pricing-product: allow custom (not translated) features

### LINKED ISSUE #796: cc-pricing-product: allow custom features without translations
## Context

On the Clever Cloud Website, we want to be able to use the `cc-pricing-product` component to showcase a product not included within the API yet, with features that were not part of the original list of features.

The pricing components filter out features that have no translation registered.

This means we cannot add features without adding new translations and releasing a new version of the component.

## Things to do:

- [ ] add `name` to the data model of features,
- [ ] use `name` if no translation is registered for a given feature code,
- [ ] add translations for features we already know we need.

florian-sanders-cc:
Fixes #796 

## What does this PR do?

- Tweaks the code of `cc-pricing-product` to benefit from type checking and states.
- Updates the `cc-pricing-product` and `cc-pricing-estimation` components to use the `name` property if the feature code is not associated with a translation.
- Updates the `fake product` story to reflect this change.
- Updates all other product stories to filter features displayed (since features not translated used to be hidden, we did not have to explicitly filter them out before this change).
- Changes the `temporality` translation in English ("temporality" does not really mean what we wanted to convey here. I left it within the code to avoid an unnecessary breaking change, it's understandable for devs I think).
- Adds `Fake product` to the list of products displayed in the `cc-pricing-page` stories.
- Fixes the dollar story of pricing page.
- Adds a visible label + legend in the `cc-pricing-product-consumption` component.
- Fixes content overflowing from `cc-pricing-estimation` in the `cc-pricing-page` stories.

## How to review?

- Check both commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-pricing-new/allow-custom-product-features/index.html) (all pricing product stories + all the pricing page stories) : 
  - the `fake plan` story should now display new features but other stories should be the same as before.
  - the `dollar` pricing page story should now have a pricing header select set to dollar instead of euro.
  - the `cc-pricing-product-consumption` component (bottom of the stories) should have visible labels and legends for quantity/size and units (these should also be translated).
  - the label for the temporality dropdown in `cc-pricing-header` should now be `Unit of time` in English.
  - when you add many different products to the estimation, the content is not overflowing anymore within the `cc-pricing-estimation`, it's scrollable.

<details>
<summary>Archived - Issue with plan features vs product features when used without the smart component</summary>

> ## Questions/Notes
> 
> If you add a custom product to the estimation the custom features are not displayed in the estimation.
> You can test this by adding a plan from the `fake product` at the bottom of a pricing page story.
> 
> 
> I tried to make these part of the estimation details but it's tricky because:
> 
> - We manipulate two pieces of data related to features:
>   - `productFeatures` which is the list of features we want to display. It does not contain values, only the features code, type and name.
>   - plan `features` which are the actual features related to a plan. It contains features code, type, value and maybe a name (or maybe not). => When products are set from the API, plan features always have a `name`. When they are set manually (custom product), developers are allowed not to set any `name` (the reason for that is that if we force devs to set a name both in `productFeatures` and `plan` features, they may set different name for the same feature code by mistake...).
> - We display custom features within the `cc-pricing-product` component if they have a `name` property with a value.
> - We check whether the feature has a name by checking the `productFeatures` since plan `features` may not have a `name`.
> - The list of plans that `cc-pricing-estimation` manages come from plan `features`.
> - This component does not have any access to `productFeatures`.
> 
> This raises two issues:
>   - The `cc-pricing-estimation` cannot use the same strategy as `cc-pricing-product` and check if the feature is translated or has a name because custom product plan `features` may not have a name.
>   - The `cc-pricing-estimation` does not filter features based on `productFeatures` because it does not have any access to it.
>     - => we could add `productFeatures` to its properties but it would have to be set manually because it's not watched by any smart component.
>     - => we could filter plan `features` based on `productFeatures` before sending them to `cc-pricing-estimation` but this is not the most readable pattern (so many places to filter these: within render, sub-renders, right before adding plan...).
>     - => this is only an issue with custom products. Products coming from the API are formatted by methods within `product.js` and we filter plan features from there so `cc-pricing-product` and `cc-pricing-estimation` only deal with features that have already been filtered.

=> We decided to force the use of a `name` property on plan features as well as product features because:
  - this is what the API already does,
  - this means we don't have to change the current code,
  - the issue only happens when using the `cc-pricing-product` without a smart component and setting custom features, which should be very uncommon.
If the issue turns out to be an actual problem, we discussed a solution that would consist in:
- changing plan features into a key (code) value model
- relying on the `productFeatures` to get the `name` and the `type`.
- adding `name` and `type` to the info that is passed to `cc-pricing-estimation` upon adding a plan to it.
</details>

### DIFF
--- a/src/components/cc-pricing-estimation/cc-pricing-estimation.js
+++ b/src/components/cc-pricing-estimation/cc-pricing-estimation.js
@@ -149,8 +149,16 @@ export class CcPricingEstimation extends LitElement {
    * @return {string|void} the translated feature name if a translation exists or nothing if the translation does not exist
    */
   _getFeatureName (feature) {
-    if (feature != null && FEATURES_I18N[feature.code] != null) {
+    if (feature == null) {
+      return '';
+    }
+
+    if (FEATURES_I18N[feature.code] != null) {
       return FEATURES_I18N[feature.code]();
+    }
+
+    if (feature.name != null) {
+      return i18n('cc-pricing-estimation.feature.custom', { featureName: feature.name });
     }
   }
 
@@ -471,16 +479,15 @@ export class CcPricingEstimation extends LitElement {
           </summary>
           ${hasFeatures ? html`
             <dl class="plan__features">
-              ${plan.features?.map((feature) => {
-                if (AVAILABLE_FEATURES.includes(feature.code)) {
+              ${plan.features
+                ?.filter((feature) => AVAILABLE_FEATURES.includes(feature.code) || feature.name != null)
+                ?.map((feature) => {
                   return html`
                     <div class="plan__features__feature">
                       <dt>${this._getFeatureName(feature)}</dt>
                       <dd>${this._getFeatureValue(feature)}</dd>
                     </div>
                   `;
-                }
-                return '';
               })}
             </dl>
           ` : ''}
--- a/src/components/cc-pricing-product/cc-pricing-product.js
+++ b/src/components/cc-pricing-product/cc-pricing-product.js
@@ -111,8 +111,16 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
    * @return {string|undefined} the translated feature name if a translation exists or nothing if the translation does not exist
    */
   _getFeatureName (feature) {
-    if (feature != null && FEATURES_I18N[feature.code] != null) {
+    if (feature == null) {
+      return '';
+    }
+
+    if (FEATURES_I18N[feature.code] != null) {
       return FEATURES_I18N[feature.code]();
+    }
+
+    if (feature.name != null) {
+      return feature.name;
     }
   }
 
@@ -254,7 +262,7 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
   _renderProductPlans (productName, productPlans, productFeatures) {
     // this component is not rerendering very often so we consider we can afford to sort plans and filter the features here.
     const sortedPlans = [...productPlans].sort((a, b) => a.price - b.price);
-    const filteredProductFeatures = [...productFeatures].filter((feature) => AVAILABLE_FEATURES.includes(feature.code) || feature.name != null);
+    const filteredProductFeatures = productFeatures.filter((feature) => AVAILABLE_FEATURES.includes(feature.code) || feature.name != null);
 
     // We don't really have a good way to detect when the component should switch between big and small mode.
     // Also, when this component is used several times in the page, it's better if all instances switch at the same breakpoint.
--- a/src/components/cc-pricing-product/cc-pricing-product.js
+++ b/src/components/cc-pricing-product/cc-pricing-product.js
@@ -130,7 +130,7 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
       case 'boolean':
         return i18n('cc-pricing-product.type.boolean', { boolean: feature.value === 'true' });
       case 'boolean-shared':
-        return i18n('cc-pricing-product.type.boolean-shared', { boolean: feature.value === 'shared' });
+        return i18n('cc-pricing-product.type.boolean-shared', { shared: feature.value === 'shared' });
       case 'bytes':
         return (feature.code === 'memory' && feature.value === '0')
           ? i18n('cc-pricing-product.type.boolean-shared', { shared: true })
--- a/src/components/cc-pricing-product/cc-pricing-product.js
+++ b/src/components/cc-pricing-product/cc-pricing-product.js
@@ -223,10 +223,10 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
    * Adds the product name to the given plan.
    * Dispatches a `cc-pricing-product:add-plan` event with the plan as its payload.
    *
+   * @param {string} productName - the name of the product to which the plan is attached
    * @param {Plan} plan - the plan to be added to the estimation
    */
-  _onAddPlan (plan) {
-    const productName = this.product.name;
+  _onAddPlan (productName, plan) {
     dispatchCustomEvent(this, 'add-plan', { productName, ...plan });
   }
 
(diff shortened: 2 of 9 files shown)


src/components/cc-pricing-product/cc-pricing-product.js
```diff
@@ -130,7 +140,7 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
       case 'boolean':
         return i18n('cc-pricing-product.type.boolean', { boolean: feature.value === 'true' });
       case 'boolean-shared':
-        return i18n('cc-pricing-product.type.boolean-shared', { boolean: feature.value === 'shared' });
+        return i18n('cc-pricing-product.type.boolean-shared', { shared: feature.value === 'shared' });
```
florian-sanders-cc, line 143: this is to fix a legacy error, the related i18n line is as follows:
```js
  'cc-pricing-product.type.boolean-shared': ({ shared }) => `${shared ? 'Shared' : 'Dedicated'}`,
```

roberttran-cc (commented):
Very nice job! :muscle: 

Nothing to say about the code.

However, I've got an overflow issue (Chrome and Firefox) in the `cc-pricing-page` default story, at the level of the Redis product: when the browser is moderately narrow (not using the preview size control of Storybook), the table content is bleeding horizontally. It does not happen in the current production version (ping me if you need more details).

(and you probably saw it but an unused variable in `src/components/cc-pricing-estimation/cc-pricing-estimation.js` causes the CI to fail :wink:).

florian-sanders-cc:
:wave: @roberttran-cc thanks a lot for your review.
It actually revealed some issues that I had not noticed after my latest modifications.

It should be fixed now but could you check the stories again? 
I did not find the overflowing so it could still be there.

If you could also check again the `fix(cc-pricing-product)` and `fix(cc-pricing-estimation)` commit it would be great :hugs:

roberttran-cc (commented):
LGTM! The visual bug does not appear anymore! :muscle: 

There's just a wandering `console.log` to remove before we're good to go!

src/components/cc-pricing-product/cc-pricing-product.stories.js
```diff
@@ -85,7 +99,7 @@ export const error = makeStory(conf, {
     product: { state: 'error' },
   }],
 });
-
+console.log({ yep: getFullProductAddon('es-addon', addonFeatures) });
```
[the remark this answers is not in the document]
florian-sanders-cc: fixed :sweat_smile:

Thanks @roberttran-cc!

roberttran-cc (approved):

src/components/cc-pricing-estimation/cc-pricing-estimation.js
```diff
@@ -149,9 +149,17 @@ export class CcPricingEstimation extends LitElement {
...
+      return '';
+    }
+
+    if (FEATURES_I18N[feature.code] != null) {
       return FEATURES_I18N[feature.code]();
     }
+
+    if (feature.name != null) {
+      return i18n('cc-pricing-estimation.feature.custom', { featureName: feature.name });
+    }
```
pdesoyres-cc, line 162: Maybe we should always try to return something instead of `void`: use `feature.code` as a fallback.

```suggestion
    return i18n('cc-pricing-estimation.feature.custom', { featureName: feature.name ?? feature.code });
```

src/components/cc-pricing-product/cc-pricing-product.js
```diff
@@ -111,9 +111,17 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
...
+      return '';
+    }
+
+    if (FEATURES_I18N[feature.code] != null) {
       return FEATURES_I18N[feature.code]();
     }
+
+    if (feature.name != null) {
+      return feature.name;
+    }
```
pdesoyres-cc, line 124: Maybe we should always try to return something instead of `void`: use `feature.code` as a fallback.

```suggestion
    return feature.name ?? feature.code;
```

pdesoyres-cc (changes requested):
Nice job Florian!

I have some minor suggestions.

Also, I'm wondering why we have introduced the `name` property on the `Feature` entity. I think we could use the `code` property also for custom features and say, if we have a translation for the code use it, otherwise use the code directly.

We have a clash in type `Feature` which is present in two files `cc-addon-features.types.d.ts` and `common.types.d.ts`.
This perturbs WebStorm in resolving the right type in `cc-pricing-product.js`.
Could we rename the `Feature` type in `cc-addon-features.types.d.ts` into `AddonFeature` ?

hsablonniere (approved):
I looked at the code and preview. LGTM.

src/components/cc-pricing-estimation/cc-pricing-estimation.js
florian-sanders-cc: The issue is that the `cc-pricing-estimation` filters features that have no `name` or no translation registered for a given feature code.

```js
plan.features
  ?.filter((feature) => AVAILABLE_FEATURES.includes(feature.code) || feature.name != null)
  ?.map((feature) => {
    return html`
      <div class="plan__features__feature">
        <dt>${this._getFeatureName(feature)}</dt>
        <dd>${this._getFeatureValue(feature)}</dd>
      </div>
    `;
})
```
To make this change useful, we'd have to remove the filter and let everything go through.
But to be able to do this, the `cc-pricing-product` would need to clean the features it sends to the `cc-pricing-estimation` and remove the ones that are not part of `productFeatures`. 

We discussed this refactoring (plan features model rework + `cc-pricing-product` not relying only on its smart component to filter features) with Hubert and Mathieu and decided not to go down this road for now.

The main reason is that this is only useful when the component is not used with a smart component, which is very rare (only for 1 or 2 products on the website, where they will make sure to use `name` properties anyway).

src/components/cc-pricing-product/cc-pricing-product.js
florian-sanders-cc: (see `cc-pricing-estimation` comment for reasons why I chose not to change this even though it was a good idea)

florian-sanders-cc:
> We have a clash in type Feature which is present in two files cc-addon-features.types.d.ts and common.types.d.ts.
> This perturbs WebStorm in resolving the right type in cc-pricing-product.js.
> Could we rename the Feature type in cc-addon-features.types.d.ts into AddonFeature ?

@pdesoyres-cc I'll open a quick win issue for this and handle it directly in a separate PR :+1:

Galimede (approved):
GG @florian-sanders-cc :muscle:

PR #818, opened 2023-08-09
feat: add component graph CLI and graph to the CEM components description

### LINKED ISSUE #806: introduce a component graph usage
There has been a few time when refactoring of our code where we wanted to know if a component had some other components as a child. Or for a component we wanted to know what where the components that made use of it. So we thought having a graph that would represent that would be cool.

To test that we could do a POC that would show a simple terminal/script graph with a rollup plugin. 

We should be able to have two outputs:
* one that gives for a component all its children (e.g: `cc-env-var-form` contains `cc-env-var-editor-simple`, `cc-env-var-editor-expert`...)
* one that gives for a component its usage in another component. (e.g: cc-notice is used by `cc-addon-backups`, `cc-article-list`...)

Next we could:
* output the whole thing as a `graphviz` file
* handle recursive children components (parent component has a child component that contains a grand child component...)
...

Galimede:
Fixes #806 

## TODO

- [ ] Figure out how to deal with ESLint top await module error
- [ ] Add more documentation
- [ ] Check what can be improved


tasks/component-usage.js
```diff
@@ -0,0 +1,163 @@
...
+}
+else if (all != null) {
+  maxLevel = all;
+}
+
+function isComponent (input) {
+  return input.startsWith('cc-') || input === 'global';
+}
+
+let componentsInput = '';
```
pdesoyres-cc, line 139: I'm not very confortable with this variable which can be either a string or an array of string. And, then is passed to function that can not handle both types (`getDependantsLsStyle` only handle string while `getComponentsTree` only handle array).

pdesoyres-cc (changes requested):
Really nice work Mathieu!

I really like how you included this inside the Storybook documentation! Well done :clap:

Could we consider rename the files:
- component-usage.js -> component-usage-cli.js
- fetch-graph.js -> component-usage-graph.js

package.json
```diff
@@ -58,6 +58,7 @@
     "@shoelace-style/shoelace": "^2.5.0",
     "chart.js": "^3.5.0",
     "chartjs-plugin-datalabels": "^2.0.0",
+    "es-module-lexer": "^1.3.0",
```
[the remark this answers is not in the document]
Galimede: Yes it should, thanks for noticing :+1:

tasks/component-usage.js
```diff
@@ -0,0 +1,163 @@
...
+    break;
+  case 'uses':
+    console.log(getDependantsLsStyle(componentsInput, graph, maxLevel));
+    break;
+  case 'used-by':
+    console.log(getComponentsTree(componentsInput, graph, 'dependencies', maxLevel));
+    break;
+}
+
+process.exit(0);
```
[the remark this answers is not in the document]
Galimede: We could get rid of it but I think it's cleaner to leave it that way. WDYT?

tasks/component-usage.js
Galimede: I was thinking of making it an array every time and for the 'uses' mode I'll pass `componentsInput[0]` to the `getDependantsLsStyle` function. WDYT?

tasks/component-usage-graph.js
```diff
@@ -0,0 +1,174 @@
+import fs from 'fs';
+import path from 'path';
+import chalk from 'chalk';
+import { DepGraph } from 'dependency-graph';
+import { init, parse } from 'es-module-lexer';
+import { getComponentFiles } from '../rollup/rollup-common.js';
+
+/**
+ *
+ * @returns {DepGraph} - returns a component `DepGraph`.
```
florian-sanders-cc, line 10: **nitpick:** to satisfy Typescript:
```suggestion
 * @returns {Promise<DepGraph>} - returns a component `DepGraph`.
```

tasks/component-usage-graph.js
```diff
@@ -0,0 +1,174 @@
...
+      const importerRelativeDir = path.dirname(importerRelative);
+      const sourceRelative = path.join(importerRelativeDir, imported.n);
+
+      if (!componentsGraph.hasNode(importerRelative)) {
+        componentsGraph.addNode(importerRelative);
+      }
+
+      const isImporterInList = componentsFileList.includes(importerRelative);
+      const isImportedInList = componentsFileList.includes(sourceRelative);
+      const isImportedLocal = imported.n.startsWith('./') || imported.n.startsWith('../');
```
florian-sanders-cc, line 38: **question:** I'm confused by this line I don't understand why we need to check this :thinking: (it's not a request for change, just a request for help :sweat_smile:)

tasks/component-usage-graph.js
```diff
@@ -0,0 +1,174 @@
...
+
+      if (isImporterInList && isImportedInList && isImportedLocal && !componentsGraph.hasNode(sourceRelative)) {
```
florian-sanders-cc, line 40: **suggestion:** looks like `isImporterInList && isImportedInList && isImportedLocal` is common to both your conditions (l40 and l44), maybe we could move these into their own `if`?

I see two options but I might be wrong:

```js
if (isImporterInList && isImportedInList && isImportedLocal) {
  if (componentsGraph.hasNode(sourceRelative)) {
    componentsGraph.addNode(sourceRelative);
  }

  componentsGraph.addDependency(importerRelative, sourceRelative);
}
```
or (really not sure about this one, I never use `continue` :eyes:)
```js
if (!isImporterInList || !isImportedInList || !isImportedLocal) {
  continue;
}

if (!componentsGraph.hasNode(sourceRelative)) {
  componentsGraph.addNode(sourceRelative);
}

componentsGraph.addDependency(importerRelative, sourceRelative);
```
or just move `isImporterInList && isImportedInList && isImportedLocal` to a `const` (but I don't know how to name it) and reuse it:
```js
const areInListAndLocal = isImporterInList && isImportedInList && isImportedLocal;

if (areInlistAndLocal && !componentsGraph.hasNode(sourceRelative)) {
  componentsGraph.addNode(sourceRelative);
}

if (areInListAndLocal) {
  componentsGraph.addDependency(importerRelative, sourceRelative);
}
```

cem/add-dependencies-in-description.js
```diff
@@ -0,0 +1,39 @@
...
+
+const graph = await getComponentsGraph();
+
+export default function addDependenciesInDescription () {
+
+  return {
+    name: 'add-dependencies-in-description',
+    moduleLinkPhase ({ moduleDoc }) {
+
+      const isSmartComponentFile = !moduleDoc.path.includes('smart');
```
florian-sanders-cc, line 19: **suggestion:** `smart` could be used within the name of the component, for instance `cc-smart-form`. It's unlikely but just in case I would go for a more specific test:

```suggestion
      const isSmartComponentFile = !moduleDoc.path.includes('.smart.js');
```

florian-sanders-cc: **question:** I'm confused because the const name is `isSmartComponentFile`, meaning that it should be true if the component file is for a smart component but the value indicates the contrary :thinking:

florian-sanders-cc (approved):
Very nice work @Galimede it's impressive!

I just had a few questions (not requests for changes) and nitpicks :+1: 

About the doc in storybook, I really like it and I think it is great that you've managed to include this into this quick win.

Later on, when we have time, we should revisit how we display the graphs in storybook because it is not accessible but it is a very low priority issue to be honest, really good job again!

tasks/component-usage-graph.js
Galimede: On some previous iterations the option 1 was the one implemented. Though, I didn't stick with it as I don't like having multiple if nested if I can avoid them. However, I'll change it for option 3 if that's okay for you :wink:

tasks/component-usage-graph.js
Galimede: We have to do this check because packages are dependencies thus not within our source files.
Here's an example: every component need `Lit` so they will have that line `import { css, html, LitElement } from 'lit';`. 
From the import we know that as there's no sign of a path here it's a dependency and we don't have to take this into account. I hope it's clearer for you that way :pray:

cem/add-dependencies-in-description.js
Galimede: > question: I'm confused because the const name is isSmartComponentFile, meaning that it should be true if the component file is for a smart component but the value indicates the contrary 🤔

Oops, it's a mistake, nice catch :wink: 

> suggestion: smart could be used within the name of the component, for instance cc-smart-form. It's unlikely but just in case I would go for a more specific test:
Will do the change :wink:

tasks/component-usage-cli.js
```diff
@@ -0,0 +1,159 @@
...
+  return input.startsWith('cc-') || input === 'global';
+}
+
+const graph = await getComponentsGraph();
+
+let component = '';
+let mode = null;
+let maxLevel = 1;
+
+if (uses != null && uses === 'global') {
```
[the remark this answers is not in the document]
Galimede: Hmm, I guess that's a safety check that was made from a previous iteration. I think we don't need that redundant check, I will remove it. :wink:

tasks/component-usage-graph.js
florian-sanders-cc: Yes no problem, even your version was ok for me :+1:

tasks/component-usage-graph.js
florian-sanders-cc: Ah yes I gues it now, perfect explanation! Looks good to me :+1:

tasks/component-usage.js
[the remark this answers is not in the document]
pdesoyres-cc: I've got nothing against that. Let's keep it

tasks/component-usage-graph.js
```diff
@@ -53,56 +54,59 @@ export async function getComponentsGraph () {
...
- * @param graph
- * @param maxLevel
+ * @param {string} componentPath
+ * @param {DepGraph} graph
+ * @param {number} [maxLevel=Infinity]
  * @returns {string}
  *
+ * This function differs from `getComponentsTree` for various reasons:
+ * * It only takes a component path as a param instead of a list of components paths
+ * * It forms a specific `npm ls` like output
```
pdesoyres-cc, line 80: In JSDoc, the description of the function is always above @param and @return tags.

tasks/component-usage-graph.js
```diff
@@ -111,39 +115,42 @@ export function getDependantsLsStyle (componentPath, graph, maxLevel = Infinity)
...
- * @param mode
- * @param maxLevel
+ * @param {string[]} componentsPath
+ * @param {DepGraph} graph
+ * @param {'dependencies'|'dependants'} mode
+ * @param {number} [maxLevel=Infinity]
  * @returns {string|null}
+ *
+ * This function allows you to get either the dependencies or the dependants of one or more components given its path list.
+ * It will work whether you provide it with an array of one component path or a full list.
```
pdesoyres-cc, line 125: fix: In JSDoc, the description of the function is always above @param and @return tags.

tasks/component-usage-graph.js
```diff
@@ -111,39 +115,42 @@ export function getDependantsLsStyle (componentPath, graph, maxLevel = Infinity)
 
 /**
  *
- * @param inputs
- * @param graph
- * @param mode
- * @param maxLevel
+ * @param {string[]} componentsPath
+ * @param {DepGraph} graph
+ * @param {'dependencies'|'dependants'} mode
```
pdesoyres-cc, line 120: fix: mode is optional with default value:

```suggestion
 * @param {'dependencies'|'dependants'} [mode='dependencies']
```

pdesoyres-cc (changes requested):
Well done! Almost there, I've got a few nitpicks on JSDoc.

tasks/component-usage.js
pdesoyres-cc: ok

pdesoyres-cc (changes requested):
Do not forget to link the issue in the commit messages

roberttran-cc (commented):
Very very nice job: definitively a very useful tool.

I only have a minor feedback in the code. Also, it could (nitpick) be nice to display a feedback when no dependant/dependency is found.

pdesoyres-cc (approved):
LGTM. NIce work @Galimede

hsablonniere (commented):
Hey @Galimede, well done :clap:

It's a really good addition to our toolchain :heart_eyes:

## Comments

* I only created `suggestions` comments, so nothing blocking, feel free to apply them or not

## The UX

* I'm not sure about the usage or red to highlight the component in the tree view. npm uses yellow and it feels more like a highlight than an error.
* A bit sad that you didn't use the lib npm uses to have proper trees but we'll revisit this later :wink:
* There's not documentation in the project
  * We could write a small markdown document, referenced in the guides
  * Or we could add this as an npm script and document it's existence in `docs/references/tasks.reference.md`, the details are already in the `-h` command :+1:

hsablonniere (approved):
All good for me @Galimede, well done :clap:

PR #820, opened 2023-08-28
docs(breaking-changes): add breaking change policy

### LINKED ISSUE #695: RFC: breaking changes and design tokens
# Context 

Following the addition of a design token for focus styles (see #685), we discussed whether it could be considered a breaking change or not.

Since we introduced a theme that could be customized by people using our components, it does mean that when we release new tokens, people may have to update their own theme.

In the end it seems we don't think it's a breaking change because it adds a new way to customize how clever components look / behave. If you don't customize it, you get a black outline so this does not really break anything but it does look a lot different than before :thinking:.

# Things to discuss 

* Discuss and write down our criteria for breaking changes in a specific document.
* Add a way to warn developers about theme updates.

florian-sanders-cc:
Fixes #695

### DIFF
--- a/docs/guides/breaking-change-policy.md
+++ b/docs/guides/breaking-change-policy.md
@@ -1,0 +1,72 @@
+---
+kind: '📌 Docs'
+---
+
+# Breaking change policy
+
+When a new version of our components is released, we need to consider the potential impact of changes made and how we label them.
+
+> Breaking changes are changes that can potentially break an integration.
+> Additive changes are changes that should not break an integration.
+
+*Source: [GitHub's Breaking Changes](https://docs.github.com/en/rest/overview/breaking-changes)*
+
+This is why our project is based on [Semantic Versioning](https://semver.org/) (or Semver) to define version number: a major version introduces at least one breaking change.
+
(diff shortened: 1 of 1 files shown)


docs/guides/breaking-change-policy.md
```diff
@@ -0,0 +1,70 @@
...
+* slots,
+* events,
+* CSS custom properties,
+* methods,
+* host `display` property.
+
+Therefore, the following should be considered breaking changes:
+
+* renaming or deleting one of those elements,
+* modifying the type of an attribute or property and the signature of an event or method.
```
pdesoyres-cc, line 34:
```suggestion
* modifying the type of an attribute or property,
* modifying the signature of a public method,
* modifying the type of the object included in an event.
```

pdesoyres-cc (commented):
GG, very nice doc.

docs/guides/breaking-change-policy.md
roberttran-cc: Thanks for the suggestion: doc updated!

pdesoyres-cc (approved):
Perfect

PR #823, opened 2023-08-28
fix(tests): add another resize loop error message to the ignore list

### LINKED ISSUE #821: Tests: resize observer loop error makes tests fail
## Context

It looks like Chrome now uses the same error message as Firefox for the `resize loop` error (see #808).

When we merge `cc-logs`, we'll have a proper way to handle this but in the meantime, we need to add this error message to the ignore list.

### LINKED ISSUE #822: `cc-tile-metrics`: missing `Filler` import and registration for chat.js
## Context

When fixing the tests I realized we were getting warnings about a missing chartjs plugin import & registration (`Filler`).
image

Everything seems to be working fine anyway but I'll add the missing import & registration for this plugin.

florian-sanders-cc:
Fixes #821
Fixes #822

## How to review?
 
**Disclaimer:** the browser still logs `null` everytime a resize loop error happens. I tried to remove it but failed because we're patching the fake `window.onerror` of `Mocha` and it seems tricky. I stopped searching for a solution since we have a proper fix on its way and I don't think it's worth more efforts in the meantime as long as tests are fixed.

- 2 reviewers should be enough for this one,
- review the commits,
- check in local if tests are running fine (still verbose but I have no solutions for this),
- check that `cc-tile-metrics` is the same as in prod.

### DIFF
--- a/src/components/cc-tile-metrics/cc-tile-metrics.js
+++ b/src/components/cc-tile-metrics/cc-tile-metrics.js
@@ -1,4 +1,4 @@
-import { BarController, BarElement, CategoryScale, Chart, LinearScale, Title, Tooltip } from 'chart.js';
+import { BarController, BarElement, CategoryScale, Chart, LinearScale, Title, Tooltip, Filler } from 'chart.js';
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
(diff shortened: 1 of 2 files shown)


roberttran-cc (approved):

Galimede (approved):
LGTM, GG! :muscle:

Issue #398, opened 2022-03-16
design tokens


florian-sanders-cc:
As the team and the project grow, we need to define design tokens so that we can build a consistent and easy to maintain design system.

List of design tokens we need to define (most critical ones first) :
1. - [ ] Colors (WIP)
	  - Background-colors
	  - Element / Foreground-colors
	  - Text-colors
	  - Gradients
2.  - [ ] Typography
	  - Font families
	  - Font size
	  - Font styles & letter case
	  - Line height
3. - [ ] Sizes
	  - Spacers
	  - Padding / Inset
	  - Common Sizes for icons, etc.
4.- [ ]  Border styles
5.  - [ ] Elevation / shadow
6. - [ ] Opacity
7. - [ ] Motion
	  - Timing / Duration / delay
	  - Curves
8. - [ ] Icons

This list is a trimmed down version of what can be found in the very thorough [Naming Tokens in Design Systems](https://medium.com/eightshapes-llc/naming-tokens-in-design-systems-9e86c7444676) article by Nathan Curtis.

hsablonniere:
This list is good :+1:

PR #830, opened 2023-09-12
feat(cc-popover): init component

### LINKED ISSUE #699: cc-button: implement parts of ARIA API
# Context

There are some situations where we need to add ARIA attributes to make a button accessible.
To name a few:

- `aria-expanded` for [disclosure patterns](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) (Show / Hide)
- `aria-controls` for the same pattern as above. => we cannot do it yet, we need to wait for `Cross root ARIA - export ID` (https://github.com/WICG/aom/blob/gh-pages/exportid-explainer.md)
- `aria-pressed` for toggle button patterns.

The only way to do this is to mimic the ARIA API by adding props to our `cc-button`.
We'll need to make sure these props are properly namespaced to avoid clashing with actual ARIA.

(Once Accessibility Object Model is available :roll_eyes: we'll be able to simplify this but we cannot really wait for it)

### LINKED ISSUE #829: cc-popover: init component
## Context

This component is needed for the logs controller component. We want the different options of the logs component to be located in a popover because it provides an easy way to access options, and we don't have to rely on a more complicated modal dialog component.

Also, the native [popover API](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) is not yet supported everywhere.

This component is made of:

* a trigger: A `<cc-button>`.
* a content: The default slot.

## To be discussed

`isOpen` is a public property that can be set from the outside.
Changing this property from outside the component won't trigger the `cc-popover:open` or `cc-popover:close` events.
Maybe, this is something we would like to discuss in sync with the team.

## Inputs

We don't need to provide all the features the `<cc-button>` has. We will expose only some useful ones.

### Attributes/properties

| `accessibleName` | `string`                     | `null`              | `true`  | Accessible name to set on the `<cc-button>`           |
| `hideText`               | `boolean`                 | `false`            | `false` | Whether the `<cc-button>` should hide text            |
| `icon`                       | `IconModel`            | `null`              | `false` | The icon to set on the `cc-button`                              |
| `isOpen`                  | `boolean`                | `false`             | `false` | Whether the content is shown on screen                |
| `position`                | `PopoverPosition` | `bottom-left` | `false` | The position of the content relative to the button |

### Slots

| Name  | Description |
|-------|-------------|
| `default`        | The popover content                           |
| `button-text` | The text to set on the `<cc-button>` |

### CSS custom property

| `--cc-popover-z-index` | `number` | `999` | The z-index to set on the popover content wrapper|

## Outputs

| `cc-popover:open` | `null` | Fires whenever the popover is opened |
| `cc-popover:close` | `null` | Fires whenever the popover is closed   |

## Type Definitions

```typescript
export interface IconModel {
  content: string;
}

export type PopoverPosition = 'bottom-right' |  'bottom-left' |  'top-right' |  'top-left';
```

## Stories & states

* default: withIconOnly,
* withButtonText,
* withButtonTextAndIcon

pdesoyres-cc:
Fixes #829

Fixes #699

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -49,6 +49,8 @@ export class CcButton extends LitElement {
   static get properties () {
     return {
       accessibleName: { type: String, attribute: 'accessible-name' },
+      a11yExpanded: { type: Boolean, attribute: 'a11y-expanded', reflect: true },
+      a11yPressed: { type: Boolean, attribute: 'a11y-pressed', reflect: true },
       circle: { type: Boolean },
       danger: { type: Boolean },
       delay: { type: Number },
@@ -72,6 +74,12 @@ export class CcButton extends LitElement {
 
     /** @type {string|null} Forces the values of the `aria-label` and `title` attributes on the `button` element. CAUTION: The accessible name should always start with the visible text if there is one. For instance "add to estimation - NodeJS XS" */
     this.accessibleName = null;
+
+    /** @type {null|boolean} Sets aria-expanded on the inner `button` element. */
+    this.a11yExpanded = null;
+
+    /** @type {null|boolean} Sets aria-pressed on the inner `button` element. */
+    this.a11yPressed = null;
 
     /** @type {boolean} Sets button UI to a circle form when in `hide-text` and `image` mode. */
     this.circle = false;
@@ -241,6 +249,8 @@ export class CcButton extends LitElement {
         @click=${this._onClick}
         title="${ifDefined(this._getTitle())}"
         aria-label="${ifDefined(this._getAriaLabel())}"
+        aria-expanded="${ifDefined(this.a11yExpanded ?? undefined)}"
+        aria-pressed="${ifDefined(this.a11yPressed ?? undefined)}"
       >
         <!--
           When delay mechanism is set, we need a cancel label.
--- a/src/components/cc-popover/cc-popover.js
+++ b/src/components/cc-popover/cc-popover.js
@@ -1,0 +1,264 @@
+import '../cc-button/cc-button.js';
+import { css, html, LitElement } from 'lit';
+import { createRef, ref } from 'lit/directives/ref.js';
+import { dispatchCustomEvent, EventHandler } from '../../lib/events.js';
+
+/**
+ * @typedef {import('../common.types.js').IconModel} IconModel
+ * @typedef {import('./cc-popover.types.js').PopoverPosition} PopoverPosition
+ */
+
+/**
+ * A component displaying a floating content next to a button element.
+ *
+ * ## Details
+ *
+ * ### Button element
+ *
+ * The `button` is the element that will trigger the display of the floating content.
+ * This element is a `<cc-button>`. The `cc-button:click` event will trigger the popover toggle.
+ *
+ * ### Popover content
+ *
+ * The `content` is the element that will be displayed next to the `button` element.
+ * This element must be placed in the default slot.
+ * It will be placed according to the `position` property.
+ *
+ * ## Accessibility
+ *
+ * The component places necessary aria attribute on the `button` element:
+ *
+ * * `aria-expanded`: set to `true` when popover is open, `false` when it is closed.
+ *
+ * When the popover is open, pressing `esc` will close the popover and focus the `button` element.
+ *
+ * ## Usage
+ *
+ * ```html
+ * <cc-popover>
+ *   <span slot="button-content">Click me</span>
+ *   <div>This is a content to be displayed when button is clicked.</div>
+ * </cc-popover>
+ * ```
+ *
+ * @cssdisplay block
+ *
+ * @event {CustomEvent} cc-popover:open - Fires whenever the popover is opened.
+ * @event {CustomEvent} cc-popover:close - Fires whenever the popover is closed.
+ *
+ * @slot - The area containing the content of the popover.
+ * @slot button-content - The area containing the button content.
+ *
+ * @cssprop {Size} --cc-popover-gap - Sets the gap between the button and the floating area (default 0.4em).
+ * @cssprop {Number} --cc-popover-z-index - Sets the z-index of the floating content (defaults: `999`).
+ */
+export class CcPopover extends LitElement {
+  static get properties () {
+    return {
+      accessibleName: { type: String, attribute: 'accessible-name' },
+      hideText: { type: Boolean, attribute: 'hide-text' },
+      icon: { type: Object },
+      isOpen: { type: Boolean, attribute: 'is-open', reflect: true },
+      position: { type: String },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {string|null} Sets the accessibleName property of the underlying `cc-button` element. CAUTION: The accessible name should always start with the visible text if there is one. */
+    this.accessibleName = null;
+
+    /** @type {boolean} Whether the button text should be hidden. */
+    this.hideText = false;
+
+    /** @type {IconModel|null} Sets the button icon. */
+    this.icon = null;
+
+    /** @type {boolean} Whether the popover is opened */
+    this.isOpen = false;
+
+    /** @type {PopoverPosition} Sets the position of the popover relative to the `button` element. */
+    this.position = 'bottom-left';
+
+    /** @type {Ref<HTMLElement>} */
+    this._contentRef = createRef();
+
+    /** @type {Ref<HTMLElement>} */
+    this._buttonRef = createRef();
+
+    this._onOutsideClickHandler = new EventHandler(window, 'click', (event) => {
+      const contentElement = this._contentRef.value;
+      if (contentElement != null && !event.composedPath().includes(contentElement)) {
+        this.close();
+      }
+    });
+
+    this._onEscapeKeyHandler = new EventHandler(this, 'keydown', (event) => {
+      if (event.key === 'Escape') {
+        this.close();
+      }
+    });
+
+    // Opening a popover must close the last opened popover.
+    let lastOpenedPopover = null;
+    this._onCcPopoverOpenHandler = new EventHandler(window, 'cc-popover:open', (event) => {
+      // We cannot use event.target because events that happen in shadow DOM and when caught from outside the shadow DOM,
+      // have the host element as the target (and not the real target element inside the shadow DOM).
+      const popover = event.composedPath()[0];
+
+      if (popover !== this) {
+        lastOpenedPopover = popover;
+      }
+      else {
+        lastOpenedPopover?.close(false);
+        lastOpenedPopover = null;
+      }
+    });
+  }
+
+  // region Public methods
+
+  /**
+   * Opens the popover.
+   */
+  open () {
+    if (!this.isOpen) {
+      this.isOpen = true;
+      dispatchCustomEvent(this, 'open');
+    }
+  }
+
+  /**
+   * Closes the popover.
+   * @param {boolean} [shouldFocus = true] Whereas the button should be focused. This applies only if the popover was opened.
+   */
+  close (shouldFocus = true) {
+    if (this.isOpen) {
+      this.isOpen = false;
+      if (shouldFocus) {
+        this.focus();
+      }
+      dispatchCustomEvent(this, 'close');
+    }
+  }
+
+  /**
+   * Toggle the popover display.
+   */
+  toggle () {
+    if (this.isOpen) {
+      this.close();
+    }
+    else {
+      this.open();
+    }
+  }
+
+  /**
+   * Moves the focus on the button.
+   */
+  focus () {
+    this._buttonRef.value?.focus();
+  }
+
+  // endregion
+
+  // region Lit lifecycle
+
+  updated (changedProperties) {
+    if (changedProperties.has('isOpen')) {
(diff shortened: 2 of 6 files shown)


src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  /**
+   * Closes the popover.
+   */
+  close () {
+    if (this.isOpen) {
+      this.isOpen = false;
+      this._buttonRef.value?.focus();
```
florian-sanders-cc, line 118: **suggestion:** considering this is a public method that could be used from outside in case we don't foresee, I wonder if we could add a `shouldMoveFocus` boolean flag (`true` by default) to this methods and only move focus when it's true.

For instance, imagine we programatically close several popovers at once, where should the focus go?
Same idea, imagine we have a button that is used to close all popovers at once, the focus should actually remain on that button.

These cases are not very likely/common but I think it's an easy to do so we should go for it.

florian-sanders-cc (changes requested):
Well done @pdesoyres-cc this component feels easy to use and it's very useful!

I've suggested a few fixes but mostly trivial issues and nitpicks.

My main concern is how we handle `aria-expanded` & `aria-controls` attributes with the `cc-button`.
Let's discuss in sync during the team meeting?

One last thought: we could add a `focus()` method to focus the inner button programmatically like we do on most interactive components :+1:.

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+      ` : null}
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          position: relative;
```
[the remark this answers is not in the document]
roberttran-cc: Nice catch! But I'm not sure if I would make this change. :thinking: 

Your suggestion will slightly complicate the DOM: it raises the question of how much effort we put in preventing wrong usage of the component regarding the cost of the potential trade-off?

roberttran-cc (approved):
LGTM!

I wonder if we'll' need to implement some smart behavior (like https://floating-ui.com/) in the future but, for now, it's very nice. :+1:

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+  toggle () {
+    if (this.isOpen) {
+      this.close();
+    }
+    else {
+      this.open();
+    }
+  }
+
+  updated (_changedProperties) {
```
hsablonniere, line 135: nitpick: `_changedProperties` does not need to start with a `_` (WebStorm autocomplete I guess)

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+        
+        .content {
+          position: absolute;
+          z-index: var(--cc-popover-z-index, 999);
+          padding: 0.5em;
+          border: 1px solid var(--cc-color-border-neutral, #aaa);
+          background-color: var(--cc-color-bg-default, #fff);
+          border-radius: 0.25em;
+          box-shadow: 0 2px 4px rgb(38 38 38 / 25%),
+            0 5px 15px rgb(38 38 38 / 25%);
```
hsablonniere, line 194: question: do we have tokens for this?

src/components/cc-popover/cc-popover.js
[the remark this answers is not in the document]
hsablonniere: Two very good feedbacks :muscle:

@florian-sanders-cc Technically, the behaviour would only break if someone puts `position: static`, this is very unlikely. It would still work with `absolute` or `fixed`.

@roberttran-cc Seems like the DOM would only get a `div` wrapper. I don't see this as complicated.

I'm 51% leaning for the let's keep it like this.

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+    super.disconnectedCallback();
+    this._onOutsideClickHandler.disconnect();
+    this._onEscapeKeyHandler.disconnect();
+  }
+
+  render () {
+    return html`
+      <cc-button
+        ${ref(this._buttonRef)}
+        aria-expanded=${this.isOpen ? 'true' : 'false'}
```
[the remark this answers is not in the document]
hsablonniere: I think we already had the conversation about do we need a real `cc-button` when we only want the default styles. We also use the icon and not using it while keeping the same styles is not that easy.

> We have an issue about this subject (https://github.com/CleverCloud/clever-components/issues/699), I guess we need to discuss how we want to handle this 🤔.

We need to make progress on #699, @florian-sanders-cc do you have bandwidth to tackle this?

hsablonniere (changes requested):
* Can we add a story where the default slot contains a focusable element to test the escape mechanism?
* We set the content of the button with the `button-text` slot but why a slot it if's not HTML and not an attribute with just text?
* If we have two popovers in the same page, I think clicking one button should close the other popover

src/components/cc-popover/cc-popover.js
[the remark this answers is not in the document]
florian-sanders-cc: @pdesoyres-cc is on it, we discussed it this week and he opened a great RFC to sum up our options to tackle this subject:
#849 :+1:

src/components/cc-popover/cc-popover.js
florian-sanders-cc: we don't, it's in our list of things to tackle here: #398

src/components/cc-popover/cc-popover.js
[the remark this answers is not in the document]
pdesoyres-cc: I prefer to keep using the `cc-button` and tackle #699 to have good accessibility support.

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          position: relative;
+          display: block;
+
+          --gap: 0.4em;
```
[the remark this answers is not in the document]
pdesoyres-cc: I didn't want to make this a public custom prop.
Because this value is used by two CSS rules, I just wanted a variable to avoid repeating things.

However, making it a public custom prop is also a good idea so I'll go for that.

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,217 @@
...
+        }
+        
+        .content {
+          position: absolute;
+          z-index: var(--cc-popover-z-index, 999);
+          padding: 0.5em;
+          border: 1px solid var(--cc-color-border-neutral, #aaa);
+          background-color: var(--cc-color-bg-default, #fff);
+          border-radius: 0.25em;
```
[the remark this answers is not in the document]
pdesoyres-cc: good catch

src/components/cc-popover/cc-popover.js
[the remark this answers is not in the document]
pdesoyres-cc: I'll go for a `div` wrapper

src/components/cc-popover/cc-popover.js
pdesoyres-cc: Very good suggestion. Thank you

pdesoyres-cc:
@hsablonniere 

> Can we add a story where the default slot contains a focusable element to test the escape mechanism?

I can do that.
But the Escape mechanism is already testable without it (or maybe I misunderstood).

> We set the content of the button with the button-text slot but why a slot it if's not HTML and not an attribute with just text?

This is to respect the `<cc-button>` spec which allows setting HTML in its default slot.
Maybe the slot name is a bad choice.
If you think, we don't need HTML for the button of a popover, I will go for a simple prop instead.
What do you think?

> If we have two popovers in the same page, I think clicking one button should close the other popover

That makes sense.

hsablonniere:
> I can do that. But the Escape mechanism is already testable without it (or maybe I misunderstood).

OK :+1: 

> This is to respect the `<cc-button>` spec which allows setting HTML in its default slot. Maybe the slot name is a bad choice. If you think we don't need HTML for the button of a popover, I will go for a simple prop instead.

I see, forget what I said and let's keep this slot name :+1:

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,264 @@
...
+      icon: { type: Object },
+      isOpen: { type: Boolean, attribute: 'is-open', reflect: true },
+      position: { type: String },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {string|null} Sets the accessibleName property of the underlying `cc-button` element. CAUTION: The accessible name should always start with the visible text if there is one. For instance "add to estimation - NodeJS XS" */
```
Galimede, line 69: question: `For instance "add to estimation - NodeJS XS".` Is this not relevant here?

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,264 @@
...
+
+  // region Public methods
+
+  /**
+   * Opens the popover.
+   */
+  open () {
+    if (!this.isOpen) {
+      this.isOpen = true;
+      // this._lastOpenedPopover?.close(false);
```
Galimede, line 127: question: is this a leftover?

Galimede (approved):
Nice work as usual! :muscle: 

LGTM! :+1: 

Beside my two quick questions that I left in the comments, I had another one in mind. When you test the popover with a smaller display, the popover content goes off-screen. So I was wondering if:
* Is it something that we're aware of? If so, do we have a way to fix it?
* Is it a Storybook related problem?

pdesoyres-cc:
@Galimede 

> When you test the popover with a smaller display, the popover content goes off-screen. So I was wondering if:
>
> * Is it something that we're aware of? If so, do we have a way to fix it?
> * Is it a Storybook related problem?

I am aware of this.

It's because Storybook uses iframes.

This is something we don't need to tackle right now.

src/components/cc-popover/cc-popover.js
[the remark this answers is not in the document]
pdesoyres-cc: This PR will contain the fix for #699

src/components/cc-button/cc-button.js
```diff
@@ -232,6 +240,8 @@ export class CcButton extends LitElement {
 
     const tabIndex = this.skeleton ? -1 : null;
 
+    console.log('render button', this.a11yExpanded);
```
florian-sanders-cc, line 243: **nitpick:** forgot a console log here :wink:

florian-sanders-cc (changes requested):
Gj @pdesoyres-cc, apart from a small console.log to be removed, it's all good!

roberttran-cc (approved):

Galimede (approved):

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * When the popover is open, pressing `esc` will close the popover and focus the `button` element.
+ *
+ * ## Usage
+ *
+ * ```html
+ * <cc-popover>
+ *   <span slot="button-text">Click me</span>
+ *   <div>This is a content to be displayed when button is clicked.</div>
+  * </cc-popover>
```
hsablonniere, line 41: nitpick: we can remove the space here

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * * `aria-expanded`: set to `true` when popover is open, `false` when it is closed.
+ *
+ * When the popover is open, pressing `esc` will close the popover and focus the `button` element.
+ *
+ * ## Usage
+ *
+ * ```html
+ * <cc-popover>
+ *   <span slot="button-text">Click me</span>
```
hsablonniere, line 39: issue: the slot is named `button` now

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,263 @@
...
+    });
+
+    this._onEscapeKeyHandler = new EventHandler(this, 'keydown', (event) => {
+      if (event.key === 'Escape') {
+        this.close();
+      }
+    });
+
+    let lastOpenedPopover = null;
+    this._onCcPopoverOpen = new EventHandler(window, 'cc-popover:open', (event) => {
```
hsablonniere, line 104: nitpick: maybe this should also have a `Handler` suffix like the other

src/components/cc-popover/cc-popover.js
```diff
@@ -0,0 +1,263 @@
...
+    `;
+  }
+
+  static get styles () {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: block;
+
```
hsablonniere, line 224: nitpick: should we remove this empty line or is it the formatter?

src/components/cc-popover/cc-popover.js
pdesoyres-cc: I'll rename that to "button-content"

src/components/cc-popover/cc-popover.js
pdesoyres-cc: No, we can't. Stylelint you know...

PR #844, opened 2023-09-22
Resize observer mixin: move to a reactive controller

### LINKED ISSUE #551: Rewrite the `withResizeObserver` mixin as a reactive controller
(title an description edited to reflect the latest proposition)

## Context

* We created the `withResizeObserver` mixin to hide the boilerplate of using `ResizeObserver` and provide nicers APIs for our components who need to be responsive
* Lit introduced the concept of [reactive controllers](https://lit.dev/docs/api/controllers/#ReactiveController)
  * This is clearly a better approach/solution than mixins
* The lit-labs project provides a [ready to use reactive controller](https://github.com/lit/lit/blob/main/packages/labs/observers/src/resize-controller.ts) to use the `ResizeObserver` API in a component

## About the [`ResizeController`](https://github.com/lit/lit/blob/main/packages/labs/observers/src/resize-controller.ts) from lit-labs

🙂 Pros:

* It doesn't do much things we don't need
* It seems very smart (just one `ResizeObserver` for many instances of the same component but multiple `observe` calls)

🤬 Cons:

* It doesn't provide the nice CSS API (with the automatic attributes on breakpoints) we have in our mixin
  * Most of those could be replaced by [CSS container queries](https://caniuse.com/css-container-queries) but the support is not good enough
* It just provides the `ResizeObserverEntry[]` but as we experienced it, it's not enough for us

## Proposition

* Create our own reactive controller reusing the code from our mixin
  * Try to see if we can reuse lit-labs's `ResizeController` or if we can get inspiration from it
* Replace all our mixin usage with the reactive controller
* Remove the mixin (is it a breaking change if it was not exposed / documented for public usage?)

WDYT?

florian-sanders-cc:
Fixes #551 

## Context

We decided to go with our own `resizeController` but with an API that does not rely on public properties:

- the controller should accept a `breakpoints` parameter,
- if `breakpoints != null` => it should add the relevant attributes,
- the controller should `requestUpdate` the `host` everytime a resize happens,
- if we need the `width` or `height` of the component, we can compute it within the `render` (or `willUpdate` if we need the size in several sub-renders).

## How to review?

- Review each commit,
- Review the [Resize Controller story](https://clever-components-preview.cellar-c2.services.clever-cloud.com/resize-observer/from-mixin-to-controller/index.html?path=/story/%F0%9F%95%B9%EF%B8%8F-controllers-resizecontroller--default-story),
- Review the [Resize Controller doc](https://clever-components-preview.cellar-c2.services.clever-cloud.com/resize-observer/from-mixin-to-controller/index.html?path=/docs/%F0%9F%95%B9%EF%B8%8F-controllers-resizecontroller--default-story),
- Review each impacted component using the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/resize-observer/from-mixin-to-controller/index.html), try to resize it and check that it does not break and it does not throw any error in the console:
  - cc-invoice-list
  - cc-invoice-table
  - cc-map
  - cc-orga-member-card
  - cc-overview
  - cc-pricing-product
  - cc-pricing-product-consumption
  - cc-tile-metrics
  - cc-tile-requests
  - cc-zone-input
- Don't forget to check the ADR! 

## TODO

- [x] Code the resize controller,
- [x] Provide a doc + a story for it,
- [x]  Replace `withResizeObserver` mixin with the `ResizeController` everywhere,
- [x] Write a short ADR about this.

<details>
<summary>Archived suggestions</summary>

In #551, we decided to rewrite the `resizeObserver` mixin and use a reactive controller instead.

The `resizeObserver` mixin has two features that we need:

- update the component internal `_size` prop with the current `width` of the component, everytime it is resized (we rely on a public `onResize` method to do so),
- set and remove attributes based on publicly declared `breakpoints` on the component everytime it is resized.

We want the `resizeController` to be able to do at least the same.

There are several ways to do so:

- building our own `resizeController` and using the component public methods and properties like we do with the mixin (first commit),
- relying on the lit-labs `resizeController` and using the component public methods and properties (second commit).
- relying on the lit-labs `resizeController` and using its API (`value` property of the observer + `willUpdate` lifecycle hook) (third commit).

All these options work, you can check the `cc-orga-member-card` / `cc-orga-member-list` (data loaded stories) since this component relies on both features (attribute & size prop).

</details>


src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -82,22 +87,6 @@ export class CcOrgaMemberCard extends LitElement {
...
-     * used to:
-     * - wrap buttons when the component width is below 740.
-     * - switch to a vertical card layout when the width is below 580.
-     * - show buttons and badges below each other when the width is below 350.
-     *
-     * @protected
-     */
-    this.breakpoints = {
-      width: [BREAKPOINT_TINY, BREAKPOINT_SMALL, BREAKPOINT_MEDIUM],
-    };
```
florian-sanders-cc, line 94: using the `lit-labs` controller API, we no longer need to declare a public "protected" `breakpoints` property.

The controller will trigger the trigger the `willUpdate` lifecycle hook everytime a resize happens.

src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -106,8 +95,12 @@ export class CcOrgaMemberCard extends LitElement {
     /** @type {Ref<CcSelect>} */
     this._roleRef = createRef();
 
-    /** @type {string} Set by `withResizeObserver` mixin. See the `onResize` method for more info. */
-    this._size = '';
+    /** @type {number|null} set within the `willUpdate` lifecycle hook triggered everytime a resize happens */
+    this._size = null;
+
+    this._componentWidthObserver = new ResizeController(this, {
+      callback: getWidthOnResize,
```
florian-sanders-cc, line 102: This is actually not mandatory.
Since we get the `width` of the component by using `element.getBoundingClientRect()`, we could just set up the controller with no callback and do:

```js
  this._size = this.getBoundingClientRect();
```

within `willUpdate` directly :thinking:.

florian-sanders-cc:
We decided to go with our own `resizeController` but with an API that does not rely on public properties:

- the controller should accept a `breakpoints` parameter,
- if `breakpoints != null` => it should add the relevant attributes,
- the controller should `requestUpdate` the `host` everytime a resize happens,
- if we need the `width` or `height` of the component, we can compute it within the `render` (or `willUpdate` if we need the size in several sub-renders).

- [ ] TODO Write a short ADR about this

src/controllers/resize-controller.stories.js
```diff
@@ -0,0 +1,328 @@
...
+        name: 'Florian',
+        team: 'Dog 🐶',
+      },
+      {
+        name: 'Robert',
+        team: 'Dog 🐶',
+      },
+      {
+        name: 'Pierre',
+        team: 'NC 🏳️',
```
pdesoyres-cc, line 36: I have :frog: in my pond ;)

pdesoyres-cc (changes requested):
Nice work @florian-sanders-cc

src/controllers/resize-controller.js
```diff
@@ -0,0 +1,69 @@
...
+    this._host = host;
+
+    /** @type {Breakpoints|undefined} if defined sets attributes `w-lt-${breakpoint}` for each breakpoint */
+    this._breakpoints = breakpoints;
+
+    this._observer = new ResizeObserver(() => this._onResize());
+  }
+
+  _onResize () {
+    const { width } = this._host.getBoundingClientRect();
```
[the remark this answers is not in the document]
florian-sanders-cc: @pdesoyres-cc yes very good idea. This is actually what the Lit Labs implementation does but I don't know why I didn't go this way :thinking: 

Anyway, I'll let others review and suggest stuff but I'll probably go for this :+1: (and I'll also restore the comment as you suggest)

src/controllers/resize-controller.stories.js
florian-sanders-cc: Frogs are cool too, I'll add it to the story :joy:

src/controllers/resize-controller.stories.js
```diff
@@ -0,0 +1,328 @@
...
+  static get properties () {
+    return {
+      breakpoints: { type: Object },
+    };
+  }
+
+  constructor () {
+    super();
+
+    this._teams = [
```
HeleneAmouzou, line 17: praise: great idea ! :smile_cat:

HeleneAmouzou (commented):
GG @florian-sanders-cc ! 
There is just one small issue I can not solve: when I review the impacted components using the preview and try to resize it, sometimes the 350px size throw an error in the console ("Uncaught DOMException: Index or size is negative or greater than the allowed amount"), but as soon as I refresh the error is gone and when I switch again from other sizes to 350px it does not always re-throw it.

src/controllers/resize-controller.js
```diff
@@ -0,0 +1,69 @@
...
+        `ResizeController error: browser does not support ResizeObserver.`,
+      );
+      return;
+    }
+
+    host.addController(this);
+
+    this._host = host;
+
+    /** @type {Breakpoints|undefined} if defined sets attributes `w-lt-${breakpoint}` for each breakpoint */
```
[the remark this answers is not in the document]
florian-sanders-cc: it's been forgotten, good catch :smile: thanks!

src/controllers/resize-controller.js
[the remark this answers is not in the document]
hsablonniere: Agreed with @pdesoyres-cc, I thought we decided that the ResizeController class would expose the `width` via a public getter like Lit does.

src/components/cc-invoice-table/cc-invoice-table.js
```diff
@@ -46,20 +45,16 @@ export class CcInvoiceTable extends withResizeObserver(LitElement) {
     /** @type {Invoice[]|null} Sets the list of invoices. */
     this.invoices = null;
 
-    /** @type {number|null} */
-    this._width = null;
-  }
-
-  onResize ({ width }) {
-    this._width = width;
+    new ResizeController(this);
```
hsablonniere, line 48: issue: let's keep a reference to the controller so we can get its `width` property

```js
this._resizeController = new ResizeController(this);
```

src/components/cc-invoice-table/cc-invoice-table.js
```diff
@@ -46,20 +45,16 @@ export class CcInvoiceTable extends withResizeObserver(LitElement) {
...
   }
 
   render () {
 
+    const { width } = this.getBoundingClientRect();
     // NOTE: This value is arbitrary, we don't have a better solution for now
     // It's a bit more than the width of the table in french (which is the largest) and with both links (download and pay)
     // The table width is mostly stable since the with of the amount is fixed and the rest is almost always the same number of characters
-    const bigMode = (this._width > 700);
+    const bigMode = (width > 700);
```
hsablonniere, line 57: issue: let's use the controller reference for its width property:

```js
const bigMode = (this._resizeController.width > 700);
```

src/components/cc-map/cc-map.js
```diff
@@ -313,10 +308,13 @@ export class CcMap extends withResizeObserver(LitElement) {
   }
 
   render () {
-
     const noHeatmapPoints = (!this.error && this.mode === 'heatmap' && this.heatmapPoints?.length === 0);
     const errorMode = this.loading ? 'loading' : 'info';
 
+    if (this._map != null) {
+      this._map.invalidateSize();
+    }
```
hsablonniere, line 316: question: I tried but I don't understand why it works compared to the `onResize () { this._map.invalidateSize(); }`

src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -53,13 +53,13 @@ const BREAKPOINT_TINY = 350;
...
  */
-export class CcOrgaMemberCard extends withResizeObserver(LitElement) {
+export class CcOrgaMemberCard extends LitElement {
 
   static get properties () {
     return {
       authorisations: { type: Object },
       member: { type: Object },
-      _size: { type: String, state: true },
+      _width: { type: Number, state: true },
```
hsablonniere, line 62: issue: we don't need this

src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -103,8 +91,19 @@ export class CcOrgaMemberCard extends withResizeObserver(LitElement) {
...
+    /**
+     * @type {ResizeController}
+     * used to:
+     * - wrap buttons when the component width is below 740.
+     * - switch to a vertical card layout when the width is below 580.
+     * - show buttons and badges below each other when the width is below 350.
+     */
+    new ResizeController(this, {
+      width: [BREAKPOINT_TINY, BREAKPOINT_SMALL, BREAKPOINT_MEDIUM],
+    });
```
hsablonniere, line 106: issue: let's keep a reference to the controller so we can get its `width` property

```js
this._resizeController = new ResizeController(this);
```

src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -231,6 +225,10 @@ export class CcOrgaMemberCard extends withResizeObserver(LitElement) {
     });
   }
 
+  willUpdate () {
+    this._width = this.getBoundingClientRect().width;
+  }
```
hsablonniere, line 230: issue: we don't need this

src/components/cc-orga-member-card/cc-orga-member-card.js
```diff
@@ -231,6 +225,10 @@ export class CcOrgaMemberCard extends withResizeObserver(LitElement) {
...
+
   render () {
 
     const waiting = this.member.state === 'updating' || this.member.state === 'deleting';
```
hsablonniere, line 234: issue: we can use the width from the controller directly inside the render

```js
const width = this._resizeController.width;
```

src/components/cc-pricing-product-consumption/cc-pricing-product-consumption.js
```diff
@@ -89,16 +88,10 @@ export class CcPricingProductConsumption extends withResizeObserver(LitElement)
...
-    this._size = null;
-
     /** @type {SectionStates} Sets the state of the section. It is modified everytime the quantity is changed, the section is shown / hidden, or if the unit value changes. */
     this._sectionStates = {};
-  }
 
-  /* Used by the `withResizeObserver` mixin. */
-  onResize ({ width }) {
-    this._size = width;
+    new ResizeController(this);
```
hsablonniere, line 94: issue: let's keep a reference to the controller so we can get its `width` property

```js
this._resizeController = new ResizeController(this);
```

src/components/cc-pricing-product-consumption/cc-pricing-product-consumption.js
```diff
@@ -360,9 +353,10 @@ export class CcPricingProductConsumption extends withResizeObserver(LitElement)
    * @param {PricingSection[]} sections
    */
   _renderLoaded (sections) {
+    const { width } = this.getBoundingClientRect();
```
hsablonniere, line 356: issue: let's use the controller reference for its width property:

```js
const width = this._resizeController.width;
```

src/components/cc-pricing-product/cc-pricing-product.js
```diff
@@ -99,12 +98,7 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
     */
     this.temporalities = DEFAULT_TEMPORALITY_LIST;
 
-    /** @type {number|null} Set by the `withResizeObserver` mixin. The width of the component in `px`. See the `onResize` method for more info. */
-    this._size = null;
-  }
-
-  onResize ({ width }) {
-    this._size = width;
+    new ResizeController(this);
```
hsablonniere, line 101: issue: let's keep a reference to the controller so we can get its `width` property

```js
this._resizeController = new ResizeController(this);
```

src/components/cc-pricing-product/cc-pricing-product.js
```diff
@@ -263,13 +257,15 @@ export class CcPricingProduct extends withResizeObserver(LitElement) {
    * @param {Feature[]} productFeatures - the list of features to display
    */
   _renderProductPlans (productName, productPlans, productFeatures) {
+
+    const { width } = this.getBoundingClientRect();
```
hsablonniere, line 261: issue: let's use the controller reference for its width property:

```js
const width = this._resizeController.width;
```

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -252,6 +248,9 @@ export class CcTileMetrics extends withResizeObserver(LitElement) {
 
     const panel = this._getCurrentPanel();
 
+    this._cpuChart?.resize();
+    this._memChart?.resize();
+
```
hsablonniere, line 253: question: I tried but I don't understand why it works compared to the `onResize` + `this.*chart.resize()`

src/components/cc-tile-requests/cc-tile-requests.js
```diff
@@ -72,21 +72,8 @@ export class CcTileRequests extends withResizeObserver(LitElement) {
...
-    }
-    if (width >= 380 && width < 540) {
-      this._barCount = 8;
-    }
-    if (width >= 540) {
-      this._barCount = 12;
-    }
-    this._refreshChart();
-    this.requestUpdate();
+    new ResizeController(this);
```
hsablonniere, line 76: issue: let's keep a reference to the controller so we can get its `width` property

```js
this._resizeController = new ResizeController(this);
```

src/components/cc-tile-requests/cc-tile-requests.js
```diff
@@ -233,6 +220,22 @@ export class CcTileRequests extends withResizeObserver(LitElement) {
     });
   }
 
+  willUpdate () {
+    const { width } = this.getBoundingClientRect();
```
hsablonniere, line 224: issue: let's use the controller reference for its width property:

```js
const width = this._resizeController.width;
```

src/controllers/resize-controller.stories.js
hsablonniere: So nice! :heart:

docs/adr/adr-0024-from-resize-mixin-to-resize-controller.md
```diff
@@ -0,0 +1,272 @@
...
+- A straightforward API. We want to pass the breakpoints directly to our controller and we want the controller to trigger the update lifecycle because we think it's a really great idea from Lit Labs.
+
+### Limits
+
+- Moving away from the mixin means that this code only works with Lit Elements. The mixin could be used with native elements but the reactive controller cannot. This is something we accept because we actually work only with Lit Elements and we needed a solution for this usecase.
+- The Lit Labs implementation allowed to observe several elements from the same controller but this is something we don't need.
+
+## What's next?
+
+When container queries and containers units support is good enough, the breakpoints feature of the controller will be obsolete.
```
hsablonniere, line 262: suggestion: can we add a link to container queries from MDN or something?

hsablonniere (changes requested):
Nice work @florian-sanders-cc, especially the ADR :clap:

Like @pdesoyres-cc, I think the Controller should expose the `width`. I tried to annotate all files according for this change but you'll have to update the story, the doc and the ADR :pray:

It's clearly a nitpick but could you move the commit that removes the mixin at the end (or just before the ADR)?

---

FYI, even if we just read and not write, each time we call `getBoundingClientRect()`, we trigger a layout/reflow. It's always better not to call it too much so if the component asks the controller, it's better.

List of stuffs that trigger a layout/reflow:

https://gist.github.com/paulirish/5d52fb081b3570c81e3a

src/controllers/resize-controller.js
```diff
@@ -0,0 +1,84 @@
...
+        (this.width < breakpoint)
+          ? this._host.setAttribute(ltAttr, '')
+          : this._host.removeAttribute(ltAttr);
+      });
+    }
+
+    this._host.requestUpdate();
+
+    if (this._callback != null) {
+      this._host.updateComplete.then(() => {
```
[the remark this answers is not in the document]
florian-sanders-cc: Good point! I'll change that, thanks :wink:

src/controllers/resize-controller.js
```diff
@@ -33,19 +33,25 @@ export class ResizeController {
...
     this._widthBreakpoints = options?.widthBreakpoints;
 
     this._observer = new ResizeObserver(() => this._onResize());
   }
 
+  get width () {
+    return this._width;
+  }
+
+  set width (newWidth) {
```
pdesoyres-cc, line 51: fix: Oh no, you don't want anybody but the controller itself to change the width. This is the whole point of making width private with only a getter on it. (This makes the width readonly from outside this class)

src/controllers/resize-controller.js
```diff
@@ -33,19 +33,25 @@ export class ResizeController {
...
+    this._width = newWidth;
+  }
+
   _onResize () {
     // NOTE: We could use entries[0].borderBoxSize.inlineSize but not supported in Chrome, Safari or polyfill
     this.width = this._host.getBoundingClientRect().width;
```
pdesoyres-cc, line 57: fix: use the private property directly
```suggestion
    this._width = this._host.getBoundingClientRect().width;
```

src/controllers/resize-controller.js
```diff
@@ -68,9 +74,7 @@ export class ResizeController {
     this._host.requestUpdate();
 
     if (this._callback != null) {
-      this._host.updateComplete.then(() => {
-        this._callback();
-      });
+      this._callback(this.width);
```
pdesoyres-cc, line 77: nitpick: you could use the internal property instead of the getter
```suggestion
this._callback(this._width);
```

src/controllers/resize-controller.js
```diff
@@ -68,9 +74,7 @@ export class ResizeController {
     this._host.requestUpdate();
 
     if (this._callback != null) {
```
pdesoyres-cc, line 76: nitpick: you can simplify that with :
```
this._callback?.(this._width);
```

pdesoyres-cc (changes requested):
Hey, it's me again! We're nearly there.

src/controllers/resize-controller.js
florian-sanders-cc: ahhh yes of course :sweat_smile:  thanks

florian-sanders-cc:
@pdesoyres-cc Thanks for your patience :joy: should be ok now? :eyes:

pdesoyres-cc (approved):
Well done Florian!

florian-sanders-cc:
> There is just one small issue I can not solve: when I review the impacted components using the preview and try to resize it, sometimes the 350px size throw an error in the console ("Uncaught DOMException: Index or size is negative or greater than the allowed amount"), but as soon as I refresh the error is gone and when I switch again from other sizes to 350px it does not always re-throw it.

@HeleneAmouzou I couldn't reproduce this issue so I'll merge but when you are back we can test again together and if we still reproduce, we can open an issue and about this :+1:

PR #847, opened 2023-09-27
fix(cc-toggle / cc-select): add and reflect name attributes

### LINKED ISSUE #581: cc-toggle / cc-select: add a `name` prop to populate the name attribute of these form controls
The `cc-toggle` component implements a hard coded name attribute.
We need the `name` attribute to be populated from a prop so that it can be retrieved in a `FormData` when we move to this.
This is a breaking change.

The `cc-select` element implements a `name` prop that is not used within the component :facepalm:.
This prop should populate a `name` attribute on the `select` element.

HeleneAmouzou:
Fixes #581

### DIFF
--- a/src/components/cc-toggle/cc-toggle.js
+++ b/src/components/cc-toggle/cc-toggle.js
@@ -51,6 +51,7 @@ export class CcToggle extends LitElement {
       inline: { type: Boolean, reflect: true },
       legend: { type: String },
       multipleValues: { type: Array, attribute: 'multiple-values', reflect: true },
+      name: { type: String, reflect: true },
       subtle: { type: Boolean },
       value: { type: String, reflect: true },
     };
@@ -78,6 +79,11 @@ export class CcToggle extends LitElement {
 
     /** @type {string[]} Enables multiple mode and sets the selected values. */
     this.multipleValues = null;
+
+    /** @type {string|null} Sets `name` attribute on native `<input>` element.
+     * If left `null`, the name of the native `<input>` will be 'toggle'. In order to navigate through a group of inputs using the arrow keys, each `<input>` must have the same `name` value.
+     */
+    this.name = null;
 
     /** @type {boolean} Uses a more subtle display mode, less attractive to the user's attention. */
     this.subtle = false;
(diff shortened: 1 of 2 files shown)


src/components/cc-toggle/cc-toggle.js
```diff
@@ -130,7 +134,7 @@ export class CcToggle extends LitElement {
           ${repeat(this.choices, ({ value }) => value, ({ label, image, value }) => html`
             <input
               type=${type}
-              name="toggle"
+              name=${this.name ?? 'toggle'}
```
pdesoyres-cc, line 137: I think you put a fallback to 'toggle' to avoid a breaking change.
I would prefer to introduce a breaking change and have a homogeneous way of handling name attributes on all our input components.

If you don't know what to do with this comment, let's wait for other reviewers to give their thoughts about that.

pdesoyres-cc (commented):
Nice first PR Hélène :clap:

I have just one comment about breaking change or not.

src/components/cc-toggle/cc-toggle.js
```diff
@@ -51,6 +51,7 @@ export class CcToggle extends LitElement {
       inline: { type: Boolean, reflect: true },
       legend: { type: String },
       multipleValues: { type: Array, attribute: 'multiple-values', reflect: true },
+      name: { type: String, reflect: true },
```
hsablonniere, line 54: issue(mild): The reflect is needed when we use the property in CSS selector. In this situation, we can remove it.

hsablonniere (changes requested):
Well done @HeleneAmouzou :+1: 

suggestion: I don't think we need the comment about name defaulting to toggle at two places. One is enough, but maybe it's just me.

src/components/cc-toggle/cc-toggle.js
florian-sanders-cc: I would prefer we keep reflecting it.

The name attribute is useful in a native context where element internals is not supported to retrieve elements participating in a form:

```js
const myFormElements = form.querySelectorAll([name])
```

We also use the name attribute in our experimentations with @pdesoyres-cc :wink:

pdesoyres-cc (approved):
LGTM. This is perfect

HeleneAmouzou:
> suggestion: I don't think we need the comment about name defaulting to toggle at two places. One is enough, but maybe it's just me.

I will just keep the two comments, it seems clearer to me.

hsablonniere (approved):
All good!

florian-sanders-cc (approved):
LGTM! Gg @HeleneAmouzou !!

PR #853, opened 2023-10-06
fix(cc-icon): improve SVG aria attributes behavior

### LINKED ISSUE #770: `cc-icon`: error when `.icon` and `.accessibleName` props are not set or null
## Context

When both the `.icon` and `.accessibleName` props are null or not set, the component still tries to set ARIA attributes on its inner `<svg>` element which does not exist.

It throws the following error: 
```js
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'setAttribute')
    at m.updated (cc-icon.js:71:13)
    at m._$AE (reactive-element.js:1370:10)
    at m.performUpdate (reactive-element.js:1335:12)
    at m.scheduleUpdate (reactive-element.js:1253:17)
    at m._$Ej (reactive-element.js:1225:25)
```

This is an edge case but it's also very easy to fix so why not?

HeleneAmouzou:
* only try to update SVG aria attributes if SVG is in the DOM
* don't create several `<title>` tags when `accessibleName` changes
* apply aria attributes on SVG when `icon` changes

Fixes #770

### DIFF
--- a/src/components/cc-icon/cc-icon.js
+++ b/src/components/cc-icon/cc-icon.js
@@ -65,23 +65,25 @@ export class CcIcon extends LitElement {
   }
 
   updated (changedProperties) {
-    if (!this.skeleton && changedProperties.has('accessibleName')) {
-      const svg = this.shadowRoot.querySelector('svg');
+    const shouldPatchSvg = changedProperties.has('accessibleName') || changedProperties.has('icon');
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (approved):
LGTM. Well done Hélène.

florian-sanders-cc (approved):
Gg @HeleneAmouzou, LGTM!

PR #859, opened 2023-10-11
fix(cc-pricing-product): filter out addon plans without zones

### LINKED ISSUE #858: `cc-pricing-product`: filter out products with no `zone`
## Context

We have disabled `MongoDB` `dev` plans but the pricing API still returns it.

In the console, we rely on the fact that this plan is no longer attached to any `zone` to filter it.

We want to do the same within our pricing components. If a plan has no `zone`, then we shouldn't display it.

florian-sanders-cc:
Fixes #858

## Context

We have disabled MongoDB dev plans but the pricing API still returns it.

In the console, we rely on the fact that this plan is no longer attached to any zone to filter it.

We want to do the same within our pricing components. If a plan has no zone, then we shouldn't display it.

## How to review?

- review the commit,
- run locally and go to [localhost:6006/demo-smart/cc-pricing-page/](http://localhost:6006/demo-smart/cc-pricing-page/). => The "MongoDB" section should no longer display the `dev` plan.

### DIFF
--- a/demo-smart/cc-pricing-page/index.html
+++ b/demo-smart/cc-pricing-page/index.html
@@ -127,6 +127,13 @@
         </div>
 
         <div class="product">
+          <cc-smart-container context='{"productId":"mongodb-addon"}'>
+            <h2>MongoDB</h2>
+            <cc-pricing-product mode="addon" action="add"></cc-pricing-product>
+          </cc-smart-container>
+        </div>
+
+        <div class="product">
             <cc-smart-container context='{ "productId": "cellar" }'>
               <h2>Cellar</h2>
               <cc-pricing-product-consumption></cc-pricing-product-consumption>
--- a/src/lib/product.js
+++ b/src/lib/product.js
@@ -1,8 +1,10 @@
 export function formatAddonProduct (addonProvider, priceSystem, selectedFeatures) {
+  // We filter out add-ons that are not attached to any zone. This is sometimes done on dev plans to disable them.
+  const addonPlansWithZones = addonProvider.plans.filter((plan) => plan.zones.length > 0);
   return {
     name: addonProvider.name,
     productFeatures: formatAddonFeatures(addonProvider.features, selectedFeatures),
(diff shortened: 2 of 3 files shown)


src/lib/product.js
```diff
@@ -1,8 +1,10 @@
 export function formatAddonProduct (addonProvider, priceSystem, selectedFeatures) {
+  // we filter addons that are not attached to any zone. This removes the `MongoDB dev` plan
```
hsablonniere, line 2:
```suggestion
  // We filter out add-ons that are not attached to any zone. This is sometimes done on dev plans to disabled them.
```

demo-smart/cc-pricing-page/index.html
```diff
@@ -126,6 +126,13 @@ <h2>Redis</h2>
             </cc-smart-container>
         </div>
 
+        <div class="product">
+          <cc-smart-container context='{"productId":"mongodb-addon"}'>
+            <h2>MongoDB</h2>
+            <cc-pricing-product mode="addon" action="add"></cc-pricing-product>
+          </cc-smart-container>
+      </div>
```
Galimede, line 134: nitpick(non-blocking): the indentation feels off here :thinking:

demo-smart/cc-pricing-page/index.html
Galimede, line 1: question: do we want to add a link to this demo-page somewhere on the `demo-smart/index.html` page? I went there first before noticing that there was no direct link to the pricing one. :sweat_smile:

Galimede (approved):
GG @florian-sanders-cc, LGTM! :muscle: 

I just left a nitpick but feel free to tackle it or not as it's only a demo file. :wink:

demo-smart/cc-pricing-page/index.html
florian-sanders-cc: Good idea!

demo-smart/cc-pricing-page/index.html
florian-sanders-cc: good catch :wink:

PR #868, opened 2023-10-18
cc-article-list: fix XML parser

### LINKED ISSUE #867: cc-article-listl: fix XML parser
As our new site has been released, the RSS feed has changed a little bit causing errors. We need to investigate and change our XML parser method so that the smart component can work properly again.

Galimede:
Fixes #867 

## How to review?

* Check the UI
  * Go to `https://clever-components-preview.cellar-c2.services.clever-cloud.com/fix-article-smart/index.html?path=%2Fstory%2F%F0%9F%9B%A0-homepage-cc-article-list-%F0%9F%92%A1-smart--page` 
  * Check that the component load correctly
  * Check both english and french
  * (With the new RSS the description is longer, I can't really do much about that sadly)
* Review the code except the sandbox commit. I thought we could use it but the smart page can act as a sandbox for us :wink:

### DIFF
--- a/src/lib/xml-parser.js
+++ b/src/lib/xml-parser.js
@@ -9,24 +9,33 @@
  * @returns {Article[]}
  */
 export function parseRssFeed (xmlStr, limit = 9) {
+
+  if (limit <= 0) {
+    return [];
+  }
+
   const xmlStrTrimmed = xmlStr.trim();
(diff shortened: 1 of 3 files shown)


florian-sanders-cc (approved):
LGTM, gg @Galimede!

pdesoyres-cc (approved):
LGTM. Well done @Galimede

PR #873, opened 2023-10-23
fix(cc-img): populate alt value

### LINKED ISSUE #726: cc-img: use `text` value as `alt` value for the inner `<img>` tag
# Context

The `cc-img` component has a `text` prop that is used to display a text when the image is not loaded.
The `alt` attribute of the `` tag inside the component is always empty and there is no way to set it from outside the component.

# Solutions

Either use `text` value as `alt` value (when there is no error), or introduce a separate `accessible-name` prop.

HeleneAmouzou:
Fixes #726

### DIFF
--- a/src/components/cc-header-orga/cc-header-orga.js
+++ b/src/components/cc-header-orga/cc-header-orga.js
@@ -66,7 +66,7 @@ export class CcHeaderOrga extends LitElement {
         ` : ''}
 
         ${!this.error ? html`
-          <cc-img class="logo" ?skeleton=${skeleton} src=${ifDefined(orga.avatar)} text=${initials}></cc-img>
+          <cc-img class="logo" ?skeleton=${skeleton} src=${ifDefined(orga.avatar)} accessible-name=${initials}></cc-img>
           <div class="details">
             <div class="name ${classMap({ skeleton })}">${orga.name}</div>
             ${orga.cleverEnterprise ? html`
--- a/src/components/cc-img/cc-img.js
+++ b/src/components/cc-img/cc-img.js
@@ -20,7 +20,7 @@ export class CcImg extends LitElement {
     return {
       skeleton: { type: Boolean, reflect: true },
       src: { type: String },
-      text: { type: String },
+      accessibleName: { type: String, attribute: 'accessible-name' },
       _error: { type: Boolean, state: true },
       _loaded: { type: Boolean, state: true },
     };
@@ -36,7 +36,7 @@ export class CcImg extends LitElement {
     this.src = null;
 
     /** @type {string|null} Sets short fallback text to display when the image cannot be loaded or if `src` is not defined and `skeleton` is `false`. */
-    this.text = null;
+    this.accessibleName = null;
 
     /** @type {boolean} */
     this._error = false;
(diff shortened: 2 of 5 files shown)


src/components/cc-img/cc-img.js
```diff
@@ -20,7 +20,7 @@ export class CcImg extends LitElement {
     return {
       skeleton: { type: Boolean, reflect: true },
       src: { type: String },
-      text: { type: String },
+      accessibleName: { type: String },
```
Galimede, line 23: suggestion: maybe we could add an attribute name `attribute: 'accessible-name` as it's two words?
```suggestion
      accessibleName: { type: String, attribute: 'accessible-name' },
```

Galimede (approved):
Just a suggestion otherwise LGTM! :+1:  :muscle:

src/components/cc-img/cc-img.stories.js
```diff
@@ -22,24 +22,24 @@ const conf = {
 
 export const defaultStory = makeStory(conf, {
   items: [
-    { text: 'OMG' },
-    { text: 'OMG', skeleton: true },
-    { text: 'OMG', skeleton: true, src: 'http://placekitten.com/200/200' },
+    { accessibleName: 'OMG' },
+    { skeleton: true },
```
florian-sanders-cc, line 26: **suggestion:** I remember we removed `text` / `accessibleName` here to see if the component behaved well when the prop is `undefined` / `null`.

Do we want to create a specific story for that maybe?

A copy of the `defaultStory` with no `accessibleName` in the items?

src/components/cc-header-orga/cc-header-orga.js
```diff
@@ -66,7 +66,7 @@ export class CcHeaderOrga extends LitElement {
         ` : ''}
 
         ${!this.error ? html`
-          <cc-img class="logo" ?skeleton=${skeleton} src=${ifDefined(orga.avatar)} text=${initials}></cc-img>
+          <cc-img class="logo" ?skeleton=${skeleton} src=${ifDefined(orga.avatar)} accessibleName=${initials}></cc-img>
```
florian-sanders-cc, line 69: **nitpick:** when we're dealing with `string` and other simple types, we tend to use attributes instead of properties (same for all other occurrrences)
```suggestion
          <cc-img class="logo" ?skeleton=${skeleton} src=${ifDefined(orga.avatar)} accessible-name=${initials}></cc-img>
```

florian-sanders-cc (changes requested):
Good job @HeleneAmouzou, I only have a few nitpick comments.

pdesoyres-cc (approved):
LGTM. Well done @HeleneAmouzou

florian-sanders-cc (approved):

src/components/cc-img/cc-img.js
```diff
@@ -35,8 +35,8 @@ export class CcImg extends LitElement {
     /** @type {string|null} Sets `src` attribute on inner native `<img>` element. */
     this.src = null;
 
-    /** @type {string|null} Sets short fallback text to display when the image cannot be loaded or if `src` is not defined and `skeleton` is `false`. */
-    this.text = null;
+    /** @type {string|null} Sets short fallback accessibleName to display when the image cannot be loaded or if `src` is not defined and `skeleton` is `false`. */
```
roberttran-cc, line 38: **🤓 nitpick:** Actually, I think it makes more sense to keep the word "text" for this comment.

src/components/cc-img/cc-img.js
```diff
@@ -65,14 +65,16 @@ export class CcImg extends LitElement {
...
     const isLoading = (this.src != null && !this._loaded && !this._error);
     const isSkeleton = (this.skeleton || isLoading);
-    const displayText = (this.src == null || this._error);
+    const displayAccessibleName = (this.src == null || this._error);
     return html`
-      <div class="wrapper ${classMap({ skeleton: isSkeleton, loaded: this._loaded, text: displayText })}">
-        <img src=${ifDefined(this.src ?? undefined)} @load=${this._onLoad} @error=${this._onError} alt="">
-        ${displayText ? html`
-          <div class="error-msg">${this.text}</div>
+      <div class="wrapper ${classMap({ skeleton: isSkeleton, loaded: this._loaded, accessibleName: displayAccessibleName })}">
```
roberttran-cc, line 73: **🎯 suggestion:** Most of the time, CSS class names are snake-cased so I would suggest this instead:
```suggestion
      <div class="wrapper ${classMap({ skeleton: isSkeleton, loaded: this._loaded, 'accessible-name': displayAccessibleName })}">
```

roberttran-cc (commented):
Great job! Clean & it improves the component. :ok_hand: 

I just have a small feedback and a suggestion. :eyes:

roberttran-cc (approved):
Perfect! :partying_face:

PR #878, opened 2023-11-02
feat(cc-logs-instances) init Component

### LINKED ISSUE #870: cc-loader: add an accessible name
The `cc-loader` is no more than an icon. It just has animation over it. So just like the `cc-icon` we may want to set an accessible name to it.

### LINKED ISSUE #877: cc-logs-instances: init component
## Context

The logs page needs a way to filter by instances: As a user, I want to see logs produced by this instance or this selection of instances.

Instances can be grouped in two ways:

* by some logical categories:
  * instances attached to the deployment in progress
  * running instances
  * instances that are in the process of being stopped
  * deleted instances
* by deployment

The first will be used for the 'live' mode. The latter is for the 'cold' mode.

In front of each instance, is a checkbox that allows a selection of instances.

## To be discussed

The name of the component can be discussed.

In the logs page,  we may want to allow users to filter logs on other data (like logs level, service ...). Should we have a single component that allows all those filters or should we keep this simple and allow only instance selection?

### Attributes/properties

| `state`    | `LogsInstancesState` | `{ state: 'loading' }`  | `false`    |         |

## Outputs

| `cc-logs-instances:selectionChange` | `Array<string>`    | Fires whenever the instances selection changes |

## Type Definitions

```typescript
export type LogsInstancesState = LogsInstancesStateLoading | LogsInstancesStateError | LogsInstancesStateLoaded;

export interface LogsInstancesStateLoading {
  state: "loading";
}

export interface LogsInstancesStateError {
  state: "error";
}

export interface LogsInstancesStateLoaded {
  state: "loaded";
  mode: "live" | "cold";
  instances: Array<Instance>;
  selection: Array<string>;
}

export type DeploymentState = "QUEUED" | "WIP" | "OK" | "CANCELLED" | "FAILED"

export interface Deployment {
  id: string;
  index: number;
  state: DeploymentState;
  creationDate: Date;
  instanceCount: number;
  commitId: string;
}


export type InstanceState = "BOOTING" | "STARTING" | "DEPLOYING" | "READY" | "UP" | "STOPPING" | "DELETED";
export type InstanceKind = "BUILD" | "RUN";

export interface Instance {
  id: string;
  name: string;
  index: number;
  deployment: Deployment;
  state: InstanceState;
  creationDate: Date;
  deletionDate: Date;
  flavor: string;
  kind: InstanceKind;
}
```

pdesoyres-cc:
Fixes #877 

Fixes #870

### DIFF
--- a/src/components/cc-logs-instances/cc-logs-instances.js
+++ b/src/components/cc-logs-instances/cc-logs-instances.js
@@ -1,0 +1,758 @@
+import '../cc-datetime-relative/cc-datetime-relative.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-notice/cc-notice.js';
+import '../cc-toggle/cc-toggle.js';
+import { css, html, LitElement } from 'lit';
+import {
+  iconRemixCompassDiscoverLine as iconHeaderDeploying,
+  iconRemixPlayCircleLine as iconHeaderRunning,
+  iconRemixStopCircleLine as iconHeaderStopping,
+  iconRemixCloseCircleLine as iconHeaderDeleted,
+  iconRemixGhostLine as iconHeaderGhost,
+  iconRemixCompassDiscoverLine as iconDeploymentWip,
+  iconRemixCheckboxCircleFill as iconDeploymentSucceeded,
+  iconRemixCloseCircleFill as iconDeploymentCancelled,
+  iconRemixErrorWarningFill as iconDeploymentFailed,
+  iconRemixCompassDiscoverLine as iconInstanceDeploying,
+  iconRemixPlayCircleLine as iconInstanceRunning,
+  iconRemixStopCircleLine as iconInstanceStopping,
+  iconRemixCloseCircleLine as iconInstanceDeleted,
+  iconRemixToolsLine as iconBuildInstance,
+} from '../../assets/cc-remix.icons.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
+import { i18n } from '../../lib/i18n.js';
+import { groupBy } from '../../lib/utils.js';
+
+/**
+ * @param {Instance} i1
+ * @param {Instance} i2
+ * @return {number}
+ */
+const INSTANCE_SORT_ORDER = (i1, i2) => {
+  const number = i1.index - i2.index;
+  if (number !== 0) {
+    return number;
+  }
+  const time = i1.creationDate.getTime() - i2.creationDate.getTime();
+  if (time !== 0) {
+    return time;
+  }
+  return i1.kind === 'BUILD' ? -1 : 1;
+};
+
+/**
+ * @param {GhostInstance} i1
+ * @param {GhostInstance} i2
+ * @return {number}
+ */
+const GHOST_INSTANCE_SORT_ORDER = (i1, i2) => {
+  return i1.id.localeCompare(i2.id);
+};
+
+/** @type {Array<InstanceState>} */
+const INSTANCE_DEPLOYING_STATES = ['BOOTING', 'STARTING', 'DEPLOYING', 'READY'];
+/** @type {Array<InstanceState>} */
+const INSTANCE_RUNNING_STATES = ['UP', 'MIGRATION_IN_PROGRESS', 'TASK_IN_PROGRESS', 'BUILDING'];
+/** @type {Array<DeploymentState>} */
+const DEPLOYMENT_WIP_STATES = ['QUEUED', 'WORK_IN_PROGRESS'];
+
+/**
+ * @typedef {import('./cc-logs-instances.types.js').LogsInstancesState} LogsInstancesState
+ */
+
+/**
+ * Allows users to select some instances.
+ *
+ * * When using `live` mode, instances are grouped by logical state.
+ * * When using `cold` mode, instances are grouped by deployment.
+ *
+ * @cssdisplay flex
+ *
+ * @event {CustomEvent<Array<string>>} cc-logs-instances:selection-change - Fires whenever the instances selection changes
+ */
+export class CcLogsInstances extends LitElement {
+
+  static get properties () {
+    return {
+      state: { type: Object },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {LogsInstancesState} State of the component. */
+    this.state = {
+      state: 'loading',
+    };
+  }
+
+  _isSelected (instanceId) {
+    return this.state.state === 'loaded' && this.state.selection != null && this.state.selection.includes(instanceId);
+  }
+
+  _onInstanceClick (e) {
+    const instanceId = e.target.id;
+    if (this._isSelected(instanceId)) {
+      this.state = {
+        ...this.state,
+        selection: this.state.selection.filter((i) => i !== instanceId),
+      };
+    }
+    else {
+      this.state = {
+        ...this.state,
+        selection: [...(this.state.selection ?? []), instanceId],
+      };
+    }
+
+    dispatchCustomEvent(this, 'selection-change', this.state.selection);
+  }
+
+  render () {
+    if (this.state.state === 'error') {
+      return html`
+        <div class="wrapper wrapper--center">
+          <cc-notice intent="warning" message="${i18n('cc-logs-instances.loading.error')}"></cc-notice>
+        </div>
+      `;
+    }
+
+    if (this.state.state === 'loading') {
+      return html`
+        <div class="wrapper wrapper--center">
+          <cc-loader a11y-name="${i18n('cc-logs-instances.loading.loader')}"></cc-loader>
+        </div>
+      `;
+    }
+
+    if (this.state.mode === 'cold') {
+      return this._renderColdMode(this.state.instances);
+    }
+
+    return this._renderLiveMode(this.state.instances);
+  }
+
+  /**
+   * @param {Array<Instance|GhostInstance>} instances
+   */
+  _renderColdMode (instances) {
+    const { ghost: ghostInstances, real: realInstances } = groupBy(instances, (instance) => instance.ghost ? 'ghost' : 'real');
+
+    return html`
+      <fieldset class="section section--cold">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.cold.header'),
+        })}
+        ${realInstances?.length > 0
+          ? this._renderInstancesGroupedByDeployment(instances, false)
+          : html`<div class="empty">${i18n('cc-logs-instances.cold.empty')}</div>`
+        }
+      </fieldset>
+      ${this._renderGhostInstances(ghostInstances)}
+    `;
+  }
+
+  /**
+   * @param {Array<Instance|GhostInstance>} instances
+   */
+  _renderLiveMode (instances) {
+    /** @type {Array<Instance>} */
+    const deployingInstances = [];
+    /** @type {Array<Instance>} */
+    const runningInstances = [];
+    /** @type {Array<Instance>} */
+    const stoppingInstances = [];
+    /** @type {Array<Instance>} */
+    const deletedInstances = [];
+
+    const { ghost: ghostInstances, real: realInstances } = groupBy(instances, (instance) => instance.ghost ? 'ghost' : 'real');
+
+    (realInstances || [])
+      .sort(INSTANCE_SORT_ORDER)
+      .forEach((instance) => {
+        if (DEPLOYMENT_WIP_STATES.includes(instance.deployment.state)) {
+          deployingInstances.push(instance);
+        }
+        else if (instance.state === 'DELETED') {
+          deletedInstances.push(instance);
+        }
+        else if (instance.state === 'STOPPING') {
+          stoppingInstances.push(instance);
+        }
+        else {
+          runningInstances.push(instance);
+        }
+      });
+
+    return html`
+      ${this._renderDeployingInstances(deployingInstances)}
+      ${this._renderRunningInstances(runningInstances)}
+      ${this._renderStoppingInstances(stoppingInstances)}
+      ${this._renderDeletedInstances(deletedInstances)}
+      ${this._renderGhostInstances(ghostInstances)}
+    `;
+  }
+
+  /**
+   * @param {Array<Instance>} instances
+   */
+  _renderDeployingInstances (instances) {
+    if (instances.length === 0) {
+      return '';
+    }
+
+    return html`
+      <fieldset class="section section--deploying">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.deploying.header'),
+          icon: iconHeaderDeploying,
+          commit: instances[0].deployment.commitId,
+        })}
+        <div class="section-content instances instances--with-state">
+          ${this._renderInstances(instances, true)}
+        </div>
+      </fieldset>
+    `;
+  }
+
+  /**
+   * @param {Array<Instance>} instances
+   */
+  _renderRunningInstances (instances) {
+    if (instances.length === 0) {
+      return html`
+        <fieldset class="section section--running">
+          ${this._renderHeader({
+            title: i18n('cc-logs-instances.running.header'),
+            icon: iconHeaderRunning,
+          })}
+          <div class="empty">
+            ${i18n('cc-logs-instances.running.empty')}
+          </div>
+        </fieldset>
+      `;
+    }
+
+    return html`
+      <fieldset class="section section--running">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.running.header'),
+          icon: iconHeaderRunning,
+          commit: instances[0].deployment.commitId,
+        })}
+        <div class="section-content instances">
+          ${this._renderInstances(instances, false)}
+        </div>
+      </fieldset>
+    `;
+  }
+
+  /**
+   * @param {Array<Instance>} instances
+   */
+  _renderStoppingInstances (instances) {
+    if (instances.length === 0) {
+      return '';
+    }
+
+    return html`
+      <fieldset class="section section--stopping">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.stopping.header'),
+          icon: iconHeaderStopping,
+          commit: instances[0].deployment.commitId,
+        })}
+        <div class="section-content instances">
+          ${this._renderInstances(instances, false)}
+        </div>
+      </fieldset>
+    `;
+  }
+
+  /**
+   * @param {Array<Instance>} instances
+   */
+  _renderDeletedInstances (instances) {
+    if (instances.length === 0) {
+      return '';
+    }
+
+    return html`
+      <fieldset class="section section--deleted">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.deleted.header'),
+          icon: iconHeaderDeleted,
+        })}
+        ${this._renderInstancesGroupedByDeployment(instances, false)}
+      </fieldset>
+    `;
+  }
+
+  /**
+   * @param {Array<GhostInstance>|null} instances
+   */
+  _renderGhostInstances (instances) {
+    if (instances == null || instances.length === 0) {
+      return '';
+    }
+
+    return html`
+      <fieldset class="section section--ghost">
+        ${this._renderHeader({
+          title: i18n('cc-logs-instances.ghost.header'),
+          icon: iconHeaderGhost,
+        })}
+
+        <div class="section-content">
+          <cc-notice intent="info" message="${i18n('cc-logs-instances.ghost.notice')}"></cc-notice>
+          <div class="instances instances--ghost">
+            ${instances.sort(GHOST_INSTANCE_SORT_ORDER).map((instance) => html`
+            <label class="instance" for="instance-${instance.id}">
+              <input type="checkbox"
+                     id="instance-${instance.id}"
+                     .checked=${this._isSelected(instance.id)}
+                     @change=${this._onInstanceClick}>
+              <span class="instance-id">${instance.id}</span>
+            </label>
+          `)}
+          </div>
+        </div>
+      </fieldset>
+    `;
+  }
+
+  _renderHeader ({ title, icon, commit }) {
+    return html`
+      <legend class="section-header">
+        ${icon != null ? html`
+          <cc-icon class="section-header-icon" .icon=${icon}></cc-icon>
+        ` : ''}
+        <span class="section-header-title">${title}</span>
+        ${commit != null ? this._renderCommit(commit) : ''}
+      </legend>
+    `;
+  }
+
+  /**
+   * @param {Array<Instance>} instances
+   * @param {boolean} renderState
+   */
+  _renderInstancesGroupedByDeployment (instances, renderState) {
+    const groups = groupBy(instances, (instance) => instance.deployment?.id);
+
+    return html`
+      <div class="section-content deployments">
+        ${Object.values(groups)
+          // deployment date is the date of the deployment of the first instance in the group
+          .map((instances) => ({ deploymentDate: instances[0].deployment.creationDate, instances }))
+          // sort by deployment date
+          .sort((o1, o2) => o1.deploymentDate - o2.deploymentDate)
+          // render group of instances
+          .map(({ instances }) => html`
+            <fieldset class="deployment">
+              ${this._renderDeploymentDetails(instances[0].deployment)}
+              <div class="instances">
+                ${this._renderInstances(instances, renderState)}
+              </div>
+            </fieldset>
+          `)}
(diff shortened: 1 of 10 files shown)


florian-sanders-cc (changes requested):
Gg @pdesoyres-cc!

I left a few comments, the most important ones are about the `cc-loader`, the rest is fairly trivial.

A few more global comments:

- For the `empty` state, maybe some italic text. Although we don't use italic in tiles so maybe we prefer without it? (I prefer italic but it's not very important)
- Semantically, I think we need to group each set of checkboxes within a `<fieldset>` with the `header` as a `<legend>`. We can discuss it in sync if my comment is not clear :wink: 
- The headings in `Deployments in progress` feel weird to me: "Deployed 1 minute ago"  with a loader on the left :thinking: Maybe we could go for "Deploying" but I guess this would be weird to say "1 min ago" right next to it. I don't know yet, I'll try to find a better suggestion.
- I have to say the red cross icons make me want to click them to remove the deleted instances from the UI. I would go for no icon with these but I'm not sure and it might very well just be me.

src/translations/translations.fr.js
```diff
@@ -469,6 +469,31 @@ export const translations = {
...
+  'cc-logs-instances.instance.state.booting': `Instance en cours de boot`,
+  'cc-logs-instances.instance.state.deleted': `Instance supprimée`,
+  'cc-logs-instances.instance.state.deploying': `Instance en cours de déploiement`,
+  'cc-logs-instances.instance.state.ready': `Instance prête`,
+  'cc-logs-instances.instance.state.starting': `Instance en cours de démarrage`,
+  'cc-logs-instances.instance.state.stopping': `Instance en cours d'arrêt`,
+  'cc-logs-instances.instance.state.up': `Instance démarrée`,
+  'cc-logs-instances.loading.error': `Une erreur est survenue pendant le chargement des instances.`,
+  'cc-logs-instances.loading.loader': `Instances en cours de chargement`,
+  'cc-logs-instances.running.empty': `Aucun instance démarrée`,
```
HeleneAmouzou, line 493:
```suggestion
  'cc-logs-instances.running.empty': `Aucune instance démarrée`,
```

HeleneAmouzou (approved):
I just have a small typo suggestion, but otherwise GG ! 👏

About the comments @florian-sanders-cc made :
I agree on the `Deployments in progress` part, it also feels weird to me to have a loader next to "Deployed X minutes ago", but I don't have an alternative to suggest...

src/components/cc-logs-instances/cc-logs-instances.types.d.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface LogsInstancesStateLoaded {
+  state: "loaded";
+  mode: "live" | "cold";
+  instances: Array<Instance>;
+  selection: Array<string>;
+}
+
+export type DeploymentState = "QUEUED" | "WIP" | "OK" | "CANCELLED" | "FAILED"
```
Galimede, line 19: nitpick: semicolon is missing at the end of line.

src/components/cc-logs-instances/cc-logs-instances.types.d.ts
```diff
@@ -0,0 +1,44 @@
...
+  state: "loading";
+}
+
+export interface LogsInstancesStateError {
+  state: "error";
+}
+
+export interface LogsInstancesStateLoaded {
+  state: "loaded";
+  mode: "live" | "cold";
```
Galimede, line 14: question: In the rest of the file, the unions are in a dedicated type, why not here?

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,567 @@
...
+
+/**
+ * Allows users to select some instances.
+ *
+ * * When using `live` mode, instances are grouped by logical state.
+ * * When using `cold` mode, instances are grouped by deployment.
+ *
+ * @cssdisplay flex
+ *
+ * @event {CustomEvent} cc-logs-instances:selectionChange - Fires whenever the instances selection changes
```
Galimede, line 52: suggestion: It seems that the selection is an `Array<string>` maybe we could change `CustomEvent` to `CustomEvent<Array<string>>` or `CustomEvent<string[]>`, WDYT?

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,567 @@
...
+  static get properties () {
+    return {
+      state: { type: Object },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {LogsInstancesState} */
```
Galimede, line 65: nitpick: on other components we provide a description next to the type. For the state we tend to do something like `Sets the state of...` or `State of the ...`, could we do that here too?

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,567 @@
...
+
+    const accessibleName = getAccessibleName(state);
+
+    if (state === 'BOOTING' || state === 'STARTING' || state === 'DEPLOYING') {
+      return html`
+        <cc-loader accessible-name="${accessibleName}" 
+                   class="instance_state"></cc-loader>`;
+    }
+
+    if (state === 'READY') {
```
Galimede, line 400: thought: maybe we could have avoided these cascade ifs. 
If we:
* had a function to retrieve the icon with the state given
* Either have the state lowercased to put dinamycally the class.
* Or have the class name in the CSS uppercased.

Anyway, that's just a thought, maybe I'm wrong and there's something I am not aware of, don't worry about it. :wink:

example:

```js
return html`
  <!-- Icon could have been retrieved with a dedicated function -->
  <cc-icon .icon=${getIconState(state)}
                 accessible-name="${accessibleName}"
                 class="instance_state instance_state--${state.toLowerCase()}"></cc-icon>`;
`;
```

or if the class was in uppercase

```js

return html`
  <!-- Icon could have been retrieved with a dedicated function -->
  <cc-icon .icon=${getIconState(state)}
                 accessible-name="${accessibleName}"
                 class="instance_state instance_state--${state}"></cc-icon>`;
`;
```

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,324 @@
+import './cc-logs-instances.js';
+import { makeStory } from '../../stories/lib/make-story.js';
+import { enhanceStoriesNames } from '../../stories/lib/story-names.js';
+
+/**
+ * @typedef {import('./cc-logs-instances.types.d.ts').Instance} Instance
```
Galimede, line 6: question: just curious does that help the IDE doing that here?

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,324 @@
...
+ * @typedef {import('./cc-logs-instances.types.d.ts').Deployment} Deployment
+ */
+
+function randomCommit () {
```
Galimede, line 10: question: `cc-header-app` uses a hard coded String in its story file, maybe we could put this a lib?

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,324 @@
...
+  return randomString(40, 'abcdef0123456789');
+}
+function randomString (length = 8, alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz') {
```
Galimede, line 13: nitpick: could you break a line between the two functions?
question: I've had to use a similar function and I could use it for `cc-product-card` instead of what I did, could we put it in a lib file too?

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,324 @@
...
+    commitId: randomCommit(),
+    index: 1,
+    instanceCount: 3,
+  },
+};
+
+/**
+ * @type {Array<Instance>}
+ */
+const instances = [
```
Galimede, line 107: nitpick: if its a constant maybe we could uppercase the var?

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,324 @@
...
+
+const conf = {
+  component: 'cc-logs-instances',
+};
+
+/**
+ *
+ * @type {{[key: string]: Deployment}}
+ */
+const deployments = {
```
Galimede, line 45: nitpick: if its a constant maybe we could uppercase the var?

Galimede (changes requested):
Nice work as usual Pierre! :muscle: 

I've just left a few questions/nitpicks.

I do agree on what has been said above too.

Also, I don't know if i'm the only one, but I find the spacing between the checkboxes and the "header" or the "deployment title" a bit weird.

src/translations/translations.fr.js
```diff
@@ -469,6 +469,31 @@ export const translations = {
...
+  'cc-logs-instances.cold.header': `Instances`,
+  'cc-logs-instances.commit.accessible-name': `Commit Git`,
+  'cc-logs-instances.deleted.header': `Instances supprimées`,
+  'cc-logs-instances.deploying.header': `Déploiement en cours`,
+  'cc-logs-instances.deployment.deployed': `Déployée`,
+  'cc-logs-instances.deployment.state.cancelled': `Déploiement annulé`,
+  'cc-logs-instances.deployment.state.failed': `Déploiements en échec`,
+  'cc-logs-instances.deployment.state.ok': `Déploiement réussi`,
+  'cc-logs-instances.deployment.state.wip': `Déploiement en cours`,
+  'cc-logs-instances.instance.state.booting': `Instance en cours de boot`,
```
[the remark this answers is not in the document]
pdesoyres-cc: good question. I'm not sure of the response but I would say:

pdesoyres-cc:
@Galimede 

> Also, I don't know if i'm the only one, but I find the spacing between the checkboxes and the "header" or the "deployment title" a bit weird.

What do you mean by "weird"?

src/components/cc-logs-instances/cc-logs-instances.stories.js
pdesoyres-cc: yep

src/components/cc-logs-instances/cc-logs-instances.stories.js
pdesoyres-cc: I'll do that

Galimede:
> @Galimede
> 
> > Also, I don't know if I'm the only one, but I find the spacing between the checkboxes and the "header" or the "deployment title" a bit weird.
> 
> What do you mean by "weird"?

Sorry I should have been more explicit.
I find the alignment weird for the checkboxes of the instances relative by the content above them `header` or `Deployed x ago` 


I've made an example below of what I the alignment could have been in my mind.
I've tried to align the checkboxes with the `deployed 3 minutes ago` for the `deployment in progress part`. For the others I have tried to align them with the title of the header `running instances` and `stopping instances` respectively.
Though doing that makes two alignments, so I'm not sure that's a solution.  

What I mean overall is that I would have seen them more aligned as you would have bullets points aligned in a list like this:
* deployment in progress
  * deployed 3 minutes ago
    * instance 1
    * instance 2
    * instance 3
* running instances
  * instance 1
  * instance 2
* stopping instances  
  * instance 1

image


I don't know if it's clearer for you, though I understand I might go too far here and that might be hard to make so don't worry too much about it. :hugs:

florian-sanders-cc:
First off, the redesign looks really good!

The info is easier to read in this version.
The previous version was okay too but when I compare them, I realize that this one is cleaner and clearer so well done!

Some things I didn't notice during my first review:

- As discussed before, I find the index info confusing, I would remove it,
- I don't really understand the `ready` vs `running` state but the icons do convey the right meaning.

Apart from that, LGTM!

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -182,11 +182,17 @@ export class CcLogsInstances extends LitElement {
     }
 
     return html`
-      ${this._renderHeader(i18n('cc-logs-instances.deploying.header'), iconHeaderDeploying)}
-      ${this._renderDeploymentDetails(instances[0].deployment)}
-      <div class="instances">
+      <div class="section section--deploying" role="group" aria-labelledby="deploying-instances">
```
florian-sanders-cc, line 185: **fix:** unless we have a good reason not to, we should use `<fieldset>` + `<legend>` instead of ARIA.

Did you face issues with those?

florian-sanders-cc (changes requested):
Forgot to check the code sorry.

I think we should stick to `<fieldset>` + `<legend>` as much as we can because the support for the ARIA technique does have very small drawbacks (see [ADR 25 - Screen reader tests - VoiceOver iOS](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%93%8C-architecture-decision-records-adr-0025-moving-away-from-fieldset-and-legend--page)).

src/components/cc-logs-instances/cc-logs-instances.js
pdesoyres-cc: I didn't even try.
I find it much easier to use aria that `<fieldset>` + `<legend>` for which I would need to reset the default CSS.
Also, we recently changed to aria in the `cc-toggle`, so I thought it would be nice to uniformize in other components too.

If you think I'm wrong, I'll change that.

florian-sanders-cc: I definitely get that and I agree that it's easier to use this technique but since the support on iOS is slightly better with `<fieldset>` + `<legend>`, I think we should stick with that as much as we can.

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,653 @@
...
+               id="${instance.id}"
+               .checked=${this._isSelected(instance.id)}
+               @change=${this._onInstanceClick}>
+          ${renderState ? this._renderInstanceState(instance) : ''}
+          <span class="instance-index">[${instance.index}]</span>
+          <span class="instance-name">${instance.name}</span>
+          ${instance.kind === 'BUILD' ? html`
+            <cc-icon .icon=${iconBuildInstance}
+                     a11y-name="${i18n('cc-logs-instances.build.a11y-name')}"
+                     class="instance state"></cc-icon>
```
roberttran-cc, line 340: **🔨 issue:** `<cc-icon>` has an `instance` CSS class, as well as the wrapping `<label>`. There's no effect right now but I think it's not intended and issues could arise later. I think it's safe to simply remove it.

```suggestion
            <cc-icon .icon=${iconBuildInstance}
                     a11y-name="${i18n('cc-logs-instances.build.a11y-name')}"
                     class="state"></cc-icon>
```

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -0,0 +1,328 @@
...
+  createInstance(DEPLOYMENTS.wip, 1, 'RUN', 'READY', 'Rich medicham'),
+  createInstance(DEPLOYMENTS.ok1, 0, 'RUN', 'UP', 'Easy hydreigon'),
+  createInstance(DEPLOYMENTS.ok1, 1, 'RUN', 'UP', 'Long delphox'),
+  createInstance(DEPLOYMENTS.ok1, 2, 'RUN', 'STOPPING', 'Curly honchkrow'),
+  createInstance(DEPLOYMENTS.ok1, 0, 'BUILD', 'DELETED', 'Rich medicham'),
+  ...FAILED_DEPLOYMENT,
+  ...CANCELLED_DEPLOYMENT,
+];
+
+export const defaultStory = makeStory(conf, {
```
roberttran-cc, line 138: **🎯 suggestion:** We may need  an instance in ghost mode, to test proper rendering.

roberttran-cc (changes requested):
Great job! I'm looking forward the whole log UI put together. :star_struck: 

I only have a few suggestions and feedback.

Ideally, we should split #870 in a separate PR but I guess it's not that big of a deal, this PR is still perfectly readable.

test/utils.test.js
```diff
@@ -50,6 +50,20 @@ describe('clampNumber function', () => {
   });
 });
 
+describe('randomString function', () => {
+  it('should be empty if length is 0', () => {
+    expect(randomString(0)).to.equal('');
+  });
+
+  it('should have a size equal to the given length', () => {
+    expect(randomString(10).length).to.equal(10);
```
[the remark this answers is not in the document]
pdesoyres-cc: Not sure it has better coverage: it will check only for one size.

src/components/cc-logs-instances/cc-logs-instances.stories.js
pdesoyres-cc: You're right, but after thinking about it a bit more, I think we should discuss it with others:
- In which section should ghost instances be placed? (running or deleted)
- Do we really want to show this information to the user?

Galimede (commented):
Nice work, I really like the redesign :star_struck:!

## UI

There seems to be a problem with the text and the header when no instances are running as you can see in the screenshot below. I thought at first that it was only a Firefox problem but it does the same on Chrome.
image

Mathieu from the future here, I've checked the `first deployment` story and it also seems weird :thinking: 
image


While I think, the `cc-notice` here doens't feel right in the center
image
I do think that for the loader it would be better if it was placed right in the center
image

I've also seen that for these, you use a wrapper, they don't seem to use the whole space available, is there a reason?

I find this icon (see below) very confusing, what does it represent? :thinking: 
image

pdesoyres-cc:
@Galimede 

* I fixed the empty state problem (this was due to the horrible <fieldset> + <legend> thing)
* Both loader and error notice are vertically aligned at the top. (I think I prefer it to be at the top, but I would be ok to put it at the center too).
* Loader is horizontally aligned at the center.
* The wrapper takes the whole width and defines margins for the notice.

roberttran-cc (approved):

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+};
+
+/**
+ * @param {GhostInstance} i1
+ * @param {GhostInstance} i2
+ * @return {number}
+ */
+const GHOST_INSTANCE_SORT_ORDER = (i1, i2) => {
+  return i1.id.localeCompare(i2.id);
+};
```
hsablonniere, line 52: praise: nice way to move the sorter functions :+1:

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+          <cc-loader a11y-name="${i18n('cc-logs-instances.loading.loader')}"></cc-loader>
+        </div>
+      `;
+    }
+
+    if (this.state.mode === 'cold') {
+      return html`
+        <fieldset class="section section--cold" role="group" aria-labelledby="cold-logs">
+          ${this._renderColdMode(this.state.instances)}
+        </fieldset>
```
hsablonniere, line 135: question: why is the fieldset not in the subrender?

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+          .map(([_, instances]) => {
+            return html`
+            <fieldset class="deployment">
+              ${this._renderDeploymentDetails(instances[0].deployment)}
+              <div class="instances">
+                ${this._renderInstances(instances, renderState)}
+              </div>
+            </fieldset>
+          `;
+          })}
```
hsablonniere, line 362: suggestion: in this kind of template, removing the arrow body simplifies the code (IMO)

```suggestion
          .map(([_, instances]) => html`
            <fieldset class="deployment">
              ${this._renderDeploymentDetails(instances[0].deployment)}
              <div class="instances">
                ${this._renderInstances(instances, renderState)}
              </div>
            </fieldset>
          `)}
```

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+   * @param {Array<Instance>} instances
+   * @param {boolean} renderState
+   */
+  _renderInstancesGroupedByDeployment (instances, renderState) {
+    const groups = groupBy(instances, (instance) => instance.deployment?.id);
+
+    return html`
+      <div class="section-content deployments">
+        ${Object.entries(groups)
+          .sort(([, instances1], [, instances2]) => instances2[0].deployment.creationDate - instances1[0].deployment.creationDate)
```
hsablonniere, line 352: nitpick: this one with the comas in square brackets it not easy to read

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+               a11y-name="${icon.a11yName}"
+               class="instance-state ${icon.class}"></cc-icon>
+    `;
+  }
+
+  _renderCommit (commit) {
+    return html`
+      <span class="commit">
+        ${commit.substring(0, 7)}
+      </span>`;
```
hsablonniere, line 521: nitpick

```suggestion
    return html`
      <span class="commit">
        ${commit.substring(0, 7)}
      </span>
    `;
```

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+    return html`
+      <span class="instance-index">
+        ${instance.kind === 'BUILD' ? html`
+            <cc-icon .icon=${iconBuildInstance}
+                     a11y-name="${i18n('cc-logs-instances.build.a11y-name')}"></cc-icon>
+          ` : ''}
+        ${instance.kind !== 'BUILD' ? html`
+          ${instance.index}
+          ` : ''}
+      </span>`;
```
hsablonniere, line 534: nitpick:

```suggestion
    return html`
      <span class="instance-index">
        ${instance.kind === 'BUILD' ? html`
          <cc-icon .icon=${iconBuildInstance}
                         a11y-name="${i18n('cc-logs-instances.build.a11y-name')}"></cc-icon>
          ` : ''}
        ${instance.kind !== 'BUILD' ? instance.index : ''}
      </span>
    `;
```

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+   * @param {Array<Instance>} instances
+   * @param {boolean} renderState
+   */
+  _renderInstances (instances, renderState) {
+    return html`
+      ${
+        instances.sort(INSTANCE_SORT_ORDER)
+          .map((instance) => this._renderInstance(instance, renderState))
+      }
+    `;
```
hsablonniere, line 377: nitpick:

```suggestion
    return instances
      .sort(INSTANCE_SORT_ORDER)
      .map((instance) => this._renderInstance(instance, renderState));
```

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+          <cc-notice intent="info" message="${i18n('cc-logs-instances.ghost.notice')}"></cc-notice>
+          <div class="instances instances--ghost">
+            ${instances.sort(GHOST_INSTANCE_SORT_ORDER).map((instance) => html`
+            <label class="instance" for="${instance.id}">
+              <input type="checkbox"
+                     id="${instance.id}"
+                     .checked=${this._isSelected(instance.id)}
+                     @change=${this._onInstanceClick}>
+              <span class="instance-id">${instance.id}</span>
+            </label>
```
hsablonniere, line 322: question: to we really need the `for` and `id` if we wrap the input inside the label?

question: should we prefix the raw instance ID with `instance-`?

src/components/cc-logs-instances/cc-logs-instances.js
```diff
@@ -0,0 +1,755 @@
...
+   */
+  _renderInstance (instance, renderState) {
+    return html`
+      <label class="instance" for="${instance.id}">
+        <input type="checkbox"
+               id="${instance.id}"
+               .checked=${this._isSelected(instance.id)}
+               @change=${this._onInstanceClick}>
+        <span class="instance-name">${instance.name}</span>
+      </label>
```
hsablonniere, line 393: question: to we really need the `for` and `id` if we wrap the input inside the label?

question: should we prefix the raw instance ID with `instance-`?

hsablonniere (approved):
Nice work @pdesoyres-cc :+1: 

I like what you've done with the icons. I think they're too small but it's a good iteration, we can move forward with this and adjust later.

I added a few suggestions/nitpicks/questions but nothing blocking. You can apply them (or not) but you can move forward with this :+1: 

praise: this was not an easy component to organize and the different subrenders help a lot!

PR #882, opened 2023-11-09
cc-header-orga: tweak the design to add a slot

florian-sanders-cc:


src/components/cc-header-orga/cc-header-orga.js
```diff
@@ -3,185 +3,215 @@ import '../cc-img/cc-img.js';
...
-        .hotline_number:active {
-          box-shadow: none;
+          flex: 0 0 auto;
         }
 
-        .hotline_number cc-badge {
-          /* Prevent space below badge because of text lines */
-          display: flex;
-          text-decoration: underline;
+        ::slotted([slot='footer']) {
```
florian-sanders-cc, line 211: **Note:** I felt clever at the time but I'm not sure it's a good idea :laughing: 

We want the footer to have a style (grey background, same padding as the body for the content, a subtle top border).
But we only want this "footer" to be displayed when there is a slotted content.
If we set something like: 
```
<div class="footer">
  <slot name="footer"></slot>
</div>
```

The `div.footer` is always in the DOM and its style (border, padding, background) is visible even when there is no slotted content.
We could listen for slot changes to toggle a `hidden` attribute on the `div.footer` but I wanted to see if we could avoid this.

The current solution does work but it has some drawbacks: if, from outside this component, you define styles on the `<div slot="footer">`, they will override the styles defined here. It makes overriding the styles here very simple but maybe it's actually too easy and it also means you may override these by mistake :thinking:.

src/components/cc-header-orga/cc-header-orga.types.d.ts
```diff
@@ -0,0 +1,17 @@
+export type HeaderOrgaState = HeaderOrgaStateLoaded | HeaderOrgaStateLoading | HeaderOrgaStateError;
+
+interface HeaderOrgaStateLoaded {
+  state: 'loaded',
```
pdesoyres-cc, line 4: nitpick: should end with a semi-colon

src/components/cc-header-orga/cc-header-orga.types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  name: string;
+  avatar?: string;
+  cleverEnterprise?: boolean;
+  emergencyNumber?: number; 
+}
+
+interface HeaderOrgaStateLoading {
+  state: 'loading',
```
pdesoyres-cc, line 12: nitpick: should end with a semi-colon

src/components/cc-header-orga/cc-header-orga.types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+interface HeaderOrgaStateError {
+  state: 'error',
```
pdesoyres-cc, line 16: nitpick: should end with a semi-colon

src/components/cc-header-orga/cc-header-orga.js
pdesoyres-cc: I'm ok with this drawback as soon as it is explained in the @slot JSDoc section

src/components/cc-header-orga/cc-header-orga.stories.js
```diff
@@ -2,54 +2,158 @@ import './cc-header-orga.js';
...
+  `,
+  items: [{
+    orga: DEFAULT_ORGA,
+    innerHTML: `
+      <div slot="footer">
+        <p>The slotted content container has a default styling: <code>background-color</code>, <code>padding</code> and <code>border-top</code>.</p>
+        <p>You may change any of these by styling the tag on which you have added the <code>slot="footer"</code> attribute.</p>
+        <p><strong>Caution:</strong> the default styling was designed to remain consistent with the header body and avoid drawing too much attention to it. You should stick to the default styling as much as possible and stay as close as possible to it. This example is only here to demo that anything can be customized.
+      </div>
+    `,
```
pdesoyres-cc, line 139: excellent explanation

pdesoyres-cc (approved):
I've not much to say so well done Florian!

hsablonniere (approved):
Hey @florian-sanders-cc, good job on this :+1:

Small nitpick, the change on `src/components/cc-header-orga/cc-header-orga.stories.js` of `loaded` to `error` should be in the first commit but who cares.

Galimede:
LGTM, GJ @florian-sanders-cc! :muscle: :+1:

Galimede (approved):

HeleneAmouzou (approved):
LGTM ! :tada:

PR #890, opened 2023-11-22
Add label style customization for cc-input-text & similar components

### LINKED ISSUE #888: Open `<label>` style customization in cc-input-text & similar components
## Proposal

Adding component-level CSS custom properties for the following rules:
- `color`
- `font-size`
- `font-weight`

Potential caveats: `padding/margin` and `height` properties.

roberttran-cc:
Fixes #888.

# What to review

Is the feature properly working & rendered on screen.
Does the feature introduces visual regression (excluding "Visual breaking changes").

# How to review

Updated components:
- `cc-input-text`
- `cc-input-number`
- `cc-input-date`
- `cc-select`
- `cc-toggle`

UI review:
- for the introduced feature, a dedicated story "Custom label style" has been added, displaying relevant cases ;
- for regression check, compare stories between the preview & the current release.

Code review: there is one commit per component so it's quite straightforward.

<details>
<summary>(previous topic on "visual breaking changes")</summary>
# Visual breaking changes

Actually, this feature needed a change in the component global layout, impacting many cases but especially when label is displayed inline.

Thus introducing potential "visual breaking changes" (to be discussed).

## The root cause

In inline mode, when the label part (including the "required" text) is big enough, it introduces a gap between the input and the content below (either the help or the error text).

image

## The fix

The solution was to change the grid `grid-template-areas` property from...

```css
  grid-template-areas: 
    'label input'
    '. help'
    '. error';
```
...to
```css
  grid-template-areas: 
    'label input'
    'label help'
    'label error';
```

## The impacts

But to ensure a relatively homogeneous layout in all cases, the `grid` layout in inline mode is now based on the following rule:

```css
:host([inline]) {
  align-items: baseline;
}
```

Causing a slight difference in the "Inline" and "Inline (error and help messages)" stories:

[inline.webm](https://github.com/CleverCloud/clever-components/assets/102976382/f2b0c452-3db3-463f-8f65-04b467e26b2e)

[inline-with-error-and-help-messages.webm](https://github.com/CleverCloud/clever-components/assets/102976382/1584411e-d071-4358-ad12-9d337e78a36d)

But causing a significant difference in the "Inline (required)" story, the overall component taking ultimately more height:

[inline-with-required.webm](https://github.com/CleverCloud/clever-components/assets/102976382/b6857761-8b3f-418f-84df-072e1065cfca)

> [!IMPORTANT]  
> WDYT?

# What's left to do

Maybe consider adding a custom property for the label color.

If the feature is validated, the same feature should also be implemented in the following components:

- [x] cc-input-date
- [x] cc-input-number
- [x] cc-select
- [x] cc-toggle

Also test on each browser.
</details>

### DIFF
--- a/src/components/cc-input-number/cc-input-number.stories.js
+++ b/src/components/cc-input-number/cc-input-number.stories.js
@@ -172,6 +172,63 @@ export const customWidth = makeStory(conf, {
     }),
 });
 
+const customBaseItems = [
+  { label: 'The label', value: 100 },
+  { label: 'The label', value: 100, required: true },
+  { label: 'The label', value: 100, controls: true },
+  { label: 'The label', value: 100, required: true, controls: true },
+];
+
+export const customLabelStyle = makeStory({ ...conf, displayMode: 'block' }, {
+  // language=CSS
+  css: `
+    cc-input-number {
+      --cc-input-label-color: #475569;
+      --cc-input-label-font-size: 1.2em;
+      --cc-input-label-font-weight: bold;
+      font-size: 1.25em;
+      max-width: 32em;
+    }
+    cc-input-number:nth-of-type(${customBaseItems.length + 'n'}) {
+      margin-block-end: 2em;
+    }
+  `,
+  items: [
+    ...customBaseItems,
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="help">Must be an integer</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="help">Must be an integer</p><p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="help">Must be an integer</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="help">Must be an integer</p><p slot="error">You must enter a value</p>`,
+    })),
+  ],
+});
+
 export const allFormControls = allFormControlsStory;
 
 export const simulation = makeStory(conf, {
--- a/src/components/cc-input-text/cc-input-text.stories.js
+++ b/src/components/cc-input-text/cc-input-text.stories.js
@@ -226,6 +226,68 @@ export const tagsWithClipboard = makeStory(conf, {
   items: tagsItems.map((p) => ({ ...p, clipboard: true })),
 });
 
+const customBaseItems = [
+  { label: 'The label', placeholder: 'No value yet...' },
+  { label: 'The label', placeholder: 'No value yet...', required: true },
+  { label: 'The label', placeholder: 'No value yet...', multi: true },
+  { label: 'The label', placeholder: 'No value yet...', multi: true, required: true },
+  { label: 'The label', placeholder: 'No value yet...', multi: true, value: 'Simple value\nOther lines...' },
+  { label: 'The label', placeholder: 'No value yet...', multi: true, required: true, value: 'Simple value\nOther lines...' },
+];
+
+export const customLabelStyle = makeStory(conf, {
+  // language=CSS
+  css: `
+    cc-input-text {
+      --cc-input-label-color: #475569;
+      --cc-input-label-font-size: 1.2em;
+      --cc-input-label-font-weight: bold;
+      font-size: 1.25em;
+    }
+    cc-input-text:nth-of-type(${customBaseItems.length + 'n'}) {
+      margin-block-end: 2em;
+    }
+    cc-input-text[multi] {
+      /* Necessary because the container is a display:flex */
+      width: 100%;
+    }
+  `,
+  items: [
+    ...customBaseItems,
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="help">Must be at least 7 characters long</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      innerHTML: `<p slot="help">Must be at least 7 characters long</p><p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="help">Must be at least 7 characters long</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="error">You must enter a value</p>`,
+    })),
+    ...customBaseItems.map((item) => ({
+      ...item,
+      inline: true,
+      innerHTML: `<p slot="help">Must be at least 7 characters long</p><p slot="error">You must enter a value</p>`,
+    })),
+  ],
+});
+
 export const allFormControls = allFormControlsStory;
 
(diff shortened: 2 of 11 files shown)


pdesoyres-cc:
In fact, I prefer this new alignment when in inline mode.

pdesoyres-cc (approved):
LGTM! Well done Bob.

florian-sanders-cc:
@roberttran-cc have you considered going for size props instead of (or in addition to) custom properties?

I feel we could provide more guidelines and more design consistency by designing 2 or 3 sizes to choose from :thinking:.

I've seen some design systems / component libraries do that and the selected size both increases the label and the input (+ its content).
Example with Shoelace:

Shoelace example


This would also make the alignment issue a little bit better for `inline` as we would make the input grow with the label.

About the alignment changes, I prefer the "old" alignment but `inline` mode is tricky, discouraged, especially with `required`. It works best outside forms (stuff that doesn't need to be submitted like UI filters, etc.), or in a form with 1 field.

When there is only one field, we don't necessarily need to add the `required` visually so in the end `inline` + `required` is very unlikely to happen.

To sum it up: I'm ok with the new alignment because I think the `inline + required` mode should not block us.

roberttran-cc:
@florian-sanders-cc 

> I feel we could provide more guidelines and more design consistency by designing 2 or 3 sizes to choose from 🤔.

I think that could be a great idea (did not consider this option at all) but I'm not sure if we can introduce this notion within this PR, mainly because we don't have typography design tokens yet. But I think it's a subject worth talking about.

> as we would make the input grow with the label.

Actually, it's possible by changing the `font-size` on the component. The CSS custom properties allow to separate the label size from the input.

> When there is only one field, we don't necessarily need to add the required visually so in the end inline + required is very unlikely to happen.

Did not think of that! Thanks for this insight.

HeleneAmouzou (approved):
LGTM, I like the new alignment ! :clap:

hsablonniere (approved):
Nice work @roberttran-cc :clap: 

I prefer the new alignement, it works well.

About @florian-sanders-cc suggestion on size prop vs CSS custom prop. I'll let you both figure out the right decision but I'm on the "let's go step by step" side of things. As you explained it, we don't have enough tokens yet (fonts, sizes, weights...) and maybe not enough hindsights on the big picture.

Galimede (approved):
+1 I like the new alignment too!

src/components/cc-select/cc-select.js
```diff
@@ -20,6 +20,10 @@ import { i18n } from '../../lib/i18n.js';
  *
  * @fires {CustomEvent<string>} cc-select:input - Fires the `value` whenever the `value` changes.
  *
+ * @cssprop {Color} --cc-input-label-color - The color for the input's label (defaults: `inherit`).
+ * @cssprop {FontSize} --cc-input-label-font-size - The font-size for the input's label (defaults: `inherit`).
+ * @cssprop {FontWeight} --cc-input-label-font-weight - The font-weight for the input's label (defaults: `normal`).
```
florian-sanders-cc, line 25: **nitpick:**
```suggestion
 * @cssprop {Color} --cc-input-label-color - The color for the select's label (defaults: `inherit`).
 * @cssprop {FontSize} --cc-input-label-font-size - The font-size for the select's label (defaults: `inherit`).
 * @cssprop {FontWeight} --cc-input-label-font-weight - The font-weight for the select's label (defaults: `normal`).
```

src/components/cc-select/cc-select.js
```diff
@@ -273,6 +284,7 @@ export class CcSelect extends LitElement {
           background-color: var(--cc-color-bg-primary, #000);
           clip-path: polygon(100% 0%, 0 0%, 50% 100%);
           content: '';
+          pointer-events: none;
```
florian-sanders-cc, line 287: **praise:** nice catch & fix :sunglasses:

src/components/cc-input-number/cc-input-number.stories.js
```diff
@@ -172,6 +172,61 @@ export const customWidth = makeStory(conf, {
...
+  css: `cc-input-number {
+    --cc-input-label-color: #475569;
+    --cc-input-label-font-size: 1.2em;
+    --cc-input-label-font-weight: bold;
+    font-size: 1.25em;
+    max-width: 32em;
+  }
+  cc-input-number:nth-of-type(${customBaseItems.length + 'n'}) {
+    margin-block-end: 2em;
+  }`,
```
florian-sanders-cc, line 193: **nitpick:**
```suggestion
  css: `
    cc-input-number {
      --cc-input-label-color: #475569;
      --cc-input-label-font-size: 1.2em;
      --cc-input-label-font-weight: bold;
      font-size: 1.25em;
      max-width: 32em;
    }
    cc-input-number:nth-of-type(${customBaseItems.length + 'n'}) {
      margin-block-end: 2em;
    }
  `,
```

src/components/cc-input-text/cc-input-text.stories.js
```diff
@@ -226,6 +226,66 @@ export const tagsWithClipboard = makeStory(conf, {
...
+    --cc-input-label-font-weight: bold;
+    font-size: 1.25em;
+  }
+  cc-input-text:nth-of-type(${customBaseItems.length + 'n'}) {
+    margin-block-end: 2em;
+  }
+  cc-input-text[multi] {
+    /* Necessary because the container is a display:flex */
+    width: 100%;
+  }`,
```
florian-sanders-cc, line 252: **nitpick:**
```suggestion
  css: `
  cc-input-text {
    --cc-input-label-color: #475569;
    --cc-input-label-font-size: 1.2em;
    --cc-input-label-font-weight: bold;
    font-size: 1.25em;
  }
  cc-input-text:nth-of-type(${customBaseItems.length + 'n'}) {
    margin-block-end: 2em;
  }
  cc-input-text[multi] {
    /* Necessary because the container is a display:flex */
    width: 100%;
  }
`,
```

src/components/cc-toggle/cc-toggle.js
```diff
@@ -38,6 +38,9 @@ import { dispatchCustomEvent } from '../../lib/events.js';
  * @cssprop {FontWeight} --cc-toggle-font-weight - Sets the value of the font weight CSS property (defaults: `bold`).
  * @cssprop {Filter} --cc-toggle-img-filter - A CSS filter to apply on images of all choices (defaults: `none`). It must be defined directly on the element.
  * @cssprop {Filter} --cc-toggle-img-filter-selected - A CSS filter to apply on images of selected choices (defaults: `none`). It must be defined directly on the element.
+ * @cssprop {Color} --cc-input-label-color - The color for the input's label (defaults: `inherit`).
+ * @cssprop {FontSize} --cc-input-label-font-size - The font-size for the input's label (defaults: `inherit`).
+ * @cssprop {FontWeight} --cc-input-label-font-weight - The font-weight for the input's label (defaults: `normal`).
```
florian-sanders-cc, line 43: **nitpick/discuss:**
```suggestion
 * @cssprop {Color} --cc-input-label-color - The color for the group's legend (defaults: `inherit`).
 * @cssprop {FontSize} --cc-input-label-font-size - The font-size for the group's legend (defaults: `inherit`).
 * @cssprop {FontWeight} --cc-input-label-font-weight - The font-weight for the group's legend (defaults: `normal`).
```

florian-sanders-cc (commented):
Gg @roberttran-cc for this PR + also fixing the select little issue which I didn't even know existed with the arrow icon :eyes: 

My feedback is mostly nitpick but some of them are related to a non trivial subject for which I don't have answers: how do we name these custom properties?

Until now, custom props on `cc-input` components were prefixed with `cc-input-` since they were shared only by the `cc-input-` components.
The ones we're adding here are also used within `cc-select` and `cc-toggle`. 
It feels like the `cc-input` namespace does not quite work for these. We could introduce something like `cc-form-fields-...`? I'm not sure :thinking: 

Also, it's a nitpick but for `cc-toggle` these custom props actually impact the `legend` instead of the `labels` of the items. 
I do get why we'd want to stick with a simple common custom prop that would work for all of our form fields but I'm not sure it's the right move here. (I'm not sure does not mean I don't agree, there are advantages & drawbacks to this choice :thinking:)
We could go for `cc-toggle-legend-...` for the `cc-toggle`.

This seems trivial / nitpicky but personally I wouldn't expect `cc-input-label-font-size` to override the legend font size of `cc-toggle`'s inner legend. 
This is very subjective since I make a clear distinction between legends & labels while I know most devs don't :thinking:.

I don't have a strong opinion about this at the moment, if you think it's better to prioritize ease of use with a common custom prop, it makes sense.

pdesoyres-cc (approved):
LGTM. Nice improvement.

florian-sanders-cc (approved):
All good @roberttran-cc thanks for handling this and creating + thinking about the RFC related to this issue :muscle:

PR #894, opened 2023-11-29
deprecate accessible-name in favor of a11y-name

### LINKED ISSUE #893: Deprecate `accessible-name` property in favor of `a11y-name`
This is part of a property renaming process.

The `accessible-name` will still be available but user will be encouraged to use the new name.

Impacted components:

* [x] cc-button
* [x] cc-icon
* [x] cc-img
* [x] cc-popover
* [x] cc-badge

pdesoyres-cc:
Fixes #893

## What does this PR do?

* 4 commits for `accessible-name` => `a11y-name`
  * cc-button
  * cc-icon
  * cc-img
  * cc-popover
* 1 commit for  `icon-accessible-name` => `icon-a11y-name`
  * cc-badge
* 1 commit for using the new `a11yName` everywhere in our codebase

## How it does?

* add property `a11yName` (and keep the `accessibleName` one)
* replace all usage of `this.accessibleName` by `this.a11yName`
* add getter/setter on `accessibleName` (but use this.a11uName inside them)
* add JSDoc about deprecation on the setter

## How to review?

* check the code
* check the storybook doc for
  * cc-button
  * cc-icon
  * cc-img
  * cc-popover
  * cc-badge
* check stories (especially on `cc-button` and `cc-icon` which have a specific story about accessible name)

2 reviewers should be enough.


src/components/cc-button/cc-button.stories.js
```diff
@@ -157,12 +158,25 @@ As you can see here, \`hide-text\` can only be used if there is an \`image\` or
...
+      margin-top: 2em;
+      margin-bottom: 0.5em;
+    }
+  `,
   dom: (container) => {
     container.innerHTML = `
+        <cc-notice intent="warning"><span slot="message">The <code>accessible-name</code> attribute is deprecated in favor of <code>a11y-name</code></span></cc-notice>
         <p>The accessible name can be checked by using the accessibility inspector of your browser.</p>
         <p>You may also hover the buttons because we populate the <code>title</code> attribute with the same value.</p>
+
```
florian-sanders-cc, line 172: **suggestion:** this is debatable and it's only my opinion but since we're testing the deprecate process: I would remove documentation for deprecated stuff.

I would keep the notice (nice one by the way), but I would remove the whole example and explanation about `accessible-name`.

We keep `accessible-name` for people who are already using it but we don't want new people to use it. We should only promote the new way.

src/components/cc-img/cc-img.js
```diff
@@ -65,16 +78,16 @@ export class CcImg extends LitElement {
...
 
   render () {
-    const altValue = this.accessibleName ?? '';
+    const altValue = this.a11yName ?? '';
     const isLoading = (this.src != null && !this._loaded && !this._error);
     const isSkeleton = (this.skeleton || isLoading);
     const displayAccessibleName = (this.src == null || this._error);
     return html`
-      <div class="wrapper ${classMap({ skeleton: isSkeleton, loaded: this._loaded, 'accessible-name': displayAccessibleName })}">
+      <div class="wrapper ${classMap({ skeleton: isSkeleton, loaded: this._loaded })}">
```
roberttran-cc, line 86: **👏 praise:** Nice cleanup!

roberttran-cc (approved):
Good job, LGTM! :ok_hand:

HeleneAmouzou (approved):
LGTM !

PR #897, opened 2023-11-30
fix(cc-toggle): fix extra margin on top of cc-toggle

### LINKED ISSUE #896: cc-toggle displays extra margin when inside a grid and when inline
When a` <cc-toggle>` with an inlined label is inside a gride area, an extra margin is displayed on top of it.

image

```html
<div class="grid">
  <div class="area1"><cc-toggle inline legend="legend" choices='[{ "label": "on", "value": "on" },{ "label": "off", "value": "off" }]'></cc-toggle></div>
  <div class="area2">area 2</div>
  <div class="area3">area 3</div>
</div>
```
```css
.grid {
  display: grid;
  grid-template-areas:
            'area1 area3'
            'area2 area3';
  grid-template-columns: 250px 1fr;
  grid-template-rows: max-content 1fr;
  gap: 1em;
}

.area1 {
  grid-area: area1;
  background-color: indianred;
}
.area2 {
  grid-area: area2;
  background-color: steelblue;
}
.area3 {
  grid-area: area3;
  background-color: darkseagreen;
}
```

pdesoyres-cc:
Fixes #896 

## What does this PR do?

Fixes the extra margin above the component when using `inline` and when inside a grid area.

## How to review?

You cannot see the problem until you use the component in the exactly explained situation.
See the issue details to have an example of code that reproduces the bug.

To help you, here is a story you can add to the code to see the problem, and to check if it has been fixed:
```javascript
export const grid = makeStory(conf, {
  // language=CSS
  css: `
    .grid {
      display: grid;
      grid-template-areas:
            'area1 area3'
            'area2 area3';
      grid-template-columns: 250px 1fr;
      grid-template-rows: max-content 1fr;
      gap: 1em;
    }

    .area1 {
      grid-area: area1;
      background-color: indianred;
    }
    .area2 {
      grid-area: area2;
      background-color: steelblue;
    }
    .area3 {
      grid-area: area3;
      background-color: darkseagreen;
    }
  `,
  dom: (container) => {
    container.innerHTML = `
        <div class="grid">
          <div class="area1"><cc-toggle inline legend="legend" choices='[{ "label": "on", "value": "on" },{ "label": "off", "value": "off" }]'></cc-toggle></div>
          <div class="area2"><cc-toggle legend="legend" choices='[{ "label": "on", "value": "on" },{ "label": "off", "value": "off" }]'></cc-toggle></div>
          <div class="area3"><cc-toggle inline choices='[{ "label": "on", "value": "on" },{ "label": "off", "value": "off" }]'></cc-toggle></div>
        </div>
    `;
  },
});
```

Also, check for regression on the following components:
```
─ cc-addon-option
│  ├─ cc-toggle
├─ cc-env-var-form
│  ├─ cc-toggle
├─ cc-invoice-list
│  ├─ cc-toggle
├─ cc-logsmap
│  ├─ cc-toggle
├─ cc-pricing-product-consumption
│  ├─ cc-toggle
```

Finally, check for legend alignment in the `All form controls` story.

2 reviewers should be enough.

### DIFF
--- a/src/components/cc-toggle/cc-toggle.js
+++ b/src/components/cc-toggle/cc-toggle.js
@@ -15,7 +15,7 @@ import { dispatchCustomEvent } from '../../lib/events.js';
  *
  * * This component does not replace regular usage of radio/checkbox inputs in forms.
  * * It works well in toolbars or filter panels.
- * * The single mode mode (default) works well to toggle a component between two (or more) modes.
+ * * The single mode (default) works well to toggle a component between two (or more) modes.
  *
  * ## Details
  *
@@ -129,9 +129,11 @@ export class CcToggle extends LitElement {
         : this.value === value;
     };
 
+    const hasLegend = this.legend != null && this.legend.length > 0;
+
     return html`
-      <fieldset>
-        <legend>${this.legend}</legend>
+      <div role="group" aria-labelledby=${ifDefined(hasLegend ? 'legend' : undefined)} class="group">
+        ${hasLegend ? html`<div id="legend">${this.legend}</div>` : ''}
         <div class="toggle-group ${classMap(classes)}">
           ${repeat(this.choices, ({ value }) => value, ({ label, image, value }) => html`
             <!--
@@ -156,7 +158,7 @@ export class CcToggle extends LitElement {
             </label>
           `)}
         </div>
-      </fieldset>
+      </div>
     `;
   }
 
@@ -174,40 +176,21 @@ export class CcToggle extends LitElement {
 
           display: inline-flex;
         }
-
-        /* RESET */
-
-        fieldset {
-          display: inline-block;
-          min-width: 0;
-          padding: 0;
-          border: 0;
-          margin: 0;
+        
+        .group {
+          display: flex;
+          flex-direction: column;
+          gap: 0.35em;
         }
 
-        /* RESET */
-
-        legend {
-          max-width: 100%;
-          padding: 0;
-          color: inherit;
-          line-height: inherit;
-          white-space: normal;
+        :host([inline]) .group {
+          flex-direction: row;
+          align-items: center;
+          gap: 1em;
         }
-
-        legend:not(:empty) {
-          padding-bottom: 0.35em;
+        
+        #legend {
           line-height: 1.25em;
-        }
-
-        :host([inline]) legend {
-          width: max-content;
-          padding: 0;
-          margin-right: 1em;
-          /* cannot use flex to change legend position. Only solution is float. */
-          float: left;
-          /* used to vertically center the floated element. */
-          line-height: var(--height);
         }
 
         .toggle-group {


roberttran-cc (approved):
LGTM!

Question: do we have an idea of why the original extra space? I tried to understand but it was a failure. :dizzy_face:

pdesoyres-cc:
Nope, I could not understand wy neither

src/components/cc-toggle/cc-toggle.js
```diff
@@ -172,13 +172,12 @@ export class CcToggle extends LitElement {
           --cc-toggle-img-filter-selected: none;
           --height: 2em;
 
-          display: inline-flex;
+          display: block;
```
florian-sanders-cc, line 175: **suggestion:** I didn't realize it when we discussed in sync but most form elements have a default display that is `inline-block`.

Maybe we should respect that ? Also, changing the `display` property of an element is a breaking change so we may need to rethink that and make sure there is no other solution.

I wonder if you face the same issue with `cc-input-text` and `inline` mode?

florian-sanders-cc: I'll test a few things and I'll get back to you as soon as possible

florian-sanders-cc (changes requested):
Since this is a breaking change and most form elements (+ our own `cc-` form components) have a default display that is `inline` or `inline-*`, we should make sure we have considered other solutions before going for this change.

I'll test a few things this morning :wink:

florian-sanders-cc:
If we switch the `fieldset` display to `flex` or `inline-flex` and add `align-items: center`, it fixes the issue without changing the `host` display.

This solution is not very good because I don't understand why it works. But this is the same with the `host` display change, I don't understand why it fixes the issue.

This just shows again that the `<fieldset>` / `<legend>` thing is a mess. I went for the `float: left` solution because it was the only solution, even though it's undocumented (see https://morgan.cugerone.com/blog/how-to-make-a-fieldset-legend-inline/).

I think it's time to move away from `fieldset` / `legend` for this usecase and go for ARIA (`<div>` with `role="group"` and `aria-labelledby="legendId"`).

We'll dicuss it in sync but if we do that, it would be great to write a very shot ADR about it. I think we should always write ADRs when we decide to move away from native HTML to use ARIA instead. I'm ok to do both :wink:.

src/components/cc-toggle/cc-toggle.js
```diff
@@ -174,40 +176,21 @@ export class CcToggle extends LitElement {
...
-          width: max-content;
-          padding: 0;
-          margin-right: 1em;
-          /* cannot use flex to change legend position. Only solution is float. */
-          float: left;
-          /* used to vertically center the floated element. */
-          line-height: var(--height);
+        
+        #legend {
+          height: 1.25em;
```
florian-sanders-cc, line 193: **suggestion:** I think you want `line-height` here instead so that it's exactly the same look as our labels in `cc-input`. 

This is a very minor issue and it's actually quite annoying to maintain this kind of stuff, we'll see how we can improve that when we decide on a way to share styles between form components.

florian-sanders-cc (changes requested):
Gj @pdesoyres-cc, I only pointed out a tiny spacing issue to adjust but apart from that, it's good!

florian-sanders-cc (approved):

PR #901, opened 2023-12-04
docs(ADR): add ADR about `<fieldset>` & `<legend>`

florian-sanders-cc:
## Context

We faced an issue with weird spaces added on top of the `<cc-toggle>` component when used in a grid context.

We decided to rely on `ARIA` for semantics instead of `<fieldset>` / `<legend>` so we can style it in a more robust way.

This ADR explains this and contains screen reader tests that prove this method is ok for accessibility.

## TODO

- [x]  test with VoiceOver on iOS and update the results.

### DIFF
--- a/docs/adr/adr-0025-moving-away-from-fieldset-legend.md
+++ b/docs/adr/adr-0025-moving-away-from-fieldset-legend.md
@@ -1,0 +1,151 @@
+---
+kind: '📌 Architecture Decision Records'
+---
+# ADR 0025: Moving away from `fieldset` and `legend`
+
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (approved):
Excellent writing :+1:

roberttran-cc (approved):
Perfect, thanks for this ADR. :100:

hsablonniere (approved):
Nice writing, good job on the screen reader tests :+1:

Galimede:
This ADR is really nice, GJ! :heart_eyes:

PR #903, opened 2023-12-07
chore: handle beta components when firing custom events 

### LINKED ISSUE #902: Handle beta components when firing custom events
When using `dispatchCustomEvent` utility function, custom events fired from a component `<cc-xxx>` are prefixed by `cc-xxx:`.

For beta components, it would be easier to rename them as non-beta if we could take into account the naming convention when dispatching custom event.
Custom events fired from `<cc-xxx-beta>` element should be prefixed by `cc-xxx:` instead of `cc-xxx-beta:`.

pdesoyres-cc:
Fixes #902 

### What does this PR do?

When dispatching custom events, we check if the element is a beta component (`^cc-.+-beta$`).
If it the case, the name of the event will be prefixed by the component name **without** the `-beta` suffix.

There are 2 commits:

* The first one adds unit tests on `dispatchCustomEvent` function.
* The second one implements the feature itself.

### How to review?

* check the code
* With Storybook, with Action addon, check that [this <cc-logs-beta> story](https://clever-components-preview.cellar-c2.services.clever-cloud.com/event-beta/index.html?path=/story/%F0%9F%9A%A7-beta-%F0%9F%9B%A0-logs-cc-logs-beta--data-loaded-with-many-metadata-with-wrap-lines) emits the right events when
  * scrolling to the bottom: `onCcLogsFollowChange: "false"`
  * scrolling up: `onCcLogsFollowChange: "true"`

2 reviewers should be enough.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "@commitlint/cli": "^17.1.2",
         "@commitlint/config-conventional": "^17.1.0",
         "@custom-elements-manifest/analyzer": "^0.6.4",
+        "@esm-bundle/chai-as-promised": "^7.1.1",
         "@open-wc/dev-server-hmr": "^0.1.3",
         "@open-wc/testing": "^3.1.6",
         "@rollup/plugin-alias": "^3.1.2",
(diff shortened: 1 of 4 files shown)


florian-sanders-cc (approved):
Nothing to say, this is great!

Good job, especially for adding the tests as well :muscle:

PR #907, opened 2023-12-08
fix(i18n-sanitize): handle cases where origin equals empty string

### LINKED ISSUE #906: Automatic Tests fail with chrome > 120
## Context

With chrome > 120, the `i18n-sanitize` tests fail.

They fail because the sanitize function adds `target` and `rel` attributes on the sanitized anchor element while it did not with previous chrome versions.

The `sanitize` function tests if the origin of the sanitized link (`node.origin`) equals `window.location.origin`.
If origins are different, it adds the `target="_blank"` and `rel="..."` attributes.

It looks like `node.origin` for a sanitized link with `<a href="/foo">toto</toto>` now results in en empty string while it resulted to `http://localhost...` (or whatever) before.

This seems to only happen with the `sanitize` function and for the moment, I suspect it might be related to the fact that we return a document fragment / template.

florian-sanders-cc:
This fixes an issue with Chrome >= 120

Please read the issue for more info about the bug. This will make the solution easier to understand.

Fixes #906

## How to review?

- review the commit,
- run the tests locally with chrome >= 120 on this branch vs on the main branch.

Thanks @Galimede for your help when trying to reproduce this issue and for trying to find the reason in the chrome changelog :wink:

### DIFF
--- a/src/lib/i18n-sanitize.js
+++ b/src/lib/i18n-sanitize.js
@@ -77,7 +77,9 @@ export function sanitize (statics, ...params) {
         // If link has href and external origin => force rel and target
         if (node.tagName === 'A' && node.getAttribute('href') != null) {
           node.classList.add('sanitized-link');
-          if (node.origin !== window.location.origin) {
+          // Chrome > 120 returns an empty string for an anchor element with an absolute url like `href=/foo` when it's in a template DOM element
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (approved):
LGTM

Galimede (approved):
LGTM, thanks for the fix! :+1:

PR #909, opened 2023-12-12
feat(cc-logs-control): init component

### LINKED ISSUE #908: cc-logs-control: init component
## Context

This component is a wrapper around a `<cc-logs>` component and adds the ability for a user to toggle some options.

Options are encapsulated into a `<cc-popover>` element.

### Attributes/properties

| Property            | Type                                   | Default value | Mandatory | Description                                                        |
|---------------------|----------------------------------------|---------------|-----------|--------------------------------------------------------------------|
| `dateDisplay`       | `DateDisplay`                          | `null`        | `false`   | The date display to pass to `<cc-logs>`                            |
| `filter`            | `Array<MetadataFilter>`                | `null`        | `false`   | The filter to pass to `<cc-logs>`                                  |
| `follow`            | `Boolean`                              | `null`        | `false`   | The initial follow state to pass to `<cc-logs>`                    |
| `limit`             | `Number`                               | `null`        | `false`   | The maximum number of logs in memory to pass to `<cc-logs>`        |
| `logs`              | `Array<Log>`                           | `null`        | `false`   | The initial array of logs to pass to `<cc-logs>`                   |
| `metadataDisplay`   | `{[key: string]: LogsMetadataDisplay}` | `null`        | `false`   | The metadata the user can toggle (and their initial display state) |
| `metadataRenderers` | `{[key: string], MetadataRenderer}`    | `null`        | `false`   | The metadata renderer to pass to `<cc-logs>`                       |
| `palette`           | `LogsControllerPalette`                | `null`        | `false`   | The name of the color palette to apply on `<cc-logs>`              |
| `stripAnsi`         | `Boolean`                              | `null`        | `false`   | The stripAnsi to pass to `<cc-logs>`                               |
| `timezone`          | `Timezone`                             | `null`        | `false`   | The timezone to pass to `<cc-logs>`                                |
| `wrapLines`         | `Boolean`                              | `null`        | `false`   | The warpLines to pass to `<cc-logs>`                               |

### Slots

| Name  | Description |
|-------|-------------|
| `header` | A zone on top of the `<cc-logs>` element |

## Outputs

| `cc-logs-control:optionChange` | `LogsControlOption`    | Fires a `LogsControlOption` whenever an `option` changes. |

## Type Definitions

```typescript
type DateDisplay = 'none' | 'datetime-iso' | 'time-iso' | 'datetime-short' | 'time-short';

interface Log {
  id: string;
  date: Date;
  message: string;
  metadata: Array<Metadata>;
}

interface Metadata {
  name: string;
  value: string;
}

interface MetadataFilter {
  metadata: string;
  value: string;
}

type MetadataRenderer = MetadataRenderingProvider | MetadataRendering;

type MetadataRenderingProvider = (metadata: Metadata) => MetadataRendering;

interface MetadataRendering {
  hidden?: boolean;
  intent?: MetadataIntent;
  showName?: boolean;
  size?: 'auto' | number;
  strong?: boolean;
  text?: string;
}

type MetadataIntent = 'neutral' | 'info' | 'success' | 'warning' | 'danger';

interface LogsMetadataDisplay {
  label: string;
  hidden: boolean;
}

type LogsControlPalette = "default" | "One Light" | "Tokyo Night Light" | "Night Owl" | "Everblush" | "Hyoob"

interface LogsControlWrapLinesOption {
  name: "wrap-lines";
  value: boolean;
}

interface LogsControlDateDisplayOption {
  name: "date-display";
  value: DateDisplay;
}

interface LogsControlTimezoneOption {
  name: "timezone";
  value: Timezone;
}

interface LogsControlPaletteOption {
  name: "palette";
  value: LogsControlPalette;
}

interface LogsControlStripAnsiOption {
  name: "strip-ansi";
  value: boolean;
}

interface LogsControlDisplayedMetadataOption {
  name: "metadata-display";
  value: { [key: string]: boolean };
}

type LogsControlOption =
  LogsControlWrapLinesOption
  | LogsControlDateDisplayOption
  | LogsControlTimezoneOption
  | LogsControlPaletteOption
  | LogsControlStripAnsiOption
  | LogsControlDisplayedMetadataOption;

type Timezone = 'local' | 'UTC';
```

pdesoyres-cc:
Fixes #908

## What does this PR do?

3 commits:

* refactor: extract and rework some methods that help for creating fake logs for <cc-logs-*> stories and sandbox
* upgrade remix icon to get a new icon from the library
* implement the new component itself

## How to review?

* Overall design
* Code

A review from everyone would be great.


src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+              <input id="metadata-${name}"
+                     type="checkbox"
+                     data-name="${name}"
+                     @change=${this._onMetadataChange}
+                     .checked=${!display.hidden}> ${display.label}
+            </label>
+          `;
+        })}
+      </div>
+      </div>
```
florian-sanders-cc, line 427: **fix:** you got a `</div>` here

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: grid;
+          align-items: center;
+          column-gap: 0.35em;
+          grid-template-areas: 
+            'header-slot scroll-button options-popover'
+            'logs        logs          logs';
```
florian-sanders-cc, line 441: **praise:** I like this way of formating `grid-template-areas` :sunglasses:

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+
+        .options-group {
+          display: grid;
+          margin: 1em 0.75em 0.75em;
+          grid-template-columns: minmax(max-content, 1fr);
+          row-gap: 0.75em;
+        }
+
+        input[type='checkbox'] {
+          margin: 0;
```
florian-sanders-cc, line 497: **fix:** it looks like inputs do not scale when resizing the font so I suggest adding these to fix the issue (feel free to change the values of course)
```suggestion
          margin: 0;
          width: 1em;
          height: 1em;
```

florian-sanders-cc (changes requested):
Well done @pdesoyres-cc this component looks great and the code is easy to read :+1: 

Most of my comments are nitpicks as usual and here is one more:

- I think we should change the icon for the follow button. I always think I'm going to be able to download the logs and since this button is icon only, we should avoid ambiguous icons. I couldn't find a really good fit within the remix icons project but [the scroll down examples form the Noun Project](https://thenounproject.com/browse/icons/term/scroll-down/) look decent enough to me. Maybe we could create an issue within the Remix project?

pdesoyres-cc:
[let see how it goes](https://github.com/Remix-Design/RemixIcon/issues/704)

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+ * * strip ansi
+ * * date display
+ * * timezone
+ * * displayed metadata
+ *
+ * Options are encapsulated into a `<cc-popover>` element.
+ *
+ * @cssdisplay grid
+ *
+ * @event {CustomEvent<LogsControllerOption>} cc-logs-controller:optionChange - Fires a `LogsControllerOption` whenever an `option` changes.
```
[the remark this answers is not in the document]
pdesoyres-cc: Lets discuss this with the team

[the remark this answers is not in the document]
pdesoyres-cc: ok, lets go for kebab-case

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+    /** @type {boolean} Whether to strip ANSI from log message. */
+    this.stripAnsi = false;
+
+    /** @type {Timezone} The timezone to use when displaying date. */
+    this.timezone = 'UTC';
+
+    /** @type {boolean} Whether to wrap long lines. */
+    this.wrapLines = false;
+
+    /** @type {Ref<CcLogs>} */
```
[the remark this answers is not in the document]
pdesoyres-cc: This is an edge case that our homemade type checking cannot handle. It will complain that the `CcLogs` import is not used.

src/components/cc-logs-controller/cc-logs-controller.stories.js
```diff
@@ -0,0 +1,59 @@
...
+    {
+      follow: true,
+      metadataDisplay: metadataDisplay,
+      metadataRenderers: CUSTOM_METADATA_RENDERERS,
+      logs: createFakeLogs(100, { longMessage: true, ansi: true }),
+    },
+  ],
+});
+
+export const withoutMetadataDisplay = makeStory(conf, {
```
HeleneAmouzou, line 46: **question**: I am not sure to understand why do we need the case without metadata display ?

HeleneAmouzou (commented):
GG @pdesoyres-cc ! :clap: 
The visual is very nice !  :sparkles:  I went through the code and I have just left one question. Also about the `cc-logs-controller`, when I saw the scroll to bottom icon I though it was a button to download the logs. I haven't noticed it before and I don't know if I am the only one to have this impression or if there is another icon more appropriate. Otherwise, LGMT !

src/components/cc-logs-controller/cc-logs-controller.stories.js
pdesoyres-cc: This is to show that it is possible to not let the user change the metadata display, and that in that case, the whole section disappears from the UI.

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+    /** @type {{[key: string], MetadataRenderer}} The resolved metadata renderers to use for displaying metadata. */
+    this._resolvedMetadataRenderers = {};
+  }
+
+  /* region Public methods */
+
+  /**
+   * Forces scroll to bottom.
+   */
+  scrollToBottom () {
```
hsablonniere, line 141: question: do we need this to be public?

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+  _getPaletteLabel (palette) {
+    if (palette === 'default') {
+      return i18n('cc-logs-controller.palette.default');
+    }
+    return palette;
+  }
+
+  /* endregion */
+
+  willUpdate (_changedProperties) {
```
hsablonniere, line 282: nitpick: `_changedProperties` => `changedProperties`

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+    if (_changedProperties.has('metadataRenderers') || _changedProperties.has('metadataDisplay')) {
+      this._resolvedMetadataRenderers = this._resolveMetadataRenderers(this.metadataRenderers, this.metadataDisplay);
+    }
+  }
+
+  render () {
+    return html`
+      <div class="header-slot">
+        <slot name="header"></slot>
+      </div>
```
hsablonniere, line 292: issue: can we add a story to show off this feature

or maybe we can remove this feature for now as we don't need it yet

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+        ${this._renderDisplayOptions()}
+        ${this._renderDateOptions()}
+        ${this._renderMetadataOptions()}
+      </div>
+    `;
+  }
+
+  _renderDisplayOptions () {
+    const PALETTE_CHOICES = Object.entries(PALETTES)
+      .map(([p]) => ({ label: this._getPaletteLabel(p), value: p }));
```
hsablonniere, line 335: question: can/should we move this to the beginning of the file (and do it just once)?

This also means removing the `_getPaletteLabel`.

We do something similar here: https://github.com/CleverCloud/clever-components/blob/master/src/components/cc-pricing-estimation/cc-pricing-estimation.js#L22-L33

src/components/cc-logs-controller/cc-logs-controller.js
```diff
@@ -0,0 +1,504 @@
...
+                 .checked=${this.wrapLines}> ${i18n('cc-logs-controller.wrap-lines')}
+        </label>
+      </div>`;
+  }
+
+  _renderDateOptions () {
+    const DATE_DISPLAY_CHOICES = DATE_DISPLAYS
+      .map((d) => ({ label: this._getDateDisplayLabel(d), value: d }));
+    const TIMEZONE_CHOICES = TIMEZONES
+      .map((z) => ({ label: this._getTimezoneLabel(z), value: z }));
```
hsablonniere, line 370: question: can/should we move this to the beginning of the file (and do it just once)?

This also means removing the `_getDateDisplayLabel` and `_getTimezoneLabel`.

We do something similar here: https://github.com/CleverCloud/clever-components/blob/master/src/components/cc-pricing-estimation/cc-pricing-estimation.js#L22-L33

src/components/cc-logs-controller/cc-logs-controller.stories.js
```diff
@@ -0,0 +1,59 @@
...
+      border-radius: 0.2em;
+      padding: 0.5em;
+      height: 800px;
+    }
+  `,
+};
+
+const metadataDisplay = {
+  ip: {
+    label: 'Afficher l\'adresse IP',
```
hsablonniere, line 27: issue: can we use english in stories?

src/components/cc-logs-controller/cc-logs-controller.stories.js
```diff
@@ -0,0 +1,59 @@
...
+    hidden: true,
+  },
+  level: {
+    label: 'Afficher le niveau de log',
```
hsablonniere, line 31: issue: can we use english in stories?

src/components/cc-logs-controller/cc-logs-controller.stories.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const conf = {
+  component: 'cc-logs-controller-beta',
+  beta: true,
+  // language=CSS
+  css: `
+    cc-logs-controller-beta {
+      border: 1px solid #ddd;
+      border-radius: 0.2em;
```
hsablonniere, line 18: issue: I'm not sure we want to display such a border, it may be confusing for reviewers, thinking it's part of the component

hsablonniere (changes requested):
Hey @pdesoyres-cc, good job on this!! :clap: 

I added a few remarks and here are some global notes:

* Now that I see the component, I'm not sure `cc-logs-options` is the right name
* Will we put the future input for text filtering in this component or through the slot?
* I was about to say "why do we need this component, why is it not directly in the main component ?"
  * Looking at the code, I understand why and it's a good way to isolate stuffs, even if it means mirroring a lot of properties
* Can we have a story to show of the fact that we can select other options than the default ones?

src/components/cc-logs-controller/cc-logs-controller.js
pdesoyres-cc: not really needed

src/components/cc-logs-controller/cc-logs-controller.js
pdesoyres-cc: I don't want to call the i18n function outside the instance of the component.

src/components/cc-logs-controller/cc-logs-controller.js
pdesoyres-cc: I don't want to call the i18n function outside the instance of the component.

hsablonniere (changes requested):
Thanks for your feedbacks and new commits :+1: 

I still have questions:

* I'm not sure `cc-logs-options` is the right name but I don't want `controller` either, WDYT?
* Will we put the future input for text filtering in this component or through the slot?
* Can we have a story to show of the fact that we can select other options than the default ones?

pdesoyres-cc:
> I'm not sure cc-logs-options is the right name but I don't want controller either, WDYT?

- I can propose `cc-logs-control` so that we don't clash with the MVC design pattern's concept of _controller_.
- I can propose `cc-logs-wrapper` but I don't like it very much as it doesn't say what it does.
- I can propose `cc-logs-tuner`, `cc-logs-manager`, `cc-logs-proxy`, `cc-logs-adapter`

> Will we put the future input for text filtering in this component or through the slot?

My first idea for the slot was to put the date range input, but there was not enough room.
Also, I wanted to put some loading indicators here, but finally, I found a better place.
The question is: does the text-filtering apply to addon logs, application logs, and access logs? I would say yes.

Having said that, we could get rid of the slot and plan the addition of the text-filtering input.

> Can we have a story to show of the fact that we can select other options than the default ones?

yes, I'll do that.

Galimede (approved):
Nice work Pierre, LGTM! :muscle:

hsablonniere:
After a discussion with @pdesoyres-cc 

* About the name: `cc-logs-control` seems to be the better idea
* Maybe we cold rename `cc-logs` into `cc-logs-viewer`
  * Pierre will decide on this one
* Let's remove the slot, the text input will be part of this component

florian-sanders-cc:
:wave: Sorry that I did not notice it ealier: we're missing a focus style for checkboxes
Chrome
black focus outline on chrome for checkboxes
double focus outline on firefox for checkboxes

This should do the trick:
```css
        input:focus {
          outline: var(--cc-focus-outline, #000 solid 2px);
          outline-offset: var(--cc-focus-outline-offset, 2px);
        }
```

Apart from that, LGTM!
About the name, we could also go for `cc-logs-settings`? But I'm fine with `cc-logs-control` too :wink:

roberttran-cc (changes requested):
Congrats for the awesome work: there's an issue with a non-working story but LGTM for the rest!

src/components/cc-logs-control/cc-logs-control.stories.js
```diff
@@ -0,0 +1,135 @@
...
+  items: [
+    {
+      follow: true,
+      metadataRenderers: CUSTOM_METADATA_RENDERERS,
+      logs: createFakeLogs(100, { longMessage: true, ansi: true }),
+    },
+  ],
+});
+
+export const withCustomHeader = makeStory(conf, {
```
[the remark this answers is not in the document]
pdesoyres-cc: I removed the custom header slot. This story is no longer relevant.

Thanks

PR #910, opened 2023-12-13
feat(cc-product-card): init component

### LINKED ISSUE #891: cc-product-card: init component
## Context

With our current work on a new creation tunnel for our console, we have decided to split it in several steps. 
This component will be part of the first step which consists of selecting the product you want to create.
This will be a card that will allow you to select a product and display basic infos such as a description, a title, an icon and some keywords. In the future you should be able to put this product as favorite within the card, but that's not a priority for now.

### Attributes/properties

| `name`    | `String` | `''`        | `true`    | name of the product        |
| `icon`    | `String` | `''`        | `true`    | icon of the product        |
| `description`    | `String` | `''`        | `true`    | description of the product        |
| `link`    | `String` | `''`        | `true`    | link to the next step of the creation tunnel        |
| `keywords`    | `Array<Keyword>` | `[]`        | `false`    | keywords/tags of the products        |
| `favorite`    | `boolean` | `false`        | `false`    | is the product favorite (TBD afterwards)        |
## Outputs

| `cc-product-card:select` | `String`    |    This event could serve for analytics     |
| `cc-product-card:favorite` | `Object`    |    Will emit the name and a boolean   (TBD afterwards)  |

## Type Definitions

```typescript
interface Keyword {
  value: string;
  hidden: boolean;
}
```
## Stories & states

If we have a product that comes from the API, we might have to have a skeleton and error mode.
Though, we'll need to find out if we want to do it here or in the list.

* Skeleton/loading
* Loaded
* Error
* Simulation

## UX writing & i18n

For the link to the next step we had the idea of giving it the "name": "create". While I thought this was a good idea at first, at this step we will only "select" the product that we want for the next step.  With this in mind `select` might be a better choice.

| Key | EN value | FR value |
|--------|--------|--------|
| `cc-product-card.select` | Select | Sélectionner |

Examples from a prototype
javascript-prototype
elixir-prototype
elastic-prototype

Galimede:
Fixes #891

## What does this PR do?

This PR introduces the first component for the first step of our new creation tunnel.

## How to review?

* Check the preview, feel free to give your thoughts on the padding, alignment and so on.
  * There's a lot of stories to cover cases that could happen. (Especially with product from the API)
  * If you checkout the branch, you can go to `/sandbox` in your local Storybook. It will let you have a glimpse of the products from the API. You can also customize the product details if you want to play with the card. It's not mandatory though :wink: 
 * Check the code

A review by everyone would be appreciated :slightly_smiling_face: 
The component is quite simple, it should be easy to review :wink:


src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * @typedef {import('./cc-product-card.types.js').Keyword} Keyword
+ */
+
+/**
+ * A component displaying some information about a product in a card.
+ *
+ * @cssdisplay block
+ *
+ * @event {CustomEvent} cc-product-card:select - Fires the product name when the product is selected.
```
pdesoyres-cc, line 17: nitpick: add detail type:

```suggestion
 * @event {CustomEvent<string>} cc-product-card:select - Fires the product name when the product is selected.
```

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+ *
+ */
+export class CcProductCard extends LitElement {
+
+  static get properties () {
+    return {
+      description: { type: String },
+      icon: { type: String },
+      link: { type: String },
+      keywords: { type: Array },
```
pdesoyres-cc, line 27: nitpick:should be defined before `link` property

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * @event {CustomEvent} cc-product-card:select - Fires the product name when the product is selected.
+ *
+ */
+export class CcProductCard extends LitElement {
+
+  static get properties () {
+    return {
+      description: { type: String },
+      icon: { type: String },
```
pdesoyres-cc, line 25: nitpick: most of the time we use `icon` property to receive an `IconModel`. Here it is an url, so maybe, we should rename this property into `iconUrl` (or `logoUrl` which is what comes from the API)

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+    this.description = null;
+
+    /** @type {string|null} The icon of the product */
+    this.icon = null;
+
+    /** @type {string|null} The link where the user will be redirected when the product is selected  */
+    this.link = null;
+
+    /** @type {Keyword[]} The keywords representing the product */
+    this.keywords = [];
```
pdesoyres-cc, line 45: nitpick: this property should be defined before `link` property.

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+    super();
+
+    /** @type {string|null} The description of the product */
+    this.description = null;
+
+    /** @type {string|null} The icon of the product */
+    this.icon = null;
+
+    /** @type {string|null} The link where the user will be redirected when the product is selected  */
+    this.link = null;
```
pdesoyres-cc, line 42: nitpick: `link` is more about a visible representation on screen. Here this property is a URL, so maybe we should rename it into `url`.

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+    dispatchCustomEvent(this, 'select', this.name);
+  }
+
+  render () {
+
+    const hasKeywords = this.keywords.length > 0;
+
+    // language=HTML
+    return html`
+      <div class="${hasKeywords ? 'wrapper' : 'wrapper-no-keywords'} wrapper-base">
```
pdesoyres-cc, line 61: nitpick: this could be implemented with `classMap`:

```js
classMap({'wrapper-base': true, wrapper: hasKeyword, 'wrapper-no-keywords': !hasKeyword})
```

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+      }
+        
+      /* endregion */
+        
+      /* region grid-items */
+
+      cc-img {
+        width: 2em;
+        height: 2em;
+        border-radius: 0.2em;
```
pdesoyres-cc, line 140: question: why not using `border-radius: var(--cc-border-radius-default, 0.25em);` ?

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+        color: var(--cc-color-text-weak);
+        font-size: 0.8em;
+        grid-area: description;
+        line-height: 1.5em;
+      }
+
+      .select {
+        padding: 0.25em;
+        border: 1px solid;
+        border-color: var(--cc-color-bg-primary);
```
pdesoyres-cc, line 166: nitpick: maybe add a default value (#333)

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,198 @@
...
+      .keywords {
+        display: flex;
+        flex-wrap: wrap;
+        gap: 0.5em;
+        grid-area: keywords;
+      }
+
+      .description {
+        align-self: center;
+        color: var(--cc-color-text-weak);
```
pdesoyres-cc, line 157: nitpick: maybe add a default value (#333)

src/components/cc-product-card/cc-product-card.types.d.ts
```diff
@@ -0,0 +1,4 @@
+interface Keyword {
```
pdesoyres-cc, line 1: nitpick: I think it's better to export it.

src/components/cc-product-card/cc-product-card.stories.js
```diff
@@ -0,0 +1,147 @@
...
+  items: productInMultipleWidth({
+    icon: 'https://placekitten.com/202/202',
+    name: 'My product',
+    description: 'my very long very long very long very long very long very long very long very long description',
+  }),
+});
+
+export const longDescriptionAndLongName = makeStory(conf, {
+  items: productInMultipleWidth({
+    icon: 'http://placekitten.com/202/202',
```
pdesoyres-cc, line 110: nitpick: use https

src/components/cc-product-card/cc-product-card.stories.js
```diff
@@ -0,0 +1,147 @@
...
+    name: 'My product with a very very very very very very very long name',
+    description: 'my very long very long very long very long very long very long very long very long description',
+    keywords: generateRandomKeywords(3),
+  }),
+});
+
+export const longDescriptionAndLongNameWithoutKeywords = makeStory(conf, {
+  items: productInMultipleWidth({
+    icon: 'http://placekitten.com/202/202',
```
pdesoyres-cc, line 119: nitpick: use https

src/components/cc-product-card/cc-product-card.stories.js
```diff
@@ -0,0 +1,147 @@
...
+    name: 'My product with a very very very very very very very long name',
+    description: 'my very long very long very long very long very long very long very long very long description',
+  }),
+});
+
+export const longDescriptionAndLongNameWithManyKeywords = makeStory(conf, {
+  items: productInMultipleWidth({
+    icon: 'http://placekitten.com/202/202',
```
pdesoyres-cc, line 127: nitpick: use https

src/components/cc-product-card/cc-product-card.stories.js
```diff
@@ -0,0 +1,147 @@
...
+  component: 'cc-product-card',
+};
+
+const conf = {
+  component: 'cc-product-card',
+};
+
+export const defaultStory = makeStory(conf, {
+  items: productInMultipleWidth({
+    icon: 'https://placekitten.com/202/202',
```
pdesoyres-cc, line 32: nitpick: maybe add a `const` and use it on every story

src/components/cc-product-card/generate-random-keywords.js
```diff
@@ -0,0 +1,10 @@
+export function generateRandomKeywords (number) {
+  const keywords = [{ value: 'hidden', hidden: true }];
+
+  for (let i = 0; i < number; i++) {
+    const str = Math.random().toString(36).substring(2, 7);
```
pdesoyres-cc, line 5: suggestion: Maybe, you could extract this random string generator into `src/lib/utils.js` which already contains some random related functions.

sandbox/cc-product-card/cc-product-card-sandbox.js
```diff
@@ -0,0 +1,113 @@
...
+import '../../src/components/cc-block/cc-block.js';
+import '../../src/components/cc-input-number/cc-input-number.js';
+import '../../src/components/cc-input-text/cc-input-text.js';
+import '../../src/components/cc-product-card/cc-product-card.js';
+import { generateRandomKeywords } from '../../src/components/cc-product-card/generate-random-keywords.js';
+import '../../src/components/cc-toggle/cc-toggle.js';
+import '../../src/components/cc-select/cc-select.js';
+
+const DEFAULT_PRODUCT = {
+  icon: 'http://placekitten.com/202/202',
```
pdesoyres-cc, line 12: nitpick: use https

sandbox/cc-product-card/cc-product-card-sandbox.js
```diff
@@ -0,0 +1,113 @@
...
+
+    return html`
+      <cc-select
+        label="API products"
+        value="default"
+        .options="${API_PRODUCTS_SELECT}"
+        @cc-select:input="${this._onSelect}"
+      ></cc-select>
+      <cc-input-number value=${this._componentWidth} label="component width (em)" @cc-input-number:input="${this._onWidthInput}"></cc-input-number>
+      <cc-block state="close">
```
pdesoyres-cc, line 86: nitpick: I'm not sure the cc-block is necessary. Also, it took me while to understand that we could modify this input.

sandbox/cc-product-card/cc-product-card-sandbox.js
```diff
@@ -0,0 +1,113 @@
...
+        <div slot="title">Products details</div>
```
pdesoyres-cc, line 87: nitpick: Use singular as there is only one product displayed at a time.

pdesoyres-cc (changes requested):
First stone of a huge project! Well done Mathieu.

On the overall design, I find it simple and clear. I'm not totally sold about the button link alignement, especially when the card is large. Maybe it could be aligned on the right. Anyway, It is ok like that for a first iteration.

I'm wondering if we should consider this component as `beta`.

I let you some little nitpicks that you may consider or not.

src/components/cc-product-card/cc-product-card.js
```diff
@@ -36,14 +37,14 @@ export class CcProductCard extends LitElement {
     this.description = null;
 
     /** @type {string|null} The icon of the product */
```
pdesoyres-cc, line 39:
```suggestion
/** @type {string|null} The url of the icon of the product */
```

src/components/cc-product-card/cc-product-card.js
```diff
@@ -36,14 +37,14 @@ export class CcProductCard extends LitElement {
...
-    this.icon = null;
-
-    /** @type {string|null} The link where the user will be redirected when the product is selected  */
-    this.link = null;
+    this.iconUrl = null;
 
     /** @type {Keyword[]} The keywords representing the product */
     this.keywords = [];
 
+    /** @type {string|null} The link where the user will be redirected when the product is selected  */
```
pdesoyres-cc, line 45:
```suggestion
    /** @type {string|null} The url where the user will be redirected when the product is selected  */
```

src/lib/utils.js
```diff
@@ -92,18 +92,35 @@ export function asyncMap (array, asyncCallback) {
...
-  * Returns the currency symbol corresponding to the given currency.
-  *
-  * @param {string} currency - the currency to get the symbol for
-  */
+ * Returns a list of random keywords.
+ *
+ * @param {number} number
+ * @returns {Array<{value: string, hidden: boolean}>}
+ */
+export function generateRandomKeywords (number) {
```
pdesoyres-cc, line 100: Sorry, I was not clear enough in my previous feedback. I didn't mean to move all the function to this `utils.js` file.
I was only talking about a `randomString` function: something like that:

```js
export function randomString(length) {
  return Math.random().toString(36).substring(2, length);
}
```

The `generateRandomKeyworks` function can be kept in the file you previously had because it is very specific to the component you are creating.

Sorry again for the confusion.

pdesoyres-cc (changes requested):
It's me again, I've got some nitpicks and one point on which we didn't understand each other (sorry for that).

pdesoyres-cc (approved):
LGTM, well done Mathieu.

florian-sanders-cc (approved):
Gg @Galimede this component looks good and the code is great.

Thanks for adding the list story, it helped a lot.

One of the suggestions I've added is not an easy subject so feel free to ask the team for their opinion :+1: 

I have one more suggestion: we could improve `cc-badge` with `dimmed + neutral` to make it a little bit more subtle and in your component, reduce the font-size for `cc-badge` (`0.85em` for instance).

Here is a before / after so you can decide if it's worth considering / discussing:

roberttran-cc (approved):
Looking good! :fire:

About the code, I only have a non-blocking suggestion.

About the UI, I can't find a way where the "Select" button looks great so... I tend to think that the solution is maybe get rid of it and the whole tile becoming a clickable button.
Actually, this matter should be discussed synchronously.

GG nevertheless, looking forward this component in motion in the new tunnel! :rocket:

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,206 @@
...
+ * @typedef {import('./cc-product-card.types.js').Keyword} Keyword
+ */
+
+/**
+ * A component displaying some information about a product in a card.
+ *
+ * @cssdisplay block
+ *
+ * @event {CustomEvent<string>} cc-product-card:select - Fires the product name when the product is selected.
+ *
```
hsablonniere, line 19: nitpick: remove empty line

hsablonniere (changes requested):
Hey @Galimede, well done on this one :clap: 

A few nitpicks and suggestions here and there and one issue for the event.

* I like the look and I agree with Florian's suggestions about the badges
* I'm not sure we need the beta status on this one but it's your call

hsablonniere (approved):
:+1: Hey Mathieu, I really like this buttonless version.

I like the hover and focuses states. Let's go for this, well done.

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        .wrapper:hover {
+          border-color: var(--cc-color-border-neutral-focused, #777);
+          cursor: pointer;
+        }
+
+        .wrapper:focus-within {
+          border: 2px solid var(--cc-color-border-neutral, #aaa);
+          border-color: var(--cc-color-border-neutral-focused, #777);
```
florian-sanders-cc, line 128: **nitpick:** we could set the right color directly within the `border` shothand
```suggestion
          border: 2px solid var(--cc-color-border-neutral-focused, #777);
```

src/components/cc-product-card/cc-product-card.js
```diff
@@ -0,0 +1,187 @@
...
+              return (!keyword.hidden) ? html`
+                <li>
+                  <cc-badge class="keyword-badge">${keyword.value}</cc-badge>
+                </li>
+              ` : '';
+            })}
+          </ul>
+        ` : ''}
+        <a href="${this.url}" class="select">${i18n('cc-product-card.select')}</a>
+      </div>
```
florian-sanders-cc, line 75: **fix:** the hidden link works for screen reader users but it causes issues for people browsing with voice only.

The link is called "select" but there is not visible text to guess that you have to say "select" to activate this card.

The best option would be to remove this hidden link and use a link on the heading instead.
We can still make it span through the whole card with a similar technique, we can discuss it in sync if you want.

If we do this, people may activate this link just by saying their command ("activate", "click", whatever) and the name of the product "NodeJS" for instance.
Sorry that I forgot about this subject when we discussed it in sync the other day :sweat_smile: 
```suggestion
      <div class="wrapper ${classMap({ 'wrapper--no-keywords': !hasKeywords })}">
        <cc-img src="${this.iconUrl}"></cc-img>
        <a href="${this.url}" class="name">${this.name}</a>
        <p class="description">${this.description}</p>
        ${hasKeywords ? html`
          <ul class="keywords">
            ${this.keywords.map((keyword) => {
              return (!keyword.hidden) ? html`
                <li>
                  <cc-badge class="keyword-badge">${keyword.value}</cc-badge>
                </li>
              ` : '';
            })}
          </ul>
        ` : ''}
      </div>
```

florian-sanders-cc (changes requested):
A small nitpick and a fix for a11y that I should have thought about when we discussed this subject in the first place (sorry :smile:).

Apart from that, good job, it looks good and I agree with going buttonless, it looks better this way :+1:

PR #914, opened 2023-12-18
chore(stylelint): rename config config file

### LINKED ISSUE #913: Stylelint config: change the config filename to match the official recommandations
## Context

Our Stylelint config name is `stylelint.config.json`.

It does not match any of the possible config names listed in the [Stylelint Docs](https://stylelint.io/user-guide/configure).

This means some editors / IDEs / LSPs may fail to detect the config file.
For instance, with VSCode and SublimeText/Neovim with the Stylelint LSP, we have to specify the config file path manually.

This is not big deal but since it's easy and it has no bad consequences, we should rename it to one of the standard names, for instance:

`.stylelintrc.json`

florian-sanders-cc:
Fixes #913

TLDR; this PR renames the config file to match the Stylelint standards so that all editors/LSPs may detect the config file without having to manually set the config path.

## How to review?

- 1 reviewer should be more than enough,
- review the commit,
- pull the branch and ensure in local that your WebStorm still picks up Stylelint errors + that it still fixes them when needed (the easiest way to test is by messing with property order in a selector).

I have already tested the CI to make sure it still worked:
Stylelint error within the CI


Galimede (approved):
LGTM, GG! :muscle:

PR #919, opened 2023-12-21
refactor(cc-addon-linked-apps): migrate to the new smart component design

pdesoyres-cc:
Fixes #918 (which is part of #708)
Also fixes a part of #917.

## What is this?

* rework the properties to use our new way of implementing component states (this is a breaking change)
* migration of the smart component with the new design (no RxJs)
* remove incoming confusion with Instance type (see #917).

## How to review?

* code
* check for regressions

PR #928, opened 2023-12-22
refactor(cc-tile-instances)!: rework properties to make impossible state impossible

### LINKED ISSUE #917: Tackle confusion on multiple Instance types
When working on the new logs view, I discovered that some other components already use `Application` and `Instance` concepts.
The `cc-logs-instances` component used inside the new logs view displays a list of Instances.
Those instances come from the V4 API that doesn't have the same shape as the one already designed for other components.

This creates confusion between those types.

## The actual situation

3 types are involved:

```ts
interface Application {
  name: string;
  link: string;
  instance: Instance;
  zone: Zone;
}

interface Instance {
  flavourName: string;
  count: number;
}

interface InstancesState {
  running: Instance[];
  deploying: Instance[];
}
```

All those three types are located in `common.types.d.ts`.

2 components are involved:

* `cc-addon-linked-apps`
  * uses `Application`
  * uses the field `variant` of the application instance even though it is not defined on the `Instance` interface.
  * It uses none of the fields defined in the `Instance` interface.
  * This component is fed by its associated smart component
  * It takes its data from `/v2/organisations/${owner_id}/addons/${addon_id}/applications` API.
* `cc-tile-instances`
  * uses `InstancesState`
  * It is used directly in the console (no smart component)
  * Data are taken from `/v2/organisations/${owner_id}/instances` API
  * Data from this API is transformed to get
    * the instance count (`count` property)
    * the `flavorName` is fed with `flavor.name`

To sum up, we have  
* `cc-addon-linked-apps` component which doesn't use any of the fields defined in the `Instance` interface.
   Instead, it uses an undefined field.
* `cc-tile-instances` component which uses something converted from the instances API.

### `cc-addon-linked-apps`

* create its own `types.d.ts` file
* create a `LinkedApplication` interface

```ts
interface LinkedApplication {
  name: string;
  link: string;
  variantName: string;
  variantLogoUrl: string;
  zone: Zone;
}
```

### `cc-tile-instances`

* move types into its own `types.d.ts` file
* rename `Instance` type to `InstanceState`

### LINKED ISSUE #927: cc-tile-instances: migrate to new state
This is about making impossible state impossible.

pdesoyres-cc:
Fixes #927
Also fixes a part of #917.

### What is this?

* rework the properties to use our new way of implementing component states (this is a breaking change)
* remove confusion on Instance type (see #917)

### How to review?

* check the code
* check for regression

### Todo

* [x] merge #919 and then, remove Instance interface

### DIFF
--- a/src/components/cc-tile-instances/cc-tile-instances.js
+++ b/src/components/cc-tile-instances/cc-tile-instances.js
@@ -18,22 +18,12 @@ const statusIcon = {
   deploying: iconStarting,
 };
 
-/** @type {InstancesState} */
-const SKELETON_INSTANCES = {
-  running: [],
-  deploying: [],
-};
-
 /**
- * @typedef {import('../common.types.js').InstancesState} InstancesState
+ * @typedef {import('./cc-tile-instances.types.js').TileInstancesState} TileInstancesState
  */
 
 /**
  * A "tile" component to display current status of running and deploying instances for a given app.
- *
- * ## Details
- *
- * * When `instances` is nullish, a loader is displayed.
  *
  * @cssdisplay grid
  */
@@ -41,19 +31,15 @@ export class CcTileInstances extends LitElement {
 
   static get properties () {
     return {
-      error: { type: Boolean, reflect: true },
-      instances: { type: Object },
+      state: { type: Object },
     };
   }
 
   constructor () {
     super();
 
-    /** @type {boolean} Displays an error message. */
-    this.error = false;
-
-    /** @type {InstancesState|null} Sets the current state of running and deploying instances. */
-    this.instances = null;
+    /** @type {TileInstancesState} Sets the current state of running and deploying instances. */
+    this.state = { type: 'loading' };
   }
 
   _getStatusLabel (type) {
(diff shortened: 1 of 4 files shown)


src/components/common.types.d.ts
```diff
@@ -175,11 +175,6 @@ interface Instance {
   count: number;
 }
 
-interface InstancesState {
-  running: Instance[];
-  deploying: Instance[];
-}
-
```
roberttran-cc, line 182: **❔ question:** actually, should we remove all unused interface in this PR? Including the one related to #919?

src/components/cc-tile-instances/cc-tile-instances.js
```diff
@@ -95,30 +98,16 @@ export class CcTileInstances extends LitElement {
...
         <div class="tile_message">${i18n('cc-tile-instances.empty')}</div>
       ` : ''}
-
-      ${this.error ? html`
-        <div class="tile_message">
-          <div class="error-message">
-            <cc-icon .icon="${iconAlert}" a11y-name="${i18n('cc-tile-instances.error.icon-a11y-name')}" class="icon-warning"></cc-icon>
-            <p>${i18n('cc-tile-instances.error')}</p>
-          </div>
+
```
roberttran-cc, line 104: **🤓 nitpick:**

```suggestion

```

roberttran-cc (approved):
GG, the whole set of PR on states are so nice. :heart: 

Some non-blocking feedback.

src/components/cc-tile-instances/cc-tile-instances.js
```diff
@@ -18,42 +18,28 @@ const statusIcon = {
...
 
   constructor () {
     super();
 
-    /** @type {boolean} Displays an error message. */
-    this.error = false;
-
-    /** @type {InstancesState|null} Sets the current state of running and deploying instances. */
-    this.instances = null;
+    /** @type {TileInstancesState|null} Sets the current state of running and deploying instances. */
```
florian-sanders-cc, line 41: **nitpick:**
```suggestion
    /** @type {TileInstancesState} Sets the current state of running and deploying instances. */
```

florian-sanders-cc (changes requested):
Apart from a small nitpick, all good! :+1:

hsablonniere (approved):
Nice work @pdesoyres-cc, congrats :clap:

I tried a different approach for the render method in a branch:

* the title is only declared once instead of three times
* the type is enforced (is it?)

https://github.com/CleverCloud/clever-components/blob/9ee2dea21ce69f429573b109f8f859c9492bcbe3/src/components/cc-tile-instances/cc-tile-instances.js#L55-L119

I'm not sure it's a better solution. Feel free to use it, rework it or dismiss it :wink:

src/components/common.types.d.ts
pdesoyres-cc: I can't remove it in this PR because it is used by Application type which needs to be removed in #919 .
We need to merge #919 and when rebasing on master, I will remove this interface.
Thanks for the reminder.

pdesoyres-cc:
@hsablonniere I like your proposal.

Galimede (approved):
LGTM, GG!

HeleneAmouzou (approved):

roberttran-cc (approved):

PR #929, opened 2023-12-22
refactor(cc-invoice)!: migrate to the new smart component design

### LINKED ISSUE #920: cc-invoice: Migrate to the new smart component design
This is part of the migration effort settled in #708

pdesoyres-cc:
Fixes #920

### What is this?

* new state design: no impossible state
* new smart: no RxJs
* add demo in demo-smart

### How to review

* check code
* check for regression

### DIFF
--- a/src/components/cc-invoice/cc-invoice.js
+++ b/src/components/cc-invoice/cc-invoice.js
@@ -1,8 +1,8 @@
-import { css, html, LitElement } from 'lit';
 import '../cc-block/cc-block.js';
-import '../cc-notice/cc-notice.js';
 import '../cc-html-frame/cc-html-frame.js';
 import '../cc-icon/cc-icon.js';
+import '../cc-notice/cc-notice.js';
+import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 import { iconRemixFileTextLine as iconFile } from '../../assets/cc-remix.icons.js';
@@ -10,17 +10,16 @@ import { i18n } from '../../lib/i18n.js';
 import { skeletonStyles } from '../../styles/skeleton.js';
 import { ccLink, linkStyles } from '../../templates/cc-link/cc-link.js';
 
-/** @type {Invoice} */
+/** @type {Partial<InvoiceStateLoaded>} */
 const SKELETON_INVOICE = {
   emissionDate: '2020-01-01',
   number: '????????????',
-  type: 'INVOICE',
   status: 'PENDING',
-  total: { currency: 'EUR', amount: 10.00 },
+  amount: 10.00,
 };
 
 /**
- * @typedef {import('../common.types.js').Invoice} Invoice
+ * @typedef {import('./cc-invoice.types.js').InvoiceState} InvoiceState
  */
 
 /**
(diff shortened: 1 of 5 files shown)


roberttran-cc (approved):

src/components/cc-invoice/cc-invoice.js
```diff
@@ -32,49 +32,64 @@ export class CcInvoice extends LitElement {
...
-
-    const number = this.number ?? SKELETON_INVOICE.number;
-    const date = invoice.emissionDate;
-    const amount = invoice.total.amount;
+    const invoiceNumber = this._getInvoiceNumber();
+    const skeletonNumber = invoiceNumber == null;
+    const number = skeletonNumber ? SKELETON_INVOICE.number : invoiceNumber;
+    const skeleton = skeletonNumber || (this.invoice.state === 'loading');
+    const invoice = skeleton ? SKELETON_INVOICE : this.invoice.invoice;
+    const title = `${i18n('cc-invoice.title')} ${number}`;
```
florian-sanders-cc, line 66: **suggestion:** I think we could simplify a little bit by doing this + some modifications within the HTML (see my next comment)
```suggestion
    const invoiceNumber = this._getInvoiceNumber();
    const skeleton = (this.invoice.state === 'loading');
    const skeletonNumber = invoiceNumber == null;
    const number = skeletonNumber ? SKELETON_INVOICE.number : invoiceNumber;
    const invoice = skeleton ? SKELETON_INVOICE : this.invoice.invoice;
```

src/components/cc-invoice/cc-invoice.js
```diff
@@ -32,49 +32,64 @@ export class CcInvoice extends LitElement {
...
-            `) : ''}
-          </cc-html-frame>
-        ` : ''}
-        ${this.error ? html`
+      <cc-block .icon=${iconFile} class=${classMap({ 'has-errors': this.invoice.state === 'error' })}>
+        <div slot="title">
+          ${skeletonNumber ? html`
+            ${i18n('cc-invoice.title')} <span class="skeleton">${number}</span>
+          ` : title}
+        </div>
```
florian-sanders-cc, line 74: **suggestion:** 
```suggestion
        <div slot="title">
          ${i18n('cc-invoice.title')}
          <span class=${classMap({ skeleton: skeletonNumber })}>
            ${number}
          </span>
        </div>
```

florian-sanders-cc (changes requested):
I have a small suggestion and a question:
I feel like the `invoice number` thing is adding a bit of complexity / code just to make sure it is displayed when in skeleton mode as long as it has a value.
I get why you want it to be part of the state so I wonder if we could decide to just always put the number in skeleton mode when `state === loading`  even if we already have its value?

I'm not sure to be honest, it could just as well remain like that :+1:

src/components/cc-invoice/cc-invoice.js
pdesoyres-cc: Thank you for your suggestion, but we need the `title` for the iframe title. So in fact, this will not simply things so much.

pdesoyres-cc:
I think I will leave it like that. I don't want to change the initial behavior.

The error state needs the number to display the appropriate title so there is also a complexity in the error state that I can't tackle.

src/components/cc-invoice/cc-invoice.types.d.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+interface InvoiceStateError {
+  state: 'error';
+  number: string;
+}
+
+interface InvoiceStateLoaded {
+  state: 'loaded';
+  invoice: Invoice;
```
hsablonniere, line 17: suggestion: in the following situations, we avoided the `invoice.invoice` by naming the property `value` but it was for lists. I think we should rename `invoice` to `value`. WDYT?

* src/components/cc-pricing-header/cc-pricing-header.types.d.ts
  * `PricingZoneStateLoaded` with `value: Zone[]`
* src/components/cc-email-list/cc-email-list.types.d.ts
  * `EmailListStateLoaded` with `value: EmailList`
* src/components/cc-tcp-redirection-form/cc-tcp-redirection-form.types.d.ts
  * `TcpRedirectionFormStateLoaded` with `value: TcpRedirectionState[]`

hsablonniere (changes requested):
Well done on this one @pdesoyres-cc :clap: 

I made a suggestion on the naming of the type.

I noticed that we won't ever need to use this component without knowing the invoice number but it seems cleaner to handle it anyway like you did :+1:

hsablonniere:
FYI: I had similar concerns as @roberttran-cc regarding the invoice number but I came to the conclusion that we really need to display the invoice number as soon as possible.

HeleneAmouzou (approved):

src/components/cc-invoice/cc-invoice.js
```diff
@@ -32,49 +31,64 @@ export class CcInvoice extends LitElement {
...
-
-    /** @type {Invoice|null} Sets the invoice. */
-    this.invoice = null;
+    /** @type {InvoiceState} Sets the invoice state. */
+    this.state = { type: 'loading' };
+  }
 
-    /** @type {string|null} Sets the invoice number. */
-    this.number = null;
+  _getInvoiceNumber () {
```
Galimede, line 45: nitpick: I guess we can remove this now?

src/components/cc-invoice/cc-invoice.types.d.ts
```diff
@@ -0,0 +1,20 @@
+export type InvoiceState = InvoiceStateLoading | InvoiceStateError | InvoiceStateLoaded;
+
+interface InvoiceStateLoading {
+  type: 'loading';
+  number?: string;
+}
+
+interface InvoiceStateError {
+  type: 'error';
+  number: string;
```
Galimede, line 10: question: is there a reason to not be optional here while it is on `loading`?

Galimede (changes requested):
Thanks for the changes Pierre, it's more clear now! :muscle: 

I've just left a tiny nitpick otherwise that's great, GG!

src/components/cc-invoice/cc-invoice.types.d.ts
pdesoyres-cc: Yes, the error comes from remote call which needs the number. So, there is always a number.

roberttran-cc (approved):

PR #931, opened 2023-12-22
refactor(cc-article-list)!: migrate to the new smart component design

### LINKED ISSUE #921: cc-article-list: Migrate to the new smart component design
This is part of the migration effort settled in #708

### LINKED ISSUE #922: cc-invoice-list: Migrate to the new smart component design
This is part of the migration effort settled in #708

pdesoyres-cc:
Fixes #921 

## What is this?

* new state design: no impossible state
* new smart: no RxJs
* add demo in demo-smart

## How to review?

* check code
* check for regression

### DIFF
--- a/demo-smart/index.html
+++ b/demo-smart/index.html
@@ -56,9 +56,9 @@
   <li>
     <a class="definition-link" href="?definition=cc-addon-linked-apps.smart">cc-addon-linked-apps.smart</a>
   </li>
-  <!--  <li>-->
-  <!--    <a class="definition-link" href="?definition=cc-article-list.smart">cc-article-list.smart</a>-->
-  <!--  </li>-->
+  <li>
+    <a class="definition-link" href="?definition=cc-article-list.smart">cc-article-list.smart</a>
+  </li>
   <!--  <li>-->
   <!--    <a class="definition-link" href="?definition=cc-env-var-form.smart-config-provider&context=config-provider">cc-env-var-form.smart-config-provider</a>-->
   <!--  </li>-->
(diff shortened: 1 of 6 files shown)


roberttran-cc (approved):

src/components/cc-article-list/cc-article-list.types.d.ts
```diff
@@ -1,4 +1,19 @@
...
+  state: 'loading';
+}
+
+interface ArticleListStateError {
+  state: 'error';
+}
+
+interface ArticleListStateLoaded {
+  state: 'loaded';
+  articles: Array<Article>;
```
florian-sanders-cc, line 13: **suggestion/question:** what do you think about changing `articles` to `data`? 
This would avoid `this.articles.articles` inside the component :eyes: 

It could also be applied to the `cc-invoice` PR.

florian-sanders-cc (approved):
Small question about property naming but apart from that it's all good :wink:

demo-smart/index.html
```diff
@@ -118,6 +118,7 @@
 <div class="context-buttons">
   <button data-context='{}'>null</button>
   <button data-context='{"apiConfig":{}}'>bad API config</button>
+  <button data-context='{"lang":"en", "limit": 5}'>article-liste</button>
```
florian-sanders-cc, line 121: **nitpick:** ```suggestion
  <button data-context='{"lang":"en", "limit": 5}'>article-list</button>
```

src/components/cc-article-list/cc-article-list.types.d.ts
pdesoyres-cc: I do agree that this is not very nice.
This is something that occurs very often when designing the new state. 
Maybe you're suggestion should be applied to all states.
We should maybe discuss that with the rest of the team.

pdesoyres-cc: Let's go for `value` like for other components

Galimede (changes requested):
Historically `limit` was optional. With how the smart works now it makes it no longer possible. While it's not much of an inconvenience here, (the console provides a limit anyway), can you change the doc `cc-article-list.md` to reflect it and so that the examples may work again?

HeleneAmouzou (approved):

roberttran-cc (approved):

PR #935, opened 2023-12-28
fix(cc-badge): use `iconA11yName` in `iconAccessibleName` getter/setter

### LINKED ISSUE #934: cc-badge: `iconAccessibleName` getter and setter point to `a11yName` instead of `iconA11yName`
## Context

The getter and setter in the `cc-badge` component both point to the `a11yName` property which does not exist in the `cc-badge` component.

They should point to `iconA11yName`.

florian-sanders-cc:
Fixes #934

## How to review?

- Reviewing the commit should be enough,


pdesoyres-cc (approved):
Good catch. LGTM

PR #940, opened 2024-01-11
build(cdn): remove generated icon files from rollup config

### LINKED ISSUE #898: cc-icons: tree shaking issue with the CDN build
## Context

Components that rely on the `cc-icon` seem to trigger the fetch of all icons instead of just the ones they are using.

In the CDN bundle, it looks like all icons are regrouped within a `cc-remix.icons.hash.js` file. 

We need to investigate and find a solution so that components only fetch icons they need.

roberttran-cc:
Fixes #898

# How to review

- Checkout branch
- Run `(export VERSION=12.0.0 && npm run cdn-release:build)` task
- Check the `dist-cdn/cc-remix.icons-{HASH}.js` file contents
- It should no more include all Remix icons and only include icons actually used in our Clever Components (43 icons)

# Related impacts

Also impacts the NPM build: remix icons are no more "officially" included. More precisely, they are still available but not at the project root. This behavior is intended.

# Reviewers

Two will be sufficient.

### DIFF
--- a/rollup/rollup-common.js
+++ b/rollup/rollup-common.js
@@ -77,7 +77,6 @@ export function minifyStylesheet (stylesheet) {
 export function getMainFiles () {
 
   const mainFilesPatterns = [
-    `${SOURCE_DIR}/assets/*.icons.js`,
     `${SOURCE_DIR}/components/**/*.js`,
     `${SOURCE_DIR}/lib/i18n.js`,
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (approved):
LGTM. Well Done @roberttran-cc

hsablonniere (approved):
Well done @roberttran-cc, I just tried it locally and it's perfect! :+1: :clap:

PR #952, opened 2024-02-05
Storybook/upgrade with vite

### LINKED ISSUE #887: storybook: migrate to 7.X using `@storybook/builder-vite`
## Context

In #856, we tried to migrate to storybook 7.X using the `Web storybook builder`. 

We are blocked because it does not support the `autodocs` feature yet.

We want to try migrating using the vite builder.

## What works

- everything related to storybook works. Most of the changes described in #856 also work with the vite builder. It makes sense because most of these changes are related to the fact we are now directly interacting with Storybook and the Vite ecosystem is very close to Rollup so our rollup plugins worked without any modifications for both build and dev.
- I had to change the `WDS` plugin that handles CEM generation to make it work with vite. Fairly simple, we rely on `resolve` and `load` lifecycle hooks as described here: https://vitejs.dev/guide/api-plugin#importing-a-virtual-file. Note that I did not really respect the conventions (imported file should follow this naming convention: `virtual:name`) because it would also mean using this system for the build which is not really necessary in our case.

## What doesn't work

- We cannot serve anything else than storybook. The vite builder is used as a middleware of the storybook CLI which relies on express. The vite builder only handles requests related to the iframe:
https://github.com/storybookjs/builder-vite/blob/main/packages/builder-vite/index.ts#L27

The `Web storybook builder` is also used as a middleware (this is the storybook builder way) but it handles all requests:
https://github.com/modernweb-dev/web/blob/master/packages/storybook-builder/src/index.ts#L121

I'm not 100% sure of all of this yet.

As a workaround, we could use another command to run a vite server when we want to serve our demos but it really sucks and it does not even work yet... (it tries to find an entry, goes in the test files and fails to resolve some dependencies :bomb:).

=> Managed to make it work thanks to the undocumented `middleware.js` solution (https://github.com/storybookjs/storybook/pull/435).

TODO:

- [x] investigate `runtime.js:4 Component not found in custom-elements.json: Doc` warning with doc stories. 
- [x] fix/test HMR i18n,
- [x] fix oauth for demo-smart,
- [x] clean dependencies,
- [x] move files where they should be.

florian-sanders-cc:
Fixes #887 

## What does this PR do?

- removes `@web/dev-server` & `@web/dev-server-storybook` (`storybook-prebuilt`),
- installs storybook 7.0 with the `storybook/builder-vite`,
- adapts the config to match our needs.

## How to review?

- Review each commit, and focus specifically on:
  - `chore(storybook): add markdown stories`,
  - `chore(storybook): implement smart auth for smart.md stories`.
- You may skip batch modifications, I'll deal with that,
- Run storybook locally and check:
  - random stories:
    - press "d" within a story and test that the `action` and the `controls` feature still work,
    - press "i" and check that the language switching feature still works,
    - use the controls above the story and check that they still work (especially the `size` control).
  - random docs (autogenerated & md docs),
  - `cc-orga-member-list - smart` story => should contain the component working just fine at the bottom of the docs,
  - demo-smart,
  - sandbox.
- Check random stories / docs on the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/storybook/upgrade-with-vite/index.html?path=/docs/readme--docs) as well.

I've left a few comments (on GitHub) with notes / questions :wink: 

## TODO Flo:

- [x] remove `cc-orga-member-list.smart.md/js` modifications once the review phase is over.
- [x] dispatch feedback edits inside commits.


.storybook/manager.js
```diff
@@ -1,5 +1,6 @@
 import { addons } from '@storybook/manager-api';
 import { create } from '@storybook/theming';
+import { enhanceStoryName } from '../src/stories/lib/story-names';
```
florian-sanders-cc, line 3: You may skip all component files (batch modifications) but please do review the `story-names` file :wink:

.storybook/main.js
```diff
@@ -25,6 +25,9 @@ const config = {
     name: "@storybook/web-components-vite",
     options: {},
   },
+  docs: {
+    autodocs: "tag",
```
florian-sanders-cc, line 38: **Question:** At the beginning, I could not set this to `true` because `controller` and `template` docs were not recognized as "autogenerated" docs.

Setting this to `true` would have generated an auto doc in addition to our markdown doc that is supposed to replace it.

Since I've managed to make Storybook recognize our docs as true docs that may replace `autogenerated` docs, we can now use `true` here instead of manually adding `tags: ['autodocs']` in every story default export.

Still, I'm not sure this is a good idea because maybe one day we'll create a story that requires no doc?

Feel free to discuss this here and / or ping me to discuss it in sync.

src/stories/lib/markdown-to-csf.js
```diff
@@ -0,0 +1,107 @@
...
+
+  const csfScript = `
+    import { MarkdownDocs } from '/src/stories/lib/markdown-docs.jsx';
+    import React from 'react';
+
+    export default {
+      title: '${title}',
+      parameters: {
+        docs: {
+          page: () => React.createElement(MarkdownDocs, { html: ${parsedHTML} }),
```
florian-sanders-cc, line 56: Using an external `jsx` component is not mandatory but it makes the file easier to read so that's the option I prefer.

If you want to revert this back to what we did before (see https://github.com/CleverCloud/clever-components/blob/master/src/stories/lib/markdown.cjs#L72), this should be doable so don't hesitate to say it here :wink:

src/stories/lib/markdown-docs.jsx
```diff
@@ -0,0 +1,13 @@
...
+// several docs rely on `cc-notice`
+import '../../components/cc-notice/cc-notice.js';
+
+export function MarkdownDocs ({ html }) {
+  const htmlContent = { __html: html };
+
+  return (
+    <div className="markdown-body sb-unstyled" dangerouslySetInnerHTML={htmlContent}></div>
+  );
+}
```
florian-sanders-cc, line 13: we could make this a class component but since this is react, I followed the conventions.

src/stories/lib/markdown-docs.jsx
```diff
@@ -0,0 +1,13 @@
+// we disable this rule because this file is only meant to be processed by Storybook. It is not part of our npm / CDN bundle.
+// eslint-disable-next-line import/no-extraneous-dependencies, no-unused-vars
+import React from 'react';
+// several docs rely on `cc-notice`
```
florian-sanders-cc, line 4: Bear in mind that this wasn't a problem before because our components were all loaded automatically when landing on any page of our storybook.

Now that stories are lazy loaded, components are too, so `cc-notice` is only available if you have already encountered a story where it is imported.

This is why we import it here.

.storybook/preview.js
```diff
@@ -27,6 +27,53 @@ const availableLanguages = Object
...
+          '📌 Architecture Decision Records',
+          '🧬 Atoms',
+          '🧬 Molecules',
+          '*',
+          '🔀 Mixins',
+          '🕹️ Controllers',
+          '♻️ Templates',
+        ]
+      },
+    },
```
florian-sanders-cc, line 76: as explained in sync, we are forced to use all of this inline and we cannot use a const or anything.
If you want to see what happens, you may try in local to place this in a variable and run storybook, it should trigger an error message during the startup.

src/components/cc-orga-member-list/cc-orga-member-list.smart.md
```diff
@@ -59,4 +61,7 @@ interface ApiConfig {
 }'>
     <cc-orga-member-list></cc-orga-member-list>
 <cc-smart-container>
-```
\ No newline at end of file
+```
+<cc-smart-container context='{ "ownerId": "orga_3547a882-d464-4c34-8168-add4b3e0c135" }'>
+  <cc-orga-member-list></cc-orga-member-list>
+</cc-smart-container>
```
florian-sanders-cc, line 67: TODO @florian-sanders-cc, this is only here for testing, remove once the review is done.

src/components/cc-orga-member-list/cc-orga-member-list.stories.js
```diff
@@ -1,4 +1,4 @@
-import './cc-orga-member-list.js';
+import './cc-orga-member-list.smart.js';
```
florian-sanders-cc, line 1: this is actually not necessary because the `injectAuthForSmartComponentsPlugin` already does the import.
This will be removed anyway.
TODO @florian-sanders-cc

sandbox/index.html
```diff
@@ -58,6 +58,20 @@
...
+        "lit": "../node_modules/lit/index.js",
+        "lit/": "../node_modules/lit/",
+        "@lit/reactive-element": "../node_modules/@lit/reactive-element/reactive-element.js",
+        "@lit/reactive-element/": "../node_modules/@lit/reactive-element/",
+        "lit-html": "../node_modules/lit-html/lit-html.js",
+        "lit-html/": "../node_modules/lit-html/",
+        "lit-element": "../node_modules/lit-element",
+        "lit-element/lit-element.js": "../node_modules/lit-element/lit-element.js"
+      }
+    }
```
florian-sanders-cc, line 73: **Question**: As discussed in sync, do we keep this or do we move the sandboxes somewhere inside `src/stories`?

We should probably also open an RFC to discuss how we want to be able to test our components from now on but I still need a quick solution / decision so we don't block this PR :thinking:

tsconfig-typecheck-jsdoc.json
```diff
@@ -16,6 +16,7 @@
     "noImplicitThis": true,
     "noPropertyAccessFromIndexSignature": false,
     "target": "ESNext",
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "skipLibCheck": true,
```
florian-sanders-cc, line 20: :sweat_smile: look at the commit desc for more info.
Anyway, we need to merge #948 before we merge this one.

TODO @florian-sanders-cc remove this commit once #948 is merged.

pdesoyres-cc (approved):
Impressive work!

.storybook/preview.js
```diff
@@ -27,6 +27,53 @@ const availableLanguages = Object
...
+            'Release',
+            'Browser support',
+            'Tools',
+            'Resources',
+          ],
+          '📌 Architecture Decision Records',
+          '🧬 Atoms',
+          '🧬 Molecules',
+          '*',
+          '🔀 Mixins',
```
hsablonniere, line 71: question: do we still have mixins?

.storybook/preview.js
```diff
@@ -27,6 +27,53 @@ const availableLanguages = Object
...
+            'Previews',
+            'Release',
+            'Browser support',
+            'Tools',
+            'Resources',
+          ],
+          '📌 Architecture Decision Records',
+          '🧬 Atoms',
+          '🧬 Molecules',
+          '*',
```
hsablonniere, line 70: suggestion: if we want to keep the same order, we should add the `Beta` section at the bottom

src/stories/lib/smart-auth-plugin.js
```diff
@@ -0,0 +1,61 @@
...
+            API_OAUTH_TOKEN: '${API_OAUTH_TOKEN}',
+            API_OAUTH_TOKEN_SECRET: '${API_OAUTH_TOKEN_SECRET}',
+            OAUTH_CONSUMER_KEY: '${OAUTH_CONSUMER_KEY}',
+            OAUTH_CONSUMER_SECRET: '${OAUTH_CONSUMER_SECRET}',
+          },
+        });
+      `;
+
+      return code;
+    }
```
hsablonniere, line 59: suggestion: can we factorize this a bit and simplify the code? The only difference seems to be Warp10 and (for now) both could/should use it.

hsablonniere (dismissed):
Wow, impressive work, congrats @florian-sanders-cc :clap:

* Kudos for the attention to the explanations in comments everywhere
* So nice to say goodbye to all those manual `enhanceStoryName`
* The default font inside the components is not the same as before
  * It's `Nunito Sans`, why this change?
  * It's a bit too thin, WDYT?
* The beta ribbon on stories is a bit weird
  * It's padded in preview canvas
  * It goes over the component in the Docs page (it was already the case before)
* The beta section in the menu used to be at the bottom before controllers
* We used to have a CC favicon, it's a Storybook favicon now
* Can you add the `tags: ['autodocs'],` to the example file?
* The `/demo-smart/` page has some component that don't work

src/components/cc-logs-instances/cc-logs-instances.stories.js
```diff
@@ -22,6 +23,7 @@ function getDeploymentDate (sinceDay) {
 }
 
 export default {
+  tags: ['autodocs'],
```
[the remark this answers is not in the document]
florian-sanders-cc: About your first question, as discussed in sync it's because I forgot to update newly added files from the last rebase.

About your second, you are right, nice catch :+1:

.storybook/preview-head.html
```diff
@@ -1,77 +1,91 @@
...
-  font-size: 13px;
-  font-weight: bold;
-  margin: 2rem -1rem 1rem;
-  padding-bottom: 0.25rem;
-  padding-left: 1rem;
-}
+  /* Story viewer in docs mode */
+  .docs-story .story-shadow-container {
+    box-shadow: none;
+  }
```
florian-sanders-cc, line 48: Basically, we wanted to have some padding so that `beta` ribbons would overlap too much with the component in auto-generated docs.
Before (in prod)
image
Now 
image

roberttran-cc (approved):

.storybook/main.js
```diff
@@ -35,6 +35,9 @@ const config = {
     options: {},
   },
   docs: {
+    // This makes Storybook auto-generate docs for every story with the `autodocs` tag
```
pdesoyres-cc, line 38: :praise: This is perfect

pdesoyres-cc (approved):
I left some comments. We are almost there. Thanks again to you for this huge work!

docs/cc-example-component.stories.js
```diff
@@ -4,6 +4,8 @@ import { makeStory, storyWait } from '../stories/lib/make-story.js';
 import { enhanceStoriesNames } from '../stories/lib/story-names.js';
```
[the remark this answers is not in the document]
florian-sanders-cc: Indeed, you had already mentioned it and Bob as well but it looks like I forgot about it.
Should be good now :+1: 
Thanks

pdesoyres-cc (approved):
LGTM

PR #954, opened 2024-02-15
chore: remove RxJS dependency

pdesoyres-cc:


hsablonniere (approved):
Si is so nice!!! Thank you @pdesoyres-cc :+1:

florian-sanders-cc (approved):
Gj, well done for updating the docs too :muscle:

PR #957, opened 2024-02-20
feat: add error event from `sendToApi` calls

### LINKED ISSUE #956: add an event when an error is catched from `sendToApi` calls
Currently, when there's an error from an API call that we make from our components ,there's no way to catch that error from outside. We should dispatch an event with the error.

Galimede:
Fixes #956


florian-sanders-cc (approved):
LGTM! Gg @Galimede :+1:

pdesoyres-cc (approved):
LGTM. Well done @Galimede

hsablonniere (approved):
LGTM :+1: Well done @Galimede :clap:

PR #958, opened 2024-02-21
feat(cc-product-list): init component

### LINKED ISSUE #892: cc-product-list: init component
## Context

With our current work on a new creation tunnel for our console, we have decided to split it in several steps.
This component will be part of the first step which consists of selecting the product you want to create.
This component will contains a list of products (`cc-product-card`). You should be able to filter them with a search bar and possibly some filters. In the future, you should be able to have you favorite products and some suggestions on this list.

### Attributes/properties

| `categories`    | `Array<string>` | `[]`        | `true`    |  List of product categories        |
| `products`    | `Array<Product>` | `[]`        | `true`    |  List of products        |
## Outputs

| `cc-product:select` | `String`    |  For analytics purpose        |

## Type Definitions

```typescript
interface Product {
  name:  string;
  description: string;
  icon: string;
  keywords: Array<Keyword>;
  category: string;
}
```

```typescript
interface Keyword {
  value: string;
  hidden: boolean;
}
```

## Data/APIs

* `GET api.clever-cloud.com/v2/products/instances`
* `GET api.clever-cloud.com/v2/products/addonproviders`

## Stories & states

* Skeleton/loading
* Loaded
* Error
* Simulation

## UX writing & i18n

TBD

Example from prototype

image

Galimede:
Fixes #892

* This PR init a component that handles the step 1 of the future creation tunnel.

## How to review?
 * Check the overall design
 *  Check the code

### DIFF
--- a/src/components/cc-product-list/cc-product-list.js
+++ b/src/components/cc-product-list/cc-product-list.js
@@ -1,0 +1,234 @@
+import { css, html, LitElement } from 'lit';
+import { i18n } from '../../lib/i18n.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { linkStyles } from '../../templates/cc-link/cc-link.js';
+import '../cc-product-card/cc-product-card.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-input-text/cc-input-text.js';
+import { ProductsController } from './products-controller.js';
+
+/**
+ * @typedef {import('lit').PropertyValues<CcProductList>} CcProductListPropertyValues
+ * @typedef {import('./cc-product-list.types.js').ProductsCategory} ProductsCategory
+ * @typedef {import('lit').TemplateResult<1>} TemplateResult
+ */
+
+/**
+ * A component that displays a list of products in sections and allows to search and filter through it.
+ *
+ * @cssdisplay block
+ */
+export class CcProductList extends LitElement {
+
+  static get properties () {
+    return {
+      categoryFilter: { type: String, attribute: 'category-filter' },
+      productsByCategories: { type: Array, attribute: 'products-by-categories' },
+      textFilter: { type: String, attribute: 'text-filter' },
+    };
+  }
+
+  constructor () {
+    super();
+
+    /** @type {string|null} a string to prefilter by a given category */
+    this.categoryFilter = null;
+
+    /** @type {ProductsCategory[]} the list of products in their respective categories.  */
+    this.productsByCategories = [];
+
+    /** @type {string} a string to presearch through the list */
+    this.textFilter = '';
+
+    /** @type {ProductsController} */
+    this._productsCrtl = new ProductsController(this);
+  }
+
+  /**
+   * @param {Event & { target: HTMLInputElement }} e
+   */
+  _onCategoryChange (e) {
+    this.categoryFilter = e.target.value;
+  }
+
+  /**
+   * @param {Object} event
+   * @param {string} event.detail
+  */
+  _onSearchInput ({ detail: value }) {
+    this.textFilter = value;
+  }
+
+  /**
+   * @param {CcProductListPropertyValues} changedProperties
+  */
+  willUpdate (changedProperties) {
+
+    if (changedProperties.has('productsByCategories')) {
+      this._productsCrtl.productsByCategories = this.productsByCategories;
+    }
+
+    if (changedProperties.has('categoryFilter')) {
+      this._productsCrtl.toggleCategoryFilter(this.categoryFilter);
+    }
+    if (changedProperties.has('textFilter')) {
+      this._productsCrtl.textFilter = this.textFilter;
+    }
+  }
+
+  render () {
+
+    const categories = this._productsCrtl.getCategories();
+
+    return html`
+      <div class="search-form">
+        <cc-input-text
+          label="${i18n('cc-product-list.search-label')}"
+          value="${this.textFilter ?? ''}"
+          @cc-input-text:input="${this._onSearchInput}"></cc-input-text>
+        <fieldset class="category-filter">
+          <legend class="visually-hidden">${i18n('cc-product-list.filter-category-legend')}</legend>
+          ${this._renderCategory(i18n('cc-product-list.all-label'), 'all', this._productsCrtl.getCurrentCategory() === 'all')}
+          ${categories.map((c) => this._renderCategory(c.categoryName, c.categoryName, c.toggled))}
+        </fieldset>
+      </div>
+      <div class="products">
+        ${this._renderProductsByCategories()}
+      </div>
+    `;
+  }
+
+  /**
+   * @param {string} label
+   * @param {string} value
+   * @param {boolean} isToggled
+   * @returns {TemplateResult}
+   */
+  _renderCategory (label, value, isToggled) {
+    const id = label.replace(' ', '-');
+    return html`
+      <input 
+        type="radio"
+        id="${id}"
+        .value=${value}
+        name="category-filter"
+        class="visually-hidden" 
+        @change=${this._onCategoryChange}>
+      <label for="${id}">
+        <cc-badge
+          weight="${(isToggled ? 'dimmed' : 'outlined')}"
+          intent="info"
+        >${label}
+        </cc-badge>
+      </label>
+    `;
+  }
+
+  _renderProductsByCategories () {
+
+    const productsByCategories = this._productsCrtl.getFilteredProductsByCategories();
+
+    if (productsByCategories.length === 0) {
+      return html`
+        <p class="search-empty">${i18n('cc-product-list.search-empty')}</p>
+      `;
+    }
+
+    return productsByCategories.map((productsCategory) => html`
+      <div class="category">
+        <div class="category-title">
+          ${productsCategory.icon != null ? html`
+            <cc-icon .icon="${productsCategory.icon}" class="category-icon"></cc-icon>
+          ` : ''}
+          <div class="category-name">${productsCategory.categoryName}</div>
+        </div>
+        <div class="category-products">
+          ${productsCategory.products.map((p) => html`
+            <cc-product-card
+              name="${p.name}"
+              description="${p.description}"
+              .keywords="${p.keywords ?? []}"
+              icon-url="${p.iconUrl}"
+              url="${p.url}"
+            ></cc-product-card>
+          `)}
+        </div>
+      </div>
+    `);
+  }
+
+  static get styles () {
+    return [
+      accessibilityStyles,
+      linkStyles,
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+
+        .category-name {
+          font-size: 1.75em;
+          font-weight: bold;
+        }
+
+        .category-products {
+          display: grid;
+          gap: 1em;
+          grid-template-columns: repeat(auto-fill, minmax(20em, 1fr));
+        }
+
+        .category-title {
+          display: flex;
+          align-items: center;
+          margin: 2.5em 0 1.5em;
+          gap: 0.5em;
+        }
+
+        .category-filter {
+          display: flex;
+          flex-wrap: wrap;
+          padding: 0;
+          border: none;
+          gap: 0.25em;
+        }
+
+        label:hover {
+          cursor: pointer;
+
+          --cc-color-bg-primary: var(--cc-color-text-primary-strong);
+        }
+
+        input[type='radio']:focus-visible + label {
+          /**
+          * FIXME:
+          * remove this once the "border-radius" of the "cc-badge" component is set on its host instead of its wrapper
+          * see https://github.com/CleverCloud/clever-components/issues/990.
+          */
+          border-radius: 1em;
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: var(--cc-focus-outline-offset, 2px);
+        }
+
+        .category-icon {
+          width: 1.75em;
+          height: 1.75em;
+        }
+
+        .search-empty {
+          margin: 1.5em 0 0;
+          font-style: italic;
+        }
+
+        .search-form {
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-product-list', CcProductList);
--- a/src/components/cc-product-list/cc-product-list.stories.js
+++ b/src/components/cc-product-list/cc-product-list.stories.js
@@ -1,0 +1,652 @@
+import './cc-product-list.js';
+import { makeStory } from '../../stories/lib/make-story.js';
+
+export default {
+  tags: ['autodocs'],
+  title: '🛠 Creation Tunnel/<cc-product-list>',
+  component: 'cc-product-list',
+};
+
+const conf = {
+  component: 'cc-product-list',
+};
+
+const PRODUCTS = [
+  {
+    categoryName: 'Applications',
+    products: [
+      {
+        name: 'JavaScript',
+        description: 'Deploy your JavaScript/TypeScript and Meteor applications on Node.js with automatic dependency management (via npm ou yarn).',
+        keywords: [
+          {
+            value: 'Node.js',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/nodejs.svg',
+        url: '',
+        apiId: 'node',
+      },
+      {
+        name: 'PHP',
+        description: 'Deploy your PHP applications and static sites on Apache2 + PHP-FPM with automatic dependency management (via composer).',
+        keywords: [
+          {
+            value: 'Web',
+            hidden: false,
+          },
+          {
+            value: 'Symfony',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/03/php-4.svg',
+        url: '',
+        apiId: 'php',
+      },
+      {
+        name: 'Python',
+        description: 'Deploy your Python and Django applications on Nginx + WSGI with automatic dependency management (via pip) and Celery tasks support.',
+        keywords: [
+          {
+            value: 'Flask',
+            hidden: false,
+          },
+          {
+            value: 'Django',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/python.svg',
+        url: '',
+        apiId: 'python',
+      },
+      {
+        name: 'Ruby',
+        description: 'Deploy your Ruby applications on Nginx + Puma with automatic dependency management (via Rake, gem...) and Sidekiq tasks support.',
+        keywords: [
+          {
+            value: 'On Rails',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/ruby-2.svg',
+        url: '',
+        apiId: 'ruby',
+      },
+      {
+        name: 'Go',
+        description: 'Deploy your Go applications with automatic dependency management (via go mod, go build...).',
+        keywords: [
+          {
+            value: 'Google',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/go-2.svg',
+        url: '',
+        apiId: 'go',
+      },
+      {
+        name: 'Static',
+        description: 'Deploy your PHP applications and static sites on Apache2 + PHP-FPM with automatic dependency management (via composer).',
+        keywords: [
+          {
+            value: 'Apache',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/static-apache-3-1.svg',
+        url: '',
+        apiId: 'static-apache',
+      },
+      {
+        name: 'Rust',
+        description: 'Deploy your Rust applications with automatic dependency management (via Cargo).',
+        keywords: [
+          {
+            value: 'Low level',
+            hidden: false,
+          },
+          {
+            value: 'crab',
+            hidden: true,
+          },
+          {
+            value: '🦀',
+            hidden: true,
+          },
+          {
+            value: 'Ferris',
+            hidden: true,
+          },
+        ],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/08/rust-2.svg',
+        url: '',
+        apiId: 'rust',
+      },
+      {
+        name: 'Scala + SBT',
+        description: 'Deploy your JVM based applications (Java, Scala, GraalVM...) with automatic build and dependency management (Maven, Gradle, SBT, Ant...).',
+        keywords: [
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/scala.svg',
+        url: '',
+        apiId: 'sbt',
+      },
+      {
+        name: 'Java Jar',
+        description: 'Deploy and run any Java ARchive file with automatic build. Some configuration files or environment variable are needed to specify the the JAR path.',
+        keywords: [
+          {
+            value: '☕',
+            hidden: true,
+          },
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/java-jar.svg',
+        url: '',
+        apiId: 'jar',
+      },
+      {
+        name: 'Java maven',
+        description: 'Deploy your yur Java Maven projects with automatic build and dependency management. Some configuration files or environment variable are needed.',
+        keywords: [
+          {
+            value: 'Java',
+            hidden: false,
+          },
+          {
+            value: '☕',
+            hidden: true,
+          },
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/maven.svg',
+        url: '',
+        apiId: 'maven',
+      },
+      {
+        name: 'Java War',
+        description: 'Deploy and run your JEE applications packaged as WAR or EAR with the container you want (Tomcat, Jetty, Glassfish, JBoss, Paraya…).',
+        keywords: [
+          {
+            value: '☕',
+            hidden: true,
+          },
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/java-war.svg',
+        url: '',
+        apiId: 'war',
+      },
+      {
+        name: 'Play Framework 1',
+        description: 'Deploy your Java application with Play framework in version 1. Play framework’s goal is to ease development while sticking with Java.',
+        keywords: [
+          {
+            value: 'Java',
+            hidden: false,
+          },
+          {
+            value: '☕',
+            hidden: true,
+          },
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/play1.svg',
+        url: '',
+        apiId: 'play1',
+      },
+      {
+        name: 'Play Framework 2',
+        description: 'Deploy your based applications with Play 2.0, a Java and Scala framework that integrates the components and APIs you need for modern development.',
+        keywords: [
+          {
+            value: 'Java',
+            hidden: false,
+          },
+          {
+            value: 'Scala',
+            hidden: false,
+          },
+          {
+            value: '☕',
+            hidden: true,
+          },
+          {
+            value: 'JVM',
+            hidden: false,
+          },
+        ],
+        iconUrl: 'https://assets.clever-cloud.com/logos/play2.svg',
+        url: '',
+        apiId: 'play2',
+      },
+      {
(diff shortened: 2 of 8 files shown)


src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,383 @@
...
+          },
+        ],
+      );
+    });
+
+    it('should return all the product if null is provided', function () {
+
+      const pl = [sqlProducts, fakeProducts];
+
+      productsListCrtl.productsList = pl;
```
hsablonniere, line 136: suggestion: you can assign it the array directly

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,128 @@
...
+    const regex = RegExp(input, 'i');
+
+    for (const productSection of plByCategory) {
+
+      /** @type {ProductSection} */
+      const filteredProductSection = { categoryName: productSection.categoryName, products: [] };
+
+      if (productSection.icon != null) {
+        filteredProductSection.icon = productSection.icon;
+      }
```
hsablonniere, line 103: suggestion: if we add a new property on the Category type, we'll also have to handle it, it would be a bit better if the filtering was copying anything without really knowing much.

hsablonniere:
@Galimede I don't know if it's because you marked it as draft but I forgot to send a message with my review so here it is:

Well done @Galimede, I added some comments/suggestions, especially on the tests. I would like to do some pair programming together to discuss and maybe apply them. WDYT?

Also, your test does not use the spy. You never check if the `requestUpdate` is called, if I remove the `this._host.requestUpdate()` from the implem the tests still pass.

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  getProductsList () {
+    return this._currentProductsList;
+  }
+
+  /**
+   * @param {ProductSection[]} productsList
+   */
+  set productsList (productsList) {
```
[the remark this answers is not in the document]
Galimede: No, we don't have to do that. If the component gets a new `productsList`, the component will have to rerender. Before rendering again it will go in the `willUpdate` and the controller will be updated with the new `productsList` so there's no need to trigger an update again.

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+    const categories = this._productsCrtl.getCategories();
+
+    return categories.map((c) => html`
+      <cc-badge
+        weight="${(c.toggled ? 'dimmed' : 'outlined')}"
+        intent="info"
+        @click=${() => this._onBadgeClick(c.categoryName)}>
+        ${c.categoryName}
+      </cc-badge>
+    `);
```
pdesoyres-cc, line 105: nitpick: you're creating a new function at each render which is less efficient than doing this like that:

```suggestion
<cc-badge
        weight="${(c.toggled ? 'dimmed' : 'outlined')}"
        intent="info"
        @click=${this._onBadgeClick}>
        data-category="${c.categoryName}"
        ${c.categoryName}
      </cc-badge>
```

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  _renderProductsList () {
+
+    const categoryDataList = this._productsCrtl.getCategoryDataList();
+
+    return categoryDataList.map((categoryData) => html`
+      <div class="category">
+        <div class="category-title">
+          ${categoryData?.icon != null ? html`
```
pdesoyres-cc, line 115: `categoryData` cannot be null

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+          <div class="category-name">${categoryData.categoryName}</div>
+        </div>
+        <div class="category-products">
+          ${categoryData.products.map((p) => html`
+            <cc-product-card
+              name="${p.name}"
+              description="${p.description}"
+              .keywords="${p.keywords ?? []}"
+              icon-url="${p.iconUrl}"
+              url="${p?.url}"
```
pdesoyres-cc, line 127: nitpick: `p` cannot be null

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+
+    /** @type {ProductsController} */
+    this._productsCrtl = new ProductsController(this);
+  }
+
+  /**
+   * @param {string} categoryName
+   */
+  _onBadgeClick (categoryName) {
+    this._productsCrtl.filterCategory(categoryName);
```
pdesoyres-cc, line 50: question: Why not just setting the `filterCategory` property and the `willUpdate` will take care of applying the filter?

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * @param {Object} event
+   * @param {string} event.detail
+  */
+  _onSearchInput ({ detail: value }) {
+    this._productsCrtl.search(value);
```
pdesoyres-cc, line 58: question: Why not just setting the `filterInput` property and the `willUpdate` will take care of applying the filter?

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * @typedef {import('./cc-product-list.types.js').CategoryData} CategoryData
+ * @typedef {import('./cc-product-list').CcProductList} CcProductList
+ * @typedef {import('./cc-product-list.types.js').Product} Product
+ * @typedef {import('./cc-product-list.types.js').CategoryFilter} CategoryFilter
+ */
+
+export class ProductsController {
```
pdesoyres-cc, line 8: nitpick: rearrange the method order (public methods above)

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  getCategories () {
+    return this._categoriesFilters;
+  }
+
+  /**
+   * @param {string} categoryName
+   */
+  filterCategory (categoryName) {
```
pdesoyres-cc, line 52: nitpick: it's not easy to see that this is a toggle. Maybe think about renaming the method? Or add some comments?

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,132 @@
...
+  _getProductsByCurrentCategory () {
+    return (this._currentCategoryNameFilter !== 'all')
+      ? this._categoryDataList.filter(({ categoryName }) => categoryName === this._currentCategoryNameFilter)
+      : this._categoryDataList;
+  }
+
+  /**
+   * @param {string|null} searchInput
+   */
+  search (searchInput) {
```
pdesoyres-cc, line 107: nitpick: cc-product-lit component is using `filterInput`. Here we have `search` and `searchInput`. For more clarity, maybe we should align the naming.

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,132 @@
...
+   */
+  set categoryDataList (categoryDataList) {
+
+    this._categoryDataList = categoryDataList;
+    this._currentCategoryDataList = this._categoryDataList;
+    this._categoriesFilters = this._categoryDataList.map(({ categoryName }) => ({ categoryName, toggled: false }));
+    this._currentCategoryNameFilter = 'all';
+  }
+
+  getCategories () {
```
pdesoyres-cc, line 45: nitpick: Maybe rename that into `getCategoryFilters()`

src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,1483 @@
...
+  return {
+    name,
+    description,
+    iconUrl: 'https://example.com/icon.png',
+    url: 'https://example.com',
+    keywords,
+  };
+}
+
+describe('', function () {
```
pdesoyres-cc, line 29:
```suggestion
describe('ProductsController', function () {
```

src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,1483 @@
...
+  let requestUpdateSpy;
+
+  /** @type {ProductsController} */
+  let productsCrtl;
+
+  beforeEach(() => {
+    requestUpdateSpy = hanbi.spy();
+
+    // @ts-expect-error - TypeScript wants a CcProductList however it's not relevant here.
+    productsCrtl = new ProductsController({
```
pdesoyres-cc, line 41:
```suggestion
    productsCtrl = new ProductsController({
```

src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,1483 @@
...
+          categoryName: 'aaa',
+          products: [
+            generateProduct({ name: 'one' }),
+            generateProduct({ name: 'two' }),
+            generateProduct({ name: 'three' }),
+          ],
+        },
+      ];
+
+      productsCrtl.search('one  three');
```
pdesoyres-cc, line 186:
```suggestion
      productsCrtl.search('  one  three  ');
```

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  _getProductsByCurrentCategory () {
+    return (this._currentCategoryNameFilter !== 'all')
+      ? this._categoryDataList.filter(({ categoryName }) => categoryName === this._currentCategoryNameFilter)
+      : this._categoryDataList;
+  }
+
+  /**
+   * @param {string|null} searchInput
```
pdesoyres-cc, line 105: Add a comment warning that if a filter by category is also to be applied, this method should be called after calling `filterCategory()`.

pdesoyres-cc (commented):
Well done Mathieu. I left some few comments and nipicks.

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,187 @@
...
+    super();
+
+    /** @type {CategoryData[]} the list of products in their respective categories.  */
+    this.categoryDataList = [];
+
+    /** @type {string} a string to presearch through the list */
+    this.filterInput = '';
+
+    /** @type {'all'|string} a string to prefilter by a given category */
+    this.filterCategory = 'all';
```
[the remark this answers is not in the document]
Galimede: I thought it was more explicit this way. We could switch to null, but I also want to have the thoughts from the rest of the team about this.

Galimede:
Note: I haven't made any modifications to the naming yet. I'll wait for more feedback on this matter before doing any naming change.

src/translations/translations.en.js
```diff
@@ -784,6 +784,9 @@ export const translations = {
   //#region cc-product-card
   'cc-product-card.select': (name) => `${name} - select this product`,
   //#endregion
+  // #region cc-product-list
+  'cc-product-list.search-label': `Search a product`,
```
florian-sanders-cc, line 788: **fix:** without "for", it kinda means "to examine" which is not what you want here
```suggestion
  'cc-product-list.search-label': `Search for a product`,
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+export const filterInput = makeStory(conf, {
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterInput: 'cellar',
+    },
+  ],
+});
+
+export const unrelevantFilterInput = makeStory(conf, {
```
florian-sanders-cc, line 396: **nitpick:** 
```suggestion
export const irrelevantFilterInput = makeStory(conf, {
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterInput: 'unrelevant',
```
florian-sanders-cc, line 400: **nitpick:** 
```suggestion
      filterInput: 'irrelevant',
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+export const filterCategory = makeStory(conf, {
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'Storage',
+    },
+  ],
+});
+
+export const unrelevantFilterCategory = makeStory(conf, {
```
florian-sanders-cc, line 414: **nitpick:** 
```suggestion
export const irrelevantFilterCategory = makeStory(conf, {
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'unrelevant',
+    },
+  ],
+});
+export const unrelevantFilterCategoryButRelevantInput = makeStory(conf, {
```
florian-sanders-cc, line 422: **nitpick:** 
```suggestion
export const irrelevantFilterCategoryButRelevantInput = makeStory(conf, {
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'unrelevant',
```
florian-sanders-cc, line 426: **nitpick:** 
```suggestion
      filterCategory: 'irrelevant',
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'Storage',
+      filterInput: 'cellar',
+    },
+  ],
+});
+
+export const unrelevantFilterCategoryAndUnrelevantInput = makeStory(conf, {
```
florian-sanders-cc, line 442: **nitpick:** 
```suggestion
export const irrelevantFilterCategoryAndUnrelevantInput = makeStory(conf, {
```

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'unrelevant',
+      filterInput: 'unrelevant',
```
florian-sanders-cc, line 447: **nitpick:** 
```suggestion
      filterCategory: 'irrelevant',
      filterInput: 'irrelevant',
```

florian-sanders-cc (changes requested):
Great job @Galimede I don't have much to say about the code except an accessibility issue (however a fairly important one).
It's really clean and easy to read, well done!

### Feedback about visual/UX stuff

- We could increase the margin between headings and their content. I tried with `1.5em` bottom and `2.5em` top, felt good but it's just an opinion.
- We could use the same color as `cc-block` headings `var(--cc-color-text-primary-strongest)`. But maybe the next `cc-block` won't use that anymore? I don't know. I don't mind if it remains as it is, we lack common heading tokens/styles anyway (font-size, color, etc.) so it's a bigger subject and it shouldn't prevent us from going forward.
- We could make the "Search for a product" bigger now that we have merged #890. I havn't tested so I don't know how it would look but it feels like category headings are big so it would be okay to have a big label to introduce the filter section :thinking: 
- Heptapod logo is broken,
- Do we want a placeholder icons for products with no icons ?
- We could add a sentence when there are not results matching our filters :eyes: This message could even invite people to suggest new products to the support team like "blabla maybe there's a typo or this product is not part of our catalog yet. blabla feel free to contact us because we're always on the look for new products blabla" => this is way too long but you get the idea. Could be discussed with David / Marketing team? 
- This empty state could remind people of their filter criteria (no results for "sqql" in the "application" category" => this reminds people that maybe they wanted to search for "sql" in the "storage" category.
- The "php" logo is cropped.

### Accessibility stuff

- `onClick` events on `cc-badge` means category filters are inaccessible for keyboard and screen reader users. 
- Since category filters are mutually exclusive: we need the semantics to reflect that. The best option would be: radio buttons.
- But this is not possible with the current design: radio buttons cannot really "unselected" (well, they can, but people don't expect it so it's not the best experience).
- We could go for buttons with `aria-pressed` attributes but this would not convey the "exclusive options" semantics.
- Actually, this is not just a semantics issue, even as a sighted mouse user, I did not expect category filters to be mutually exclusive (but this is highly subjective).
- Suggestion: add a "all" option and rely on radio buttons. This way everybody knows how to use the filters from the start: all for all products, one of the other options for one of the categories.
- We can discuss/code it in sync, even with the rest of the team if you feel it's necessary :wink: 
- We could add a live region `aria-live` to warn screen reader users about the search results :thinking: We should try the `aria-live="polite" and see how it sounds with screen readers (not very long to test, we can do it in sync).

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+        name: 'Haskell',
+        description: 'Deploy your Haskell applications with automatic dependency management (via Stack).',
+        keywords: [{ value: 'Functional', hidden: false }],
+        iconUrl: 'https://cdn.clever-cloud.com/uploads/2023/06/haskell.svg',
+        url: '',
+        apiId: 'haskell',
+      },
+      {
+        name: 'JavaScript',
+        description: 'Distributed, advanced, and high-performance messaging and event streaming system.',
```
roberttran-cc, line 101: **💭 thought:** My whole view of JS just changed! :smiley:

roberttran-cc (approved):
Looking good!

A few non-blocking feedback: the component is overall very well-made. And shootout to the test suite. :muscle:

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,450 @@
...
+        apiId: 'fs-bucket',
+      },
+    ],
+  },
+  {
+    categoryName: 'SaaS',
+    icon: saasIcon,
+    products: [
+      {
+        name: 'Jenkins',
```
HeleneAmouzou, line 339: **question**: is it normal that Jenkins has no description ?

HeleneAmouzou (approved):
LGTM ! I've just left one non-blocking question ! :clap:

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        label:hover {
+          cursor: pointer;
+
+          --cc-color-bg-primary: var(--cc-color-text-primary-strong);
+        }
+
+        input[type='radio']:focus-visible + label {
+          /* TODO: change this if we add a custom prop to the cc-badge */
```
florian-sanders-cc, line 206: **suggestion:** 
```suggestion
          /* 
           * FIXME:
           * remove this once the border-radius of the `cc-badge` component is set on its host instead of its wrapper
           * see https://github.com/CleverCloud/clever-components/issues/990.
           */
```

src/translations/translations.fr.js
```diff
@@ -797,6 +797,12 @@ export const translations = {
   //#region cc-product-card
   'cc-product-card.select': (name) => `${name} - sélectionner ce produit`,
   //#endregion
+  //#region cc-product-list
+  'cc-product-list.all-label': `Tout`,
+  'cc-product-list.filter-category-legend': `Filtrer par catégorie`,
+  'cc-product-list.search-empty': `Aucun produit n'a été trouvé selon vos critères de recherche.`,
```
florian-sanders-cc, line 803: **suggestion:** ```suggestion
  'cc-product-list.search-empty': `Aucun produit ne correspond à vos critères de recherche.`,
```

**Review by florian-sanders-cc — APPROVED** [review]
I like the renaming suggestions from Pierre.
The rest looks good to me!
Thanks for your patience and your hard work on a11y :wink: 

The component looks & works great :sunglasses:

File: src/components/cc-product-list/cc-product-list.js [hunk 30]
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * A component that displays a list of products in sections and allows to search and filter through it.
+ *
+ * @cssdisplay block
+ */
+export class CcProductList extends LitElement {
+
+  static get properties () {
+    return {
+      categoryDataList: { type: Array },
```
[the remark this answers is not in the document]
hsablonniere: I had a `_categoriesWithProducts` but I think I like AND prefer @pdesoyres-cc suggestion :+1:

src/components/cc-product-list/cc-product-list.js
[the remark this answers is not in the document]
hsablonniere: Go for `null`. If we decided we can select multiple categories, it will be easier/cleaner.

src/components/cc-product-list/products-controller.js
hsablonniere: Agreed!

src/components/cc-product-list/products-controller.js
hsablonniere: I'm OK with both.

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,138 @@
...
+   */
+  constructor (host) {
+    /** @type {CcProductList} The host. */
+    this._host = host;
+
+    /** @type {CategoryData[]} The initial category data list. */
+    this._categoryDataList = [];
+
+    /** @type {CategoryData[]} category data list filtered by the category selected nor with the input */
+    this._currentCategoryDataList = [];
```
hsablonniere, line 21: nitpick: I was expecting something similar to `logs-controller` with `this._logs` and `this._logsFiltered` but I'll let you decided

src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,1483 @@
+/* eslint-env node, mocha */
```
hsablonniere, line 1: praise: the whole test suite is very impressive!!! :clap:

src/components/cc-product-list/products-controller.test.js
```diff
@@ -0,0 +1,1483 @@
...
+
+/**
+ * @param {Partial<Product>} product
+ * @returns {Product}
+ */
+function generateProduct ({
+  name = '',
+  description = '',
+  keywords = [],
+}) {
```
hsablonniere, line 19: question: is this ok to have a `@param` named `product` but it's directly destructured without being named?

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * @param {CcProductList} host
+   */
+  constructor (host) {
+    /** @type {CcProductList} The host. */
+    this._host = host;
+
+    /** @type {CategoryData[]} The initial category data list. */
+    this._categoryDataList = [];
```
hsablonniere, line 18: suggestion: I would go for a similar naming as the component, like @pdesoyres-cc suggested : `productsByCategories`

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,233 @@
...
+        name="category-filter"
+        class="visually-hidden" 
+        @change=${this._onCategoryChange}>
+      <label for="${id}">
+        <cc-badge
+          weight="${(isToggled ? 'dimmed' : 'outlined')}"
+          intent="info"
+        >${label}
+        </cc-badge>
+      </label>`;
```
hsablonniere, line 124: nitpick: the backtick should go on a new line

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,614 @@
...
+});
+
+export const filterInput = makeStory(conf, {
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterInput: 'cellar',
+    },
+  ],
+});
```
hsablonniere, line 558: issue: I think the story is outdated, the cellar product is no longer in the list

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,614 @@
...
+
+export const filterCategoryAndInput = makeStory(conf, {
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'Storage',
+      filterInput: 'cellar',
+    },
+  ],
+});
```
hsablonniere, line 604: issue: I think the story is outdated, the cellar product is no longer in the list, same for the storage category

src/components/cc-product-list/cc-product-list.stories.js
```diff
@@ -0,0 +1,614 @@
...
+
+export const filterCategory = makeStory(conf, {
+  items: [
+    {
+      categoryDataList: PRODUCTS,
+      filterCategory: 'Storage',
+    },
+  ],
+});
```
hsablonniere, line 576: issue: I think the story is outdated, the storage category is no longer in the list

hsablonniere (changes requested):
Hey @Galimede, congrats on this work!! :clap: 

I wrote a few notes and I agree with @pdesoyres-cc suggestion. I'll let you handle @florian-sanders-cc feedbacks but I really like this one:

> We could use the same color as cc-block headings var(--cc-color-text-primary-strongest). But maybe the next cc-block won't use that anymore? I don't know. I don't mind if it remains as it is, we lack common heading tokens/styles anyway (font-size, color, etc.) so it's a bigger subject and it shouldn't prevent us from going forward.

We should try with the cc-block blue title color.

## Bug

I noticed a bug in the storybook:

* type "pg" in input filter
  * no results
* click on "databases" category
  * all databases are displayed
* focus input filter
  * type a space and remove it
  * no results again

If it comes from the controller, make sure you create a unit test to reproduce the bug and then fix the bug to pass the test :wink:

src/components/cc-product-list/products-controller.test.js
Galimede: yes you can, TS will understand it. However, `Partial` is important here as we don't provide everything from the `Product` type.

florian-sanders-cc (approved):

src/components/cc-product-list/cc-product-list.js
```diff
@@ -0,0 +1,235 @@
...
+  willUpdate (changedProperties) {
+
+    if (changedProperties.has('productsByCategories')) {
+      this._productsCrtl.productsByCategories = this.productsByCategories;
+    }
+
+    if (changedProperties.has('filterCategory')) {
+      this._productsCrtl.toggleCategory(this.filterCategory);
+      this._productsCrtl.search(this.searchInput);
+    }
```
hsablonniere, line 75: issue: I was expecting the `this._productsCrtl` to remember the last `searchInput`.

This logic should be inside the ProductController. Whether you do search then toggle category or the other way around should work in the controller.

hsablonniere (changes requested):

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   *
+   * @param {ProductsByCategory[]} productsByCategories
+   * @param {string} textFilter
+   * @returns {ProductsByCategory[]|null}
+   */
+  _search (productsByCategories, textFilter) {
+    const textFilterFormatted = textFilter?.toLowerCase().trim();
+
+    if (textFilterFormatted == null || textFilterFormatted === '') {
```
pdesoyres-cc, line 135: You can use the `isStringEmpty` utility

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,148 @@
...
+   */
+  getFilteredProductsByCategories () {
+
+    const productsByCategoriesToggled = this._getProductsByCurrentCategory();
+    const productsByCategoriesTextFiltered = this._search(productsByCategoriesToggled, this._currentTextFilter);
+
+    // We either return:
+    // - the list of products filtered by the text and category
+    // - the list of products filtered only by the current category if the text filter is empty or null
+    return productsByCategoriesTextFiltered ?? productsByCategoriesToggled;
```
pdesoyres-cc, line 43: nitpick:
- the `_search` method could have no parameters.
- the `_search` method should have all that logic.
- So that you could simplify that by `return this._search()`
- Having said that, do we really need the internal _search method? All the code in the _search method could be inlined here since the _search is used only here.

pdesoyres-cc (approved):
Well done Mathieu. This is really nice. I just added two nitpicks, do what you want with them.

src/components/cc-product-list/products-controller.js
```diff
@@ -0,0 +1,148 @@
...
+    this._host = host;
+
+    /** @type {ProductsByCategory[]} The initial products category list. */
+    this._productsByCategories = [];
+
+    /** @type {CategoryFilter[]} Categories and their current state.  */
+    this._categoriesFilters = [];
+
+    /** @type {string|null} Current category selected.  */
+    this._currentCategoryNameFilter = null;
```
hsablonniere, line 24: praise: I was about to say "what about juste `currentCategoryFilter` but we would be confused about whether it's a category object or a string.

src/components/cc-product-list/products-controller.js
hsablonniere: :+1 Agreed with @pdesoyres-cc, let's remove `_search` and move the code here.

hsablonniere (changes requested):
Well done @Galimede :+1

I agreed with @pdesoyres-cc on a detail and added a small comment but you can handle them and merge without me :wink:

hsablonniere (approved):
LGTM :+1
Well done :tada

PR #969, opened 2024-02-28
fix(smart-component): protect from abortController nullity when smart component is disconnected

### LINKED ISSUE #968: smart-component: abort controller can be undefined when `onDisconnect` is called before `onContextUpdate`
The abort controller associated with a smart component is created when the `onContextUpdate` callback is called.
If the `onDisconnect` callback is called before that, we must ensure that the abort controller is not undefined before using it.

pdesoyres-cc:
Fixes #968


florian-sanders-cc (approved):
LGTM! Thanks for the fix :muscle:

roberttran-cc (approved):

Galimede (approved):
LGTM!

PR #976, opened 2024-03-05
Components/remove deprecated accessible name prop

florian-sanders-cc:
## Context

With the 12.0.0 release, we have decided to deprecate the `accessibleName` property & `accessible-name` attribute (see #894).

With the 13.0.0, we are now removing this prop / attribute.

This PR removes the prop / attribute from components.

## How to review?

- review every commit,
- 2 people should be enough for this one.

**Note**: I've checked the console project and there is no usage of `accessibleName` / `accessible-name`. This means that we don't need to adapt the console to this breaking change when integrating the `13.0.0` release @pdesoyres-cc :+1:

### DIFF
--- a/docs/adr/adr-0022-new-icon-system.md
+++ b/docs/adr/adr-0022-new-icon-system.md
@@ -189,18 +189,18 @@ Using `role="img"` and `aria-label` attributes on a custom element does not seem
 
 Since we could not rely on the custom element itself, we had to rely on the `<svg>` element and manipulate it depending on its context.
 
-If the developer does not provide any `accessible-name` prop, the component adds an `aria-hidden="true"` attribute to the inner `<svg>` element.
-If the developer does provide an `accessible-name` prop, the component adds to the inner `<svg>` element:
+If the developer does not provide any `a11y-name` prop, the component adds an `aria-hidden="true"` attribute to the inner `<svg>` element.
+If the developer does provide an `a11y-name` prop, the component adds to the inner `<svg>` element:
 
 * A `role="img"` attribute,
-* An `aria-label` which value matches the value of the `accessible-name` prop,
-* A `<title>` element which value matches the value of the `accessible-name` prop.
+* An `aria-label` which value matches the value of the `a11y-name` prop,
+* A `<title>` element which value matches the value of the `a11y-name` prop.
 
 The addition of the `<title>` element is not necessary but this is the native HTML way for informative `<svg>` elements so we decided to keep it.
 
 Example of an informative `<cc-icon>`:
 ```html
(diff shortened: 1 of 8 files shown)


pdesoyres-cc (approved):
Well done, thank you for taking that.

There is an inversion in 2 commit messages:
`af48d0a` is about `accessibleName` property while `b7aac49` is about `iconAccessibleName` property.

Galimede (approved):
LGTM, GG! :muscle: 

Just a few questions: 
* Not every component that has a `a11y-name` includes a note in their respective docs,  should we uniform it later on?
* Also `cc-icon` has a section title for its accessibility guidelines which is `**Accessibility guidelines:**` whereas in `cc-popover` it is `## Accessibility`. Maybe, we should decide which one we want to use. Or we can also decide to leave it to developer preference.

florian-sanders-cc:
@Galimede 

>     * Not every component that has a `a11y-name` includes a note in their respective docs,  should we uniform it later on?

Do you have an example? 
Every component has a doc associated to their `a11yName` property within the constructor which should be enough for most cases.
The docs are not always the same because the usecase and what the property actually generates differs depending on the inner HTML of the component (setting `a11yName` on `cc-icon` triggers the addition of different attributes compared to setting on `cc-img` for instance).

>     * Also `cc-icon` has a section title for its accessibility guidelines which is `**Accessibility guidelines:**` whereas in `cc-popover` it is `## Accessibility`. Maybe, we should decide which one we want to use. Or we can also decide to leave it to developer preference.

That's a good point, we could open a quick win issue about this but we'd also need to agree on what to use and write it in some of our docs for later :thinking:

florian-sanders-cc:
> @Galimede
> 
> > ```
> > * Not every component that has a `a11y-name` includes a note in their respective docs,  should we uniform it later on?
> > ```
> 
> Do you have an example? Every component has a doc associated to their `a11yName` property within the constructor which should be enough for most cases. The docs are not always the same because the usecase and what the property actually generates differs depending on the inner HTML of the component (setting `a11yName` on `cc-icon` triggers the addition of different attributes compared to setting on `cc-img` for instance).
> 
> > ```
> > * Also `cc-icon` has a section title for its accessibility guidelines which is `**Accessibility guidelines:**` whereas in `cc-popover` it is `## Accessibility`. Maybe, we should decide which one we want to use. Or we can also decide to leave it to developer preference.
> > ```
> 
> That's a good point, we could open a quick win issue about this but we'd also need to agree on what to use and write it in some of our docs for later 🤔

Discussed this with @Galimede in sync, we'll create an issue to standardize `Accessibility` sections for components when it's relevant.

PR #980, opened 2024-03-11
feat(cc-logs-application-view): init component

### LINKED ISSUE #964: cc-logs-control: add header slot

### LINKED ISSUE #965: cc-logs: do not collapse spaces in log messages

### LINKED ISSUE #967: cc-logs-application-view: init component
## Context

A component that displays logs for Clever applications.

### Attributes/properties

| `dateRangeSelection`    | `DateRangeSelection` | `null`        | `false`    | The date range selection         |
| `options`    | `LogsApplicationViewOptions` | `null`        | `false`    | TODO        |
| `overflowWatermarkOffset`    | `Number` | `10`        | `false`    | The number of logs before the limit from which the user will be asked to accept or discard logs limit overflow.        |
| `selectedInstances`    | `Array<string>` | `[]`        | `false`    | The array of instances ids that should be selected.    |
| `state`    | `LogsApplicationViewState` | `LogsApplicationViewStateLoadingInstances`  | `false`    | State of the component        |
| `limit`    | `Number` | `1000`        | `false`    | The maximum number of logs to display. `null` for no limit.        |

## Outputs

| `cc-logs-application-view:date-range-change` | `DateRange`    | Fires the `range` whenever the `range` changes. |
| `cc-logs-application-view:pause` | `DateRange`    | Fires when the pause button is clicked. |
| `cc-logs-application-view:resume` | `DateRange`    | Fires when the resume button is clicked. |

## Type Definitions

```typescript
export interface LogsApplicationViewStateLoadingInstances {
  type: 'loadingInstances';
}

export interface LogsApplicationViewStateErrorInstances {
  type: 'errorInstances';
}

export interface LogsApplicationViewStateLogs {
  type: 'connectingLogs' | 'receivingLogs'| 'logStreamPaused' | 'logStreamEnded' | 'errorLogs';
  instances: Array<Instance | GhostInstance>;
  selection: Array<string>;
}


export type LogsApplicationViewState = LogsApplicationViewStateLoadingInstances | LogsApplicationViewStateLogs | LogsApplicationViewStateErrorInstances;

export type DateRange = {
  since: string;
  until?: string;
}

export interface DateRangeSelectionLive {
  type: 'live',
}

export type DateRangeSelectionPredefinedDefinition = 'lastHour' | 'last4Hours' | 'last7Days' | 'today' | 'yesterday';

export interface DateRangeSelectionPredefined {
  type: 'predefined',
  def: DateRangeSelectionPredefinedDefinition,
}

export interface DateRangeSelectionCustom {
  type: 'custom',
  since: string;
  until: string;
}

export type DateRangeSelection = DateRangeSelectionLive | DateRangeSelectionPredefined | DateRangeSelectionCustom;

export type DateRangeSelectionMenuEntry = 'live' | 'custom' | DateRangeSelectionPredefinedDefinition;

export interface LogsApplicationViewOptions {
  'date-display': DateDisplay,
  'metadata-display': {
    instance: boolean
  },
  palette: LogsControlPalette,
  'strip-ansi': boolean,
  timezone: Timezone,
  'wrap-lines': boolean,
}

```

## Data/APIs

* `GET api.clever-cloud.com/v4/orchestration/organisations/${params.id}/applications/${params.appId}/instances`
* `GET api.clever-cloud.com/v4/orchestration/organisations/${params.id}/applications/${params.appId}/instances/${params.instanceId}`
* `GET api.clever-cloud.com/v4/orchestration/organisations/${params.id}/applications/${params.appId}/deployments/${params.deploymentId}`
* `GET api.clever-cloud.com/v2/organisations/${params.id}/applications/${params.appId}/deployments/${params.deploymentId}`
* `GET api.clever-cloud.com/v4/logs/organisations/${this._ownerId}/applications/${this._appId}/logs`

### LINKED ISSUE #983: cc-logs-application-view: add a fullscreen button mode
This button is displayed next to the options popover.

### LINKED ISSUE #1019: cc-logs-application-view: loading state is not visible when using dark theme
The loading state displays a `<cc-loader>` hover, the `<cc-logs>` component. When using a dark theme, the loading message is not visible.

pdesoyres-cc:
Fixes #967
Fixes #965
Fixes #964
Fixes #1019
Fixes #983


florian-sanders-cc (commented):
:raised_hands: The amount of work and its quality is outstanding

Thanks for splitting everything in separate commits, this helped a lot for the review.
Thanks for improving the typechecking of several common dependencies + improving the sandboxes :wink: 

First: I have 0 blocking comment for the release of the component in beta. All of my comments are either nitpicks or things that require some work and can be done later on.

Here are a few comments about UX and a11y in general:

## Date range section

### Date range type selection

We should replace the `cc-popover` with a component that implements the [ARIA Combobox Design Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/). 
This is no easy task so I suggest we do like we have done for the `cc-pricing-header` component and rely on the [Shoelace Dropdown component](https://shoelace.style/components/dropdown). 

Last time I checked, it was using the proper ARIA attributes + the right keyboard shortcuts. I cannot assert that it's 100% compliant (I would need time to test it and study its code) but it's the best option we have right now.

In the future, we may want to produce our own but I don't think we should go there right now. This is not up to me alone to decide so we can discuss it in team meeting?

### Arrow buttons

The arrow buttons cause several a11y issues. 
The main issue is that their function is not explicit. They lack tooltips but I would also change the feature to make it more obvious since we rely on icon only buttons: move the buttons right next to each input like the "-" / "+" on the `cc-input-number` and make it so they add / remove 1 hour to the input? 
Also, we would need to add live regions to the input themselves so that users changing the value with the buttons be aware of the result of their action.

## Loading logs section

### Section heading

The "loading logs" section is very nice when dealing with cold logs but I feel it's confusing when dealing with live logs. 

The `Loading logs` heading makes me wonder if I'm actually seeing the live logs or if it's missing a few :eyes:. For live logs, I would try to go for a different heading like "Live logs" + pause button? It's not the best heading but it's all I have for now :thinking:.

### Pause / resume buttons

#### button Labels
I would go for more explicit a11y labels for the pause and resume button: "Pause log fetching / resume log fetching (or loading I don't know)".

#### Focus management

We'll need to take care of the focus in case the user is focusing the "pause" button and it disappears (cold logs: logs loading => logs loaded).

### Limit notice

When reaching the limit of logs, the warning is a decent temporary way to mitigate the issue but we'll need to do better for a11y: the focus should be moved to the notice when it appears, there should be more explanation about what "continue" implies (limit to 10 000).

I feel it might be a decent usecase for a modal. This would solve the focus issue + would give us more space to explain stuff. WDYT?

## Filter feature

Should we add a message when no result matches our filter? I'm not sure, I think it's worth testing but it's okay like that also :thinking: 

## Scroll follow feature

I forgot to test that earlier but when you slide the scroll bar cursor to go up, the follow feature is not disabled.
My description is terrible so I'll try to upload a video. I think this is normal since we haven't considered this way of interacting with the scrollbar so it's not blocking but something to consider for the future?

## A Bug I've faced

1. I loaded the live logs for `test-logs-pds`
2. I switch to "7 last days"
3. I opened the options menu and then closed it
4. I selected the first instance to see only its logs

issue notice empty + logs

florian-sanders-cc:
Forgot to say:

## Logs instances

Back again about the instance indexes thing :joy: 
I've wrapped my head around the fact we show indexes and their use but in the long term, I think we could do better for our customers: the APIs and the UIs could rely on a concept of "primary" instance (which would correspond to the instance #0). 
This way we could add a badge like "primary" or "cron" or whatever next to the instance name. 
I think it would be a better solution to the "how do we tell people which instance may contain the cron logs?" because it would only show the information we need instead of showing information not relevant for users (indexes of the other instances).

Anyway, as I said, I've wrapped my mind around it, maybe people won't really be confused about this like I was but I just wanted to suggest another idea in the long term.

src/lib/define-smart-component.js
```diff
@@ -5,31 +5,37 @@ const META = Symbol('META');
...
  *   updateComponent?: (
  *     type: string,
  *     callbackOrObject: CallbackOrObject,
  *   ) => void,
  *   signal?: AbortSignal,
  * }) => void} definition.onContextUpdate
- * @param {AbortSignal} [signal]
  */
-export function defineSmartComponent (definition, signal) {
+export function defineSmartComponent (definition) {
```
hsablonniere, line 38: issue: the signal is actually needed

Let me explain, the idea is that you use `defineSmartComponentCore(definition, signal)` to define a smart component and you can pass it a signal as the second param. When this signal is aborted, the smart component definition is removed from the manager.

We never use it ourselves but it's important that such an action can be aborted/cancelled and it's in the smart manager test suite.

Therefore, `defineSmartComponent` should accept a signal and pass it as is to `defineSmartComponentCore`. I'm not sure we need to do this in this PR, as you want.

Also, since we have one signal to abort the smart definition and one for the component disconnection, we may need two names.

src/components/cc-logs/logs-controller.js
```diff
@@ -41,24 +43,9 @@ export class LogsController {
...
-    this._filterCallback = (log) => {
-      // NOTE: [].every() is always true, whatever the callback
-      return filterValuesByMetadataNameEntries.every(([metadata, values]) => {
-        const logMetadata = log.metadata.find((m) => m.name === metadata);
-        return values.includes(logMetadata?.value);
-      });
-    };
+    this._filterCallback = (filter == null)
+      ? truePredicate
+      : (log) => this._getMetadataFilterCallback(filter.metadata)(log) && this._getMessageFilterCallback(filter.message)(log);
```
hsablonniere, line 48: issue: it seems like we will call `_getMetadataFilterCallback` and `_getMessageFilterCallback` for each log, I'm not sure if it's a perf problem but I'd rather do:

```js
if (filter == null) {
  this._filterCallback = truePredicate;
}
else {
  const matchesMetadata = this._getMetadataFilterCallback(filter.metadata);
  const matchesMessage = this._getMessageFilterCallback(filter.message);
  this._filterCallback (log) => matchesMetadata(log) && matchesMessage(log);
}
```

hsablonniere: question: also, it seems like (not sure) checking the metadata is more costly so I would suggest to match on message first, WDYT?

src/components/cc-logs/cc-logs.js
```diff
@@ -698,8 +702,11 @@ export class CcLogs extends LitElement {
       this._logsCtrl.limit = this.limit;
     }
 
-    if (changedProperties.has('filter')) {
-      this._logsCtrl.filter = this.filter;
+    if (changedProperties.has('messageFilter') || changedProperties.has('metadataFilter')) {
+      this._logsCtrl.filter = {
+        message: this.messageFilter,
+        metadata: this.metadataFilter,
+      };
```
hsablonniere, line 709: question: curious to know why the controller does not have two params for the filters like the component

hsablonniere (changes requested):
Hey @pdesoyres-cc, well done !!

I haven't reviewed the last commit yet, but I'm submitting my review so you already have some feedbacks.

sandbox/index.js
```diff
@@ -1,11 +1,47 @@
...
+const languageToggle = document.createElement('cc-toggle');
+languageToggle.choices = [
+  {
+    label: `🇬🇧 ${translationsEn.LANGUAGE}`,
+    value: langEn,
+  },
+  {
+    label: `🇫🇷 ${translationsFr.LANGUAGE}`,
+    value: langFr,
+  },
```
hsablonniere, line 34: nitpick: country flags should never be used for language pickers :wink: 

https://wplang.org/never-use-flags-language-selection/

florian-sanders-cc:
I just realized I had not tested form errors within the "date range" section:

- since there is a submit button, we should stick with error messages on submit,
- we need to warn users about the expected format, even if the value is "prefilled". The expected format needs to remain visible at all time because some users may not remember it once they start typing. We could specify the format ("YYYY-DD...") within the "help" slot?

src/components/cc-logs-application-view/cc-logs-application-view.js
```diff
@@ -0,0 +1,1252 @@
...
+          <cc-notice slot="header"
+                     intent="info"
+                     heading=${i18n('cc-logs-application-view.logs.warning.no-logs.title')}
+                     message=${i18n('cc-logs-application-view.logs.warning.no-logs.message')}
+          ></cc-notice>
+        </div>
+      ` : ''}
+
+      
+
```
roberttran-cc, line 797: **🤓 nitpick:**

```suggestion
```

src/components/cc-logs-application-view/cc-logs-application-view.stories.js
```diff
@@ -0,0 +1,50 @@
...
+  return Array(count).fill(0).map((_, i) => logFactory(i));
+};
+
+export const defaultStory = makeStory(conf, {
+  items: [
+    {
+      logs: logs(100),
+    },
+  ],
+});
```
roberttran-cc, line 50: **🔨 issue:** The story does not seem to be working (infinite load) and the parameters seem off. That may be irrelevant for now, as the effort is toward a working version in the console. :thinking:

roberttran-cc (approved):
First of all, congrats for this PR. A special praise to the smart component: the way it handles so much complexity is really well-made.

As for my review, I'll admit that it was not very thorough (I only have a few non-blocking feedback). I played with it on the staging console and I didn't see any weird behavior or bug.

Again, great job! :fire:

src/components/cc-logs-application-view/cc-logs-application-view.js
```diff
@@ -0,0 +1,1252 @@
...
+          <cc-button
+            class="date-range-apply-button"
+            ?disabled=${isDateRangeInvalid}
+            @cc-button:click=${this._onApplyCustomDateRange}
+          >${i18n('cc-logs-application-view.custom-date-range.apply')}
+          </cc-button>
+        </div>
+      </div>
+    `;
+  }
```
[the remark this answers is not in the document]
pdesoyres-cc: https://github.com/CleverCloud/clever-components/issues/1013

src/lib/define-smart-component.js
pdesoyres-cc: I noticed that this parameter is never used. That's why I removed it.

src/components/cc-logs/cc-logs.js
pdesoyres-cc: This setter could have two parameters. I would be ok with that.
But, for performance reasons, I want one single setter to have just one call to `this._updateList` method.

src/components/cc-logs-application-view/cc-logs-application-view.js
```diff
@@ -0,0 +1,1252 @@
...
+          flex: 1;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-logs-application-view-beta', CcLogsApplicationView);
+
+/**
```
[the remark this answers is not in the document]
pdesoyres-cc: I'll keep it here for the moment. I will consider moving it when the component is more stable.

src/components/cc-logs-application-view/cc-logs-application-view.js
```diff
@@ -0,0 +1,1252 @@
...
+        .overlay {
+          position: absolute;
+          top: 0;
+          left: 0;
+          width: 100%;
+        }
+        
+        .spacer {
+          flex: 1;
+        }
```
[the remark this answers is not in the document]
pdesoyres-cc: This did the trick

```css
.date-range-buttons {
  display: grid;
  align-items: center;
  grid-gap: 0.5em;
  grid-template-areas: 'left right spacer apply';
  grid-template-columns: auto auto 1fr auto;
}

.date-range-apply-button {
  grid-area: apply;
}
```

PR #985, opened 2024-03-14
chore(storybook): upgrade to 8.0 & remove react deps

florian-sanders-cc:
## What does this PR do?

- Upgrades storybook to 8.0,
- Adapts the `markdown-to-csf.js` file to the only breaking change impacting us (apparently),
- Removes the `react` & `read-dom` dependencies since they are no longer required by Storybook (they come with Storybook as it should always have been) :tada: 

## How to review?

- Check both locally & [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/storybook/upgrade-to-8/index.html?path=/docs/readme--docs):
  - all stories are still working,
  - all controls and shortcuts ("i" within a story to change the language) are still working,
  - the menu is still properly ordered,
  - should be enough?


src/stories/lib/markdown-to-csf.js
```diff
@@ -29,39 +29,20 @@ const processor = unified()
...
-  const subtitle = getSubTitle(frontmatterNode, headingNode);
-
-  const title = [kind, subtitle].filter((a) => a != null).join('/');
-
   const htmlContent = processor().processSync(markdownText).contents;
   const parsedHTML = JSON.stringify(htmlContent);
 
   const csfScript = `
     import { MarkdownDocs } from '/src/stories/lib/markdown-docs.jsx';
     import React from 'react';
```
florian-sanders-cc, line 37: **note**: even though we have removed `react` from our deps, vite still resolves it from `storybook` packages.

I don't see why it would stop working unless Storybook stops using react (but in such case, we wouldn't even need this file anyway :smile:).

Galimede (approved):
LGTM, GG! :muscle: 

That's nice to get of React as a direct dependency. :tada: 

Just quick question: 
```

attention => Storybook now collects completely anonymous telemetry regarding usage.
This information is used to shape Storybook's roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://storybook.js.org/telemetry
```

Do we want to opt-out of the Storybook telemetry? :thinking:

florian-sanders-cc:
> Do we want to opt-out of the Storybook telemetry? 🤔

Good question :+1:

roberttran-cc (approved):
LGTM! And I agree that Storybook seems much more reactive with this version. Me gusta.

pdesoyres-cc (changes requested):
Well done Flo.

Unfortunately, the stories and docs are broken for controllers and mixins.

florian-sanders-cc:
> Well done Flo.
> 
> Unfortunately, the stories and docs are broken for controllers and mixins.

Ah yes, thanks for this :muscle: I forgot to propagate the change done within `markdown-to-csf` to the story files of templates & controllers :sweat_smile:.
The default export now contains the doc render function directly so they should rely on the default export directly.

pdesoyres-cc (approved):
LGTM

hsablonniere (approved):
Well done @florian-sanders-cc, I only had a quick look at the code but I haven't tester locally. Since there are already a few reviews, I'm OK with this.

:+1:

PR #987, opened 2024-03-18
chore: protect master branch from push

pdesoyres-cc:
This PR adds protection on direct push on the master branch. It doesn't prevent developers from doing it (it can sometimes be necessary). But a confirmation will be required when pushing on the master branch.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+protected_branch='master'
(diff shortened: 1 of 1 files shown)


florian-sanders-cc:
Did not test it but the code LGTM!
Gg @pdesoyres-cc :sunglasses:

florian-sanders-cc (approved):

florian-sanders-cc:
Actually I couldn't resist testing it on another branch, works great ;)

PR #991, opened 2024-03-22
Storybook (dev server): refresh CEM when a component is updated

### LINKED ISSUE #989: Storybook: docs are not regenerated on the fly in dev mode
## Summary

There were actually two issues:

- vite was not picking up `custom-elements.json` file as a virtual file because the `/dist/custom-elements.json` file actually exists (sometimes at least),
- the CEM was not updated when a component was updated because the HMR was not aware of the relationship between the component file and the `custom-elements.json` virtual file/module.

## Context

We've recently moved to vite & storybook > 7.0.

It looks like the `generate-cem-vite-plugin` does not work as it should.
It generates the docs when launching storybook but it's not run when the doc of a component is updated (that is to say, when a component file is modified).

This seems to be related to HMR:

- the plugin is not run because the `custom-elements.json` file is not requested when HMR happens, only the component file. More precisely, it seems the browser requests the file and vite serves a cached version of the file because it's not supposed to have changed.
- the problem is the same whatever the lifecycle hook used by the plugin (transform / load).

We may need a way to hook into HMR when a component is updated and tell vite to consider the `custom-elements.json` module as "invalidated" :thinking: 

For more info, see: 

- https://vitejs.dev/guide/api-hmr.html
- https://vitejs.dev/guide/api-plugin.html#handlehotupdate

florian-sanders-cc:
Fixes #989 

## What does this PR do?

- Improve the `viteFinal` config handling. We rely on the `mergeConfig` helper from vite to avoid erasing parts of the builder config by mistake. This is on par with the Storybook docs (see [viteFinal - Storybook](https://storybook.js.org/docs/api/main-config-vite-final#options)).
- Add vite to our dev dependencies. We need to for the `mergeConfig` helper (although we could rely on the `storybook` node modules one) + the types. Note that since v8.0.0, storybook automatically adds vite to the dev dependencies so this feels like the right move anyway?
- Regenerate CEM everytime a component is updated (in dev mode only). (see #989 for more info + the comments within the code to review).

## How to review?

- Check both commits (the first commit's diff is not easy to read, you might as well just check the code itself for the config at least),
- Run storybook locally, go to a component's doc page and update the doc of the component you are viewing. The doc story should reload and you should see the modified doc.
- The "built" version (preview/prod) should not be affected by these changes so nothing to check expect that the preview has been built and is working.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -42,24 +42,40 @@ const config = {
   },
   // index markdown stories so they can be part of the generated menu and lazy loaded
   experimental_indexers: async (existingIndexers) => {
-    return [markdownIndexer, ...existingIndexers]
+    return [markdownIndexer, ...existingIndexers ?? []]
   },
-  viteFinal (config, { configType }) {
-    // transform markdown files to CSF so they can be loaded by storybook
-    config.plugins?.unshift(rollupMdToCsfPlugin());
+  async viteFinal (config, { configType }) {
+    // This dynamic import is the recommended Storybook way to
+    // import `vite`. With a static import you get a warning about CJS.
+    // see: https://storybook.js.org/docs/builders/vite#configuration
+    const { mergeConfig } = await import('vite');
+    /** @type {import('vite').InlineConfig} **/
+    let customConfig = {};
+
+    const commonPlugins = [rollupMdToCsfPlugin()];
+
+    const devModePlugins = [
+      // generate CEM on demand and serve it
+      generateCem(),
+      // add apiConfig to the smart container roots within markdown smart stories
+      injectAuthForSmartComponentsPlugin,
+    ];
 
     if (configType === 'DEVELOPMENT') {
-      // serve and process all files instead of storybook related files only
-      config.appType = 'mpa';
-
-      // generate CEM on demand and serve it
-      config.plugins?.push(generateCem());
-
-      // add apiConfig to the smart container roots within markdown smart stories
-      config.plugins?.push(injectAuthForSmartComponentsPlugin);
+      customConfig = {
+        // serve and process all files instead of storybook related files only
+        appType: 'mpa',
+        plugins: [...commonPlugins, ...devModePlugins],
+      }
     }
 
-    return config;
+    if (configType === 'PRODUCTION') {
+      customConfig = {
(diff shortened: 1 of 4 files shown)


.storybook/main.js
```diff
@@ -42,24 +42,48 @@ const config = {
...
   // index markdown stories so they can be part of the generated menu and lazy loaded
   experimental_indexers: async (existingIndexers) => {
-    return [markdownIndexer, ...existingIndexers]
+    return [markdownIndexer, ...existingIndexers ?? []]
   },
-  viteFinal (config, { configType }) {
-    // transform markdown files to CSF so they can be loaded by storybook
-    config.plugins?.unshift(rollupMdToCsfPlugin());
+  async viteFinal (config, { configType }) {
+    const { mergeConfig } = await import('vite');
```
florian-sanders-cc, line 48: **note:** I don't understand why we **have** to use a dynamic import but this is what the docs do (see [viteFinal - Storybook](https://storybook.js.org/docs/api/main-config-vite-final#options).

If you try to use a static import instead, vite throws a warning about the fact that you're using the `CJS` version of vite's API:
`The CJS build of Vite's Node API is deprecated.`

pdesoyres-cc (approved):
Thank you for all the comments that explain deeply how all that works.

Galimede:
Note: if you modify a `.d.ts` type file, it wont trigger cem again. Then, if you did and try to modify the doc of the related component in hope that it would do the trick, the type displayed will remain the same as before.

florian-sanders-cc:
> Note: if you modify a `.d.ts` type file, it wont trigger cem again. Still, if you wanted to then modify the doc of the related component after modifying the `.d.ts`, in hope that it would do the trick and reload cem, the type displayed will remain the same as before.

As discussed in sync, this issue was already present with older storybook version. The `d.ts` files are not actually seen as modules by the dev server so even if it detects a change, it does not trigger any refresh.
I've tried to solve it with the same logic as the one used within this PR but this didn't workout: the fact that these files are not recognized as dependencies of other files is the main issue.

I'll open an issue to see what we can do in the future but I'll merge this PR as is in the meantime.

Thanks @Galimede for pointing this out and testing with me :sunglasses:

PR #995, opened 2024-03-28
cc-button/cc-link: prevent activation in skeleton mode

### LINKED ISSUE #994: cc-link/cc-button (link mode): disable & force transparent color
## Context

The `cc-link` template and the `cc-button` with the link prop enabled both cause the same issue: users can still activate these elements while in skeleton.
Also, the text slotted inside the `cc-button` with the link prop enabled is visible on hover when in skeleton.

This can easily be reproduced in the `cc-addon-backups` skeleton story.

florian-sanders-cc:
Fixes #994 

## What does this PR do?

- cc-button: Prevent click events from going through in skeleton / waiting mode.
- cc-button: Make sure the text is not visible in link + skeleton mode.
- cc-link: Unset the `href` attribute in skeleton mode.

## How to review?

- 2 reviewers should be enough,
- review both commits,
- compare the [cc-addon-backups - skeleton story - Production](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-addon-cc-addon-backups--skeleton) and [cc-addon-backups - skeleton story - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/links/fix-skeleton/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-backups--skeleton).
  - In prod, when clicking the first blue skeleton within a backup row, the page is refreshed. It shouldn't be the case in preview.
  - In prod, when hovering the second blue skeleton within a backup row, the text is visible. It shouldn't be the case in preview.

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -179,7 +179,7 @@ export class CcButton extends LitElement {
     e.stopPropagation();
 
     // we need to check that because we use aria-disabled which doesn't prevent the onclick event to be fired.
-    if (this.disabled) {
+    if (this.disabled || this.skeleton || this.waiting) {
       return;
     }
 
@@ -631,6 +631,10 @@ export class CcButton extends LitElement {
 
         .cc-link .text-normal {
           font-size: 1em;
+        }
+
+        .cc-link.skeleton:hover {
+          color: transparent;
         }
       `,
     ];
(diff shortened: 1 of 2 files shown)


Galimede (commented):
Nice catch, LGTM! :muscle: 

Note: I was able to do the test on the prod SB on the `ccLink` story in skeleton mode. However, clicking on the first blue skeleton in your example with `cc-addon-backup` doesn't refresh the page on my side. :thinking: 
But that's no worry, the check is okay on the `ccLink` story :wink:

Galimede (approved):

pdesoyres-cc (approved):
LGTM. Thank you!

PR #1020, opened 2024-04-11
CEM & TypeScript: upgrade

### LINKED ISSUE #950: Investigate CEM Analyzer & TypeScript update opportunity
## Context

The `typescript` package version used in our dev dependencies is tied to the version bundled by the `CEM Analyzer` package on purpose.

We use it mainly for two things : 

- tests related to the docs (generated by the CEM Analyzer),
- typechecking within our IDEs.

This is why we use the same version as CEM Analyzer, we want the types to be the same in our docs, our tests on these docs & our typechecking tools.

The `typescript` version currencly in use in our project is `4.3.5`. It is ~2 years old and several tools are no longer supporting it (`ts-lit-plugin`, `lit-analyzer`, `commitlint`). 

For some of them, this is not a problem because they bring it in their dependencies (`commitlint`).
But `ts-lit-plugin` & `lit-analyzer` both rely on the local typescript version found within the project.

It looks like `CEM Analyzer` released a few versions in which they have upgraded typescript. We should try and upgrade these packages.

We'll have to check very carefully that our docs do not break. :monocle_face:

florian-sanders-cc:
Fixes #950 

## What does this PR do?

- Upgrades `custom-elements-manifest/analyzer` to `0.9.4`,
- Upgrades `typescript` to `5.4.2`,
- Upgrades `ts-lit-plugin` to `2.0.2`,
- Updates the `tsconfig.json` file to enable options we could not use before,
- Fixes the only breaking change that impacted us following this upgrade.

## How to review?

- Check the commits,
- Check that `ts-lit-plugin` is still working fine, for instance: removing the `cc-icon` import within the `cc-badge` should trigger an error where `<cc-icon>` is used,
- Check some doc stories using the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cem-typescript/upgrade/index.html) (we have already checked the diff + stories that may have been impacted),
- 2 reviewers (other than @Galimede & me) should be enough for this one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "@bundled-es-modules/chai": "^4.2.2",
         "@commitlint/cli": "^18.5.0",
         "@commitlint/config-conventional": "^18.5.0",
-        "@custom-elements-manifest/analyzer": "^0.6.4",
+        "@custom-elements-manifest/analyzer": "^0.9.4",
         "@esm-bundle/chai-as-promised": "^7.1.1",
         "@open-wc/dev-server-hmr": "^0.1.3",
         "@open-wc/testing": "^3.1.6",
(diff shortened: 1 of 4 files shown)


Galimede (approved):

pdesoyres-cc (approved):
LGTM. Thanks for taking that.

roberttran-cc (approved):

PR #1027, opened 2024-04-16
chore(storybook): prevent error with HMR and `sandbox` / `demo-smart`

### LINKED ISSUE #1026: Storybook/vite: error when HMR is triggered while browsing `demo-smart` / `sandbox` (anything but Storybook)
## Context

With #991, we introduced HMR for component docs.

This works well but it triggers an error when a file is modified while browsing `demo-smart` / `sandbox` (anything but Storybook).
We get the following error:

```ts
Cannot read properties of undefined (reading 'invalidationState')
    at ModuleGraph.invalidateModule (.../clever-components/node_modules/vite/dist/node/chunks/dep-B-u6xNiR.js:64130:43)
    at handleHotUpdate (.../clever-components/cem/generate-cem-vite-plugin.js:44:28)
    at handleHMRUpdate (.../clever-components/node_modules/vite/dist/node/chunks/dep-B-u6xNiR.js:65191:39)
    at async onHMRUpdate (.../clever-components/node_modules/vite/dist/node/chunks/dep-B-u6xNiR.js:64719:17)
    at async FSWatcher.<anonymous> (.../Projects/clever-components/node_modules/vite/dist/node/chunks/dep-B-u6xNiR.js:64756:9
```

This happens because:

1. the `handleHotUpdate` hook of the `generate-cem-vite-plugin` is called everytime a hot update is about to happen,
2. within this hook, we test if the path related to the hot update is a component file (which is true),
3. we invalidate the `CEM` virtual module but it does not exist, we're outside of Storybook here so the CEM is not imported by anything (no story, no storybook UI, no docs).

## Solution

Do not invalidate the `CEM` virtual module if the module is `undefined`.

florian-sanders-cc:
Fixes #1026

## What does this PR do?

- Fixes an error that happens when modifying a component file while browsing `sandbox` or `demo-smart`.

## How to review?

- On the master branch:
  - go to localhost:6006/demo-smart/,
  - modify a component (any component, it does not have to be the one you are viewing, it does not have to be related to a smart demo),
  - you should get the following error `Cannot read properties of undefined (reading 'invalidationState')`.
  - :warning: Don't forget to have your smart component env vars sourced otherwise the error will not be triggered!
- Do the same on this branch, you should get no error.

### DIFF
--- a/cem/generate-cem-vite-plugin.js
+++ b/cem/generate-cem-vite-plugin.js
@@ -39,8 +39,16 @@ export default function generateCem () {
     /** @param {import('vite').HmrContext} context */
     handleHotUpdate ({ server, file }) {
       const isComponentFileRegex = /cc-[^/]+\.js$/;
+
       if (file.match(isComponentFileRegex)) {
         const cemModule = server.moduleGraph.getModuleById(resolvedVirtualModuleId);
+
+        // When serving a component file outside of Storybook (`demo-smart` / `sandbox`),
+        // the cemModule is `undefined` because CEM is not imported by any file (no docs to display, no Storybook UI, etc.)
+        if (cemModule == null) {
+          return;
+        }
+
         server.moduleGraph.invalidateModule(cemModule);
       }
     },


Galimede (approved):
LGTM, thanks for this fix Florian! :hugs:

PR #1031, opened 2024-04-24
chore: prettier format test

Galimede:


roberttran-cc:
Thanks for this preview branch, it really helps!

Overall, I'm really looking forward the benefits of this tool, although some rules are meh but we'll be able to adjust them over time.

Random feedback:
- we should disable the rule enforcing self closing HTML tags, as they seem [unnecessary](https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags)
- we should exclude some files from this formatter:
  - `src/assets/*.icons.js`
  - `src/assets/world-110m.geo.js`
  - `src/stories/fixtures/*.js`

.prettierignore
```diff
@@ -0,0 +1,18 @@
...
+cem/test/fixtures
+
+# Ignore ESLint config file
+**/.eslintrc.cjs
+
+# Ignore MD files
+**/*.md
+
+# Ignore stories
+#src/components/*/*.stories.js
```
pdesoyres-cc, line 16: I'm in favor of that: We should not ignore stories

pdesoyres-cc (approved):
LGTM

florian-sanders-cc (commented):
LGTM

I think we should format story files as you suggest :+1:

.prettierignore
hsablonniere: Let's add them (even if it makes them a bit more verbose)

florian-sanders-cc (approved):

PR #1044, opened 2024-05-07
refactor(cc-addon-admin)!: rework properties to avoid impossible states

florian-sanders-cc:
## What does this PR do?

- Refactors the `cc-addon-admin` component to implement our new state structure,
- Implements TypeChecking within the `cc-addon-admin` component and its stories,
- Adds 2 new stories because we handle name / tag updates a little bit differently: we now set the related submit button in waiting mode ("update name" button for name update for instance) and disable other buttons while waiting,
- Updates the simulation story to reflect the new stories (simulates a name update, then a tag update, then an add-on deletion).

## How to review?

- Compare the [prod stories](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-addon-cc-addon-admin--default-story) to the [preview stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-addon-admin/state-migration/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-admin--default-story).

### DIFF
--- a/src/components/cc-addon-admin/cc-addon-admin.js
+++ b/src/components/cc-addon-admin/cc-addon-admin.js
@@ -8,16 +8,16 @@ import { dispatchCustomEvent } from '../../lib/events.js';
 import { i18n } from '../../lib/i18n.js';
 
 /**
- * @typedef {import('../common.types.js').Addon} Addon
+ * @typedef {import('./cc-addon-admin.types.js').AddonAdminState} AddonAdminState
+ * @typedef {import('./cc-addon-admin.types.js').AddonAdminStateLoaded} AddonAdminStateLoaded
+ * @typedef {import('./cc-addon-admin.types.js').AddonAdminStateLoading} AddonAdminStateLoading
+ * @typedef {import('./cc-addon-admin.types.js').AddonAdminStateSaving} AddonAdminStateSaving
+ * @typedef {import('lit').PropertyValues<CcAddonAdmin>} CcAddonAdminPropertyValues
+ * @typedef {import('lit').TemplateResult<1>} TemplateResult
  */
 
 /**
  * A component displaying the admin interface of an add-on to edit its name or delete the add-on.
- *
- * ## Details
- *
- * * When addon is nullish, a skeleton screen UI pattern is displayed (loading hint).
- *
  *
  * @cssdisplay block
  *
@@ -29,13 +29,10 @@ export class CcAddonAdmin extends LitElement {
 
   static get properties () {
     return {
-      addon: { type: Object },
-      error: { type: Boolean },
       noDangerZoneBackupText: { type: Boolean, attribute: 'no-danger-zone-backup-text' },
       noDangerZoneVmText: { type: Boolean, attribute: 'no-danger-zone-vm-text' },
-      saving: { type: Boolean },
+      state: { type: Object },
       _name: { type: String, state: true },
-      _skeleton: { type: Boolean, state: true },
       _tags: { type: Array, state: true },
     };
   }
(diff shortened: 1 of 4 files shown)


src/components/cc-addon-admin/cc-addon-admin.types.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface AddonAdminStateLoading {
+  type: 'loading';
+}
+
+export interface AddonAdminStateError {
+  type: 'error';
+}
+
+export type AddonAdminStateSaving = AddonAdminStateDeleting | AddonAdminStateUpdatingName | AddonAdminStateUpdatingTags;
```
hsablonniere, line 35: nitpick: I was a bit confused at first because this line was at the bottom. Can we move it with the other `type` at the top.

Other order proposition:

* type AddonAdminState
* interface AddonAdminStateLoaded
* interface AddonAdminStateLoading
* interface AddonAdminStateError
* type AddonAdminStateSaving
* interface AddonAdminStateDeleting
* interface AddonAdminStateUpdatingName
* interface AddonAdminStateUpdatingTags

src/components/cc-addon-admin/cc-addon-admin.js
```diff
@@ -29,143 +29,182 @@ export class CcAddonAdmin extends LitElement {
...
       </cc-block>
     `;
   }
 
+  /**
+   * @param {AddonAdminStateLoaded|AddonAdminStateLoading|AddonAdminStateSaving} state
+   * @returns {TemplateResult}
+   * @private
+   */
+  _renderContent (state) {
```
hsablonniere, line 135: note: this feels a bit weird to have a subrender method that takes a state but also rely on `this` to know what to do. It's no big deal and I don't see how we can improve the current code so, just thinking out loud ;-)

hsablonniere (changes requested):
Hey @florian-sanders-cc and @HeleneAmouzou, thank you for this :tada:

I added a few comments. I wanted to suggest to inline the subrender method but looking at its size and the number of variables it computes, I guess the current code is better.

Well done!

src/components/cc-addon-admin/cc-addon-admin.js
```diff
@@ -29,143 +29,182 @@ export class CcAddonAdmin extends LitElement {
...
+      this._name = this.state.name;
+    }
 
-    if (changedProperties.has('addon')) {
-      this._skeleton = (this.addon == null);
-      this._name = this._skeleton ? '' : this.addon.name;
-      this._tags = this._skeleton ? [] : this.addon.tags;
+    if (changedProperties.has('state') && 'tags' in this.state) {
+      this._tags = this.state.tags;
     }
```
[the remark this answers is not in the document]
florian-sanders-cc: @hsablonniere we hesitated a lot with this one but here is our mindset regarding this.

At first, we decided to check for state types but as you can see below, it becomes very verbose:
```js
    if (changedProperties.has('state') && (this.state.type === 'loaded' || this.state.type === 'deleting' || this.state.type === 'updatingName' || this.state.type === 'updatingTags')) {
      this._name = this.state.name;
    }
```

We thought it was difficult to read and we didn't like it so we decided to go for the straightforward fix: TypeScript says it might cause errors because `name` may not exist? Then let's make sure `name` exists.
This is the only case where we used that.

src/components/cc-addon-admin/cc-addon-admin.js
florian-sanders-cc: We agree with you but we don't see how to improve it so we'll leave it like this.

src/components/cc-addon-admin/cc-addon-admin.js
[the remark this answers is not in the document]
hsablonniere: Agreed, and I checked, using `includes` will break inference.

hsablonniere (approved):

PR #1050, opened 2024-05-07
refactor(cc-heptapod-info)!: rework properties to avoid impossible st…

HeleneAmouzou:
## What does this PR do?

- Refactors the `cc-heptapod-info` component to implement our new state structure,
- Implements TypeChecking within the `cc-heptapod-info` component and its stories.

## How to review?

- Compare the [prod stories]() to the [preview stories]().


src/components/cc-heptapod-info/cc-heptapod-info.js
```diff
@@ -18,40 +18,43 @@ const HEPTAPOD_LOGO_URL = 'https://assets.clever-cloud.com/logos/heptapod.svg';
...
-    /** @type {Statistics|null} Sets the usage statistics of this heptapod SaaS or `"not-used"` to display a message explaining the service is not used. */
-    this.statistics = null;
+  /**
+   * @param {HeptapodInfoState} state
+   * @return {state is (HeptapodInfoStateLoaded|HeptapodInfoStateLoading)}
+   * @private
+   */
+  _isLoadedOrLoading (state) {
+    return state.type === 'loaded' || state.type === 'loading';
   }
```
hsablonniere, line 53: nitpick: since it's a rather simple method and to improve "reading code from just one place", I'd rather move this to a variable inside the render function.

src/components/cc-heptapod-info/cc-heptapod-info.js
```diff
@@ -67,39 +70,48 @@ export class CcHeptapodInfo extends LitElement {
...
 
-        ${!this.error && isNotUsed ? html`
+        ${this.state.type === 'not-used' ? html`
           <div class="no-statistics">${i18n('cc-heptapod-info.not-in-use')}</div>
         ` : ''}
 
-        ${this.error ? html`
-          <cc-notice intent="warning" message="${i18n('cc-heptapod-info.error-loading')}"></cc-notice>
-        ` : ''}
+        ${this._isLoadedOrLoading(this.state) ? this._renderStatistics(statistics, skeleton) : ''}
```
hsablonniere, line 81: nitpick: like my other comment, I think it would benefit from inlining the subrender directly here as it's not that much code and we use it only once.

hsablonniere (changes requested):
Hey @florian-sanders-cc and @HeleneAmouzou, thank you for this!

I suggested some nitpicks which I think improve the reading of the code (could be only my opinion, I'll let you decide).

I also thought about another way to write the code:

* no isLoadedOrLoading method
* no variables in render
* we have 2 ternary operators, one for `loading`, one for `loaded`
* we keep the subrender

https://github.com/CleverCloud/clever-components/blob/cc-heptapod-info/state-migration--hubert/src/components/cc-heptapod-info/cc-heptapod-info.js#L39-L103

I prefer what's in the branch than the inline comments I made in this PR.

src/components/cc-heptapod-info/cc-heptapod-info.js
florian-sanders-cc: You cannot inline this into a variable within the `render` function because this will break TypeScript inference (at least for now, with TS 5.5 it may become ok). For such cases you have to use a type predicate or find a smarter solution. 
The solution you suggest below is smarter so I'll see what @HeleneAmouzou thinks about it but we'll probably go for that, thanks :wink:

roberttran-cc (approved):
Apart from Hubert's feedback (which I agree), LGTM!

hsablonniere (approved):

PR #1057, opened 2024-05-07
refactor(cc-tile-requests)!: rework properties to avoid impossible states

HeleneAmouzou:
## What does this PR do?

- Refactors the `cc-tile-requests` component to implement our new state structure,
- Implements TypeChecking within the `cc-tile-requests` component and its stories.

## How to review?

- Compare the [prod stories](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-overview-cc-tile-requests--default-story) to the [preview stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-tile-requests/state-migration/index.html?path=/story/%F0%9F%9B%A0-overview-cc-tile-requests--default-story).

### DIFF
--- a/src/components/cc-tile-requests/cc-tile-requests.types.d.ts
+++ b/src/components/cc-tile-requests/cc-tile-requests.types.d.ts
@@ -1,5 +1,20 @@
-type RequestsData = [
+export type RequestsData = [
   number, // Start timestamp in milliseconds. Expected to be rounded to the hour of its respective TZ.
   number, // End timestamp in milliseconds. Expected to be rounded to the hour of its respective TZ.
   number, // Number of request during this time window.
 ]
+
+export type TileRequestsState = TileRequestsStateLoaded | TileRequestsStateLoading | TileRequestsStateError;
+
+export interface TileRequestsStateLoaded {
+  type: 'loaded';
+  data: Array<RequestsData>;
+}
+
+export interface TileRequestsStateLoading  {
+  type: 'loading';
+}
+
+export interface TileRequestsStateError {
+  type: 'error';
+}
--- a/src/components/cc-tile-requests/cc-tile-requests.js
+++ b/src/components/cc-tile-requests/cc-tile-requests.js
@@ -3,6 +3,7 @@ import { BarController, BarElement, CategoryScale, Chart, LinearScale, Title, To
 import ChartDataLabels from 'chartjs-plugin-datalabels';
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { createRef, ref } from 'lit/directives/ref.js';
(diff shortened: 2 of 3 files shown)


src/components/cc-tile-requests/cc-tile-requests.types.d.ts
```diff
@@ -1,5 +1,25 @@
...
+
+export interface TileRequestsStateLoading  {
+  type: 'loading';
+}
+
+export interface TileRequestsStateError {
+  type: 'error';
+}
+
+export interface TileRequestsStateEmpty {
```
pdesoyres-cc, line 22: question: do we really need a separate state for emptiness? Couldn't we use the 'loaded' state and check that the data is empty?

question: what happens if the state is loaded and data is an empty array? Is it the same rendering than when state is 'empty'?

pdesoyres-cc, line 22: remark: initially, the `_empty` property was a private property. Emptiness should remain private so, I think that this 'empty' state should not exist. Instead, the emptiness should be (like initially) calculated from the `data` given with the 'loaded' state.

pdesoyres-cc (changes requested):
Well done, but see mmy comment bellow.

src/components/cc-tile-requests/cc-tile-requests.types.d.ts
florian-sanders-cc: This is a good question :thinking: 

Initially, we went for an `empty` state because that's what the `cc-tile-metrics` does and it's the most recent of the tile components (in fact it already had a state).

We figured we would discuss this point in sync but forgot to mention it within the discussion.

Anyway, I think you have a valid point but I'd also like @Galimede to weigh in since he built the `cc-tile-metrics` and I remember there were some hesitations about how to handle empty data :thinking: 

I would prefer we reach a consensus and apply the same logic to both components even though the API is not the same: it would be great if our UI components had the same logic and the smart would adapt to it if necessary.

florian-sanders-cc: see the other comment, I agree but I want @Galimede's opinion about it and I'd like to apply the same logic to the `cc-tile-metrics` if possible.

Galimede (approved):

florian-sanders-cc:
## Sync about `empty` states

Considering @Galimede's inputs, we have decided that it made more sense to leave the `cc-tile-metrics` component unchanged and remove the `empty` state from `cc-tile-requests`.

The main reason is that the `cc-tile-metrics` requires some logic to determine whether the state is empty or not and we prefer leaving it to the smart while `cc-tile-requests` is more straightforward so the logic may remain within the component itself.

pdesoyres-cc (approved):
LGTM

PR #1059, opened 2024-05-20
refactor(cc-tile-metrics)!: rework properties to avoid impossible states

florian-sanders-cc:
## What does this PR do?

- Refactors the `cc-tile-metrics` component to implement our new state structure,
- Implements TypeChecking within the `cc-tile-metrics` component and its stories.

## How to review?

- Compare the [prod stories](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-overview-cc-tile-metrics--default-story) to the [preview stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-tile-metrics/state-migration/index.html?path=/story/%F0%9F%9B%A0-overview-cc-tile-metrics--default-story),
- Don't forget to check the smart using the `demo-smart` page.

**Note:** There are a few changes that you should review / pay more attention to. The grafana link now has a state of its own because its data comes from a separate source. It cannot be set to error because there is a fallback link within the smart. This may not be the best idea, I'm not really sure so don't hesitate to weigh in with other ideas / suggestions.
On the contrary, the `metricsLink` is a simple property since it is not set by the smart and it does not depend on any API. This means that the metricsLink cannot be loading / in skeleton mode contrary to what we currently have in prod :thinking:.

### DIFF
--- a/src/components/cc-tile-metrics/cc-tile-metrics.js
+++ b/src/components/cc-tile-metrics/cc-tile-metrics.js
@@ -14,6 +14,7 @@ import {
 } from '../../assets/cc-remix.icons.js';
 import { ResizeController } from '../../controllers/resize-controller.js';
 import { i18n } from '../../lib/i18n.js';
+import { isStringEmpty } from '../../lib/utils.js';
 import { accessibilityStyles } from '../../styles/accessibility.js';
 import { tileStyles } from '../../styles/info-tiles.js';
 import { skeletonStyles } from '../../styles/skeleton.js';
@@ -52,7 +53,13 @@ const SKELETON_REQUESTS = Array
   });
 
 /**
- * @typedef {import('./cc-tile-metrics.types.js').MetricsState} MetricsState
+ * @typedef {import('./cc-tile-metrics.types.js').TileMetricsMetricsState} TileMetricsMetricsState
+ * @typedef {import('./cc-tile-metrics.types.js').TileMetricsGrafanaLinkState} TileMetricsGrafanaLinkState
+ * @typedef {import('./cc-tile-metrics.types.js').MetricsData} MetricsData
+ * @typedef {import('./cc-tile-metrics.types.js').Metric} Metric
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLCanvasElement>} RefCanvas
+ * @typedef {import('lit').PropertyValues<CcTileMetrics>} CcTileMetricsPropertyValues
+ * @typedef {import('chart.js').ChartData} ChartData
  */
 
 /**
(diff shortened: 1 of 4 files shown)


src/components/cc-tile-metrics/cc-tile-metrics.types.d.ts
```diff
@@ -1,30 +1,50 @@
...
 }
 
-interface MetricsStateEmpty {
-  state: 'empty';
+export interface Metric {
+  // Timestamp in ms
+  timestamp: number;
+  // Value is a percentage (e.g: 14.02)
+  value: number;
+  skeleton?: boolean | null;
```
pdesoyres-cc, line 25: fix: remove `skeleton` property

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -143,6 +156,11 @@ export class CcTileMetrics extends LitElement {
     });
   }
 
+  /**
+   * @param {Metric[]} inputData
+   * @returns {ChartData}
+   * @private
+   */
   _getChartData (inputData) {
```
pdesoyres-cc, line 164: fix: add `skeleton` argument to this function and use it instead of the `skeleton` property of the inputData

src/components/cc-tile-metrics/cc-tile-metrics.stories.js
```diff
@@ -61,39 +91,48 @@ function addTimestamp (array) {
...
+      type: 'loaded',
+      metricsData: {
+        cpuMetrics: addTimestamp(fakeMetricData(24, 25)),
+        memMetrics: addTimestamp(fakeMetricData(24, 16)),
       },
     },
   })),
 });
 
 export const skeleton = makeStory(conf, {
```
pdesoyres-cc, line 105: nitpick: could be renamed to "loading"

pdesoyres-cc (changes requested):
Well done. See below how we can simplify the `Metric` type.

src/components/cc-tile-metrics/cc-tile-metrics.js
florian-sanders-cc: This is a lot nicer indeed, thanks! :sunglasses:

src/components/cc-tile-metrics/cc-tile-metrics.js
```diff
@@ -319,18 +358,24 @@ export class CcTileMetrics extends LitElement {
           <p>${i18n('cc-tile-metrics.docs.more-metrics')}</p>
           <ul>
             <li>
-              ${ccLink(this.grafanaLink, i18n('cc-tile-metrics.grafana'), state === 'loading', i18n('cc-tile-metrics.link-to-grafana'))}
-            </li>
-            <li>
-              ${ccLink(this.metricsLink, i18n('cc-tile-metrics.metrics-link'), state === 'loading', i18n('cc-tile-metrics.link-to-metrics'))}
+              ${ccLink(grafanaLink, i18n('cc-tile-metrics.grafana'), skeleton, i18n('cc-tile-metrics.link-to-grafana'))}
             </li>
+            ${this.metricsLink != null && this.metricsLink.length > 0 ? html`
```
Galimede, line 363: suggestion: you could use the `isStringEmpty` function in our lib file instead

Galimede (approved):
LGTM, GG! :muscle: 

I've nothing to say regarding the `metricsLink` and the way we handle the state for Grafana, I think it's good as it is right now. :slightly_smiling_face:

pdesoyres-cc (approved):
LGTM

PR #1066, opened 2024-05-30
fix: update Materia branding

davlgd:
I've updated texts, let me know if you need/want me to change more references to Materia KV


florian-sanders-cc (approved):
LGTM! Thanks for creating the PR yourself @davlgd :heart:

florian-sanders-cc:
The commit message is missing the scope (`cc-addon-credentials`), it's optional but it's a bit better for the changelog so I will amend the commit and force push

florian-sanders-cc:
My bad I reviewed only the commit and forgot to check the preview: we can also update the story title from `👍 Data loaded (materia db kv)` to `👍 Data loaded (materia kv)`.
I'll force push again :innocent:

PR #1076, opened 2024-06-11
feat(cc-logs-application-view): allow store and restore of the selected date range

### LINKED ISSUE #1074: cc-logs-application-view: dispatch event when date range selection changes
Context: We want to be able to store and restore the date range selection set by the user.

The first thing to do is to dispatch an event when the date range selection changes.

As of now, an event is dispatched when the date range changes. It's ok, but with that we only have the date range {since/until}. Now we want to have the date range selection which is an object like: `{type: 'live'}` or `{type:'predefined', def: 'lastHour'}` ...

### LINKED ISSUE #1075: cc-logs-application-view: allow initialization from a date range selection passed into the smart context
Context: We want to be able to store and restore the date range selection set by the user.

When a date range selection is passed into the smart context, we should initialize the view with that date range selection.

pdesoyres-cc:
Fixes #1074
Fixes #1075

- You should see a new event when changing the date range
- Use the sandbox with the dateRangeSelection selector

### DIFF
--- a/src/components/cc-logs-application-view/cc-logs-application-view.smart.js
+++ b/src/components/cc-logs-application-view/cc-logs-application-view.smart.js
@@ -5,10 +5,11 @@ import { HttpError } from '@clevercloud/client/esm/streams/clever-cloud-sse.js';
 import { Buffer } from '../../lib/buffer.js';
 import { defineSmartComponent } from '../../lib/define-smart-component.js';
 import { sendToApi } from '../../lib/send-to-api.js';
+import { unique } from '../../lib/utils.js';
+import { dateRangeSelectionToDateRange } from './date-range-selection.js';
+import { isLive, lastXDays } from './date-range.js';
 import '../cc-smart-container/cc-smart-container.js';
 import './cc-logs-application-view.js';
-import { unique } from '../../lib/utils.js';
-import { getRangeToNow, isLive, lastXDays } from './date-range.js';
 
 /**
  * @typedef {import('../cc-logs-instances/cc-logs-instances.types.js').Instance} Instance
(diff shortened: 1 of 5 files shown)


src/components/cc-logs-application-view/cc-logs-application-view.types.d.ts
```diff
@@ -1,7 +1,7 @@
-import { GhostInstance, Instance } from "../cc-logs-instances/cc-logs-instances.types";
-import { DateDisplay } from "../cc-logs/date-display.types";
-import { LogsControlPalette } from "../cc-logs-control/cc-logs-control.types";
-import { Timezone } from "../../lib/date/date.types";
+import { GhostInstance, Instance } from "../cc-logs-instances/cc-logs-instances.types.js";
+import { DateDisplay } from "../cc-logs/date-display.types.js";
+import { LogsControlPalette } from "../cc-logs-control/cc-logs-control.types.js";
+import { Timezone } from "../../lib/date/date.types.js";
```
hsablonniere, line 4: praise: nice catch

hsablonniere (approved):
:+1: Let's go with this, I'll review the whole smart file later in the general logs review.

src/components/cc-logs-application-view/cc-logs-application-view.smart.js
```diff
@@ -336,6 +340,19 @@ class LogsApplicationViewSmartController {
 
   }
 
+  /**
+   *
```
florian-sanders-cc, line 344: **nitpick:** 
```suggestion

```

florian-sanders-cc (approved):
Apart from a very small nitpick, it looks good to me (from what I understand at least)!

HeleneAmouzou (approved):

roberttran-cc (approved):

PR #1077, opened 2024-06-11
fix(cc-logs-instances): make sure deploying instances grid layout is never broken

### LINKED ISSUE #1070: Bad instances display on deploying instances
Hi :wave: 

I was deploying a new clever app, I failed to deploy it properly several times, after a few run the list of instances shown inside the new console interface wasn't showing names properly. 

I opened the logs after the few failed run during a new deployment, after that new instances (even running) shown wrongly:
image

When refreshing with one fully running instance, I still have stuff going wrong:
image

As the bug can be reproduced on my name, i can give some requests logs or anything to help :)

pdesoyres-cc:
Fixes #1070

Since I could not reproduce the bug in real life, I just protected the grid layout from being broken when the state is not supported.

### DIFF
--- a/src/components/cc-logs-instances/cc-logs-instances.js
+++ b/src/components/cc-logs-instances/cc-logs-instances.js
@@ -491,6 +491,7 @@ export class CcLogsInstances extends LitElement {
         class: 'instance-state--deleted',
       };
     }
+    return null;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


hsablonniere (approved):
Nice catch, looking at the screenshots, I think it will fix the bug :+1:

florian-sanders-cc (approved):
LGTM!

roberttran-cc (approved):

PR #1078, opened 2024-06-12
fix(cc-addon-credentials)!: change `addonType` value for Materia KV

### LINKED ISSUE #1068: cc-addon-credentials: change addon type value to match the new materia product name
## Context

With #1066, we have changed the Materia name from `Materia DB KV` to `Materia KV`.

Since we needed a quick PR & release, we did not want to introduce a  breaking change at the time.

It would be nice to align the code to the new branding so that the component API reflects the new branding.

More precisely, we would need to change the `AddonType` value for materia from `materiadb-kv` to `materia-kv`.

This is a tiny breaking change so it would be nice to include it in a release that already contains other breaking changes (it's not a worth a major version on its own).

florian-sanders-cc:
Fixes #1068

## What does this PR do?

- Changes the `addonType` value for Materia KV from `materiadb-kv` to `materia-kv`

## How to review?

- Check the commit,
- Check the Materia KV story in preview,

### DIFF
--- a/src/components/cc-addon-credentials/cc-addon-credentials.js
+++ b/src/components/cc-addon-credentials/cc-addon-credentials.js
@@ -66,8 +66,8 @@ export class CcAddonCredentials extends LitElement {
         return i18n('cc-addon-credentials.description.kibana');
       case 'pulsar':
         return i18n('cc-addon-credentials.description.pulsar');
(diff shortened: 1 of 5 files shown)


Galimede (approved):
LGTM! :tada:

Issue #1038, opened 2024-04-26
UX Writing: dots at the end of sentences


florian-sanders-cc:
## Context

As discussed here : https://ux.stackexchange.com/questions/122147/periods-in-ux-ui-microcopy/122148#122148 (thanks @roberttran-cc)

We should consider removing dots at the end of one line "sentences" (for instance all form error / help texts).

PR #1090, opened 2024-06-20
cc-domain-management: init component

### LINKED ISSUE #1095: cc-domain-management: new component
## Context

We are reworking the domain screen within the console.

We need a component that allows users to:
- associate a new domain to an application,
  - this domain may be a custom domain like `example.com`
  - this domain may be a `cleverapps.io` domain
    - users should be warned that `cleverapps.io` domains are http only.
- view domains associated to an app,
- remove a domain associated to an app,
- mark a domain as primary domain,
- view DNS info related to the app load balancer so they may configure their DNS,
- Get basic info about certificates & access the page to add their own certificate.

florian-sanders-cc:
Fixes #1095 

## How to review?

- Check the code (since these are mostly new files, it might be more convenient to review locally),
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/init/index.html?path=/story/%F0%9F%9B%A0%EF%B8%8F-domains-cc-domain-management--default-story):
  - all stories,
  - both French & English,
  - different viewport sizes.
- Proceed with the [QA](https://docs.google.com/spreadsheets/d/1Gjb7LwqOx3ZDNDFt2C347wmF2pjk7IuFgFvaFEd2BAM/edit?gid=1267293083#gid=1267293083),
  - note that this QA should be done using the console (links to specific cases are available within the docs).
- At least 4 or 5 reviews are required for this PR.

## A few specific questions

- Component naming, should we got for `cc-domain-list` instead? It feels more consistent even though it doesn't only deal with the list of domains.
- The API sends us the domain & path concatenated into a single `fqdn` property:
  - to display the domain name and the path prefix separately and differently within the list, we need to have two separate properties (`domain` & `pathPrefix`). Internally, within the component, the `pathPrefix` can never be `null` or empty because it's always at least `/`.
  - for the API, the `pathPrefix` thing does not exist and we have to concatenate (+ encode :-1: ) everything. When no path prefix is set, there is no trailing slash within `fqdn`.
  - To sum it up, for `example.com`: 
    - API: `{ fqdn: 'example.com' }`
    - component data = `{ fqdn: 'example.com', domainName: 'example.com', pathPrefix: '/' }`
  - Question: do we keep `fqdn` naming or do we change it? At first I went with `domainWithPathPrefix` but it creates confusion because it's not an actual concatenation of `domainName + pathPrefix`. I opted to keep `fqdn` even though I don't like the naming because it highlights that this comes from the API. This is some kind of unique identifier we use to communicate with the API but we don't actually use it apart from within the smart.
- Do we go for a grid design or cards?
  - [Preview - Grid version](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/init-grid-version/index.html?path=/story/%F0%9F%9B%A0%EF%B8%8F-domains-cc-domain-management--default-story)
  - [Preview - Cards version](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/init/index.html?path=/story/%F0%9F%9B%A0%EF%B8%8F-domains-cc-domain-management--default-story)
  - Note that in the future, we'll propably add more badges (DNS config status, certificate status) and action buttons will be regrouped into 1 menu button.
  - I have tried several options for the grid version, none of which totally convinced me so far:

<details>
  <summary>No columns for badges</summary>

No columns for badges

**Issues:** 

Some people are triggered by the "misaligned" aspect. Note that the more you think about it the more you find it annoying but as a user, there are similar cases that I had never noticed as annoying, for instance in GitHub PR view:
GitHub PR view

</details>

<details>
 <summary>Column dedicated to primary badge</summary>

Grid - column dedicated to primary badge

**Issues:**

Considering "primary" can only applied to one row and that it is always the first row, it feels weird to dedicate a column to it.

</details>

<details>
  <summary>Align right for badge columns</summary>

Align right

**Issues:**

Badges are pretty far from what they are directly related to: the domain.

</details>

<details>
  <summary>My personal preference if you wanna know</summary>

1. cards
2. grid as within the preview (primary inlined with the domain name + column for badges, left aligned).
3. grid with no columns for the badges, all inlined next to the domain, GitHub's way

</details>

## Card design

@hsablonniere has suggested a few nice improvements for the card design, feel free to give your opinion about these as well

<details>
  <summary>Issue with domain names wrapping</summary>

When there's not enough space (small viewport or long domain name), the domain name wraps to a new line.
Sometimes the link icon ends up alone on a line, sometimes the path prefix ends up on a line.

We cannot really prevent that but we can improve the design so it happens less often by making the domain name span accross the whole width and position the buttons below it:
domain cards with domain  name spanning across the width and buttons below

</details>

<details>
  <summary>Gap between cards or not</summary>

When there's only one column (small viewport), should we remove the gap between cards and make it look like a "table"?
domain cards below each other without any gap between them, only borders

</details>

## TODO Flo

- [x] update smart doc with proper links
- [x] rebase once forms have been merged
- [ ] run prettier once all reviews & feedback have been processed.

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -1,0 +1,841 @@
+import { LitElement, css, html } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { createRef, ref } from 'lit/directives/ref.js';
+import { repeat } from 'lit/directives/repeat.js';
+import {
+  iconRemixDeleteBinLine as iconDelete,
+  iconRemixLockUnlockFill as iconHttpOnly,
+  iconRemixExternalLinkLine as iconLink,
+  iconRemixStarLine as iconPrimary,
+  iconRemixFlaskLine as iconTest,
+} from '../../assets/cc-remix.icons.js';
+import { LostFocusController } from '../../controllers/lost-focus-controller.js';
+import {
+  DomainParseError,
+  getDomainUrl,
+  getHostWithWildcard,
+  isTestDomain,
+  isTestDomainWithSubdomain,
+  parseDomain,
+  sortDomains,
+} from '../../lib/domain.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
+import { focusBySelector } from '../../lib/focus-helper.js';
+import { i18n } from '../../lib/i18n.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { linkStyles } from '../../templates/cc-link/cc-link.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-block-section/cc-block-section.js';
+import '../cc-block/cc-block.js';
+import '../cc-button/cc-button.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-notice/cc-notice.js';
+
+/**
+ * @typedef {import('./cc-domain-management.types.js').DomainManagementDnsInfoState} DomainManagementDnsInfoState
+ * @typedef {import('./cc-domain-management.types.js').DomainManagementListState} DomainManagementListState
+ * @typedef {import('./cc-domain-management.types.js').DomainManagementFormState} DomainManagementFormState
+ * @typedef {import('./cc-domain-management.types.js').DomainManagementFormStateIdle} DomainManagementFormStateIdle
+ * @typedef {import('./cc-domain-management.types.js').FormattedDomainInfo} FormattedDomainInfo
+ * @typedef {import('./cc-domain-management.types.js').DomainInfo} DomainInfo
+ * @typedef {import('./cc-domain-management.types.js').FormError} FormError
+ * @typedef {import('../cc-input-text/cc-input-text.js').CcInputText} CcInputText
+ * @typedef {import('lit').TemplateResult<1>} TemplateResult
+ * @typedef {import('lit/directives/ref.js').Ref<CcInputText>} RefCcInputText
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLParagraphElement>} RefHTMLParagraphElement
+ * @typedef {import('lit').PropertyValues<CcDomainManagement>} CcDomainManagementPropertyValues
+ */
+
+/**
+ * A component to manage domains associated to an application.
+ *
+ * @cssdisplay block
+ *
+ * @fires {CustomEvent<{ hostname: string, pathPrefix: string }>} cc-domain-management:add - Fires when clicking the "Add a domain" button.
+ * @fires {CustomEvent<DomainInfo>} cc-domain-management:mark-as-primary - Fires when clicking a "mark as primary" button.
+ * @fires {CustomEvent<DomainInfo>} cc-domain-management:delete - Fires when clicking a "delete" button.
+ */
+export class CcDomainManagement extends LitElement {
+  static get properties() {
+    return {
+      dnsInfoState: { type: Object, attribute: 'dns-info-state' },
+      domainFormState: { type: Object, attribute: 'domain-form-state' },
+      domainListState: { type: Object, attribute: 'domain-list-state' },
+      _sortedDomains: { type: Array, state: true },
+    };
+  }
+
+  /**
+   * @returns {DomainManagementFormStateIdle}
+   */
+  static get INIT_DOMAIN_FORM_STATE() {
+    return {
+      type: 'idle',
+      hostname: {
+        value: '',
+      },
+      pathPrefix: {
+        value: '',
+      },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {DomainManagementDnsInfoState} Sets the state of the DNS info section */
+    this.dnsInfoState = { type: 'loading' };
+
+    /** @type {DomainManagementFormState} Sets the state of the form section */
+    this.domainFormState = CcDomainManagement.INIT_DOMAIN_FORM_STATE;
+
+    /** @type {DomainManagementListState} Sets the state of the domain list section. */
+    this.domainListState = { type: 'loading' };
+
+    /** @type {RefCcInputText} */
+    this._domainInputRef = createRef();
+
+    /** @type {RefHTMLParagraphElement} */
+    this._emptyMessageRef = createRef();
+
+    /** @type {FormattedDomainInfo[]|null} */
+    this._sortedDomains = null;
+
+    new LostFocusController(this, '.delete-domain', ({ suggestedElement }) => {
+      if (suggestedElement != null) {
+        suggestedElement.focus();
+      } else {
+        this._emptyMessageRef.value?.focus();
+      }
+    });
+
+    new LostFocusController(this, '.mark-primary', () => {
+      focusBySelector(this.shadowRoot, '.primary');
+    });
+  }
+
+  /**
+   * @param {string} hostnameValue
+   * @returns {FormError|null} The error code or `null` if no error
+   */
+  _getFormError(hostnameValue) {
+    try {
+      const { pathname } = parseDomain(hostnameValue?.trim());
+
+      if (!hostnameValue.match(/.+\..+/)) {
+        return { code: 'invalid-format' };
+      }
+      if (pathname !== '/') {
+        /** We store the path so we can tell users to move this part to the "Path" input field instead. */
+        return { code: 'hostname-contains-path', pathWithinHostname: pathname };
+      }
+
+      return null;
+    } catch (error) {
+      if (error instanceof DomainParseError) {
+        return { code: error.code };
+      }
+
+      return { code: 'invalid-format' };
+    }
+  }
+
+  /**
+   * @param {FormError|null} error
+   * @returns {string}
+   */
+  _getErrorMessage(error) {
+    switch (error?.code) {
+      case 'empty':
+        return i18n('cc-domain-management.form.domain.error.empty');
+      case 'invalid-format':
+        return i18n('cc-domain-management.form.domain.error.format');
+      case 'invalid-wildcard':
+        return i18n('cc-domain-management.form.domain.error.wildcard');
+      case 'hostname-contains-path':
+        return i18n('cc-domain-management.form.domain.error.contains-path', { path: error.pathWithinHostname });
+    }
+  }
+
+  /** @param {ClipboardEvent} event */
+  _onDomainPaste(event) {
+    if (this.domainFormState.type !== 'idle' || this.domainFormState.hostname.value.trim() !== '') {
+      return;
+    }
+
+    event.preventDefault();
+
+    const clipboardValue = event.clipboardData.getData('text');
+
+    try {
+      const { hostname, pathname, isWildcard } = parseDomain(clipboardValue);
+
+      this.domainFormState = {
+        ...this.domainFormState,
+        hostname: {
+          ...this.domainFormState.hostname,
+          value: getHostWithWildcard(hostname, isWildcard),
+        },
+        pathPrefix: {
+          ...this.domainFormState.pathPrefix,
+          value: pathname.replace(/\/$/, ''),
+        },
+      };
+    } catch {
+      this.domainFormState = {
+        ...this.domainFormState,
+        hostname: {
+          ...this.domainFormState.hostname,
+          value: clipboardValue,
+        },
+      };
+    }
+  }
+
+  /** @param {SubmitEvent} e */
+  _onDomainSubmit(e) {
+    e.preventDefault();
+
+    if (this.domainFormState.type !== 'idle') {
+      return;
+    }
+
+    // we trim and we'll send this value to the API but we don't change what is displayed within the input
+    const hostnameValue = this.domainFormState.hostname.value.trim();
+    const pathPrefixValue = this.domainFormState.pathPrefix.value.trim();
+    const error = this._getFormError(hostnameValue);
+
+    if (error != null) {
+      this.domainFormState = {
+        ...this.domainFormState,
+        hostname: {
+          ...this.domainFormState.hostname,
+          error,
+        },
+      };
+
+      this._domainInputRef.value.focus();
+      return;
+    }
+
+    this.domainFormState = {
+      ...this.domainFormState,
+      hostname: {
+        value: this.domainFormState.hostname.value,
+      },
+    };
+
+    // we do this to strip off unwanted parts like query parameters for instance
+    const { hostname, pathname, isWildcard } = parseDomain(hostnameValue + pathPrefixValue);
+
+    dispatchCustomEvent(this, 'add', {
+      hostname,
+      pathPrefix: pathname,
+      isWildcard,
+    });
+  }
+
+  /** @param {FormattedDomainInfo} domainInfo */
+  _onMarkPrimary(domainInfo) {
+    const { id, hostname, pathPrefix, isWildcard, isPrimary } = domainInfo;
+    return () => dispatchCustomEvent(this, 'mark-as-primary', { id, hostname, pathPrefix, isWildcard, isPrimary });
+  }
+
+  /** @param {FormattedDomainInfo} domainInfo */
+  _onDelete(domainInfo) {
+    const { id, hostname, pathPrefix, isWildcard, isPrimary } = domainInfo;
+    return () => dispatchCustomEvent(this, 'delete', { id, hostname, pathPrefix, isWildcard, isPrimary });
+  }
+
+  /** @param {Event & { detail: string }} event */
+  _onDomainInput({ detail: value }) {
+    if (this.domainFormState.type !== 'idle') {
+      return;
+    }
+
+    this.domainFormState = {
+      ...this.domainFormState,
+      hostname: {
+        ...this.domainFormState.hostname,
+        value,
+      },
+    };
+  }
+
+  /** @param {Event & { detail: string }} event */
+  _onPathPrefixInput({ detail: value }) {
+    // add "/" at the beginning in case the user forgot to type it
+    if (value.length > 0 && !value.startsWith('/')) {
+      value = '/' + value;
+    }
+    this.domainFormState = {
+      ...this.domainFormState,
+      pathPrefix: {
+        ...this.domainFormState.pathPrefix,
+        value,
+      },
+    };
+  }
+
+  /** @param {CcDomainManagementPropertyValues} changedProperties */
+  willUpdate(changedProperties) {
+    if (
+      changedProperties.has('domainFormState') &&
+      changedProperties.get('domainFormState')?.hostname.error == null &&
+      this.domainFormState.hostname.error != null
+    ) {
+      this.updateComplete.then(() => this._domainInputRef.value?.focus());
+    }
+
+    if (changedProperties.has('domainListState') && this.domainListState.type === 'loaded') {
+      // Make sure we use a copy before using .sort() which modifies the array in place
+      this._sortedDomains = [...this.domainListState.domains].sort(sortDomains).map((domainState) => {
+        const { hostname } = domainState;
+
+        /** @type {FormattedDomainInfo} */
+        const formattedDomain = {
+          ...domainState,
+          isHttpOnly: isTestDomainWithSubdomain(hostname),
+          isTestingOnly: isTestDomain(hostname),
+        };
+
+        return formattedDomain;
+      });
+    }
+  }
+
+  render() {
+    return html`
+      <div class="wrapper">
+        <cc-block>
+          <div slot="title">${i18n('cc-domain-management.main-heading')}</div>
+
+          <cc-block-section>
+            <div class="form-info">
+              <p>${i18n('cc-domain-management.form.info.docs')}</p>
+              <p>${i18n('cc-domain-management.form.info.cleverapps')}</p>
+            </div>
+            <div>${this._renderForm(this.domainFormState)}</div>
+          </cc-block-section>
+
+          <cc-block-section>
+            <div slot="title">
+              ${i18n('cc-domain-management.list.heading')}
+              ${this.domainListState.type === 'loaded'
+                ? html` <cc-badge class="domain-count" circle>${this._sortedDomains.length}</cc-badge> `
+                : ''}
+            </div>
+
+            ${this.domainListState.type === 'error'
+              ? html`
+                  <cc-notice intent="warning" message="${i18n('cc-domain-management.list.loading-error')}"></cc-notice>
+                `
+              : ''}
+            ${this.domainListState.type === 'loading' ? html` <cc-loader></cc-loader> ` : ''}
+            ${this.domainListState.type === 'loaded' ? this._renderDomains(this._sortedDomains) : ''}
+          </cc-block-section>
+        </cc-block>
+
+        <cc-block>
+          <div slot="title">${i18n('cc-domain-management.certif.heading')}</div>
+          <div class="certif-info">
+            <p>${i18n('cc-domain-management.certif.automated')}</p>
+            <p>${i18n('cc-domain-management.certif.custom')}</p>
+          </div>
+        </cc-block>
+
+        <cc-block class="dns-info">
+          <div slot="title">${i18n('cc-domain-management.dns.heading')}</div>
+          <div>
+            <div class="dns-info__desc">${i18n('cc-domain-management.dns.desc')}</div>
+            <cc-notice intent="info" heading="${i18n('cc-domain-management.dns.info.heading')}">
+              <div slot="message">
+                <p>${i18n('cc-domain-management.dns.info.load-balancer')}</p>
+                <p>${i18n('cc-domain-management.dns.info.apex')}</p>
+              </div>
+            </cc-notice>
+          </div>
+
+          ${this.dnsInfoState.type === 'loading' ? html` <cc-loader></cc-loader> ` : ''}
+          ${this.dnsInfoState.type === 'error'
+            ? html`
+                <cc-notice intent="warning" message="${i18n('cc-domain-management.dns.loading-error')}"></cc-notice>
+              `
+            : ''}
+          ${this.dnsInfoState.type === 'loaded'
+            ? this._renderDnsInfo(this.dnsInfoState.cnameRecord, this.dnsInfoState.aRecords)
+            : ''}
+        </cc-block>
+      </div>
+    `;
+  }
+
+  /**
+   * @param {DomainManagementFormState} formState
+   * @returns {TemplateResult}
+   */
+  _renderForm({ type, hostname: domain, pathPrefix }) {
+    const isDisabled = type === 'adding';
+
+    return html`
+      <form novalidate @submit=${this._onDomainSubmit}>
+        <div class="fieldgroup">
+          <cc-input-text
+            class="fieldgroup__domain"
+            label="${i18n('cc-domain-management.form.domain.label')}"
+            required
+            .value="${domain.value}"
+            ?disabled="${isDisabled}"
+            .errorMessage=${this._getErrorMessage(domain.error)}
+            ${ref(this._domainInputRef)}
+            @cc-input-text:input=${this._onDomainInput}
+            @paste=${this._onDomainPaste}
+          >
+            <p slot="help">${i18n('cc-domain-management.form.domain.help')}</p>
+          </cc-input-text>
+          <cc-input-text
+            class="fieldgroup__path"
+            label="${i18n('cc-domain-management.form.path.label')}"
+            .value="${pathPrefix.value}"
+            ?disabled="${isDisabled}"
+            @cc-input-text:input=${this._onPathPrefixInput}
+          >
+            <p slot="help">${i18n('cc-domain-management.form.path.help')}</p>
+          </cc-input-text>
+        </div>
+        <cc-button primary ?waiting="${type === 'adding'}" type="submit"
+          >${i18n('cc-domain-management.form.submit')}</cc-button
+        >
+      </form>
+    `;
+  }
+
+  /**
+   * @param {FormattedDomainInfo[]} domains
+   * @returns {TemplateResult}
+   */
+  _renderDomains(domains) {
+    const hasDomains = domains.length > 0;
+    const isOneRowMarkingPrimary = domains.filter((domain) => domain.type === 'marking-primary').length > 0;
+    const hasHttpOnlyDomains = domains.filter((domain) => domain.isHttpOnly).length > 0;
+
+    if (!hasDomains) {
+      return html`
+        <p class="empty" tabindex="-1" ${ref(this._emptyMessageRef)}>${i18n('cc-domain-management.list.empty')}</p>
+      `;
+    }
+
+    return html`
+      ${hasHttpOnlyDomains
+        ? html`
+            <cc-notice no-icon intent="warning">
+              <div slot="message" class="http-only-info">
+                <cc-badge intent="warning" weight="outlined">
+                  <div class="badge-content">
+                    <cc-icon
+                      .icon=${iconHttpOnly}
+                      a11y-name=${i18n('cc-domain-management.list.badge.http-only.alt')}
+                    ></cc-icon>
+                    ${i18n('cc-domain-management.list.badge.http-only.text')}
+                  </div>
+                </cc-badge>
+                ${i18n('cc-domain-management.list.http-only.notice')}
+              </div>
+            </cc-notice>
+          `
+        : ''}
+
+      <div class="domains">
+        ${repeat(
+          domains,
+          ({ id }) => id,
+          (domain) => this._renderDomain(domain, isOneRowMarkingPrimary),
+        )}
+      </div>
+    `;
+  }
+
+  /**
+   * @param {FormattedDomainInfo} domainInfo
+   * @param {boolean} isOneRowMarkingPrimary
+   * @returns {TemplateResult}
+   **/
+  _renderDomain(domainInfo, isOneRowMarkingPrimary) {
+    const {
+      type: domainItemStateType,
+      hostname,
+      pathPrefix,
+      isWildcard,
+      isPrimary,
+      isHttpOnly,
+      isTestingOnly,
+    } = domainInfo;
+    const waiting = domainItemStateType === 'deleting' || domainItemStateType === 'marking-primary';
+    const isMarkingPrimaryDisabled = isOneRowMarkingPrimary || domainItemStateType === 'deleting';
+    const hostWithWildcard = getHostWithWildcard(hostname, isWildcard);
+    const domainUrl = getDomainUrl(hostname, pathPrefix, isWildcard, isHttpOnly);
+
+    return html`
+      <!--
+        Caution: the primary class is used to focus the primary domain when it changes, do not remove or change this class
+        (see LostFocusController within the constructor for more info) 
+      -->
+      <div class="domain ${classMap({ primary: isPrimary, waiting })}" tabindex="-1">
+        <span class="domain-name-with-path">
+          <!-- These tags need to remain on the same line so there is no white-space when pasting domain+path -->
+          <span>${hostWithWildcard}</span><span class="path-prefix">${pathPrefix}</span>
+          <a
+            class="domain-link"
+            href="${domainUrl}"
+            title="${i18n('cc-domain-management.list.link.title', { domainUrl })}"
+            target="_blank"
+          >
+            <span class="visually-hidden">${domainUrl}</span>
+            <cc-icon .icon=${iconLink} a11y-name="${i18n('cc-domain-management.new-window')}"></cc-icon>
+          </a>
+        </span>
+
+        <div class="badges">
+          ${isPrimary
+            ? html`
+                <cc-badge>
+                  <div class="badge-content">
+                    <cc-icon .icon=${iconPrimary}></cc-icon>
+                    ${i18n('cc-domain-management.list.badge.primary')}
+                  </div>
+                </cc-badge>
+              `
+            : ''}
+          ${isTestingOnly
+            ? html`
+                <cc-badge intent="info" weight="outlined">
+                  <div class="badge-content">
+                    <cc-icon .icon=${iconTest}></cc-icon>
+                    ${i18n('cc-domain-management.list.badge.testing-only')}
+                  </div>
+                </cc-badge>
+              `
+            : ''}
+          ${isHttpOnly
+            ? html`
+                <cc-badge intent="warning" weight="outlined">
+                  <div class="badge-content">
+                    <cc-icon
+                      .icon=${iconHttpOnly}
+                      a11y-name=${i18n('cc-domain-management.list.badge.http-only.alt')}
+                    ></cc-icon>
+                    ${i18n('cc-domain-management.list.badge.http-only.text')}
+                  </div>
+                </cc-badge>
+              `
+            : ''}
+        </div>
+
+        <div class="actions">
+          ${!isPrimary
+            ? html`
+                <cc-button
+                  class="mark-primary"
+                  primary
+                  outlined
+                  ?disabled=${isMarkingPrimaryDisabled}
+                  ?waiting="${domainItemStateType === 'marking-primary'}"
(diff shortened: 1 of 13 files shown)


src/components/cc-domain-management/cc-domain-management.smart.md
```diff
@@ -0,0 +1,74 @@
...
+title: '💡 Smart'
+---
+
+# 💡 Smart `<cc-domain-management>`
+
+## ℹ️ Details
+
+<!-- TODO: -->
+<table>
+  <tr><td><strong>Component    </strong> <td><a href="https://www.clever-cloud.com/doc/clever-components/?path=/docs/🛠-organisation-cc-orga-member-list--default-story"><code>&lt;cc-orga-member-list></code></a>
```
florian-sanders-cc, line 12: TODO @florian-sanders-cc: once we agree on the component name, edit this link and remove the todo comment.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,781 @@
...
+        suggestedElement.focus();
+      }
+      else {
+        this._emptyMessageRef?.value?.focus();
+      }
+    });
+
+    new LostFocusController(this, '.mark-primary', () => {
+      /** @type {HTMLDivElement|null} */
+      const primaryDomainCard = this.shadowRoot?.querySelector('.primary');
```
florian-sanders-cc, line 115: TODO @florian-sanders-cc : use the focusBySelector helper once merged

src/lib/utils.js
```diff
@@ -230,3 +230,57 @@ export function sleep (delay) {
...
+}
+
+/**
+ * Extracts `hostname` and `pathname` from a given `domain`
+ *
+ * @param {string} domain
+ * @return {{ hostname: string, pathname: string }}
+ */
+export function extractDomainParts (domain) {
+  const domainWithHttp = domain.match(/^http(s)?:\/\//) != null ? domain : 'https://' + domain;
```
hsablonniere, line 252: suggestion: same feedback for the useless capturing group of `(s)`.

src/translations/translations.en.js
```diff
@@ -213,6 +213,64 @@ export const translations = {
...
+  'cc-domain-management.form.domain.error.empty': `Enter a domain`,
+  'cc-domain-management.form.domain.error.format': `Enter a valid domain, for instance "example.com"`,
+  'cc-domain-management.form.domain.error.wildcard': () => sanitize`Enter a valid domain.<br>Wildcards "*" may only be used as subdomains like: "*.example.com"`,
+  'cc-domain-management.form.domain.help': () => sanitize`For instance: <code>example.com</code>, <code>*.example.com</code> or <code>example.cleverapps.io</code>`,
+  'cc-domain-management.form.domain.label': `Domain name`,
+  'cc-domain-management.form.info.cleverapps': () => sanitize`If you need a domain name for testing purposes, you may use <code>my-website.cleverapps.io</code>`,
+  'cc-domain-management.form.info.docs': () => sanitize`Associate a domain name to your application. For more information, refer to our documentation <a href="https://developers.clever-cloud.com/doc/administrate/domain-names/">Domain Names</a>.`,
+  'cc-domain-management.form.info.primary': `You can choose a primary domain. Links to your application from within the console lead to this domain and emails about your application only mention this domain.`,
```
hsablonniere, line 240: question: there are no mention of the path prefix in those texts, should we add some?

src/translations/translations.fr.js
```diff
@@ -226,6 +226,65 @@ export const translations = {
   //#region cc-doc-list
   'cc-doc-list.error': `Une erreur est survenue pendant le chargement de la documentation`,
   //#endregion
+  //#region cc-domain-management
+  'cc-domain-management.certif.automated': () => sanitize`Que vous utilisiez <code>cleverapps.io</code> ou vos propres noms de domaine avec les applications hébergées par Clever Cloud, un certificat Let's Encrypt est automatiquement généré et renouvelé pour l'accès HTTPS/TLS. Vous n'avez rien à faire, tout est automatisé. Pour les cas spécifiques, reportez vous à la page <a href="https://developers.clever-cloud.com/doc/administrate/ssl/"><span lang="en">Installing TLS Certificates</span> (en Anglais)</a>.`,
```
hsablonniere, line 230: question: I never realized we capitalized `Anglais` in french, I guess it's a rule?

src/translations/translations.en.js
```diff
@@ -213,6 +213,64 @@ export const translations = {
...
+  'cc-domain-management.certif.custom': () => sanitize`You can provide your own certificate by using the <a href="https://api.clever-cloud.com/v2/certificates/new">Clever Cloud Certificate Manager</a>.`,
+  'cc-domain-management.certif.heading': `Secure your application`,
+  'cc-domain-management.dns.a.desc': () => sanitize`<p>If you choose to use <code>A</code> records, you'll need to update them yourself.</p>
+  <p>Keep an eye on our <a href=https://developers.clever-cloud.com/changelog/">changelog</a> or check our <a href="https://developers.clever-cloud.com/api/v4/#load-balancers">v4 API documentation</a> for this.</p>`,
+  'cc-domain-management.dns.a.heading': `A records`,
+  'cc-domain-management.dns.cname.desc': () => sanitize`<p>Using a <code>CNAME</code> record is highly recommended. This automatically keeps your configuration up to date.</p>`,
```
hsablonniere, line 223: question: we don't mention something like "not for apex domains", maybe it's too much?

src/translations/translations.fr.js
```diff
@@ -226,6 +226,65 @@ export const translations = {
...
+  'cc-domain-management.certif.custom': () => sanitize`Vous pouvez fournir votre propre certificat grâce au <a href="https://api.clever-cloud.com/v2/certificates/new">Gestionnaire de certificats Clever Cloud (en Anglais)</a>.`,
+  'cc-domain-management.certif.heading': `Sécurisez votre application`,
+  'cc-domain-management.dns.a.desc': () => sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, vous devrez vous-même assurer leur mise à jour.</p>
+  <p>Pensez à surveiller notre <a href="https://developers.clever-cloud.com/changelog/">changelog (en Anglais)</a> ou à utiliser notre <a href="https://developers.clever-cloud.com/api/v4/#load-balancers">API v4 - <span lang="en">Load balancers</span> (en Anglais)</a> pour cela.</p>`,
```
hsablonniere, line 234: issue: the texts in FR/EN for the links aren't the same

hsablonniere (changes requested):
Hey @florian-sanders-cc, well done on this!!! :clap: 

I created a few comments. I looked at:

* component file
* smart file
* type file
* utils file

And used the different stories in my local Storybook + console

## About the UI

As I mentioned in our chat:

* I like the card design better than the list
* Maybe the cards could be more compact with some kind of table layout
  * Only when there's one column?
* Long domains have a weird behaviour with the icons
  * I would prefer if the domain/prefix/link was on a full width line and then the badges and buttons below
* if you mark a domain as primary, the hover state gets stuck

## Naming

* Component naming
  * I think the `-management` is good as it's more than just a list
* FQDN
  * I think we should get rid of the FQDN name. The API no longer returns FQDNs since we introduced the path prefix feature a very long time ago.
  * I also think we make as if the API already returns the right details and not just a full string, this means moving everything in the API layer (not in the component and not in the main part of the smart definition)
* For CNAME and A, we should use the word `record` and not `value`

## URL/domain manipulation

* We need to consider moving all URL/domain manipulations to a specific helper so we can move it in the client and use if in the CLI
* The component knows about `cleverapps.io` in some translations but it should not know what makes a test only domain as we will add new wildcards in the future

pdesoyres-cc (changes requested):
Very nice component, and the code/state is very clear.

### About the UI

* I prefer the card layout.
* Maybe a copy button could be helpful to copy the full domain name

### About the code

* I have mosty some suggestions and nitpicks but nothing really blocking
* maybe the component could be named `cc-domains-manager`, but I'm ok with the actual naming too.

Once again, this is an awesome work, well done Florian.

src/lib/utils.js
```diff
@@ -230,3 +230,57 @@ export function sleep (delay) {
...
+/**
+ * Checks if the domain is a `cleverapps.io` domain and if it contains a subdomain.
+ * For instance `subdomain.main.cleverapps.io` is HTTP only.
+ *
+ * @param {string} domain
+ * @return {boolean} whether the domain is a `cleverapps.io` is HTTP only or not
+ */
+export function isCleverappsDomainHttpOnly (domain) {
+  return domain.endsWith('cleverapps.io') && domain.split('.').length > 3;
+}
```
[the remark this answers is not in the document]
roberttran-cc: suggestion: :shrug: 

```js
function isDomainCleverappsWithSubdomain(domain) {
  //...
}
```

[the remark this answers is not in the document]
pdesoyres-cc: suggestion:

```
function isFreeDomain(domain) {
  //...
}
```

src/components/cc-domain-management/cc-domain-management.stories.js
```diff
@@ -0,0 +1,417 @@
...
+    '93.184.216.34',
+    '93.184.216.34',
+    '93.184.216.34',
+    '93.184.216.34',
+    '93.184.216.34',
+  ],
+};
+
+export const defaultStory = makeStory(conf, {
+  items: [{
```
HeleneAmouzou, line 71: **question** : I am not sure to understand why/how you do not need to put `domainFormState` in the items for some stories (here for the *defaultStory*, but also in *dataLoadedWithMoreDomains*, *empty*, etc.) and there is still a form in those stories ?

HeleneAmouzou (changes requested):
### Abouy my review 
I have checked the code (except for the smart files) and the stories ( english + french). I only have small suggestions and a question. Otherwise, despite it was a lot to review, I thought it was pretty nice and clear. GG ! :clap: 

### About the UI 
I prefer the grid option. 

Moreover, I think the component name `cc-domain-management` is good, and I also like the idea of @pdesoyres-cc : `cc-domains-manager`.

Galimede (changes requested):
Hiya, nice work as always! :muscle: :hugs: 

Here are some thoughts below:

* Fqdn feels really weird and not self exlanatory, I think either: 
  * We should write somewhere what that means
  * Change the naming  
* I prefer the card versions UI
* The name of the component feels self explanatory, I like it

roberttran-cc (approved):
Great work on a very nice component! :100: 

I've got a few minor feedback.

And about your questions:
- although less consistent, I prefer `cc-domain-managment` as it is closer to what it actually achieves
- I don't mind keeping `fqdn` but I see your point: maybe adding a comment can help
- grid or cards: I prefer the cards style!

roberttran-cc (approved):
(edit: duplicated comment)

src/lib/utils.js
```diff
@@ -230,3 +230,57 @@ export function sleep (delay) {
...
+
+/**
+ * Extracts `hostname` and `pathname` from a given `domain`
+ *
+ * @param {string} domain
+ * @return {{ hostname: string, pathname: string }}
+ */
+export function extractDomainParts (domain) {
```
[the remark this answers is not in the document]
florian-sanders-cc: Yes this is part of an issue I've created, I thought about tests but way too late when creating the QA :roll_eyes: 
Anyway, this helper will be moved to the clever client eventually and I'll use this opportunity to add tests there :+1:

src/translations/translations.fr.js
```diff
@@ -226,6 +226,65 @@ export const translations = {
...
+  'cc-domain-management.form.domain.label': `Nom de domaine`,
+  'cc-domain-management.form.info.cleverapps': () => sanitize`Pour un nom de domaine utilisé à des fins de tests, vous pouvez utiliser librement <code>mon-site-web.cleverapps.io</code>`,
+  'cc-domain-management.form.info.docs': () => sanitize`Associez un nom de domaine à votre application. Pour plus d'informations, reportez-vous à notre documentation <a href="https://developers.clever-cloud.com/doc/administrate/domain-names/">Domain Names (en Anglais)</a>.`,
+  'cc-domain-management.form.info.primary': `Vous pouvez ensuite définir un nom de domaine principal. Les liens d'accès à votre application depuis la console s'ouvriront sur ce domaine et les e-mails liés à cette application ne mentionneront que ce domaine.`,
+  'cc-domain-management.form.path.help': () => sanitize`Par exemple: <code>/api</code> ou <code>/blog</code>`,
+  'cc-domain-management.form.path.label': `Route`,
+  'cc-domain-management.form.submit': `Ajouter le domaine`,
+  'cc-domain-management.form.submit.error': ({ domain }) => `Une erreur est survenue lors de l'ajout du nom de domaine "${domain}"`,
```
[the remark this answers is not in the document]
florian-sanders-cc: We have an issue about this: https://github.com/CleverCloud/clever-components/issues/1038 but we haven't decided yet.

I'll go for no dots when it's a single sentence.

src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -0,0 +1,333 @@
...
+            (error) => {
+              console.error(error);
+              updateComponent('domainFormState',
+                /** @param {DomainManagementFormState} domainFormState */
+                (domainFormState) => {
+                  domainFormState.type = 'idle';
+                });
+
+              // FIXME: ask the API for a proper id to map a message to an error
+              if (error.message === 'Invalid domain' || error.message.includes('is invalid')) {
```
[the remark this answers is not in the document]
florian-sanders-cc: I don't understand what you mean by `ack` :thinking: 
Anyway, I've moved these into a `convertApiError` helper like you did for `email-list`, WDYT?

src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -0,0 +1,333 @@
...
+      /**
+       * @param {[RawDomainFromApi, RawDomainFromApi[]]} payload
+       * @returns {DomainStateIdle[]}
+       */
+      ([primaryDomain, domains]) => {
+        return domains
+          .map(({ fqdn }) => {
+
+            /** @type {DomainStateIdle} */
+            const domainState = {
```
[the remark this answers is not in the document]
florian-sanders-cc: I've removed one or two redundant local variables where I could but you'll see that some of them are still here.

There are cases where they are the only way to benefit from typechecking + refactoring features of Typescript.
For instance:
```ts
/**
 * @param {Object} params
 * @param {ApiConfig} params.apiConfig
 * @param {string} params.ownerId
 * @param {string} params.appId
 * @param {AbortSignal} params.signal
 * @returns {Promise<DomainStateIdle[]>}
 */
function getDomains ({ apiConfig, ownerId, appId, signal }) {
  return Promise.all([
    fetchPrimaryDomain(ownerId, appId, apiConfig, signal),
    fetchDomains(ownerId, appId, apiConfig),
  ])
    .then(([primaryDomain, domains]) => {
      return domains.map((domainState) => ({
        ...domainState,
        toto: 'titi',
        type: 'idle',
        isPrimary: primaryDomain?.fqdn === domainState.id,
      }));
    });
}
```
This won't trigger any TS error even though the `toto` property is invalid + you won't benefit from refactoring.
If you do this instead, you get TypeChecking + refactoring features:
```ts
/**
 * @param {Object} params
 * @param {ApiConfig} params.apiConfig
 * @param {string} params.ownerId
 * @param {string} params.appId
 * @param {AbortSignal} params.signal
 * @returns {Promise<DomainStateIdle[]>}
 */
function getDomains ({ apiConfig, ownerId, appId, signal }) {
  return Promise.all([
    fetchPrimaryDomain(ownerId, appId, apiConfig, signal),
    fetchDomains(ownerId, appId, apiConfig),
  ])
    .then(([primaryDomain, domains]) => {
      /** @type {DomainStateIdle[]} */
      const formattedDomains = domains.map((domainState) => ({
        ...domainState,
        type: 'idle',
        isPrimary: primaryDomain?.fqdn === domainState.id,
      }));

      return formattedDomains;
    });
}
```
Mostly within promises where we transform the final response and sometimes when destructuring an object.

Of course some of these occurrences may be due to improper typing of the promise in the first place but I couldn't find a better solution :shrug:

src/components/cc-domain-management/cc-domain-management.smart.js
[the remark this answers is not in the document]
pdesoyres-cc: By ack, I mean we should not have to do this horrible transformation on the UI side. I don't particularly appreciate having that in the event handling layer. So, moving that to the API layer would make this hack less horrible.
So, what you did is good.

src/components/cc-domain-management/cc-domain-management.smart.js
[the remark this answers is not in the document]
pdesoyres-cc: Thank you for the clarification. We need to have that in mind when reworking the typing of the smart component API.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,849 @@
...
+        ...this.domainFormState.domain,
+        value,
+      },
+    };
+  }
+
+  /**
+   * @param {Event & { detail: string }} event
+   */
+  _onPathPrefixInput ({ detail: value }) {
```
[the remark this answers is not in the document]
florian-sanders-cc: Forgot to add a condition, should be fixed for next review phase.
Thanks!

src/components/cc-domain-management/cc-domain-management.stories.js
florian-sanders-cc: The `domainFormState` property has a default value (empty values for every form control + no error).
This means we only have to set the `domainFormState` when we want the component to be in a different state than empty form controls.

Your comment was on point, I was setting `domainFormState` to the same value as the default in many stories where it wasn't needed.
It should be fixed now (only stories with form values / form errors should have items with `domainFormState` properties).

src/components/cc-domain-management/cc-domain-management.types.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  pathPrefix: {
+    value: string;
+  };
+}
+
+interface DomainManagementFormStateAdding {
+  type: 'adding';
+  domain: {
+    value: string;
+    error?: null;
```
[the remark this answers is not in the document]
florian-sanders-cc: sometimes you reset errors by setting them to `null`.
Also if you remove the error property from the type here and you try to do something like this : 
`this._getErrorMessage(domain.error)`

TypeScript will throw an error because you are trying to access `error` which does not exist in waiting state. Setting this to `null` fixes that :man_shrugging:

src/translations/translations.en.js
```diff
@@ -213,6 +213,64 @@ export const translations = {
...
+  'cc-domain-management.list.empty': `No domain associated to this application`,
+  'cc-domain-management.list.error-not-found.heading': `Domain not found`,
+  'cc-domain-management.list.error-not-found.text': ({ domain }) => sanitize`<p>The "${domain}" domain may have been removed by someone else after loading the list of domains.<p>
+  <p>Please <strong>refresh your page</strong> to get the updated list of domains.</p>`,
+  'cc-domain-management.list.heading': `Domain names linked to this application`,
+  'cc-domain-management.list.http-only.notice': () => sanitize`Subdomains are not covered by the SSL certificate of <code>*.cleverapps.io</code>. They do not support HTTPS.`,
+  'cc-domain-management.list.link.title': ({ domainUrl }) => `Open - ${domainUrl} - new window`,
+  'cc-domain-management.list.loading-error': `Something went wrong while loading domains associated to this application.`,
+  'cc-domain-management.list.primary.error': ({ domain }) => `Something went wrong while trying to mark "${domain}" as primary domain.`,
+  'cc-domain-management.list.primary.success': ({ domain }) => `The "${domain}" domain has been successfully marked as primary domain.`,
```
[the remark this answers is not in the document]
florian-sanders-cc: Good idea, I tried to do the same for every other occurrences.
Thanks!

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,844 @@
...
+    const waiting = domainItemStateType === 'deleting' || domainItemStateType === 'marking-primary';
+    const isMarkingPrimaryDisabled = isOneRowMarkingPrimary || domainItemStateType === 'deleting';
+    const domainUrl = getDomainUrl(fqdn, isHttpOnly);
+
+    return html`
+      <!--
+        Caution: the primary class is used to focus the primary domain when it changes, do not remove or change this class
+        (see LostFocusController within the constructor for more info) 
+      -->
+      <div class="domain ${classMap({ primary: isPrimary, waiting })}" tabindex="-1" >
```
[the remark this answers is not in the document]
florian-sanders-cc: I'm now using the new `focusBySelector` helper but it shouldn't change much / fix the issue.

I'll try to play with the component to see if I can reproduce.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,844 @@
...
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { linkStyles } from '../../templates/cc-link/cc-link.js';
+import { extractDomainParts, getDomainUrl, isCleverappsDomainHttpOnly, sortDomains } from './domains.js';
+import '../cc-button/cc-button.js';
+import '../cc-block/cc-block.js';
+import '../cc-block-section/cc-block-section.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-notice/cc-notice.js';
```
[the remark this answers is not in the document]
florian-sanders-cc: Indeed! I'll let prettier handle that right before fixing up every feedback commit because yeah, I'm lazy :sweat_smile:

.eslintrc.cjs
```diff
@@ -2,7 +2,6 @@ module.exports = {
   'extends': [
     'standard',
     'plugin:lit-a11y/recommended',
-    'prettier',
```
florian-sanders-cc, line 5: This is temporary you can ignore that

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -106,99 +116,77 @@ export class CcDomainManagement extends LitElement {
...
-  /**
-   * @param {string} domain
+   * @param {string} hostname
    * @returns {FormError|null} The error code or `null` if no error
    */
-  _getErrorCode (domain) {
-    if (domain.length === 0) {
-      return 'empty';
-    }
+  _getErrorCode (hostname) {
```
florian-sanders-cc, line 132: Dicussed in Sync with Pierre (thanks again :heart:)
The `parseDomain` helper throws errors when the domain to parse:
- is empty, 
- has invalid wildcard (`toto*.example.com` is invalid, "*" may only be used as a subdomain, it cannot be part of a subdomain)
- has invalid format.

This is why we use try / catch here.

The `hostname-contains-path` is not part of what the `parseDomain` checks because it's very specific. 
It's only a problem within this component and only when users try to enter a path within the domain input field. In other contexts, domains containing a path are fine since it's a feature Clever Cloud supports.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -106,99 +116,77 @@ export class CcDomainManagement extends LitElement {
...
-        return i18n('cc-domain-management.form.domain.error.empty');
-      case 'invalid-format':
-        return i18n('cc-domain-management.form.domain.error.format');
-      case 'invalid-wildcard':
-        return i18n('cc-domain-management.form.domain.error.wildcard');
-      case 'contains-path':
-        return i18n('cc-domain-management.form.domain.error.contains-path', { path: this._lastPathWithinDomainInputValue });
-      default:
-        return null;
+  _getErrorMessage (error) {
```
florian-sanders-cc, line 156: This has been refactored because `error` may be an object when the error is of type `hostname-contains-path`.

The error object in this context allows us to store the `pathWithinHostname` value to show it within the error. I used to run `parseDomain` twice to get the `pathWithinHostname` value (once when checking the error and once when setting the global reactive prop to store the value).

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,849 @@
...
+      this.updateComplete.then(() => this._domainInputRef.value?.focus());
+    }
+
+    if (changedProperties.has('domainListState') && this.domainListState.type === 'loaded') {
+      this._sortedDomains = this.domainListState.domains
+        .map((domain) => {
+          const { hostname, pathname } = extractDomainParts(domain.fqdn);
+
+          /** @type {FormattedDomainInfo} */
+          const formattedDomain = {
```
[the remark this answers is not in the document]
florian-sanders-cc: this is one of those cases where it's necessary :disappointed:

src/components/cc-domain-management/cc-domain-management.stories.js
HeleneAmouzou: I go it, thanks for the explanations ! :pray:

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+    }
+    catch (error) {
+      if (error instanceof DomainError) {
+        return error.message;
+      }
+
+      return 'invalid-format';
+    }
+
+  }
```
florian-sanders-cc, line 148: Dicussed in Sync with Pierre (thanks again ❤️)
The parseDomain helper throws errors when the domain to parse:

    is empty,
    has invalid wildcard (toto*.example.com is invalid, "*" may only be used as a subdomain, it cannot be part of a subdomain)
    has invalid format.

This is why we use try / catch here.

The hostname-contains-path is not part of what the parseDomain checks because it's very specific.
It's only a problem within this component and only when users try to enter a path within the domain input field. In other contexts, domains containing a path are fine since it's a feature Clever Cloud supports.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+
+  /**
+   * @param {DomainManagementFormState} formState
+   * @returns {TemplateResult}
+   */
+  _renderForm ({ type, hostname: domain, pathPrefix }) {
+    const isDisabled = type === 'adding';
+
+    return html`
+      <form novalidate @submit=${this._onDomainSubmit}>
```
pdesoyres-cc, line 402: question: you don't use our directive. Did we miss something when we designed the form framework? Or do you plan to use it in a next iteration?

HeleneAmouzou (approved):
LGTM ! Once again GG :clap:

pdesoyres-cc (approved):
Once again, very nice work! I left some nitpicks and suggestions. Do whatever you want with them.

src/lib/domain.js
```diff
@@ -0,0 +1,122 @@
+/**
```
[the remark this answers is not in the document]
florian-sanders-cc: This is the plan but I'll do it when I move these from this project to the client (mostly because I'm still a beginner with tests and it takes me some time so I'd rather not delay the component for that).
Next time I'll create tests when creating the helpers :roll_eyes:

src/lib/i18n-sanitize.js
```diff
@@ -2,7 +2,7 @@
 const AUTHORIZED_TAGS = ['STRONG', 'EM', 'CODE', 'A', 'BR', 'P'];
 
 function isAuthorizedAttribute(attributeName, tagName) {
-  return attributeName === 'title' || attributeName === 'aria-label' || (attributeName === 'href' && tagName === 'A');
+  return attributeName === 'title' || attributeName === 'aria-label' || (attributeName === 'href' && tagName === 'A') || attributeName === 'lang';
```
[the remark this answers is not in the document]
florian-sanders-cc: I didn't pay any attention to formatting because prettier will handle it (I'll run it after the review phase is over just to avoid cluttering diffs)

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+        </div>
+      </cc-block-section>
+    `;
+  }
+
+  static get styles () {
+    return [
+      accessibilityStyles,
+      linkStyles,
+      css`
```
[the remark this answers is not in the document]
florian-sanders-cc: same as formatting, when rebasing the CSS properties ordering rules changed and I decided not to fix issues before the review phase is over to avoid cluttering the diffs

src/components/cc-domain-management/cc-domain-management.js
florian-sanders-cc: Yes this is temporary, the directive landed in master right when the component was going into review so I've created an issue about this and it's probably one of the first subjects I'll tackle after shipping the component

roberttran-cc (approved):

src/components/cc-domain-management/cc-domain-management.types.d.ts
```diff
@@ -0,0 +1,89 @@
...
+  hostname: {
+    value: string;
+    error?: null;
+  };
+  pathPrefix: {
+    value: string;
+  };
+}
+
+export type FormError = 'empty' | 'invalid-format' | 'invalid-wildcard' | HostnameContainsPathError;
```
hsablonniere, line 25: question: is it a problem to have both string or object?

src/lib/domain.js
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!isWildcard && domain.includes('*')) {
+      throw new DomainError('invalid-wildcard');
+    }
+
+    throw new DomainError('invalid-format');
+  }
+}
+
+export class DomainError extends Error {
```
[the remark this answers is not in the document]
hsablonniere: * :+1: for the `super(message)`
* not really convinced with the getter as most properties on an error can already be modified so it's more code for not so useful safety

src/lib/domain.js
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Checks if the domain is a `cleverapps.io` domain and if it contains a subdomain.
+ * For instance `subdomain.main.cleverapps.io` is HTTP only.
+ *
+ * @param {string} domain
+ * @return {boolean} whether the domain is a `cleverapps.io` is HTTP only or not
+ */
+export function isTestDomainWithSubdomain (domain) {
+  return domain.endsWith('cleverapps.io') && domain.split('.').length > 3;
```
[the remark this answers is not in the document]
hsablonniere: :+1: for using `isTestDomain` but this means keeping the dot split instead of the regex (I think it's OK). Also, we should always have a `$` at the end for such regex.

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+
+/**
+ * A component to manage domains associated to an application.
+ *
+ * @cssdisplay block
+ *
+ * @fires {CustomEvent<{ hostname: string, pathPrefix: string }>} cc-domain-management:add - Fires when clicking the "Add a domain" button.
+ * @fires {CustomEvent<DomainInfo>} cc-domain-management:mark-as-primary - Fires when clicking a "mark as primary" button.
+ * @fires {CustomEvent<DomainInfo>} cc-domain-management:delete - Fires when clicking a "delete" button.
+ *
```
hsablonniere, line 58: nitpick: can we remove this "empty" line?

hsablonniere (changes requested):
Hey @florian-sanders-cc, awesome work on this!! :clap: 

Thank you team for your very interesting feedbacks :muscle: 

I added a few comments but nothing special. Also, well done for the `lib/domain.js` :+1:

src/lib/domain.js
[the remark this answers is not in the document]
florian-sanders-cc: I'll stick with `isTestDomain` + split.
We'll have to revisit this code if / when we add new test domains in the future anyway

src/components/cc-domain-management/cc-domain-management.types.d.ts
florian-sanders-cc: I don't have a definite answer for that, I felt it was weird but it does create issues.

The alternative would be to go for objects for all of these but it also feels weird to go for objects like this:
```js
{ code: "empty" }
```

This solution however makes dealing with FormErrors a little bit simpler because we know they always have the same shape...

If you're fine with that I think I'd go with that (all errors are objects with at least a `code` property).

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+      hostname,
+      pathPrefix: pathname,
+      isWildcard,
+    });
+  }
+
+  /** @param {FormattedDomainInfo} domainInfo */
+  _onMarkPrimary (domainInfo) {
+    const { id, hostname, pathPrefix, isWildcard, isPrimary } = domainInfo;
+    return () => dispatchCustomEvent(this, 'mark-as-primary', { id, hostname, pathPrefix, isWildcard, isPrimary });
```
Galimede, line 253: question: why do we need to return an arrow function here? :thinking:

src/components/cc-domain-management/cc-domain-management.types.d.ts
hsablonniere: Go for "all errors are objects" :muscle:

src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -0,0 +1,396 @@
...
+        updateComponent('domainFormState',
+          /** @param {DomainManagementFormState} domainFormState */
+          (domainFormState) => {
+            domainFormState.type = 'adding';
+          });
+
+        createNewDomain({ apiConfig, ownerId, appId, hostname, pathPrefix, isWildcard })
+          .then(() => {
+            if (isTestDomain(hostname)) {
+              notifySuccess(i18n('cc-domain-management.form.submit.success', { domain: domainWithPathAndWildcard }));
```
Galimede, line 84: question: the message is the same if it's a test domain or not. The only change seems to be the options for the notice and a `p` for the other translated message/notice below. Could you explain why the behavior is not the same? I'm not really sure I understand the point. :thinking:

src/components/cc-domain-management/cc-domain-management.smart.md
```diff
@@ -0,0 +1,68 @@
...
+kind: '🛠 Domains/<cc-domain-management>'
+title: '💡 Smart'
+---
+
+# 💡 Smart `<cc-domain-management>`
+
+## ℹ️ Details
+
+<table>
+  <tr><td><strong>Component    </strong> <td><a href="https://www.clever-cloud.com/doc/clever-components/?path=/docs/🛠-domains-cc-domain-list--default-story"><code>&lt;cc-domain-management&gt;</code></a>
```
Galimede, line 11: nitpick: the link points to `cc-domain-list` instead of `cc-domain-management`

.eslintrc.cjs
Galimede, line 1: issue: the config shouldn't have changed :thinking:

src/components/cc-domain-management/cc-domain-management.types.d.ts
Galimede: +1, it would make it clearer

Galimede (dismissed):
Hey, nice work Florian! :muscle: 

I've only left a few questions and nitpicks, but nice work on this component, it's impressive. :exploding_head:

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -0,0 +1,819 @@
...
+    if (error === 'invalid-format') {
+      return i18n('cc-domain-management.form.domain.error.format');
+    }
+    if (error === 'invalid-wildcard') {
+      return i18n('cc-domain-management.form.domain.error.wildcard');
+    }
+    if (error?.code != null && error?.code === 'hostname-contains-path') {
+      return i18n('cc-domain-management.form.domain.error.contains-path', { path: error.pathWithinHostname });
+    }
+    return null;
```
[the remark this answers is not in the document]
florian-sanders-cc: well it wasn't possible because `error` could be a `string` or an object so typical `switch` was not really appropriate (or you'd have to use the `switch(true) ...` pattern which I personally don't like :roll_eyes: )

Since I've moved to errors always being objects with codes I can now use a switch as you suggest :+1:

src/components/cc-domain-management/cc-domain-management.smart.js
florian-sanders-cc: It should behave as follows: 
- if the domain is a custom domain, the info message tells you to config and we want the message to remain visible because the domain association won't work unless you configure your DNS (same idea as when we leave the message visible when  you add a secondary email because it needs to be validated).
- if the domain is a test domain, there's nothing to do, it just works so we go for a simple success message.

src/components/cc-domain-management/cc-domain-management.smart.md
florian-sanders-cc: nice catch, I would definitely have missed this one :roll_eyes:

.eslintrc.cjs
florian-sanders-cc: Yeah this is a temporary commit to prevent eslint errors until I run the new eslint config.
I didn't want to run the new eslint config before the reviews were done to avoid cluttering the diffs.

This commit & this change is going to be removed don't worry (the commit message is actually a todo for me :smile: )

src/components/cc-domain-management/cc-domain-management.js
florian-sanders-cc: This is an equivalent pattern for this case
```tsx
          <cc-button
            class="delete-domain"
            ...
            @cc-button:click=${() => this._onDelete(domainInfo)}
          >
            ${i18n('cc-domain-management.list.btn.delete.text')}
          </cc-button>
```

You want to be able to pass arguments from the render to the callback listener.

For this case I could have sticked with the pattern above but when you also want to pass the event, TypeScript will complain about the event type and you end up with this:

```tsx
          <cc-button
            class="delete-domain"
            ...
            @cc-button:click=${
              /** @param {Event} */
              (event) => this._onDelete(event, domainInfo)}
            }
          >
            ${i18n('cc-domain-management.list.btn.delete.text')}
          </cc-button>
```

Since we don't like that we discussed it with Hubert and he suggested this solution instead. This pattern is already used here if you want another example: https://github.com/CleverCloud/clever-components/blob/master/src/components/cc-addon-backups/cc-addon-backups.js#L233

We can discuss it in sync if it's unclear and we should probably discuss it in team anyway :man_shrugging:

roberttran-cc: I may be wrong but I think there's a difference between those two implementations:
- when `_onMarkPrimary()` returns an arrow function, the `const { ... } = domainInfo;` line is executed at runtime, when the component is displayed,
- when `_onMarkPrimary()` returns nothing and simply dispatch the event, the `const { ... } = domainInfo;` line is executed when the click occurs.

It's because the HTML template differs as well:
- with the arrow function: `@cc-button:click=${this._onMarkPrimary(domainInfo)}`,
- without the arrow function: `@cc-button:click=${(event) => this._onMarkPrimary(event, domainInfo)}`.

A caveat with the arrow function is that, if the data tied to the DOM element (here, it's `domainInfo`) is updated dynamically from its initial value during the component lifetime, the data fired alongside the event will mismatch (I'm not sure though, should be tested).

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -106,99 +109,82 @@ export class CcDomainManagement extends LitElement {
...
-    }
+      if (!hostnameValue.match(/.+\..+/)) {
+        return 'invalid-format';
+      }
 
-    if (domain.includes('*') && !domain.startsWith('*.')) {
-      return 'invalid-wildcard';
-    }
+      if (pathname !== '/') {
+        /** We store the path so we can tell users to move this part to the "" input field instead. */
```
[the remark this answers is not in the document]
florian-sanders-cc: Yes indeed, should be fixed, thanks!

src/components/cc-domain-management/cc-domain-management.js
florian-sanders-cc: This is something I wanted to test as well when we initially used it within the `cc-addon-backups` but I didn't have time + I felt the only case where it might happen was if the piece of data we wanted to send was not part of a reactive property which is very unlikely :thinking:

PR #1093, opened 2024-06-21
fix(cc-tile-metrics): use the correct value for average memory usage within last 24h

### LINKED ISSUE #1092: cc-tile-metrics: displayed value for the average memory used within the last hour is incorrect
## Context

Since #1059, the `cc-tile-metrics` displays the average CPU within the last hour instead of the average memory value.

cc-tile-metrics showing the same value for last 24H memory and cpu usage even though charts highlight different values

florian-sanders-cc:
Fixes #1092

## What does this PR do?

- The component displayed the average CPU usage within last 24h instead of memory usage (see screenshot below). This PR fixes this issue.
cc-tile-metrics showing the same value for last 24H memory and cpu usage even though charts highlight different values

## How to review?

- Check the code,
- Check locally with `demo-smart`,


pdesoyres-cc (approved):
Nice catch

Galimede (approved):

Issue #674, opened 2022-12-16
Update tasks reference


hsablonniere:
Document tasks that aren't documented:

* [x] `components:check-type-imports` @Galimede 
* [x] `components:generate-icons-assets` @roberttran-cc 
* [x] `preview:get` @pdesoyres-cc 
* [x] `stylelint` @Galimede 
* [x] `stylelint:ci` @Galimede 
* [x] `stylelint:fix` @Galimede 
* [x] `test:cem` ??
* [x] `test:cem:watch` ??
* [x] `test:watch` ??
* [x] `typecheck` @florian-sanders-cc or @Galimede 

Complete some documentation with links:

* [x] `cdn-preview:ui` @pdesoyres-cc 
  * link to detailed build documentation about the CDN mode
* [x] `cdn-release:build` @pdesoyres-cc 
  * link to detailed build documentation
* [x] `cdn-release:publish` @pdesoyres-cc 
  * link to detailed build documentation about the CDN mode
* [x] `cdn-release:ui` @pdesoyres-cc 
  * link to detailed build documentation about the CDN mode
* [x] `components:build` ??
  * link to detailed build documentation

PR #1116, opened 2024-07-17
Add "typecheck:stats" task to display how many files are typechecked

hsablonniere:
Display something like this:

image

We can always adjust the categories and patterns in the future...

This PR also does some cleanup in the tasks defined in `package.json` and documented in `tasks.md`. The rest of the documentation update will be done through #674.


tasks/generate-icons-assets.js
```diff
@@ -27,9 +27,9 @@ async function run() {
  * @param {object} config config defining icons to be built
  * @param {string} config.namespace icon namespace
  * @param {string} config.glob icon files glob
- * @param {array<object>|null} config.svgoPlugins plugins to pass to SVGO
+ * @param {Array<object>|null} config.svgoPlugins plugins to pass to SVGO
```
pdesoyres-cc, line 30: nitpick: maybe split changes in that file into a separate commit

tasks/typechecking-stats.js
```diff
@@ -0,0 +1,67 @@
+import rawGlob from 'glob';
+import json5 from 'json5';
+import fs from 'node:fs/promises';
+import { promisify } from 'node:util';
+
+const globOne = promisify(rawGlob);
+
+/**
+ * @param {string[]} patterns
```
pdesoyres-cc, line 9: nitpick: I prefer `Array<string>` instead of `string[]`. Do what you want with that.

pdesoyres-cc (changes requested):
Nice. I have just a few comments.

florian-sanders-cc (approved):
Thanks for this command :heart_eyes_cat: 

I think it does a great good job and I like the categories you have selected :+1:

We may need to revisit it later on when we have many components included and we decide to rely on `exclude` and / or glob patterns.

HeleneAmouzou (approved):
Thanks and GG ! :clap: 
This PR made me discover new files in the `clever-components` project !

tasks/typechecking-stats.js
hsablonniere: I have two wishes:

* that we (the JS world) did not have both similar options in the first place
* that we (the component maintainers) always did the same

I updated to `Array<string>` :wink:

PR #1121, opened 2024-07-22
chore(test): force using en_US locale on browser

pdesoyres-cc:


Galimede (approved):
Thanks for the fix, it works now! :muscle: :tada:

PR #1131, opened 2024-07-30
fix(cc-input-text): match text and input

### LINKED ISSUE #1130: cc-input-text: fix text and area mismatch
The text and the span area where the text is supposed to be displayed are offset.

Screenshot from 2024-07-30 16-00-07

HeleneAmouzou:
Fixes #1130

### How to review ?
- Check the code.
- Check the `cc-input-text` stories in the preview.
- Check some components using `cc-input-text` (for instance `cc-addon-admin`).

2 reviewers should be enough.


florian-sanders-cc (approved):
Nice catch & nice reactivity! Thanks :rocket:

PR #1135, opened 2024-07-31
feat(cc-domain-management): display all A Records within a single input

### LINKED ISSUE #1134: cc-domain-management: make it possible to copy all A Records at once
## Context

Currently, each A Record is displayed within a separate `cc-input-text`. 

This is inherited from the legacy screen and has not been reconsidered properly to improve UX experience.

When dealing with A Records, people may want to add all records at once and the current UX makes this fairly painful.

We should explore using a single `cc-input-text` with `multi` (`textarea` mode) to contain all A Records. This way users can copy all values at once and format them properly for their registrar.

florian-sanders-cc:
Fixes #1134

## What does this PR do?

- Changes the display of A Record values within the `DNS configuration` section to show all of them within a single `<cc-input-text>` element in `multi` (`textarea` mode).
- This makes it possible for users to copy all values at once.

## How to review?

- Review the commit,
- Check the [preview (bottom of the screen, within the DNS config section)](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/textarea-for-a-records/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--data-loaded-with-http-only-domain)
- Give your opinion on whether you want to see this merged or not :
  - :+1: Yes merge this, it's better this way.
  - :-1: No, don't merge this. We should stick with separate `<cc-input-text>`  for each A record value + add a global copy button.
- Everybody should review since I want as many opinions as possible :wink:

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -592,17 +592,14 @@ export class CcDomainManagement extends LitElement {
         <div slot="title">${i18n('cc-domain-management.dns.a.heading')}</div>
         <div slot="info">${i18n('cc-domain-management.dns.a.desc')}</div>
         <div class="a-records">
-          ${aRecords.map(
-            (aRecord, index) => html`
-              <cc-input-text
-                hidden-label
-                label="${i18n('cc-domain-management.dns.a.label', { index: index + 1 })}"
-                readonly
-                clipboard
-                value=${aRecord}
-              ></cc-input-text>
-            `,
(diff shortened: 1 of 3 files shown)


roberttran-cc (approved):

pdesoyres-cc (approved):
Nice, I like it like that

HeleneAmouzou (approved):
LGTM ! :clap:

Galimede (approved):
LGTM, thanks! :muscle:

PR #1140, opened 2024-08-05
Web test runner: Mocha like reporter

florian-sanders-cc:
## What does this PR do?

- upgrades `@web/test-runner` to `0.18.3` (+ its dependencies to latest versions),
- use `summaryReporter` to get a reporter with the same style as Mocha (only when not in CI),
- filter irrelevant logs.

## How to review?

- Check the commit,
- Check the test results from the latest action, the report should look the same as before without irrelevant browser logs,
- Run tests locally to see the report (should be more detailed than usually).


roberttran-cc (approved):
Agree for having this output for local tests only!

Galimede (approved):
LGTM, that's nice! :muscle: 

+1 to keep this reporter local. :+1:

hsablonniere (approved):
LGTM :+1: 

Agreed on the "only for local tests".

pdesoyres-cc (approved):
LGTM

PR #1148, opened 2024-09-05
fix(cem-types-plugin): ignore non relative path imports

Galimede:
## What does this PR do?

* When using bare imports in the types, the plugin tried to look up in the corresponding file the subtypes. However, we don't want that and only want to use the type from the bare import.
* This PR also makes the plugin exit to `1` if an error has been detected.


florian-sanders-cc:
LGTM! Thanks for fixing this issue so quickly :rocket:

florian-sanders-cc (approved):

PR #1136, opened 2024-08-01
chore: typecheck component in CI

florian-sanders-cc:

PR #1152, opened 2024-09-12
refactor: Typecheck smart components in CI

florian-sanders-cc:
## What does this PR do?

- Fixes typechecking issues within most of smart components,
- Adds all smart components that pass typechecking to the CI typechecking.

## How to review?

- Check every commit starting from `refactor(cc-smart-container): typecheck`,
- That's it, CI takes care of the rest :wink: 

## TODO

- [x] Merge #1136 and rebase this one after that

### DIFF
--- a/src/components/cc-email-list/cc-email-list.smart.js
+++ b/src/components/cc-email-list/cc-email-list.smart.js
@@ -1,9 +1,6 @@
-import {
-  todo_addEmailAddress as addEmailAddress,
-  todo_getEmailAddresses as getEmailAddresses,
-  todo_removeEmailAddress as removeEmailAddress,
-  todo_getConfirmationEmail as sendConfirmationEmail,
-} from '@clevercloud/client/esm/api/v2/user.js';
+// prettier-ignore
+// @ts-expect-error FIXME: remove when clever-client exports types
+import { todo_addEmailAddress as addEmailAddress,todo_getEmailAddresses as getEmailAddresses,todo_removeEmailAddress as removeEmailAddress,todo_getConfirmationEmail as sendConfirmationEmail,} from '@clevercloud/client/esm/api/v2/user.js';
 import { defineSmartComponent } from '../../lib/define-smart-component.js';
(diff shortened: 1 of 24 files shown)


src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -1,9 +1,6 @@
-import {
-  todo_addEmailAddress as addEmailAddress,
-  todo_getEmailAddresses as getEmailAddresses,
-  todo_removeEmailAddress as removeEmailAddress,
-  todo_getConfirmationEmail as sendConfirmationEmail,
-} from '@clevercloud/client/esm/api/v2/user.js';
+// prettier-ignore
```
pdesoyres-cc, line 1: question: why prettier-ignore ?

src/components/cc-email-list/cc-email-list.smart.js
Galimede, line 1: typo: `cc-email-list` for the name of the commit :wink:

Galimede (approved):
LGTM! :muscle:

src/components/cc-email-list/cc-email-list.smart.js
florian-sanders-cc: :joy: brace yourself, here comes the explanation:
- I have an import from `clever-client.js`,
- I want to add a `@ts-expect-error` because it does not come with types,
- The `@ts-expect-error` must always be added right on top of the error you want to silence,
- So when the import spans through several lines, I would like the `@ts-expect-error` to be just above `from '...';`,
- When I add the comment, `prettier` removes it,
- If I try to `// prettier-ignore` and then `@ts-expect-error` within a multi-line import, `prettier` removes both comments (I guess no comment is allowed within imports whatsoever).

=> Hence I force the import to be single-line + I add `prettier-ignore` so that `prettier` does not make it multi-line and then I add `@ts-expect-error` :man-shrugging:

src/components/cc-email-list/cc-email-list.smart.js
florian-sanders-cc: nice catch thanks!

PR #1153, opened 2024-09-13
refactor(cc-block)!: rework component and introduce slot stack

### LINKED ISSUE #225: cc-block: refactor and introduce slot stack

HeleneAmouzou:
BREAKING CHANGE: the properties have changed
- `noHead`: property has been deleted
- `state`: property has been renamed to `toggle`
- `overlay`: property has been deleted

Fixes #225


pdesoyres-cc (changes requested):
Well done Hélène. This component will really help in standardizing our layouts.
See my comments below.

src/components/cc-block/cc-block.js
```diff
@@ -47,149 +55,138 @@ export class CcBlock extends LitElement {
...
+                    primary
+                    @cc-button:click=${this._onClickToggle}
+                    >${isToggleOpen ? i18n('cc-block.toggle.close') : i18n('cc-block.toggle.open')}
+                  </cc-button>
+                `
+              : ''}
+          </div>
+        </slot>
+
+        <slot name="content" class="${classMap({ hidden: areContentAndFooterHidden })}" ${hasSlottedChildren()}>
```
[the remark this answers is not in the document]
Galimede: +1, moreover there's a mention of `cc-expand` in the component documentation above, though there's no `cc-expand` now :thinking:
cf:
> * * The main section is wrapped in a `<cc-expand>` so variation of this section height will be animated.

Galimede (commented):
Nice work Hélène! :tada: 
Beside Pierre's comments I only have a suggestion. 
Can't wait to see this live! :smile:

florian-sanders-cc (changes requested):
Great job @HeleneAmouzou I've learned so much about slots by reviewing the component and trying alternative solutions.

My feedback mostly contains nitpicks and an accessibility subject we can tackle together in sync (could be done after the merge of this PR actually if you prefer).

I mostly reviewed the new `cc-block` itself so It's only a partial review to get you going, I still have to review the implementation of the new `cc-block` within components. 

Anyway, great job again :rocket:

src/components/cc-env-var-form/cc-env-var-form.stories.js
```diff
@@ -153,9 +153,12 @@ export const saving = makeStory(conf, {
     {
       appName: 'Foobar backend python',
       context: 'env-var',
-      state: { type: 'saving', validationMode: 'simple', variables: VARIABLES_FULL },
+      state: { type: 'loaded', validationMode: 'simple', variables: VARIABLES_FULL },
     },
   ],
+  onUpdateComplete: (component) => {
```
florian-sanders-cc, line 159: **nitpick:** let's add a small comment explaining why we did that so that nobody tries to remove it :sweat_smile: 
```suggestion
  // the component needs to be loaded with data for it to display this data in saving mode afterward
  onUpdateComplete: (component) => {
```

(don't hesitate to improve the text :wink:)

src/components/cc-env-var-form/cc-env-var-form.stories.js
```diff
@@ -153,9 +153,12 @@ export const saving = makeStory(conf, {
     {
```
florian-sanders-cc, line 153: **fix:** general feedback about the `no-context` story, I'm not sure it's something you wanted but the `cc-toggle` is no longer centered within this story:
Prod:
image


Preview:
image

src/components/cc-header-orga/cc-header-orga.js
```diff
@@ -5,6 +5,7 @@ import {
   iconRemixPhoneFill as iconPhone,
```
florian-sanders-cc, line 5: **fix:** in loading mode, the `cc-header-orga` story shows an empty footer, let's investigate in sync when you have time
image

src/components/cc-invoice/cc-invoice.js
```diff
@@ -51,22 +51,22 @@ export class CcInvoice extends LitElement {
...
-              <div slot="button">${ccLink(invoice.downloadUrl, i18n('cc-invoice.download-pdf'), skeleton)}</div>
-              <div class="info">
+              <div slot="header-right">${ccLink(invoice.downloadUrl, i18n('cc-invoice.download-pdf'), skeleton)}</div>
+              <div slot="content" class="info">
                 <em class=${classMap({ skeleton })}>
                   ${i18n('cc-invoice.info', { date: invoice.emissionDate, amount: invoice.amount })}
                 </em>
               </div>
-              <cc-html-frame class="frame" ?loading="${skeleton}" iframe-title="${title}">
+              <cc-html-frame slot="content" class="frame" ?loading="${skeleton}" iframe-title="${title}">
```
florian-sanders-cc, line 69: **suggestion:** in responsive, the padding around the invoice frame is gone
preview:
image

prod:
image


If you want to restore the padding, the easiest solution would be to slot into `content-body` instead of `content` :thinking: 

We can discuss this in sync if you need, I'm not sure we need the padding anyway, we can do something else.

src/components/cc-orga-member-list/cc-orga-member-list.js
```diff
@@ -363,6 +366,13 @@ export class CcOrgaMemberList extends LitElement {
         </cc-block-section>
 
         ${this.members.state === 'loaded' ? this._renderDangerZone(this.members) : ''}
+
+        <div slot="footer-right">
+          ${ccLink(
+            `${ORGA_MEMBER_DOCUMENTATION}`,
```
florian-sanders-cc, line 372: **nitpick:**
```suggestion
            ORGA_MEMBER_DOCUMENTATION,
```

src/components/cc-ssh-key-list/cc-ssh-key-list.js
```diff
@@ -220,10 +224,12 @@ export class CcSshKeyList extends LitElement {
             : ''}
         </cc-block-section>
 
-        <!-- documentation link -->
-        <cc-block-section>
-          <div class="align-end">${i18n('cc-ssh-key-list.doc.info')}</div>
-        </cc-block-section>
+        <div slot="footer-right">
+          ${ccLink(
+            `${SSH_KEY_DOCUMENTATION}`,
```
florian-sanders-cc, line 229: **nitpick:**
```suggestion
            SSH_KEY_DOCUMENTATION,
```

src/components/cc-tcp-redirection-form/cc-tcp-redirection-form.js
```diff
@@ -1,11 +1,14 @@
 import { css, html, LitElement } from 'lit';
```
florian-sanders-cc, line 1: **nitpick:** within this component, the docs link is still present within the desc. Since there's a link within the footer, I would remove it:
image

src/directives/hasSlottedChildren.js
```diff
@@ -0,0 +1,76 @@
...
+    this.eventHandler?.disconnect();
+  }
+
+  /**
+   * @param {Part} part
+   * @returns {unknown}
+   */
+  update(part) {
+    if (part.type !== PartType.ELEMENT) {
+      throw new Error('This directive must be set as on Element.');
```
florian-sanders-cc, line 40: **nitpick**:
```suggestion
      throw new Error('This directive must be set on an Element.');
```

src/templates/cc-link/cc-link.js
```diff
@@ -75,4 +75,8 @@ export const linkStyles = css`
     background-color: var(--cc-color-text-primary-weak, hsl(209deg 98% 73%));
     color: transparent;
   }
+
+  .cc-link cc-icon {
+    margin-right: 0.25em;
+  }
```
florian-sanders-cc, line 81: **question:** is it this still needed and are we sure it doesn't impact placed we don't know about?
Since you are using `display: flex` within documentation links, I would rely on `gap` and I wouldn't make `cc-link` itself add the margin (but it's not a bad idea overall, it's just hard to track which components / cases we impact so I would be cautious and wait for a `cc-link` component to introduce this kind of change).

web-dev-server.config.js
```diff
@@ -0,0 +1,134 @@
+import { hmrPlugin, presets } from '@open-wc/dev-server-hmr';
```
florian-sanders-cc, line 1: **fix:** you can remove this file, I think it's because your branch started before we migrated to Vite but it's no longer needed :wink:

florian-sanders-cc (changes requested):
I've finally reviewed every story of every component impacted (responsive included).

@HeleneAmouzou I've only caught a few issues which is very impressive considering you have changed a lot components and these changes impact CSS heavily (so it's very easy to break stuff, which you managed to avoid!).

The documentation links are really a great addition and the flexibility of the new `cc-block` is fantastic.

About the empty messages: I would go further and provide default styles for these from within the `cc-block` by styling the light DOM like you did for other cases (using `::slotted()`). This way you can get rid of all the duplicated rules you've had to add everytime you have added empty messages (color + italic).

Also, I think we should split into separate commits for each component impacted. I know it sounds annoying but the change on `cc-addon-credentials` is a breaking change so we need to make sure it's specified within the changelog and we cannot do that if it's part of a single big commit. In general I think it's better to have 1 commit per component for the changelog. We can do it in sync if you want but it's fairly easy to do and it doesn't take that long I promise :hugs:

src/components/cc-block/cc-block.js
```diff
@@ -47,196 +58,336 @@ export class CcBlock extends LitElement {
...
+          box-shadow: inset 0 6px 6px -6px rgb(0 0 0 / 40%);
+          display: none;
+          flex-wrap: wrap;
+          gap: 1em;
+          /* TODO doc */
+          justify-content: flex-end;
+          padding: 0.5em 1em;
+        }
+
+        /* TMP at least one of these slots is used: footer, footer-left, footer-right */
```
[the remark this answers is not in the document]
HeleneAmouzou: I think it was just an old comment I forgot to delete !

src/templates/cc-link/cc-link.js
HeleneAmouzou: Nice catch ! I did not think it  could have impact on other places than those components. I have removed it and add a gap directly on the `.cc-link` concerned, thank you !

PR #1154, opened 2024-09-18
feat(creation-tunnel): introduce `cc-zone-card` and `cc-zone-picker` components

Galimede:
## What does this PR do?

* This PR introduces two new components `cc-zone-card` and `cc-zone-picker`: 
  * It uses `ElementInternals` for `cc-zone-picker` and acts as a radio form input
  * It will be used as the zone choice in both the creation tunnel and later on on the information page of a product 
  * Note: there's no `skeleton` nor `error` state as there's a loader in the tunnel, thus it doesn't make sense to have those states for now

## How to review?

*  Check the code
* Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tunnel-creation/zone-select-two/index.html?path=%2Fstory%2F%F0%9F%9B%A0-creation-tunnel-cc-zone-picker--default-story):
  *  stories
  *  both French & English
   * different viewport sizes
   * keyboard input/focus


src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * @param {HTMLInputElementEvent} e
+   */
+  _onZoneSelect(e) {
+    this.value = e.target.value;
+  }
+
+  render() {
+    return html`
+      <form>
```
pdesoyres-cc, line 53: thinking: I'm not sure a `<form>` element carries the right semantic here. This component is an input and will surely already be inside a `<form>`.

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+        <fieldset @change="${this._onZoneSelect}">
+          <legend>
+            <cc-icon class="zone-legend-icon" .icon="${zoneIcon}" size="lg"></cc-icon>
+            <span class="zone-legend-text"> ${i18n('cc-zone-picker.legend')} </span>
+          </legend>
+          <div class="zone-section">
+            ${this.zonesSection.map(
+              (zoneSection) => html`
+                ${!isStringEmpty(zoneSection.title)
```
pdesoyres-cc, line 62: If section title is optional, maybe we could make the `ZoneSection.title` optional. The user won't have to pass an empty string if he/she doesn't want to display a section title.

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+    /** @type {string} The country code (e.g: FRA) */
+    this.countryCode = null;
+
+    /** @type {boolean} Whether the card should be disabled */
+    this.disabled = false;
+
+    /** @type {string} The url of the flag image */
+    this.flagUrl = null;
+
+    /** @type {ZoneImage[]} A list of images that will  displayed in the footer */
```
pdesoyres-cc, line 51: nitpick: extra space

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+export class CcZoneCard extends LitElement {
+  static get properties() {
+    return {
+      code: { type: String },
+      countryCode: { type: String, attribute: 'country-code' },
+      country: { type: String },
+      disabled: { type: Boolean, reflect: true },
+      flagUrl: { type: String },
+      images: { type: Array },
+      infra: { type: String },
```
pdesoyres-cc, line 27: question: property not used... is it really needed?

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+ * @cssdisplay flex
+ */
+export class CcZoneCard extends LitElement {
+  static get properties() {
+    return {
+      code: { type: String },
+      countryCode: { type: String, attribute: 'country-code' },
+      country: { type: String },
+      disabled: { type: Boolean, reflect: true },
+      flagUrl: { type: String },
```
pdesoyres-cc, line 25: fix: add `attribute: 'flag-url''`

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+            `
+          : ''}
+        ${this.images.map((image) => html`<cc-img src="${image.url}" a11y-name="${image.alt}"></cc-img>`)}
+      </div>
+    `;
+  }
+
+  static get styles() {
+    return [
+      skeletonStyles,
```
pdesoyres-cc, line 91: fix: skeleton style is not used

pdesoyres-cc (changes requested):
Well done @Galimede, those two components are very nice.

I'm triggered by the introduction of a new font, but maybe this is something that has already been acted during your investigations with @Bob

I also left some comments.

src/components/cc-zone-picker/cc-zone-picker.stories.js
```diff
@@ -0,0 +1,255 @@
...
+const cleverIcon = new URL('../../stories/assets/clevercloud.svg', import.meta.url);
+const cloudTempleIcon = new URL('../../stories/assets/cloudtemple.svg', import.meta.url);
+const oracleIcon = new URL('../../stories/assets/oracle.svg', import.meta.url);
+const ovhIcon = new URL('../../stories/assets/ovh.svg', import.meta.url);
+const scalewayIcon = new URL('../../stories/assets/scaleway.svg', import.meta.url);
+
+export default {
+  title: '🛠 Creation Tunnel/<cc-zone-picker>',
+  component: 'cc-zone-picker',
+};
```
florian-sanders-cc, line 14: **fix:**
```suggestion
export default {
  tags: ['autodocs'],
  title: '🛠 Creation Tunnel/<cc-zone-picker>',
  component: 'cc-zone-picker',
};
```

src/components/cc-zone-picker/cc-zone-picker.js
florian-sanders-cc: I wonder if we added this to make some tests and then forgot to remove it? Anyway, the semantics is not conveyed to Assistive Technologies unless the `form` has an accessible name but :+1: to remove it

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+ * A component that allows you to select a zone from a list of zones sections.
+ *
+ * @cssdisplay block
+ */
+export class CcZonePicker extends CcFormControlElement {
+  static get properties() {
+    return {
+      ...super.properties,
+      value: { type: String },
+      zonesSection: { type: Array },
```
florian-sanders-cc, line 30: **nitpick:**
```suggestion
      zoneSections: { type: Array },
```
or 
```suggestion
      zonesSections: { type: Array },
```

src/components/cc-zone-picker/cc-zone-picker.js
florian-sanders-cc: we could make the typing even stricter & clearer by allowing either a list of sections with `title` as mandatory or an array of one section without any `title`.
This way nobody would be tempted to pass a list of sections without titles.

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+    return html`
+      <div>
+        <input
+          class="visually-hidden"
+          type="radio"
+          name="zone"
+          .value="${zone.code}"
+          ?disabled=${zone.disabled}
+          ?checked="${isSelected}"
+          id="${zone.code}"
```
florian-sanders-cc, line 91: **fix:** since all radio inputs are part of the same group, people using a screen reader may not be aware of the section title info because they may navigate with arrows from one input to another.

This means we have to find a way to tell them "this input is part of this section".

The most straightforward thing to do would be to add an `id` attribute with a unique value on each section title (could be as simple as `section-header-1`, `section-header-2`) and then reference this `id` in an `aria-describedby` attribute here:
```suggestion
          id="${zone.code}"
          aria-describedby="${sectionHeaderId}"
```

Of course the `renderZoneCard` should receive the `sectionHeaderId`.

We can do it in sync if it's unclear :wink:

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+          margin: 0;
+          padding: 0;
+        }
+
+        cc-zone-card {
+          height: 100%;
+        }
+
+        input[type='radio']:focus-visible + label cc-zone-card {
+          border-radius: 2px;
```
florian-sanders-cc, line 165: **question:** I can't remember why we need the `border-radius` here and why its value is not a token :thinking:

src/components/cc-zone-picker/cc-zone-picker.stories.js
```diff
@@ -0,0 +1,255 @@
...
+          title: 'Public zones',
+          zones: PUBLIC_ZONES,
+        },
+      ],
+    },
+  ],
+});
+
+export const multipleSection = makeStory(conf, {
+  items: [
```
florian-sanders-cc, line 220: **suggestion:** I think you would benefit from typing this as well, for instance by adding this
```suggestion
  /** @type {Array<Partial<CcZonePicker>>} */  
  items: [
```

We need to type `makeStory` eventually so we don't have to do this but in the meantime, this can make refactoring easier + it adds some type checking.
For instance, if we decide `zoneSection` should be `zoneSections`, renaming the prop within the component should also rename the prop here :tada:

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+
+    /** @type {string} current selected zone code */
+    this.value = null;
+  }
+
+  /**
+   * @param {HTMLInputElementEvent} e
+   */
+  _onZoneSelect(e) {
+    this.value = e.target.value;
```
florian-sanders-cc, line 48: **fix:** the component should dispatch a change / input / select event to propagate the new value up to its parent just in case anyone decides to use it in a controlled way instead of relying on `ElementInternals`

florian-sanders-cc (changes requested):
:tada: Great job @Galimede this component looks great and it's very easy to use / understand.

I think it's API is dead simple (great thing) and the fact it relies on `ElementInternals` is :heart_eyes: 

I've left a few comments but we're almost there, don't hesitate to ping me if something is unclear or if you want to discuss some points :wink:

src/components/cc-zone-picker/cc-zone-picker.js
hsablonniere: I had a similar thought. We need plural on "section". Not sure about "zone". Both @florian-sanders-cc suggestions are OK.

src/components/cc-zone-picker/cc-zone-picker.js
hsablonniere: We will need this for the recap.

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
+import { css, html, LitElement } from 'lit';
+import { iconRemixCheckboxCircleFill as selectedIcon } from '../../assets/cc-remix.icons.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-badge/cc-badge.js';
```
hsablonniere, line 5: issue: not used anymore

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+    this.disabled = false;
+
+    /** @type {string} The url of the flag image */
+    this.flagUrl = null;
+
+    /** @type {ZoneImage[]} A list of images that will  displayed in the footer */
+    this.images = [];
+
+    /** @type {string} The infra name (e.g: clevercloud) */
+    this.infra = null;
```
hsablonniere, line 55: issue: not used anymore

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+    this.name = null;
+
+    /** @type {boolean} Whether the card should be selected */
+    this.selected = false;
+  }
+
+  render() {
+    return html`
+      <div class="title">
+        <div class="infra">${this.code}</div>
```
hsablonniere, line 67: suggestion: rename the CSS class to `code`

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+          flex-direction: column;
+          overflow: hidden;
+          position: relative;
+        }
+
+        input {
+          height: 100%;
+          position: relative;
+          width: 100%;
+        }
```
hsablonniere, line 107: issue: the render does not have any `input`

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+
+        :host .title {
+          flex: 1 1 auto;
+        }
+
+        :host .thumbnails {
+          flex: 0 0 auto;
+        }
```
hsablonniere, line 115: issue: no need to prefix these with `:host`

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
+import { css, html } from 'lit';
+import { CcFormControlElement } from '../../lib/form/cc-form-control-element.abstract.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-badge/cc-badge.js';
```
hsablonniere, line 5: issue: not used anymore

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+      </form>
+    `;
+  }
+
+  /**
+   * @param {ZoneItem} zone
+   * @param {string} selectedZoneCode
+   */
+  _renderZoneCard(zone, selectedZoneCode) {
+    const isSelected = zone.code === selectedZoneCode;
```
hsablonniere, line 81: suggestion: we don't really need the selected zone code so we could reduce the scope of the params

```js
/**
   * @param {ZoneItem} zone
   * @param {boolean} isCelected
   */
  _renderZoneCard(zone, isSelected) {
```

And use it like this:

```
${zoneSection.zones.map((zone) => this._renderZoneCard(zone, zone.code === this.value))}
```

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+          color: var(--cc-color-text-primary-strongest);
+          font-family: var(--cc-ff-form-legend), inherit;
+          font-size: 1.15em;
+          margin: 1em 1.85em;
+        }
+
+        .private-zones,
+        .public-zones {
+          margin: 1em 0;
+        }
```
hsablonniere, line 152: issue: not used anymore

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * @param {ZoneItem} zone
+   * @param {string} selectedZoneCode
+   */
+  _renderZoneCard(zone, selectedZoneCode) {
+    const isSelected = zone.code === selectedZoneCode;
+    return html`
+      <div>
```
hsablonniere, line 83: question: do we need this div?

src/styles/default-theme.css
```diff
@@ -1,4 +1,5 @@
 :root {
+  --cc-ff-form-legend: 'Source Sans 3';
```
hsablonniere, line 2: issue: we need a default fallback

```
--cc-ff-form-legend: 'Source Sans 3', sans-serif;
```

I wonder if we should just put `sans-serif` and let the console customize it.

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+ * A component displaying a card with relative zone information (name, code...)
+ *
+ * @cssdisplay flex
+ */
+export class CcZoneCard extends LitElement {
+  static get properties() {
+    return {
+      code: { type: String },
+      countryCode: { type: String, attribute: 'country-code' },
+      country: { type: String },
```
hsablonniere, line 23: nitpick: `country` should come first

src/components/cc-zone-card/cc-zone-card.js
```diff
@@ -0,0 +1,213 @@
...
+  constructor() {
+    super();
+
+    /** @type {string} The zone code */
+    this.code = null;
+
+    /** @type {string} The country name */
+    this.country = null;
+
+    /** @type {string} The country code (e.g: FRA) */
```
hsablonniere, line 42: issue: country codes are ISO 3166-1 alpha-2, they have 2 letters. We should mention this and use a proper example.

src/components/cc-zone-picker/cc-zone-picker.stories.js
```diff
@@ -0,0 +1,255 @@
...
+        {
+          title: 'Public zones',
+          zones: PUBLIC_ZONES,
+        },
+      ],
+    },
+  ],
+});
+
+export const multipleSection = makeStory(conf, {
```
hsablonniere, line 219: issue: can we rename the story to `multipleSections`?

hsablonniere (changes requested):
Hey @Galimede, well done on this one :+1: 

The code is really easy and nice to read and review. Good job!! :heart: 

* I like the different widths in stories for the card
* Should we add `title` for tooltips on country flags and infra logo?

pdesoyres-cc (approved):
LGTM

roberttran-cc (approved):
Great job!
I have a minor issue & a suggestion in this review.
On top of that, I'll push a commit to propose small tuning of the UI (on the spacing): it's up to you if you want to keep it!

hsablonniere (approved):

florian-sanders-cc (approved):
Great job again @Galimede, all good!

PR #1160, opened 2024-09-30
chore: move the `i18n.js` file back to the build root directory

pdesoyres-cc:


florian-sanders-cc (approved):
Simple and efficient :smile: 

Good job!

PR #1162, opened 2024-09-30
Prettier: format remaining files

florian-sanders-cc:
## What does this PR do?

- Changes the `format` & `format:check` npm scripts to make prettier run on all files starting from the project root and rely on `.prettierignore` to prevent formatting files we don't want.
- Runs the `format` command to format the few remaining files that had not been formatted before.

## How to review?

- Check the commits,
- 2 reviewers should be enough for this one but I need our Prettier expert aka @Galimede :wink:


pdesoyres-cc (approved):
LGTM

Galimede (approved):
LGTM! :muscle: 

Just a quick question though, I was thinking, if we format the other conifugration files, maybe we could format the ESLint config too and remove it from the `.prettierignore`, WDYT?

florian-sanders-cc:
> Just a quick question though, I was thinking, if we format the other conifugration files, maybe we could format the ESLint config too and remove it from the .prettierignore, WDYT?

Yes, but let's do that when we migrate ESLint since the config file will be a JS file it will be even better :+1:

PR #1170, opened 2024-10-02
feat(cc-invoice): add `currency` as part of the `state`

### LINKED ISSUE #1169: cc-invoice: support different currencies
## Context

The API sends both the `amount` and a `currency` when fetching info about an invoice.
The `cc-invoice` does support the `currency` prop and the smart component does not pass this info.

We need to add a `currency` prop and make the smart pass it down to the UI component.

florian-sanders-cc:
Fixes #1169

## What does this PR do?

- Adds a `currency` property to the loaded state of `cc-invoice`,
- Within the `cc-invoice` smart component, pass down `currency` from the API response to the UI component,
- Within the `cc-invoice-table` component, pass the `currency` property to translation functions,
- Adds stories to showcase that these components handle multiple currencies.

## How to review?

- Check the new stories (dollar currency for `cc-invoice-list`, `cc-invoice-table`, `cc-invoice`),
- Check the `demo-smart` > `cc-invoice`. You can mock an API call return USD instead of 'EUR' if you want (not mandatory since I've done it already),
- Check the code,
- 2 reviewers should be enough for this one but I need our Pricing expert (aka @Galimede) review :sunglasses:

### DIFF
--- a/src/components/cc-invoice-list/cc-invoice-list.stories.js
+++ b/src/components/cc-invoice-list/cc-invoice-list.stories.js
@@ -105,6 +105,26 @@ export const dataLoadedWithNoProcessed = makeStory(conf, {
   ],
 });
 
+export const dataLoadedWithDollars = makeStory(conf, {
+  items: [
+    {
+      /** @type {InvoiceListStateLoaded} */
+      state: {
+        type: 'loaded',
+        invoices: [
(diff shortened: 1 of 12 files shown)


pdesoyres-cc (approved):
LGTM. Well done Florian

Galimede (approved):
LGTM, thanks Florian! :muscle:

PR #1211, opened 2024-10-22
feat(cc-domain-management): remove spacing between domain and path

### LINKED ISSUE #1204: cc-domain-management: remove spacing between domain and path
## Context

When discussing the new diag DNS feature for the CLI, we have decided to display the path right next to the domain.
We should do the same within the component because it gets in the way when you want to copy domain + path.

In the long run we have already discussed the need for a copy button but we would need a subtle one and it doesn't exist within the components at the moment.

florian-sanders-cc:
Fixes #1204

## What does this PR do?

- Removes spacing between domain and path within the domain list.

## How to review?

- Check the commit,
- Check the preview (not much to look for, just check that there is no spacing between the domain and path :man_shrugging:).


HeleneAmouzou (approved):
LGTM ! :clap:

PR #1214, opened 2024-10-22
fix(cc-domain-management): allow newly added domain to be marked as primary

### LINKED ISSUE #1213: cc-domain-management: impossible to set a newly added domain as primary
## Context

Steps to reproduce:
1. Add a new domain (without and path prefix),
2. Try to set this newly added domain as primary,

You get an error because the trailing slash is not part of the body sent to the API.

florian-sanders-cc:
Fixes #1213

## What does this PR do?

- Fixes an issue when trying to set a domain that you have just associated to an app as primary.

## How to review?

- Check the commit,
- Go to `localhost:6006/demo-smart/`, `cc-domain-management`, `app-node`:

1. add a new domain (with no path prefix),
2. mark it as primary,
3. you should get no error,
4. try to mark others as primary, delete some domains, nothing should throw any error.

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.smart.js
+++ b/src/components/cc-domain-management/cc-domain-management.smart.js
@@ -107,9 +107,8 @@ defineSmartComponent({
               'domainListState',
               /** @param {DomainManagementListStateLoaded} domainListState */
               (domainListState) => {
-                const hasPathPrefix = pathPrefix != null && pathPrefix !== '/';
                 // TODO: once the API returns actual ids, this should be adapted (either refetch or use the API response)
-                const id = hasPathPrefix ? hostname + pathPrefix : hostname;
+                const id = hostname + pathPrefix;
 
                 /** @type {DomainStateIdle} */
                 const newDomain = {


src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -107,9 +107,8 @@ defineSmartComponent({
               'domainListState',
               /** @param {DomainManagementListStateLoaded} domainListState */
               (domainListState) => {
-                const hasPathPrefix = pathPrefix != null && pathPrefix !== '/';
                 // TODO: once the API returns actual ids, this should be adapted (either refetch or use the API response)
-                const id = hasPathPrefix ? hostname + pathPrefix : hostname;
```
florian-sanders-cc, line 112: I cannot remember why I did that in the first place :/ I do remember that there were conflicts with some API routes not handling the trailing slash properly but it seems to work fine without it and usually I leave a comment to explain why I did something weird so it may be something I forgot to remove?

florian-sanders-cc: I think the issue was as follows:
- when adding a new domain, I used to not send trailing slashes to the API because that's what was done before,
- so the `id` was the `hostname` without the path prefix if the path prefix was just "/",
- but at some point I decided to send trailing slashes to the API when adding a domain and I forgot to update the `id` here to reflect that.

HeleneAmouzou (approved):
LGTM ! :clap:

hsablonniere (approved):
LGTM on the code, I haven't tested the demo.

PR #1219, opened 2024-10-25
fix(cc-header-addon): revert `messages` class

HeleneAmouzou:
As noticed by @Galimede, the logos in the `cc-header-addon` footer were bigger than before since the component were adapted with the new `cc-block`. During the refactoring, I have removed the `messages` class the was applying a font-size, this PR revert it with the necessery properties to match the old footer design.

### DIFF
--- a/src/components/cc-header-addon/cc-header-addon.js
+++ b/src/components/cc-header-addon/cc-header-addon.js
@@ -119,7 +119,9 @@ export class CcHeaderAddon extends LitElement {
           </div>
         </div>
 
-        <cc-zone slot="footer-right" .state=${zoneState} mode="small-infra"></cc-zone>
+        <div slot="footer-right" class="messages">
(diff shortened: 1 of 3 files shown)


src/components/cc-header-addon/cc-header-addon.js
```diff
@@ -182,6 +184,12 @@ export class CcHeaderAddon extends LitElement {
           font-weight: bold;
         }
 
+        .messages {
+          align-items: center;
```
florian-sanders-cc, line 188: **suggestion:** move `align-items: center` to the `cc-zone` selector. It doesn't do anything here because `.messages` is not flex / grid.
Moving it to the `cc-zone` will align the zone text with the country logo as shown below

in preview:
image

with the fix:
image

florian-sanders-cc (approved):
I have a small suggestion but apart from that, LGTM!

Galimede (approved):
LGTM! :muscle:

PR #1221, opened 2024-10-25
feat(cc-zone-picker): add disabled and readonly modes

Galimede:
## What does this PR do?

* This PR introduces a `disabled` prop to disable the whole form
* This PR introduces a `readonly` prop to disable the whole form except the selected value
* This PR makes the component always select the first value if no value has been set 

## How to review?

* Check the commit
* Check the stories (disabled, partDisabled, readonly, noPreselectedValue)

### DIFF
--- a/src/components/cc-zone-picker/cc-zone-picker.js
+++ b/src/components/cc-zone-picker/cc-zone-picker.js
@@ -14,6 +14,7 @@ import { dispatchCustomEvent } from '../../lib/events.js';
  * @typedef {import('./cc-zone-picker.types.js').SingleZoneSection} SingleZoneSection
  * @typedef {import('./cc-zone-picker.types.js').ZonesSections} ZonesSections
  * @typedef {import('../../lib/events.types.js').EventWithTarget<HTMLInputElement>} HTMLInputElementEvent
+ * @typedef {import('lit').PropertyValues} CcZonePickerPropertyValues
  */
 
 /**
@@ -27,6 +28,8 @@ export class CcZonePicker extends CcFormControlElement {
   static get properties() {
     return {
       ...super.properties,
+      disabled: { type: Boolean },
+      readonly: { type: Boolean },
       value: { type: String },
       zonesSections: { type: Array, attribute: 'zones-sections' },
     };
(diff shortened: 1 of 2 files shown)


roberttran-cc (approved):

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -35,13 +38,31 @@ export class CcZonePicker extends CcFormControlElement {
...
+    /** @type {ZonesSections} array of zones sections */
+    this.zonesSections = [];
   }
 
+  /**
+   * @param {CcZonePickerPropertyValues} changedProperties
+   */
+  willUpdate(changedProperties) {
+    if (changedProperties.has('value') && this.value == null) {
+      this.value = this.zonesSections[0].zones[0].code;
```
pdesoyres-cc, line 59: fix: protect the code from weird usage. `this.zoneSections` may be `null` or empty. `this.zoneSections[0].zones` may be `null` or empty. In those cases the `value` should remain `null`.

pdesoyres-cc (changes requested):
Well done. I left just one comment.

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -35,13 +38,31 @@ export class CcZonePicker extends CcFormControlElement {
   constructor() {
     super();
 
-    /** @type {ZonesSections} array of zones sections */
-    this.zonesSections = [];
+    /** @type {boolean} whether the whole form should be disabled */
+    this.disabled = false;
+
+    /** @type {boolean} whether the form should be in readonly */
```
florian-sanders-cc, line 44: **suggestion:** let's not mention `form` here because this component is not a `form` by itself.
Maybe something like:
```suggestion
    /** @type {boolean} whether all the form controls should be disabled */
    this.disabled = false;

    /** @type {boolean} whether all the form controls should be readonly */
```

src/components/cc-zone-picker/cc-zone-picker.stories.js
```diff
@@ -201,6 +201,63 @@ export const defaultStory = makeStory(conf, {
...
+        {
+          zones: PUBLIC_ZONES,
+        },
+      ],
+      value: 'par',
+    },
+  ],
+});
+
+export const noPreselectedValue = makeStory(conf, {
```
florian-sanders-cc, line 248: **suggestion:** 
```suggestion
export const defaultValue = makeStory(conf, {
```

src/components/cc-zone-picker/cc-zone-picker.stories.js
```diff
@@ -201,6 +201,63 @@ export const defaultStory = makeStory(conf, {
   ],
 });
 
+export const partDisabled = makeStory(conf, {
```
florian-sanders-cc, line 204: **suggestion:**
```suggestion
export const partiallyDisabled = makeStory(conf, {
```

florian-sanders-cc (changes requested):
Gj :sunglasses: 

I've left a few suggestions (feel free to dismiss them, what you did is fine) and a small change for the desc of props.

florian-sanders-cc (approved):
LGTM! Gg @Galimede :sunglasses:

src/components/cc-zone-picker/cc-zone-picker.js
```diff
@@ -35,13 +38,31 @@ export class CcZonePicker extends CcFormControlElement {
...
+    /** @type {ZonesSections} array of zones sections */
+    this.zonesSections = [];
   }
 
+  /**
+   * @param {CcZonePickerPropertyValues} changedProperties
+   */
+  willUpdate(changedProperties) {
+    if (changedProperties.has('value') && this.value == null) {
+      this.value = this.zonesSections?.[0].zones[0].code;
```
pdesoyres-cc, line 59: fix: if user doesn't set any `zonesSections`, this will break because `this.zonesSections` will have a length of zero. We should check if `this.zonesSections` has got at least one item. (and this is the same for the `zones` array inside the first zone section)

pdesoyres-cc (changes requested):

pdesoyres-cc (approved):
LGTM.

PR #1229, opened 2024-11-04
style: renamed `hasSlottedChildren` to `has-slotted-children`

HeleneAmouzou:
Use of kebab-case to renamed `hasSlottedChildren` directive


florian-sanders-cc (approved):
LGTM!

PR #1237, opened 2024-11-15
fix(cc-kv-explorer): wrong gap between buttons in hash, list and set explorers

### LINKED ISSUE #1236: fix(cc-kv-explorer): wrong gap between buttons in hash, list and set explorers
Gap between buttons is set with a whitespace. But when integtrating inside the console, all the whitespaces ares removed by the bundler. This removes the gap between buttons.

We must set the gap explicitely instead of relying on whitespaces.

pdesoyres-cc:
Fixes #1236


florian-sanders-cc (changes requested):
**fix:** the commit about storybook should be `refactor` I think, it doesn't need to be part of the public changelog

PR #1250, opened 2024-11-26
fix(cc-kv-terminal): handle huge command history

pdesoyres-cc:
This PR handles huge command history which can creates a huge amount of nodes in the DOM.

I tried to rely on a `lit-virtualizer` without succeed. Instead, this PR relies on `content-visibility: auto` css property.

This property is supported in every modern browser but is quiet new. So it can happen that the browser won't support it but nothing bad will happen.

### How to test

* check the code
* verify that huge history doesn't break your browser.
  * You can use the command `command` to test large command outputs.
* verify no focus loss handling is not broken
* verify scrolling behavior is not broken

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@clevercloud/client": "^8.2.0",
         "@lit-labs/motion": "^1.0.3",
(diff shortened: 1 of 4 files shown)


pdesoyres-cc:
Closing because #1292 replace it.

PR #1263, opened 2024-12-02
fix(cc-block): add css rules to the slot name="content-body

### LINKED ISSUE #1259: cc-block-section: fix separation line display
## Context

The separation lines of the `cc-block-section`  used to display like this :

Screenshot from 2024-12-02 15-11-02

With the rework of the `cc-block` component, it is now displayed like that : 

Screenshot from 2024-12-02 15-10-55

This is due to the new slot organization. In the old version, the `cc-block-section` was used inside the default slot of the `cc-block` that was using a `cc-expand`. Now the default slot does not exist anymore and the `cc-expand` is used when the `slot="content-body"` is set, the `cc-block-section` should use the `slot="content-body"`  and not the `slot="content"` .

HeleneAmouzou:
Fixes #1259

## What this PR do ?
- adds css rules to the `slot name="content-body"` in the `cc-block` component,
- adapt the stories of the `cc-block-section` using the slot,
- adapt other components using the slot.

## How to review?
- check the code,
- check the stories on the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-block-section/fix-separation-line/index.html) and compare with the [prod](https://www.clever-cloud.com/doc/clever-components/?path=/docs/readme--docs).

### DIFF
--- a/src/components/cc-addon-admin/cc-addon-admin.js
+++ b/src/components/cc-addon-admin/cc-addon-admin.js
@@ -127,7 +127,7 @@ export class CcAddonAdmin extends LitElement {
     const shouldShowVmText = !this.noDangerZoneVmText;
 
     return html`
-      <cc-block-section slot="content">
+      <cc-block-section slot="content-body">
         <div slot="title">${i18n('cc-addon-admin.heading.name')}</div>
         <div slot="info"></div>
         <div class="one-line-form">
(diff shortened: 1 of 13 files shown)


pdesoyres-cc (approved):
perfect!

roberttran-cc (approved):

PR #1266, opened 2024-12-05
fix(cc-domain-management): restore loader & error message in the DNS config section

### LINKED ISSUE #1261: cc-domain-management: broken error & loading states for the DNS Config section
## Context

The DNS Config section is supposed to have a `loading` and `error` state but it appears that both with and without smart, this section is always in `loaded` state.

It doesn't trigger major issues, it displays static texts without DNS info but we should fix it to be consistent with the rest of the component.

florian-sanders-cc:
Fixes #1261

## What does this PR do?

The "Configure your DNS" section used to display a loader / error message depending on its state but it no longer does
image

This PR restores the error message / loader display depending on the state.

## How to review?

- Check the code,
- Check the preview ([loading](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/fix-dns-config-state/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--loading) & [error](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/fix-dns-config-state/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--error)),

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -377,10 +377,14 @@ export class CcDomainManagement extends LitElement {
             </cc-notice>
           </div>
 
-          ${this.dnsInfoState.type === 'loading' ? html` <cc-loader></cc-loader> ` : ''}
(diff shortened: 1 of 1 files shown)


Galimede (approved):
LGTM, thanks! :muscle:

PR #1275, opened 2024-12-05
refactor(cc-pricing-page-sandbox): filter out irrelevant features

florian-sanders-cc:
## What does this PR do?

- In the `cc-pricing-page` stories, removes irrelevant features that we don't want to display (and we don't really want to advertise these features anyway because the API sends incorrect info about them).

This is ISO with what we had before the last update: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/pocs/index.html?path=/story/%F0%9F%9B%A0-pricing-cc-pricing-page--default-story

## How to review?

- Check the code,
- Check the story,


Galimede (approved):
LGTM! :muscle:

PR #1278, opened 2024-12-10
feat(creation-tunnel): init plan components

Galimede:
## What does this PR do?

This PR introduces three new components for the new creation-tunnel:
* `cc-plan-item` => The base plan card
* `cc-plan-picker` => One plan picker 
* `cc-plan-configurator` => The form control element where you can refine your choice with two pickers

## How to review?

* Check the code
* Check the stories

### DIFF
--- a/src/components/cc-plan-item/cc-plan-item.js
+++ b/src/components/cc-plan-item/cc-plan-item.js
@@ -1,0 +1,177 @@
+import { css, html, LitElement } from 'lit';
+import { iconRemixCheckboxCircleFill as selectedIcon } from '../../assets/cc-remix.icons.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-icon/cc-icon.js';
+
+/**
+ * @typedef {import('./cc-plan-item.types.js').PlanDetails} PlanDetails
+ * @typedef {import('./cc-plan-item.types.js').PlanBadge} PlanBadge
+ * @typedef {import('lit').TemplateResult<1>} TemplateResult
+ */
+
+/**
+ * A component displaying a card with relative plan information and details
+ *
+ * @cssdisplay flex
+ */
+export class CcPlanItem extends LitElement {
+  static get properties() {
+    return {
+      badge: { type: Object },
+      details: { type: Array },
+      disabled: { type: Boolean, reflect: true },
+      name: { type: String },
+      selected: { type: Boolean, reflect: true },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {PlanBadge} Badge to display next to the plan name */
+    this.badge = null;
+
+    /** @type {PlanDetails[]} The details of the plan */
+    this.details = [];
+
+    /** @type {boolean} Whether the component should be disabled */
+    this.disabled = false;
+
+    /** @type {string} The plan name */
+    this.name = null;
+
+    /** @type {boolean} Whether the component should be selected */
+    this.selected = false;
+  }
+
+  render() {
+    return html`
+      <div class="title">
+        <div class="name">${this.name}</div>
+        ${this.badge != null ? html` <cc-badge .intent="${this.badge.intent}">${this.badge.content}</cc-badge> ` : ''}
+      </div>
+      <cc-icon class="icon-selected" .icon="${selectedIcon}" size="lg"></cc-icon>
+      ${this.details?.length > 0
+        ? html`<ul class="details">
+            ${this.details.map((detail) => this._renderDetail(detail))}
+          </ul>`
+        : ``}
+    `;
+  }
+
+  /**
+   * @param {PlanDetails} detail
+   * @returns TemplateResult
+   */
+  _renderDetail(detail) {
+    return html`
+      <li class="detail">
+        <span class="detail-icon"><cc-icon .icon="${detail.icon}" size="md"></cc-icon></span>
+        <span class="detail-value">${detail.value}</span>
+      </li>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        :host {
+          border: 2px solid var(--cc-color-border-neutral, #eee);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: block;
+          position: relative;
+        }
+
+        .title {
+          align-items: center;
+          column-gap: 0.375em;
+          display: inline-flex;
+          flex-wrap: wrap;
+          padding: 1em 1em 0.75em;
+        }
+
+        .name {
+          font-size: 1.25em;
+        }
+
+        :host([selected]) .name {
+          color: var(--cc-color-text-primary-strongest);
+        }
+
+        .icon-selected {
+          --cc-icon-color: var(--cc-color-bg-primary);
+
+          display: none;
+          position: absolute;
+          right: 0.5em;
+          top: 0.5em;
+        }
+
+        :host([selected]) .icon-selected {
+          display: block;
+        }
+
+        .details {
+          background-color: var(--cc-color-bg-neutral);
+          display: flex;
+          flex-direction: column;
+          gap: 0.25em 0;
+          margin: 0;
+          padding: 0.75em 1em;
+        }
+
+        :host([selected]) .details {
+          background-color: var(--cc-color-bg-primary-weak);
+        }
+
+        .detail {
+          display: inline-flex;
+          gap: 0.5em;
+        }
+
+        .detail-icon {
+          align-self: center;
+          color: var(--cc-color-text-weak);
+        }
+
+        :host([selected]) .detail-icon {
+          color: var(--cc-color-text-primary);
+        }
+
+        .detail-value {
+          color: var(--cc-color-text-weak);
+          font-size: 0.875em;
+        }
+
+        :host(:focus-visible) {
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: 2px;
+        }
+
+        :host([selected]) {
+          border-color: var(--cc-color-bg-primary);
+        }
+
+        :host([selected]) .detail-value {
+          color: var(--cc-color-text-primary-strongest);
+        }
+
+        :host([disabled]) {
+          border-color: var(--cc-color-border-neutral-disabled, #777);
+          opacity: var(--cc-opacity-when-disabled, 0.65);
+        }
+
+        :host(:hover:not([disabled])) {
+          border-color: var(--cc-color-border-neutral-hovered, #777);
+        }
+
+        :host(:not([selected], [disabled])) {
+          cursor: pointer;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-plan-item', CcPlanItem);
--- a/src/components/cc-plan-picker/cc-plan-picker.js
+++ b/src/components/cc-plan-picker/cc-plan-picker.js
@@ -1,0 +1,255 @@
+import { css, html } from 'lit';
+import { iconRemixServerLine as labelIcon } from '../../assets/cc-remix.icons.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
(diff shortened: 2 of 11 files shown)


src/components/cc-plan-item/cc-plan-item.js
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   *
+   * @param {PlanDetails} detail
+   * @returns HTMLElement
+   */
+  _renderDetail(detail) {
+    return html`
+      <div class="detail">
```
florian-sanders-cc, line 71: **nitpick:** ideally go for list items here
```suggestion
      <li class="detail">
```

src/components/cc-plan-configurator/cc-plan-configurator.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+      this._currentPlanId = currentPlan.id;
+      this._currentRelatedPlans = currentPlan.relatedPlans;
+    }
+  }
+
+  render() {
+    return html`
+      <cc-plan-picker
+        @input="${this._onChangePlan}"
```
florian-sanders-cc, line 136: **nitpick:** you're listening to the generic `input` event instead of our custom `cc-plan-picker:input` event. It works because the `input` bubbles but it would be better to listen to the documented event instead

src/components/cc-plan-configurator/cc-plan-configurator.js
```diff
@@ -0,0 +1,199 @@
...
+        name="plan"
+        .plans=${this.plans}
+        value="${this._currentPlanId}"
+        ?readonly="${this.readonly}"
+      ></cc-plan-picker>
+      ${this._currentRelatedPlans != null && this._currentRelatedPlans.length > 0
+        ? html`
+            <cc-plan-picker
+              @input="${this._onChangeRelatedPlan}"
```
florian-sanders-cc, line 145: **nitpick:** you're listening to the generic `input` event instead of our custom `cc-plan-picker:input` event. It works because the `input` bubbles but it would be better to listen to the documented event instead

florian-sanders-cc (changes requested):
Well done @Galimede the component looks good and the code is easy to read :+1: 

I left a few minor feedback and here are a few more:
- When reviewing, I felt like the `cc-plan-picker` and `cc-plan-configurator` could be merged into a single component. `cc-plan-picker` could be a subrender that is nested only when there are `relatedPlans`. It's not that big of a deal, this is a nitpick and I'll leave it up to you ;),
- Unless we're planning on using `cc-plan-picker` alone (but I don't see the usecase), I don't think we need to make `cc-plan-picker` a form element. Since its value is controlled by the parent anyway it doesn't bring any benefits.

Apart from that it's great!

src/components/cc-plan-picker/cc-plan-picker.types.d.ts
```diff
@@ -0,0 +1,10 @@
+import { PlanBadge, PlanDetails } from '../cc-plan-item/cc-plan-item.types.js';
+
+export interface PlanItem {
+  id: string;
+  badge?: PlanBadge;
+  name: string;
+  details?: PlanDetails[] | [];
```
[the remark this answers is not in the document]
Galimede: Just forgot to remove the empty array part, my bad :see_no_evil:

pdesoyres-cc (changes requested):
Well done @Galimede.

I really like the design.

I feal like you could merge `cc-plan-picker` and `cc-plan-configurator` inside one single component (that could be named `cc-plan-picker`) 

I also left some comments

src/components/cc-plan-configurator/cc-plan-configurator.stories.js
```diff
@@ -0,0 +1,50 @@
...
+ * @typedef {import('./cc-plan-configurator.types.js').ConfiguratorPlan} ConfiguratorPlan
+ * @typedef {import('./cc-plan-configurator.js').CcPlanConfigurator} CcPlanConfigurator
+ */
+
+const conf = {
+  component: 'cc-plan-configurator',
+};
+
+export const defaultStory = makeStory(conf, {
+  /** @type {Array<Partial<CcPlanConfigurator>>}  **/
```
HeleneAmouzou, line 22: nitpick: extra space and * at the end of the comment

src/components/cc-plan-configurator/cc-plan-configurator.stories.js
```diff
@@ -0,0 +1,50 @@
...
+  items: [
+    {
+      plans: DEFAULT_PLANS,
+      value: 'plan_12',
+    },
+  ],
+});
+
+export const readonly = makeStory(conf, {
+  /** @type {Array<Partial<CcPlanConfigurator>>}  **/
```
HeleneAmouzou, line 32: nitpick: extra space and * at the end of the comment

src/components/cc-plan-configurator/cc-plan-configurator.stories.js
```diff
@@ -0,0 +1,50 @@
...
+    {
+      plans: DEFAULT_PLANS,
+      readonly: true,
+      value: 'plan_12',
+    },
+  ],
+});
+
+export const noCustomizationPlans = makeStory(conf, {
+  /** @type {Array<Partial<CcPlanConfigurator>>}  **/
```
HeleneAmouzou, line 43: nitpick: extra space and * at the end of the comment

HeleneAmouzou (approved):
GG ! It looks very nice ! :clap: 
I have left some comments, mostly nitpicks.

src/components/cc-plan-picker/cc-plan-picker.types.d.ts
```diff
@@ -7,4 +7,5 @@ export interface PlanItem {
   details?: PlanDetails[];
   disabled?: boolean;
   selected?: boolean;
+  relatedPlans?: Array<Omit<PlanItem, 'relatedPlans'>>; // ???
```
florian-sanders-cc, line 10: **nitpick**: comment to be added or removed :eyes:

src/components/cc-plan-item/cc-plan-item.js
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * @typedef {import('./cc-plan-item.types.js').PlanDetails} PlanDetails
+ * @typedef {import('./cc-plan-item.types.js').PlanBadge} PlanBadge
+ * @typedef {import('lit').TemplateResult<1>} TemplateResult
+ */
+
+/**
+ * A component displaying a card with relative plan information and details
+ *
+ * @slot decorator - A decorator next to the plan name
```
florian-sanders-cc, line 15: **fix**: this slot doesn't seem to exist anymore :wink:

florian-sanders-cc (changes requested):
Don't forget to add the `cc-plan-item` & `cc-plan-picker` components to the list of components to be typechecked by CI (in `tsconfig.ci.json`) :wink: 

Apart from that I only have a few nitpicks compared to my first review!

Gg @Galimede almost there :sunglasses:

florian-sanders-cc (approved):
All good! Thanks & gg @Galimede

pdesoyres-cc (changes requested):
Well done Mathieu. Almost there, I left some comments and nitpicks.

hsablonniere (changes requested):
Well done @Galimede, the API and types are really good. I mostly have issues/nitpicks about the CSS and UI.

## cc-plan-item

* when the name of the plan is long, it does not wrap
  * we will try to avoid such situations but the component should still work properly if it happens
  * we need to story to demonstrate such a case: long plan name and custom CSS for story with a small `width` on the component (like when it's used in the picker).
* I wrote a lot of comments about your flexbox usage, but here's one that is a bit more global: the `display: flex` on the `:host` is not really used (same for the two `flex: xx` on `.title` and `.details`. They could/should be useful in the picker so all `cc-plan-item` have the same height and the `.title` takes the rest of the height`
  * This requires a change in the picker
  * The `flex: xx` on `.title` and `.details` would still need to be fixed
* the "selected" icon can sometimes go above the plan name or badge, is this a problem?

## cc-plan-picker

* like we need a story in cc-plan-item for long plan names, we need a story in this one to show how a long name will wrap, grow the height of the plan item and the other on the same row.

src/components/cc-plan-picker/cc-plan-picker.js
```diff
@@ -0,0 +1,259 @@
...
+          --cc-icon-color: var(--cc-color-text-primary);
+
+          align-self: center;
+        }
+
+        .plan-legend-text {
+          --ct-form-label-font-family: 'Source Sans 3';
+          --ct-form-label-font-size: 1.625em;
+          --ct-form-label-font-weight: 500;
+          --ct-form-input-font-size: 1.25em;
```
roberttran-cc, line 240: **❔ question:** what is the purpose of these CSS custom properties? :thinking:

src/components/cc-plan-item/cc-plan-item.js
```diff
@@ -0,0 +1,197 @@
...
+
+        :host([selected]) .detail-icon {
+          color: var(--cc-color-text-primary);
+        }
+
+        :host([selected]) .detail-value {
+          color: var(--cc-color-text-primary-strongest);
+        }
+
+        .title {
```
roberttran-cc, line 139: **💭 thought:** I feel like it's missing a text overflow policy but let's deal with it later.

src/stories/fixtures/plans.js
```diff
@@ -0,0 +1,385 @@
...
+        details: [
+          {
+            value: '2 GB disk size',
+            icon: iconDisk,
+          },
+        ],
+      },
+    ],
+  },
+
```
roberttran-cc, line 223: **🤓 nitpick:**

```suggestion
```

src/stories/fixtures/plans.js
```diff
@@ -0,0 +1,385 @@
...
+  id: 'third_party_3',
+  name: 'Third party 3',
+  details: [],
+},
+{
+  id: 'third_party_4',
+  name: 'Third party 4',
+  details: [],
+},
+];
```
roberttran-cc, line 385: **🤓 nitpick:**

```suggestion
export const THIRD_PARTY = [
  {
    id: 'third_party_1',
    name: 'Third party 1',
    details: [],
  },
  {
    id: 'third_party_2',
    name: 'Third party 2',
    details: [],
  },
  {
    id: 'third_party_3',
    name: 'Third party 3',
    details: [],
  },
  {
    id: 'third_party_4',
    name: 'Third party 4',
    details: [],
  },
];
```

roberttran-cc (commented):
Looking good! A few minor feedback on my side.

Also, I observed a weird behavior with the plan picker & keyboard focus but it may be legit:
- testing a story with related plans
- choosing another "parent" plan with the right arrow (you should just not select the last "parent" plan)
- focus is now on the new value
- pressing tab
- the focus is on the "related" plans, the first one
- pressing shift+tab to focus back to the "parent" plans
- I expect to be focused on the selected plan but I'm focused on the last "parent" plan

src/components/cc-plan-picker/cc-plan-picker.js
```diff
@@ -0,0 +1,259 @@
...
+
+  /**
+   * Find a plan by its ID in a list of plans.
+   *
+   * @param {PlanItem[]} plans - List of plans to search through
+   * @param {string} planId - ID of the plan to find
+   * @returns {PlanItem} The found plan or the first plan in the list if not found
+   * @private
+   */
+  _findPlan(plans, planId) {
```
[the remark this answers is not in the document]
Galimede: Imo, I do prefer relying on this signature rather than `this.plans`. Plus, if we wanted to move this later on, it will be easier to refactor and move.

hsablonniere (approved):
I had a quick look at the "hubert feedbacks", LGTM :+1: Well done!

pdesoyres-cc (approved):

PR #1281, opened 2024-12-12
chore(cc-tile-status-codes): filter out status codes below `100`

Galimede:
## What does this PR do?

* This PR filters out `0` status codes returned by the backend.
image
* It's a really small PR, one reviewer should be enough. :wink: 

## How to review?

* Check the code
* Test on a local storybook on the `/demo-smart` page that the `0` status codes are filtered.

### DIFF
--- a/src/components/cc-tile-status-codes/cc-tile-status-codes.smart.js
+++ b/src/components/cc-tile-status-codes/cc-tile-status-codes.smart.js
@@ -62,7 +62,15 @@ async function fetchStatusCodes({ apiConfig, signal, ownerId, appId }) {
   const warpToken = await getWarp10AccessLogsToken({ orgaId: ownerId }).then(
     sendToApi({ apiConfig, signal, cacheDelay: ONE_DAY }),
   );
-  return getStatusCodesFromWarp10({ warpToken, ownerId, appId }).then(
-    sendToWarp({ apiConfig, signal, timeout: THIRTY_SECONDS }),
-  );
+  return getStatusCodesFromWarp10({ warpToken, ownerId, appId })
+    .then(sendToWarp({ apiConfig, signal, timeout: THIRTY_SECONDS }))
+    .then((/** @type {Record<string, string>} */ data) => {
+      for (const statusCode in data) {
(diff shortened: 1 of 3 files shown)


src/components/cc-tile-status-codes/cc-tile-status-codes.smart.js
```diff
@@ -62,7 +62,16 @@ async function fetchStatusCodes({ apiConfig, signal, ownerId, appId }) {
...
+      /* @ts-expect-error - FIXME: data is of type StatusCodesBeta but doesn't include `0` */
+      (data) => {
+        // FIXME: This is temporary until we have a proper API and `0` status codes get removed from the backend response
+        if (data[0] != null) {
+          delete data[0];
+        }
+        return data;
+      },
+    );
 }
```
florian-sanders-cc, line 77: **suggestion:** I'm okay with this but I wonder if we could take this opportunity to guard against other mistakes the API could do.

This seems fairly simple to do because we can import `status` from `statuses` and use `status.message` to get the list of status codes we support. Then we can filter out keys that are not part of the status codes we support.
```suggestion
  return getStatusCodesFromWarp10({ warpToken, ownerId, appId })
    .then(sendToWarp({ apiConfig, signal, timeout: THIRTY_SECONDS }))
    .then(
      /** @param {Record<number, number>} data */
      (data) => {
        for (const property in data) {
          if (!(property in status.message)) {
            delete data[property];
          }
        }
        return data;
      },
    );
}
```

This way we could make this a `fix` commit so it's part of the changelog and we can release it on its own or wait for other commits to release several things (if we stick with a chore commit, we cannot release it on its own but maybe we were planning on waiting for other fixes anyway :man_shrugging:)

florian-sanders-cc (approved):
Thanks for handling this @Galimede!

It can be merged as it is but I suggested something to make it more "future proof" even though we plan on changing this part of the smart anyway :sweat_smile:

florian-sanders-cc (approved):
LGTM! Gg @Galimede thanks for handling that!

Issue #1273, opened 2024-12-05
eslint: investigate `eslint-plugin-import` false negatives & `eslint-plugin-import-x` alternative


florian-sanders-cc:
## Context

With #1141, we have moved to eslint 9.0 and we have disabled a few rules from the import plugin:
- `import/named` which checks named imports,
- `import/no-unresolved` which makes sure modules are resolved,
- `import/extensions` which enforces extension usage for imports.

Some of these rules have been disabled because TypeScript already checks that but as pointed out by @Galimede, if we extract this config later on so that it can be used in any Clever Cloud JS project, we cannot rely on TypeScript only and maybe these rules should be enabled (or at least maybe we should provide a way to enable them).

The issue is that `named` and `no-unresolved` have been disabled because they also return false negatives.
I couldn't understand why, it might be related to some CJS vs ESM shenanigan.

We should take time to understand the issues with these rules and find the proper settings or investigate using the fork of this plugin (`eslint-plugin-import-x`) which may be more reliable.

PR #1283, opened 2024-12-17
eslint: enable some rules from the import plugin in the generic clever cloud config

florian-sanders-cc:
## What does this PR do?

- Enables two rules from the `import` plugin in `eslint-config-clever-cloud-esm`:
  - `no-unresolved`: even though TypeScript already handles this fairly well, it's better to be safe than sorry + the config is meant to be generic and could be used in projects that don't have TypeScript enabled (thanks @Galimede for pointing that out),
  - `extensions`: same reason as above but TypeScript doesn't always handle the extension requirement (for plain imports like `import './my-component.js';`.

The `named` remains disabled because it doesn't seem to work well. We'll deal with that when we handle #1273 

## How to review?

- Checking the code should be enough,
- You can go to a component file and remove some extensions from imports if you want to see it working,
- 1 reviewer is enough for this one.


Galimede (approved):
LGTM, thanks Florian!

PR #1292, opened 2025-01-07
fix(cc-kv-terminal): handle huge command history using lit-virtualizer

pdesoyres-cc:
## What this PR do ?

This PR moves the history of commands and their results into a lit virtualizer so that browser can handle huge command history.

## How to review ?

- Check the code
- Use the command `command` to get huge result output
- Try to scroll up and check that you don't loose focus
- Check that when running command (hitting Enter), the view always scroll to bottom properly
- Verify on all browsers

### DIFF
--- a/src/components/cc-kv-terminal/cc-kv-terminal.js
+++ b/src/components/cc-kv-terminal/cc-kv-terminal.js
@@ -1,3 +1,4 @@
+import '@lit-labs/virtualizer';
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { createRef, ref } from 'lit/directives/ref.js';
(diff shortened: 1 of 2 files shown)


pdesoyres-cc:
Special thanks to @florian-sanders-cc for helping me on this ! :heart:

src/components/cc-kv-terminal/cc-kv-terminal.js
```diff
@@ -204,6 +268,22 @@ export class CcKvTerminal extends LitElement {
...
   }
 
+  /**
+   * @param {CcKvCommandContentItem} item
+   */
+  _renderItem(item) {
+    switch (item.type) {
+      case 'commandLine': {
+        const clazz = { command: true, empty: !item.hasResult };
+        return html`<div class=${classMap(clazz)}><cc-icon .icon=${iconShellPrompt}></cc-icon>${item.line}</div>`;
```
florian-sanders-cc, line 278: **suggestion:** we usually go this way for `classMap` and I find it easy to read
```suggestion
        return html`
          <div class="command ${classMap({ empty: !item.hasResult })}">
            <cc-icon .icon=${iconShellPrompt}></cc-icon>${item.line}
          </div>`;
```

src/components/cc-kv-terminal/cc-kv-terminal.js
```diff
@@ -204,6 +268,22 @@ export class CcKvTerminal extends LitElement {
...
+      }
+      case 'resultLine': {
+        const clazz = { result: true, error: !item.success, last: item.last };
+        return html`<div class=${classMap(clazz)}>${item.line}</div>`;
```
florian-sanders-cc, line 282: **suggestion:** same as above, I prefer what we usually do, even if it means the HTML spans on several lines. As long as it's properly formatted I think it's ok
```suggestion
        const dynamicClasses = { error: !item.success, last: item.last };
        return html`
          <div class="result ${classMap(dynamicClasses)}">
            ${item.line}
          </div>
        `;
```

florian-sanders-cc (approved):
Gg @pdesoyres-cc ! I checked on Chrome & Firefox.

I've left two suggestions but these are really personal preferences so feel free to ignore.

florian-sanders-cc (changes requested):
Actually I forgot to pull before testing on browsers, it seems it doesn't scroll to the last line at the moment :disappointed: 

We can discuss in sync if you need to see what's happening. It was working fine when you first sent me the commit you had modified so that's really weird?

florian-sanders-cc:
You beat the boss, seems to work fine now! Great job @pdesoyres-cc :sunglasses:

florian-sanders-cc (approved):

roberttran-cc (approved):
> If it works, it works. :shrug: 

GG for handling this!

Issue #1293, opened 2025-01-08
RFC: skeleton inside i18n translations


florian-sanders-cc:
## Context

Currently, our `skeleton` class is meant to be used directly on the element that you want to put in skeleton mode.

Imagine a translated sentence, where you want to only put a few parts (the dynamic ones) in skeleton.
A good example of that is the [`cc-invoice-list` in mobile  ](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-invoices-cc-invoice-list--loading&globals=viewport:w350)
![image](https://github.com/user-attachments/assets/994cf81f-2f2a-4ffe-b4b2-3aebe964028b)

In skeleton, the whole sentence is animated instead of only the specific words.
We also rely on the fact that the translation uses `<strong>` or `<code>` to set the background of the dynamic parts:
```css
        .invoice-list .skeleton code,
        .invoice-list .skeleton strong {
          background-color: #bbb;
          color: transparent;
        }
```

This is not very robust and is very specific to the `cc-invoice-list` case. We are "lucky" that we don't need to use `<strong>` in that sentence for a static non skeleton part.

## How to do better?

For the moment I see two options but there are probably other / better solutions.

### Allow using `class` attribute in translations

At the moment, `class` attributes added inside i18n properties are automatically stripped off by `i18n-sanitize`.
We could change this behavior and allow `class` attributes to be used within i18n properties.

This would allow translations like these:
```js
// translation file
'cc-toto-component.text': ({ data, skeleton }) => sanitize`
  <p>toto <span ${skeleton ? 'class="skeleton"' : ''}>data</span>
`,
```

- Pros:
  - it's very straightforward and easy to understand
- Cons:
  - brings more JS logic to translation strings which is something we tend to limit as much as possible,
  - opens the door to using classes freely inside i18n strings which can become a mess (it's always weird to define a class inside a component CSS, not find the class within the component render itself because people might think this class is unused).

### Add a new skeleton class

We could add a new skeleton class to be used on the parent of i18n nodes and allow a specific attribute to be used inside i18n strings.

For instance:

```js
// component file
<div class="classMap({ 'i18n-skeleton': this.state.type === 'loading' })">
  ${i18n('cc-toto-component.text', { data })}
</div>
```

```js
// translation file
'cc-toto-component.text': ({ data, skeleton }) => sanitize`
  <p>toto <span data-i18n-skeleton-item>data</span></p>
`,
```

```css
[...]

PR #1295, opened 2025-01-10
`cc-addon-elasticsearch-options`: support multi-currency and disabled billing

### LINKED ISSUE #1172: cc-addon-elasticsearch-options: support multiple currencies
## Context

The `cc-addon-elasticsearch-options` shows price estimation for Kibana / APM but the currency is hard coded to `EUR`.
- Both the price & currency should come from the new authenticated price system endpoint (this is handled by the tunnel creation within the console) => EDIT: maybe not from the price system, the estimation comes from another route.
- The component should be able to a different currency (this needs to be changed within the component).

### LINKED ISSUE #1230: cc-addon-elasticsearch-option: make `monthlyCost` optional
## Context

The `cc-addon-elasticsearch-option` component requires a `monthlyCost` for APM & Kibana. This property is used to generate the warning about estimated cost.

image

Since it can be used in a context where APM & Kibana have no cost (for instance in a console where everything is free :money_with_wings:), this property should be optional.

If no `monthlyCost` is provided, the text about cost estimation should be hidden.

We could also introduce a more sophisticated type for this with a boolean to enable / disable cost estimation.

florian-sanders-cc:
Fixes #1172 
Fixes #1230 

## What does this PR do?

- Adapts the `cc-addon-elasticsearch-options` component to support other currencies,
- Adapts the `cc-addon-elasticsearch-options` component to support cases where billing is disabled. In this case, sentences mentionning additional costs should be hidden, including in `loading` state when the cost is in skeleton mode while we fetch info.
- Both these changes led to some refactoring and **breaking changes**:
  - other `cc-addon-*-options` components do not fetch any data, they update data when you enable / disable options but there is no loading vs loaded logic.
  - on the contrary the `cc-addon-elasticsearch-options` component does fetch data and has a `loaded` vs `loading` logic:
    - we need to fetch what is the default instance of a Kibana and an APM instance,
    - if billing is enabled, we need to fetch the cost of the default instance of a Kibana and an APM instance.
  - this is why a `state` property was added. The types may look too complex for such a tiny component, feel free to suggest improvements but bear in mind that the subject is not as simple as it looks (there are comments that should help within the component & its types).

Note: the `skeleton` display is fairly broken right now, just like in `cc-invoice-list` but there's already an issue about this (see #1293)

## How to review?

- Check the code,
- Check the [stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-addon-elasticsearch-options/support-multi-currency/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-elasticsearch-options--default-story) (including the 2 new stories: `loading` & `no monthly cost`),
- 2 or 3 reviewers should be enough for this one.

### DIFF
--- a/src/components/cc-addon-elasticsearch-options/cc-addon-elasticsearch-options.types.d.ts
+++ b/src/components/cc-addon-elasticsearch-options/cc-addon-elasticsearch-options.types.d.ts
@@ -1,0 +1,28 @@
+import { ElasticAddonOption, EncryptionAddonOption, Flavor, FlavorWithMonthlyCost } from '../common.types.js';
+
+export type AddonElasticsearchOptionsState =
+  | AddonElasticsearchOptionsStateLoaded
+  | AddonElasticsearchOptionsStateLoadedWithoutMonthlyCost
+  | AddonElasticsearchOptionsStateLoading;
+
+type ElasticAddonOptionWithPaidFlavor = ElasticAddonOption<FlavorWithMonthlyCost>;
+type ElasticAddonOptionWithFreeFlavor = ElasticAddonOption<Flavor>;
+type ElasticAddonOptionWithoutFlavor = ElasticAddonOption<null>;
+
+export interface AddonElasticsearchOptionsStateLoaded {
+  type: 'loaded';
+  hasMonthlyCost: true; // enables the display of monthly cost for APM and Kibana options
+  options: Array<ElasticAddonOptionWithPaidFlavor | EncryptionAddonOption>; // since monthly cost is enabled, flavors in options need to have a `monthlyCost` object
+}
+
+export interface AddonElasticsearchOptionsStateLoadedWithoutMonthlyCost {
+  type: 'loaded';
+  hasMonthlyCost: false; // disables the display of monthly cost for APM and Kibana options
+  options: Array<ElasticAddonOptionWithFreeFlavor | EncryptionAddonOption>; // since monthly cost is disabled, flavors in options should not contain any `monthlyCost` object
+}
+
+export interface AddonElasticsearchOptionsStateLoading {
+  type: 'loading';
+  hasMonthlyCost: boolean; // since we rely on skeletons during loading phase, setting this to `true` shows a skeleton for the instance cost data
+  options: Array<EncryptionAddonOption | ElasticAddonOptionWithoutFlavor>; // since instance data and its cost are loading, options should not contain any `flavor` object
+}
--- a/src/components/cc-addon-option-form/cc-addon-option-form.js
+++ b/src/components/cc-addon-option-form/cc-addon-option-form.js
@@ -1,5 +1,6 @@
 import { css, html, LitElement } from 'lit';
 import { dispatchCustomEvent } from '../../lib/events.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
 import { linkStyles } from '../../templates/cc-link/cc-link.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-addon-option/cc-addon-option.js';
@@ -91,6 +92,7 @@ export class CcAddonOptionForm extends LitElement {
   static get styles() {
     return [
       linkStyles,
+      skeletonStyles,
       // language=CSS
       css`
         :host {
@@ -127,6 +129,17 @@ export class CcAddonOptionForm extends LitElement {
           color: var(--cc-color-text-warning);
 
           --cc-icon-size: 1.25em;
+        }
+
+        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
+        .option-warning .skeleton {
+          color: var(--cc-color-text-weak);
+        }
+
+        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
+        .skeleton strong {
+          background-color: #bbb;
+          color: transparent;
         }
       `,
     ];
--- a/src/components/cc-addon-option-form/cc-addon-option-form.js
+++ b/src/components/cc-addon-option-form/cc-addon-option-form.js
@@ -131,12 +131,12 @@ export class CcAddonOptionForm extends LitElement {
           --cc-icon-size: 1.25em;
         }
 
-        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
+        /* FIXME: Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
         .option-warning .skeleton {
           color: var(--cc-color-text-weak);
         }
 
-        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
+        /* FIXME: Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
         .skeleton strong {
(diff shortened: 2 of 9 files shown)


pdesoyres-cc (approved):
Well done Florian, I have some minor questions and nitpicks.

src/components/cc-addon-elasticsearch-options/cc-addon-elasticsearch-options.types.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  options: Array<ElasticAddonOptionWithPaidFlavor | EncryptionAddonOption>; // since monthly cost is enabled, flavors in options need to have a `monthlyCost` object
+}
+
+export interface AddonElasticsearchOptionsStateLoadedWithoutMonthlyCost {
+  type: 'loaded';
+  hasMonthlyCost: false; // disables the display of monthly cost for APM and Kibana options
+  options: Array<ElasticAddonOptionWithFreeFlavor | EncryptionAddonOption>; // since monthly cost is disabled, flavors in options should not contain any `monthlyCost` object
+}
+
+export interface AddonElasticsearchOptionsStateLoading {
```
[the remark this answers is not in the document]
florian-sanders-cc: :scream: Perfect question, that's an oversight, I need to create a story to showcase this because it could happen in real life (when billing is disabled, for air gap customers for instance).

HeleneAmouzou (approved):
GG ! :clap: 
Indeed, it is not as simple as it looks !
I just have a quesiton; as this component fetch data, what happen if the fetch did not work ? How does it handle error cases ?

florian-sanders-cc:
>  I just have a quesiton; as this component fetch data, what happen if the fetch did not work ? How does it handle error cases ?

@HeleneAmouzou this is a great question !!
I thought about adding some error state but I felt the info was not "important" enough to do so and that thanks to skeletons, we already had the important info. 
If it fails to load, then you still know that kibana / apm will start a new app and that it may cost you something (if billing is enabled).

To be honest, if it fails to load, there's a very high chance that you won't be able to create your add-on anyway (because this data comes from the instances route + the billing route. Both these routes are essential to the creation tunnel).

This is subjective so don't hesitate to say if you think we should handle the error case more specifically (for now error = stuck in loading).

src/components/cc-addon-elasticsearch-options/cc-addon-elasticsearch-options.types.d.ts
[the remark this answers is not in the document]
florian-sanders-cc: Story added, thanks :wink:

roberttran-cc (approved):

src/translations/translations.fr.js
```diff
@@ -166,20 +167,24 @@ export const translations = {
...
-  'cc-addon-elasticsearch-options.warning.apm': `Si vous activez cette option, nous allons déployer et gérer pour vous un APM server, ce qui entraînera des coûts supplémentaires.`,
-  'cc-addon-elasticsearch-options.warning.apm.details': /** @param {Flavor} flavor */ (flavor) =>
-    sanitize`Par défaut, l'app sera démarrée sur une <strong title="${formatFlavor(flavor)}">instance ${flavor.name}</strong> qui coûte environ <strong>${formatCurrency(lang, flavor.monthlyCost)} par mois</strong>. `,
-  'cc-addon-elasticsearch-options.warning.kibana': `Si vous activez cette option, nous allons déployer et gérer pour vous un Kibana, ce qui entraînera des coûts supplémentaires.`,
-  'cc-addon-elasticsearch-options.warning.kibana.details': /** @param {Flavor} flavor */ (flavor) =>
-    sanitize`Par défaut, l'app sera démarrée sur une <strong title="${formatFlavor(flavor)}">instance ${flavor.name}</strong> qui coûte environ <strong>${formatCurrency(lang, flavor.monthlyCost)} par mois</strong>.`,
+  'cc-addon-elasticsearch-options.warning.apm': `Si vous activez cette option, nous allons déployer et gérer pour vous un APM server`,
```
Galimede, line 179: nitpick: in the english version we mention `Elastic APM server`

src/translations/translations.en.js
```diff
@@ -155,20 +156,24 @@ export const translations = {
...
-  'cc-addon-elasticsearch-options.warning.kibana': `If you enable this option, we'll deploy and manage a Kibana application for you, this will lead to additional costs.`,
-  'cc-addon-elasticsearch-options.warning.kibana.details': /** @param {Flavor} flavor */ (flavor) =>
-    sanitize`By default, the app will start on a <strong title="${formatFlavor(flavor)}">${flavor.name} instance</strong> which costs around <strong>${formatCurrency(lang, flavor.monthlyCost)} per month</strong>.`,
+  'cc-addon-elasticsearch-options.warning.apm': `If you enable this option, we'll deploy and manage an Elastic APM server application for you.`,
+  'cc-addon-elasticsearch-options.warning.flavor': /** @param {Flavor} flavor */ (flavor) =>
+    sanitize`By default, the app will start on a <strong title="${formatFlavor(flavor)}">${flavor.name} instance</strong>.`,
+  'cc-addon-elasticsearch-options.warning.kibana': `If you enable this option, we'll deploy and manage a Kibana application for you.`,
```
Galimede, line 171: nitpick: we mention `Kibana application` here but in the french version we only refer to `Kibana`

src/components/cc-addon-option-form/cc-addon-option-form.js
```diff
@@ -128,6 +130,17 @@ export class CcAddonOptionForm extends LitElement {
 
           --cc-icon-size: 1.25em;
         }
+
+        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
```
Galimede, line 134: suggestion: maybe add a `TODO` or `FIXME` on the comment for later

src/components/cc-addon-option-form/cc-addon-option-form.js
```diff
@@ -128,6 +130,17 @@ export class CcAddonOptionForm extends LitElement {
...
+        .option-warning .skeleton {
+          color: var(--cc-color-text-weak);
+        }
+
+        /* Temporary workaround for cc-addon-elasticsearch-options because the skeleton parts should only be the <strong> elements but they are nested inside i18n strings */
```
Galimede, line 139: suggestion: maybe add a `TODO` or `FIXME` on the comment for later

Galimede (approved):
Nice work Florian, apart from a few nitpicks LGTM, GG! :muscle:

src/translations/translations.fr.js
florian-sanders-cc: It was already the case, I don't think it changes much and we don't need both languages to match word for word so I'll leave it as it is

src/translations/translations.en.js
florian-sanders-cc: It was already the case, I don't think it changes much and we don't need both languages to match word for word so I'll leave it as it is

src/components/cc-addon-option-form/cc-addon-option-form.js
florian-sanders-cc: good idea, I'll do that!

florian-sanders-cc:
Closing this PR because we need a simpler implementation for the console.
The current implementation assumes that options are known in advance by the console but it's not the case.
This means that the component cannot have a skeleton, it should have a loading state with a spinner.

But this is not specific to the `cc-addon-elasticsearch-options` component and should be done for all `cc-addon-*-options` components instead.
Since the creation tunnel is being reworked anyway, let's wait and see how this component evolves in the long run and make changes to it only when we need it.

I'll create a new PR to simply add the billing disabled case + multi-currency but without `loading` state so that the logic remains like other `cc-addon-*-options` components.

PR #1297, opened 2025-01-14
cc-button: avoid setting both `disabled` & `waiting` at the same time

### LINKED ISSUE #1124: cc-button: focus is lost when set to `disabled`
## Context

We have decided that the native button within the `<cc-button>` component should remain focusable while the component is `disabled`.
This is done by using `aria-disabled` instead of `disabled` on the `<button>` element.

Unfortunately, Chrome considers the focus to be lost as soon as we set the component to `disabled`.
This seems to be caused by a clash with the native `disabled` attribute / prop.

Even though Chrome should not do that, we should change the attribute / prop and we should enforce never using native attributes as props for our components (see # TODO edit to link to the separate issue).

Edit: the issue is only present when dealing with `formAssociated` components as you can see with this minimal reproduction (https://jsbin.com/munohafoqa/1/edit?html,css,js,output).
It does make sense because `formAssociated` elements are recognized as actual form elements and chrome removes focus from disabled elements (this is why we have decided not to disable the native `<button>` element within the component and use `aria-disable` instead). 

Moving away from `disabled` attributes means that disabled custom elements will now participate in forms and `_internals.willValidate()` will be `true`. 
We could remove form data entries when a component is disabled `_internals.setFormValue(null)` but this is not great + we would need to change what `willValidate` exposes (instead of directly exposing `_internals.willValidate`, we would have to expose whether it's disabled or willValidate. All of this is not great because the abstract class that handles form controls does not know of the `disabled` (or `isDisabled` property). 

## Decision & TODO

- [x] In `cc-button` add a console.error()` when both `disabled` and `waiting` are used at the same time,
- [x] Fix cases where both `disabled` and `waiting` are used at the same time,
- [ ] Create an issue on Chrome bug tracker with the minimal reproduction.

florian-sanders-cc:
Fixes #1124

## What does this PR do?

- Makes the `cc-button` throw an error if both `waiting` and `disabled` are set to `true` at the same time.
  - The question is: is it a good idea to throw an error and is it a good idea to do so in the lit lifecycle? 
  - The goal was to make it obvious that something is not right. Throwing an error is not enough because it won't make the tests fail + it won't be visible (the underlying issue is very easy to miss).
  - Throwing in the lifecycle actually breaks the component so I thought it was a good way to make it obvious that something is wrong but maybe it's not a good idea?
- Fixes all occurrences where both `waiting` and `disabled` were used at the same time on a `cc-button`.

## How to review?

- Check the commits,
- Compare the impacted component stories (there should be no visual changes):
  - [`cc-addon-admin` - Prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-addon-cc-addon-admin--default-story) vs [`cc-addon-admin` - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button/fix-waiting-vs-disabled/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-admin--default-story),
  - [`cc-domain-management` - Prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story) vs [`cc-domain-management` - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button/fix-waiting-vs-disabled/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story),
  - [`cc-email-list` - Prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-profile-cc-email-list--default-story) vs [`cc-email-list` - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button/fix-waiting-vs-disabled/index.html?path=/story/%F0%9F%9B%A0-profile-cc-email-list--default-story),
  - [`cc-env-var-form` - Prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-environment-variables-cc-env-var-form--default-story) vs [`cc-env-var-form` - Preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-button/fix-waiting-vs-disabled/index.html?path=/story/%F0%9F%9B%A0-environment-variables-cc-env-var-form--default-story).

### DIFF
--- a/src/components/cc-addon-admin/cc-addon-admin.js
+++ b/src/components/cc-addon-admin/cc-addon-admin.js
@@ -142,7 +142,7 @@ export class CcAddonAdmin extends LitElement {
           <cc-button
             primary
             ?skeleton=${isSkeleton}
-            ?disabled=${isFormDisabled}
+            ?disabled=${isFormDisabled && state.type !== 'updatingName'}
             ?waiting=${state.type === 'updatingName'}
             @cc-button:click=${this._onNameSubmit}
             >${i18n('cc-addon-admin.update')}</cc-button
@@ -166,7 +166,7 @@ export class CcAddonAdmin extends LitElement {
           <cc-button
             primary
             ?skeleton=${isSkeleton}
-            ?disabled=${isFormDisabled}
+            ?disabled=${isFormDisabled && state.type !== 'updatingTags'}
             ?waiting=${state.type === 'updatingTags'}
             @cc-button:click=${this._onTagsSubmit}
             >${i18n('cc-addon-admin.tags-update')}</cc-button
@@ -185,7 +185,7 @@ export class CcAddonAdmin extends LitElement {
           <cc-button
             danger
             ?skeleton=${isSkeleton}
-            ?disabled=${isFormDisabled}
+            ?disabled=${isFormDisabled && state.type !== 'deleting'}
             ?waiting=${state.type === 'deleting'}
             @cc-button:click=${this._onDeleteSubmit}
             >${i18n('cc-addon-admin.delete')}</cc-button
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -240,6 +240,12 @@ export class CcButton extends LitElement {
         this._cancelClick();
       }
     }
+
+    if (this.waiting && this.disabled) {
+      throw new Error(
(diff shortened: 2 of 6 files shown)


roberttran-cc (approved):
Tricky issue, thanks of taking care of it! :heart_hands:

pdesoyres-cc (approved):
Nice one!

I'm ok with the `throw Error` in the Lit liffecycle.

PR #1300, opened 2025-01-16
form controls: rely on `readonly` during API calls instead of `disabled` to avoid focus loss

### LINKED ISSUE #1102: form controls: focus lost upon toggling disabled mode
## Context

During async API calls, we disable form controls.
Since disabled form controls cannot receive focus, the focus is lost.

When users are submitting the form using the submit button, this does not create any issue because the <cc-button> element does not actually disable the inner <button> element (we use `aria-disabled` and prevent clicks from doing anything).

When users are submitting the form by pressing enter within a form control, this creates an issue because focus goes nowhere, it is lost.

We may need to consider the same solution as with the `cc-button` component and go for `aria-disabled` + make the form control unresponsive but this is trickier than with the `cc-button` because there are more interactions to disable.

## Decision

We have decided to remain fairly close to the native HTML behavior:

- Use `disabled` only when we truly want to disable a form control,
- Use `readonly` during API calls because it preserves focus.

## TODO

- [ ] update the docs of `readonly` and `disabled` props to include some guidance about when to use which,
- [ ] change all occurrences of `disabled` during an API call to `readonly`.

florian-sanders-cc:
Fixes  #1102

## What does this PR do?

- Adds text to explain when to use the `readonly` attribute instead of the `disabled attribute for form controls,
- Changes `disabled` to `readonly` in components during the API calls.

## How to review?

- Check the commits,
- Check the impacted components in preview vs prod.


pdesoyres-cc (approved):
LGTM. Thank you for that Florian.

HeleneAmouzou (approved):
LGTM ! :clap:

PR #1301, opened 2025-01-16
Use the `checked` property instead of the attribute

### LINKED ISSUE #1265: Avoid using the `checked` attribute
## Context

Within `cc-zone-picker` and `cc-plan-configurator`, we rely on radio inputs and we use the `checked` attribute to control the selection.

This has the same disadvantages of using the `value` attribute instead of the `value` property: you may end up with binding issues because just like `value`, the `checked` attribute only controls the default state/value of the element.

The issue usually comes up when you change the selected radio input by interacting with the group (selecting a different radio input with the mouse / keyboard) and then you try to change the selected radio input programmatically (for instance you try to reset the selected input to the first one within the list because the list has changed).

This can be reproduced fairly easily: https://jsbin.com/cogayuhelo/1/edit?html,js,output

## Solutions

To avoid this, we should rely on properties instead of attributes for such cases.
We already rely on properties within the `cc-toggle` and `cc-select` (actually we faced the same issue with `cc-select` and changed it later on).

We need a way to enforce this because it's very easy to forget.
The `eslint-plugin-lit` plugin already provides a rule to forbid using the `value` attribute (see [no-value-attribute](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/no-value-attribute.md)).
It doesn't handle `checked` but it also provides a `ban-attributes` rule that we can configure to add `checked` to the list (see [ban-attribute](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/ban-attributes.md).

## TODO

- [x] Discuss with the team so everyone is knows why we introduced this,
- [x] Enable the `no-value-attribute` rule (already done in #1239),
- [x] Enable the `ban-attribute` rule for `checked` 
- [x] Fix the `cc-zone-picker` component (use `.checked` instead of `?checked`
- [ ] Optionally, contribute to the eslint plugin to add a `no-checked-attribute` in addition to `no-value-attribute` (this is very very easy to do)

Thanks @Galimede for your help and discussions about this!

florian-sanders-cc:
Fixes #1265

## What does this PR do?

- Enables some ESLint rule to prevent devs from using the `checked` attribute and force them to use the `checked` property instead,
- Fixes the `cc-zone-picker` component to use the `checked` property instead of the attribute.

## How to review?

- Check the commits,
- It should be enough considering the changes, no need to do more.

### DIFF
--- a/eslint/lit/eslint-config-lit-clever-cloud.js
+++ b/eslint/lit/eslint-config-lit-clever-cloud.js
@@ -10,6 +10,7 @@ export default {
     ...litPlugin.configs['flat/recommended'].rules,
     'lit/attribute-names': 'error',
     'lit/lifecycle-super': 'error',
+    'lit/ban-attributes': ['error', 'checked'],
     'lit/no-classfield-shadowing': 'error',
     'lit/no-invalid-escape-sequences': 'error',
     'lit/no-legacy-imports': 'error',
--- a/src/components/cc-zone-picker/cc-zone-picker.js
+++ b/src/components/cc-zone-picker/cc-zone-picker.js
@@ -116,7 +116,7 @@ export class CcZonePicker extends CcFormControlElement {
         name="zone"
         .value="${zone.code}"
         ?disabled=${disabled}
-        ?checked="${isZoneSelected}"
+        .checked="${isZoneSelected}"
         id="${zone.code}"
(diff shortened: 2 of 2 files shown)


roberttran-cc (approved):

HeleneAmouzou (approved):
LGTM ! :clap:

PR #1303, opened 2025-01-17
refactor(cc-ssh-key-list)!: rework properties to avoid impossible states

### LINKED ISSUE #1163: cc-email-list: migrate to the new `state` API
## Context

We've recently migrated a lot components the our new `state` API but the `cc-email-list` still relies on the old one.
The component was in modification (Element Internals) when we worked on the migration for others so it was not processed with the others.

Now that the component supports `ElementInternals` and is typechecked, we can migrate it.

### LINKED ISSUE #1165: cc-ssh-key-list: migrate to the new `state` API
## Context

We've recently migrated a lot components the our new `state` API but the `cc-ssh-key-list` still relies on the old one.
The component was in modification (Element Internals) when we worked on the migration for others so it was not processed with the others.

Now that the component supports `ElementInternals` and is typechecked, we can migrate it.

florian-sanders-cc:
Fixes #1165

## What does this PR do?

- Migrates the `cc-ssh-key-list` component to the new state API.

## How to review?

- Check the commit,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-ssh-key-list/migrate-to-new-state/index.html?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story) vs [prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-profile-cc-ssh-key-list--default-story) stories,
- Run locally and check `demo-smart` to see if all features still work properly (I couldn't check for accounts that are linked to GitHub because it seems the CLI oAuth Consumer is not allowed to fetch data for such cases, the error is already present in master even though it works fine in the console, which makes me think it's related to the CLI oAuth Consumer :shrug: ),
- 3 reviewers would be nice (breaking change + smart impact, better safe than sorry).

### DIFF
--- a/src/components/cc-ssh-key-list/cc-ssh-key-list.js
+++ b/src/components/cc-ssh-key-list/cc-ssh-key-list.js
@@ -34,7 +34,7 @@ const SSH_KEY_DOCUMENTATION = 'https://developers.clever-cloud.com/doc/account/s
  */
 const SKELETON_KEYS = [
   {
-    state: 'idle',
+    type: 'idle',
     name: fakeString(15),
     fingerprint: fakeString(32),
   },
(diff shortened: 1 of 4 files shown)


roberttran-cc (approved):

HeleneAmouzou (approved):
I have reviewed the commits, the stories and the adding/deleting features on demo-smart, it seems to work perfectly ! :clap:

pdesoyres-cc (approved):
LGTM. Well done Florian

PR #1305, opened 2025-01-17
feat(creation-tunnel): init order summary component

### LINKED ISSUE #1304: cc-order-summary: init component
## Context

For the new creation tunnel, we need an UI to display the options of the product being currently ordered.

## To be discussed

We assume the creation button will be part of a form but maybe it won't always be the case. Moreover, it's a `cc-button` but without a `type='submit'`, it only fires an event.

I see 3 options for now:
- leave it as is,
- implement the `type='submit'`,
- remove the `cc-button` and replace it with a slot.

### Attributes/properties

| `orderSummaryState`    | `Object` | `null`        | `true`    | component main state |

### Slots

No slot.

### CSS custom property

| `--cc-order-summary-detail-font-size` | `FontSize` | `0.825em`        | The font-size for the list of details        |
| `--cc-order-summary-font-weight` | `FontWeight` | `600`        | Sets the value of the font weight CSS property        |

### CSS part

No CSS part.

## Outputs

| `create` | n/a    | Fires when the create button is clicked        |

## Type Definitions

```typescript
export interface OrderSummaryState {
  name: string;
  tags?: Array<string>;
  logoUrl?: string;
  configuration?: Array<ConfigurationItem>;
}

export interface ConfigurationItem {
  label: string;
  value: string;
}
```

## Data/APIs

No API.

## Stories & states

WIP.

roberttran-cc:
Fixes #1304 

Quite a simple review
- check the code
- check the preview

### DIFF
--- a/src/components/cc-order-summary/cc-order-summary.js
+++ b/src/components/cc-order-summary/cc-order-summary.js
@@ -1,0 +1,274 @@
+import { css, html, LitElement } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
+import { isStringBlank, isStringEmpty } from '../../lib/utils.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-button/cc-button.js';
+import '../cc-img/cc-img.js';
+
+/**
+ * @typedef {import('./cc-order-summary.types.js').ConfigurationItem} ConfigurationItem
+ * @typedef {import('./cc-order-summary.types.js').LogoInfos} LogoInfos
+ * @typedef {import('./cc-order-summary.types.js').OrderSummary} OrderSummary
+ */
+
+/**
+ * Displays a summary of a product being ordered.
+ *
+ * The UI is composed of:
+ * - global information about the product,
+ * - a configuration in the form of a list of label/value,
+ * - a button to trigger the creation,
+ * - a list of details for additional information.
+ *
+ * @cssdisplay block
+ *
+ * @fires {CustomEvent} cc-order-summary:create - Fires when the create button is clicked.
+ *
+ * @cssprop {FontSize} --cc-order-summary-detail-font-size - The font-size for the list of details (defaults: `0.825em`).
+ * @cssprop {FontWeight} --cc-order-summary-font-weight - Sets the value of the font weight CSS property (defaults: `600`).
+ *
+ * @slot detail - a single piece of information displayed under the main part of the UI. You can insert multiple detail items.
+ */
+
+export class CcOrderSummary extends LitElement {
+  static get properties() {
+    return {
+      orderSummary: { type: Object, attribute: 'order-summary' },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {OrderSummary} component main datas */
+    this.orderSummary = null;
+  }
+
+  _onCreateClick() {
+    dispatchCustomEvent(this, 'create');
+  }
+
+  render() {
+    if (this.orderSummary == null) {
+      return '';
+    }
+
+    return html`
+      <div class="title">${i18n('cc-order-summary.title')}</div>
+      <div class="summary">${this._renderSummary()}</div>
+      <div class="details-container">
+        <slot name="detail"></slot>
+      </div>
+    `;
+  }
+
+  _renderSummary() {
+    const { submitStatus } = this.orderSummary;
+    const disabled = submitStatus === 'disabled';
+    const waiting = submitStatus === 'waiting';
+    return html`
+      ${this._renderSummaryHeader()} ${this._renderSummaryBody()}
+      <div class="footer">
+        <cc-button
+          class="btn-submit"
+          type="submit"
+          primary
+          ?waiting=${waiting}
+          ?disabled=${disabled && !waiting}
+          @cc-button:click=${this._onCreateClick}
+        >
+          ${i18n('cc-order-summary.create')}
+        </cc-button>
+      </div>
+    `;
+  }
+
+  _renderSummaryHeader() {
+    const { name, tags, logo } = this.orderSummary;
+
+    const tagTplFn = (/** @type {string} */ tag) =>
+      html`<cc-badge intent="info" weight="dimmed">${tag.trim()}</cc-badge>`;
+
+    return html`
+      <div class="header">
+        ${!isStringEmpty(name)
+          ? html`<div class="header--name">${name}</div>`
+          : html`<div class="header--name header--name-empty">&hellip;</div>`}
+        ${tags?.length > 0
+          ? html`<div class="header--tags">${tags.filter((tag) => !isStringBlank(tag)).map(tagTplFn)}</div>`
+          : ``}
+        ${!isStringEmpty(logo?.url) && !isStringEmpty(logo?.alt)
+          ? html`<div class="header--logo">
+              <cc-img class="logo" src="${logo.url}" a11y-name="${logo.alt}"></cc-img>
+            </div>`
+          : ``}
+      </div>
+    `;
+  }
+
+  _renderSummaryBody() {
+    const { configuration, skeleton } = this.orderSummary;
+
+    return html`<dl class="body">
+      ${configuration?.map((configItem) => {
+        const { label, value } = configItem;
+        return html`<div class="body--item">
+          <dt class="body--label">
+            <span class="${classMap({ skeleton })}">${label}</span>
+          </dt>
+          <dd class="body--value ${classMap({ skeleton })}">
+            <span>${value}</span>
+          </dd>
+        </div>`;
+      })}
+    </dl>`;
+  }
+
+  static get styles() {
+    return [
+      skeletonStyles,
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+
+        /* region elements > reset */
+        dl {
+          margin-block: 0;
+        }
+
+        dd {
+          margin-inline-start: 0;
+        }
+        /* endregion */
+
+        /* region blocks */
+        .title {
+          color: var(--cc-color-text-weak, #404040);
+          font-weight: var(--cc-order-summary-font-weight, 600);
+          margin-block-end: 0.5em;
+          padding-inline: 0.125em;
+        }
+
+        .summary {
+          background-color: var(--cc-color-bg-neutral, #f5f5f5);
+          border: 1px solid var(--cc-color-border-neutral-weak, #e7e7e7);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: flex;
+          flex-direction: column;
+          padding: 1.5em;
+          row-gap: 2em;
+        }
+        /* endregion */
+
+        /* region elements > header */
+        .header {
+          display: grid;
+          gap: 0.25em;
+          grid-template-columns: 1fr min-content;
+        }
+
+        .header--name {
+          font-size: 1.125em;
+          font-weight: var(--cc-order-summary-font-weight, 600);
+          grid-column: 1 / 2;
+          grid-row: 1 / 2;
+          word-break: break-word;
+        }
+
+        .header--tags {
+          display: inline-flex;
+          flex-wrap: wrap;
+          gap: 0.25em;
+          grid-column: 1 / 2;
+          grid-row: 2 / 3;
+        }
+
+        .header--logo {
+          grid-column: 2 / 3;
+          grid-row: 1 / 3;
+        }
+        /* endregion */
+
+        /* region elements > body */
+        .body {
+          display: flex;
+          flex-direction: column;
+        }
+
+        .body--item {
+          align-items: baseline;
+          column-gap: 0.5em;
+          display: flex;
+        }
+
+        .body--item:not(:last-child) {
+          border-block-end: 1px dotted var(--cc-color-border-primary-weak, #ccd4dc);
+          margin-block-end: 1em;
+          padding-block-end: 1em;
+        }
+
+        .body--label {
+          color: var(--cc-color-text-weak, #404040);
+          flex: 1 1 auto;
+        }
+
+        .body--value {
+          flex: 0 1 auto;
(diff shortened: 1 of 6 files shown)


src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+      </div>
+    `;
+  }
+
+  _renderSummary() {
+    return html`
+      <div class="header">${this._renderSummaryHeader()}</div>
+      <div class="body">${this._renderSummaryBody()}</div>
+      <div class="footer">
+        <cc-button class="btn-submit" type="submit" primary @cc-button:click=${this._onCreateClick}>
```
pdesoyres-cc, line 69: question: there is no need to be able to put this button in `disabled` or `waiting` mode?

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+          ${i18n('cc-order-summary.create')}
+        </cc-button>
+      </div>
+    `;
+  }
+
+  _renderSummaryHeader() {
+    return html`${this.orderSummaryState.name != null && this.orderSummaryState.name !== ''
```
pdesoyres-cc, line 77: nitpick: you could use the `isStringEmpty` utility function

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+      ? html`<div class="header--name">${this.orderSummaryState.name}</div>`
+      : html`<div class="header--name header--name-empty">&hellip;</div>`}
+    ${this.orderSummaryState.tags?.length > 0
+      ? html`
+          <div class="header--tags">
+            ${this.orderSummaryState.tags
+              .filter((tag) => tag != null && tag.trim() !== '')
```
pdesoyres-cc, line 84: nitpick: you could use the `isStringBlank` utility function

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+          ${i18n('cc-order-summary.create')}
+        </cc-button>
+      </div>
+    `;
+  }
+
+  _renderSummaryHeader() {
+    return html`${this.orderSummaryState.name != null && this.orderSummaryState.name !== ''
+      ? html`<div class="header--name">${this.orderSummaryState.name}</div>`
+      : html`<div class="header--name header--name-empty">&hellip;</div>`}
```
pdesoyres-cc, line 79: fix: could you add a story showing this situation?

pdesoyres-cc (changes requested):
Nice component Bob. I really like the design. Praise for the css and html organization, it makes the code really clear and easy to read.

I left some minor comments but nothing blocking.

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+  render() {
+    if (this.orderSummaryState == null) {
+      return '';
+    }
+
+    return html`
+      <div class="title">${i18n('cc-order-summary.title')}</div>
+      <div class="summary">${this._renderSummary()}</div>
+      <div class="details-container">
+        <slot name="detail"></slot>
```
hsablonniere, line 59: note: I was about to say "Do we need this slot? Why can't we just put stuff below?" but I guess it's a good way to encapsulate the styles in the component.

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  _renderSummary() {
+    return html`
+      <div class="header">${this._renderSummaryHeader()}</div>
+      <div class="body">${this._renderSummaryBody()}</div>
+      <div class="footer">
+        <cc-button class="btn-submit" type="submit" primary @cc-button:click=${this._onCreateClick}>
+          ${i18n('cc-order-summary.create')}
+        </cc-button>
```
hsablonniere, line 71: note: I still think we should consider moving the button to the form and place in here with a slot. We'll see in the next iterations how in goes.

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+    return html`
+      <div class="title">${i18n('cc-order-summary.title')}</div>
+      <div class="summary">${this._renderSummary()}</div>
+      <div class="details-container">
+        <slot name="detail"></slot>
+      </div>
+    `;
+  }
+
+  _renderSummary() {
```
hsablonniere, line 64: nitpick: I would inline this in the main render to simplify the readability, maybe it's just me.

src/components/cc-order-summary/cc-order-summary.js
hsablonniere, line 79: nitpick: not sure what you can do about this but the first line is hard to read for me because of the formatting.

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+import '../cc-button/cc-button.js';
+import '../cc-img/cc-img.js';
+
+/**
+ * @typedef {import('./cc-order-summary.types.js').ConfigurationItem} ConfigurationItem
+ * @typedef {import('./cc-order-summary.types.js').OrderSummaryState} OrderSummaryState
+ */
+
+/**
+ * Displays a summary of a product being currently ordered.
```
florian-sanders-cc, line 14: **nitpick:** 
```suggestion
 * Displays a summary of a product being ordered.
```

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+      ${this.orderSummaryState.configuration?.length > 0
+        ? this.orderSummaryState.configuration.map((configItem) => {
+            const { label, value } = configItem;
+            return html`<div class="body--item">
+              <div class="body--label">${label}</div>
+              <div class="body--value">${value}</div>
+            </div>`;
+          })
+        : ``}
+    `;
```
hsablonniere, line 107: nitpick: we could simplify this like that

```suggestion
    return this.orderSummaryState.configuration?.map((configItem) => {
        const { label, value } = configItem;
        return html`<div class="body--item">
          <div class="body--label">${label}</div>
          <div class="body--value">${value}</div>
        </div>`;
    });
```

hsablonniere (approved):
Hey @roberttran-cc, well done!!! Just a few notes and nitpicks :wink:

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+ * @cssprop {FontSize} --cc-order-summary-detail-font-size - The font-size for the list of details (defaults: `0.825em`).
+ * @cssprop {FontWeight} --cc-order-summary-font-weight - Sets the value of the font weight CSS property (defaults: `600`).
+ *
+ * @slot detail - a single piece of information displayed under the main part of the UI. You can insert multiple detail items.
+ */
+
+export class CcOrderSummary extends LitElement {
+  static get properties() {
+    return {
+      orderSummaryState: { type: Object, attribute: 'order-summary-state' },
```
florian-sanders-cc, line 35: **nitpick**: looking at the prop name I expected to find some kind of `state` (although if it was a `state` it would simply be named `state` eh)

I'll suggest alternatives but it's definitely fine if you stick with the current prop name:
- `data`,
- `summary`,
- `order`,
- `summaryData`,
- `orderSummaryData`,
- `orderSummary`.

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {OrderSummaryState} component main state */
```
florian-sanders-cc, line 42: **nitpick:** same as above, I would avoid talking about `state` but it's also fine if you stick with it :wink:

src/components/cc-order-summary/cc-order-summary.js
florian-sanders-cc: **suggestion:** Actually this makes me wonder: have you considered flattening the component API and moving all these out of an object directly on the component itself?

Since we don't really deal with impossible states, we could have separate props for each piece of data :thinking:

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  _renderSummaryBody() {
+    return html`
+      ${this.orderSummaryState.configuration?.length > 0
+        ? this.orderSummaryState.configuration.map((configItem) => {
+            const { label, value } = configItem;
+            return html`<div class="body--item">
+              <div class="body--label">${label}</div>
+              <div class="body--value">${value}</div>
```
florian-sanders-cc, line 103: **suggestion:** this is a perfect usecase for `dl` `dt` & `dd`.
This could impact your CSS a little bit but it should be fairly light because now `dl` accepts `<div>` as children instead of only `<dt>` and `<dd>` so you can keep the `flex` strategy!

Feel free to ping if my comment is unclear or you want to discuss it :wink:

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+          color: var(--cc-color-text-weak);
+          font-weight: var(--cc-order-summary-font-weight, 600);
+          margin-block-end: 0.5em;
+          padding-inline: 0.125em;
+        }
+
+        .summary {
+          background-color: var(--cc-color-bg-neutral);
+          border: 1px solid var(--cc-color-border-neutral-weak);
+          border-radius: var(--cc-border-radius-default);
```
florian-sanders-cc, line 129: **nitpick:** we usually set fallback values for these althought I don't really mind if we don't (+ it's a beta component)


```suggestion
          border-radius: var(--cc-border-radius-default, 0.25em);
```

src/components/cc-order-summary/cc-order-summary.js
```diff
@@ -0,0 +1,233 @@
...
+          <div class="header--tags">
+            ${this.orderSummaryState.tags
+              .filter((tag) => tag != null && tag.trim() !== '')
+              .map((tag) => html`<cc-badge intent="info" weight="dimmed">${tag.trim()}</cc-badge>`)}
+          </div>
+        `
+      : ``}
+    ${this.orderSummaryState.logoUrl != null
+      ? html`<div class="header--logo">
+          <cc-img class="logo" src="${this.orderSummaryState.logoUrl}" title="${this.orderSummaryState.name}"></cc-img>
```
florian-sanders-cc, line 91: **fix:** since the image is the runtime logo and this info is not really present in the summary, we need an alt (and it should generate the title on its own so we can remove it here)

Unfortunately, this means you probably have to change the data model to require an alternative text in addition to the logoUrl

```suggestion
          <cc-img class="logo" src="${this.orderSummaryState.logo.url}" a11y-name="${this.orderSummaryState.logo.alt}"></cc-img>
```

florian-sanders-cc (approved):
Well done @roberttran-cc the component looks really good. 

I didn't expect it to feel so easy to read / lightweight so great job again!
Also, the code is very clean and easy to read :heart: 

Some of the points I raised will need to be handled before the final release (out of beta I mean): the logo alternative text, the data list semantics. 
Still most of them are really nitpicks or suggestion and can easily be ignored / dismissed.

I have a few accessibility observations that are non blocking:
- the alignment of labels on the left and values on the right is not screen magnifier friendly. Some people using these tools will struggle. It's non blocking because placing labels next to values is only a requirement in forms + I don't really have a better idea that would look ok in this case. Also the distance is not that big (very very subjective) :shrug:. It's just something we should avoid as much as possible but here I think it's ok?
- Placing text after (in the dom order) the create button is not fobidden but it should be discouraged because it's highly possible that people using screen readers won't even notice that it's there. Usually the "submit button" is the last part of a form and people don't expect to have to browse past it. We could move it within the DOM and keep the visual order, especially since it contains no focusable elements but let's wait and see how it evolves and what we want to show in that section exactly before deciding? Let's just keep that in mind.

Feel free to ping me if some stuff is unclear :wink:

HeleneAmouzou (approved):
Well done, the summary looks nice and it will sure be very useful ! :clap:  I did not leave any comments, but I have overall questions: 
- Have you considered using a `cc-block` to render the summary ? If so, why choosing not to use it ?
- Why does the component only have one state ? Is it planned to have a loading and an error states in the future ?

Thanks in advance !

roberttran-cc:
Thank you so much for your reviews! :heart_hands: 
They all led to some valuable modifications (more than I expected initially).

**@florian-sanders-cc & @HeleneAmouzou: about the state property**
Actually, I didn't give a thought beforehand, it was the force of habit. :sweat_smile: 
But the attempts to integrate the component in the console have shown that the state may turn out to be useful. But for the time being, I renamed the state variable but did not flatten the properties. We'll come back on this with more insights: hopefully, the component is in beta!

**@HeleneAmouzou: about the `cc-block` usage**
I actually did not think about it. :grimacing: 
The main reason probably is because the component is strongly opinionated design-wise and not exactly what pictured the prototype. Another point is that the order summary would not use all the options offered by the `cc-block` (at first glance): therefore, it would bring some unnecessary code.
But feel free to get back to me if you have something in mind!

**@florian-sanders-cc: on label alignment and screen magnifier**
It is a tricky one, and I don't have a proper solution that would fix this issue while retaining the design.
Meanwhile, I pushed some slight UI updates that may alleviate the issue (commit `fix: feedback Florian (more readable informations for a11y`): I'd love your feedback on this.
Globally, we need to discuss this topic synchronously (with Marion), as it will come back regularly.

**@florian-sanders-cc: on the text after the submit button**
Another tricky one! :sweat_smile: 
The issue here is that the details is out of the container where belong the submit button. Again, I don't have a solution I like but I'll keep it in mind and we will discuss this synchronously as well.

florian-sanders-cc (approved):
I've checked the feedback commits where I'm mentionned + the stories (quickly), it's all good for me!

Great job @roberttran-cc !!

HeleneAmouzou:
@roberttran-cc Thanks for the detailed answers ! I have checked stories and LGTM !

pdesoyres-cc (approved):

PR #1307, opened 2025-01-21
fix(cc-zone-picker): use `@input` instead of `@change` for event

Galimede:
## What does this PR do?

* This PR changes the event from `@change` to `@input`

## Why?

* A `@change` event on a fieldset won't work reliably because the fieldset element itself doesn't emit change events directly. 
* This caused some value desync while working on the creation tunnel in the console

## How to review?

* Check the code
* Check the preview
* One reviewer should be enough

### DIFF
--- a/src/components/cc-zone-picker/cc-zone-picker.js
+++ b/src/components/cc-zone-picker/cc-zone-picker.js
@@ -73,7 +73,7 @@ export class CcZonePicker extends CcFormControlElement {
 
   render() {
     return html`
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
As discussed in sync I don't really get why it works better with `@input` delegation instead of `@change` but anyway if it works it works!

edit: got it, `@change` is dispatched on blur after a value has changed and not necessarily when the value changes :exploding_head:

PR #1312, opened 2025-01-27
refactor(cc-kv-terminal): catch `disconnected` errors

### LINKED ISSUE #1311: `cc-kv-terminal`: handle cases where the component is disconnected before async scrolling is done
## Context

With #1292 we introduced a virtualizer and some workaround to make sure scrolling is handled properly in `cc-kv-terminal`.

The workaround generates some logs in tests that show that when the component is disconnected, the scrolling still tries go through and fails.

Although this is really not a big deal, we should probably try / catch this error so that we don't pollute the tests or sentry / glitchtip (when we have that setup).

test logs showing the disconnected error

florian-sanders-cc:
Fixes #1311

## What does this PR do?

- Catches `disconnected` errors inside the `updated` lifecycle of `cc-kv-terminal`. These errors seem to happen when the component is removed from the DOM but a scrolling promise is pending?

**Note**: this is a `refactor` commit and not a `fix` because the issue it fixes has not been released yet. 

## How to review?

- Check the commit,
- Run `npm run test src/components/cc-kv-terminal/cc-kv-terminal.test.js`, you should only have `Resize Observer` logs. (the resize observer logs will soon be gone, I handled this subject when refactoring a11y tests).

### DIFF
--- a/src/components/cc-kv-terminal/cc-kv-terminal.js
+++ b/src/components/cc-kv-terminal/cc-kv-terminal.js
@@ -229,9 +229,17 @@ export class CcKvTerminal extends LitElement {
    */
   async updated(changedProperties) {
     if (changedProperties.has('state') && this.state.type === 'idle') {
-      await this._scrollToBottom();
-      // don't ask me why, but we really need to call it twice to make sure it really scrolls to the bottom
-      await this._scrollToBottom();
+      try {
+        await this._scrollToBottom();
+        // don't ask me why, but we really need to call it twice to make sure it really scrolls to the bottom
+        await this._scrollToBottom();
+      } catch (error) {
+        // only catch `disconnected` errors that happen when the component is dismounted before the scrolling takes place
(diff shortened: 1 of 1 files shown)


pdesoyres-cc:
Question: When such error occurs (I mean, without your fix), does it break anything? Is it only an error log in the console but nothing really breaks?

roberttran-cc (approved):

florian-sanders-cc:
> Question: When such error occurs (I mean, without your fix), does it break anything? Is it only an error log in the console but nothing really breaks?

From what I gathered, it doesn't break anything, it's only an error log in the console.
I'm not sure how it would happen outside the context of tests but it felt so easy to fix that I decided to go through with it (this way when we setup glitchtip, we won't have to "ignore" this error).

Other components do not throw error logs in tests so I think we're fine (especially considering the low / nonexistent impact)

pdesoyres-cc (approved):

PR #1313, opened 2025-01-27
refactor(cc-grafana-info): avoid setting both `disabled` & `waiting` at the same time

### LINKED ISSUE #1124: cc-button: focus is lost when set to `disabled`

florian-sanders-cc:
Fixes #1124

## Context

- Adapts the `cc-grafana-info` component so that `disabled` & `waiting` are never set to `true` at the same time. I simply forgot to handle this component in #1297.

How am I sure no other components are missing?

The following log appears in local tests when there are issues of this type:

> `waiting` and `disabled` cannot be set to true at the same time, see https://github.com/CleverCloud/clever-components/issues/1124 for more info

screenshot of the error quoted above in the test results

This time I ran all component tests locally and searched for the message to make sure there was none left.

**Note**: ideally I'd like these errors to make tests fails but I'm not sure how to do that at the moment? If you have ideas feel free to weigh in!

**Note**: this is a `refactor` commit and not a `fix` because the issue it fixes has not been released yet. 

## How to review?

- Check the commit,
- If you want, you may run the tests locally and check that the error message is not part of the results,

### DIFF
--- a/src/components/cc-grafana-info/cc-grafana-info.js
+++ b/src/components/cc-grafana-info/cc-grafana-info.js
@@ -137,7 +137,7 @@ export class CcGrafanaInfo extends LitElement {
                   <cc-button
                     success
                     ?waiting=${isSwitchingGrafanaStatus}
-                    ?disabled=${isFormDisabled}
+                    ?disabled=${isFormDisabled && !isSwitchingGrafanaStatus}
                     @cc-button:click=${this._onEnableSubmit}
                   >
                     ${i18n('cc-grafana-info.enable-title')}
@@ -176,7 +176,7 @@ export class CcGrafanaInfo extends LitElement {
                 <div>
                   <cc-button
                     primary
-                    ?disabled=${isFormDisabled}
+                    ?disabled=${isFormDisabled && !isResetting}
                     ?waiting=${isResetting}
                     @cc-button:click=${this._onResetSubmit}
                   >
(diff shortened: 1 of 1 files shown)


roberttran-cc (approved):

pdesoyres-cc:
The only way I see to make test fail in that situation, is to write dedicated unit tests

pdesoyres-cc (approved):

florian-sanders-cc:
> The only way I see to make test fail in that situation, is to write dedicated unit tests

yeah I figured the same but it felt overkill :thinking: 
We'll see after the refactoring of a11y tests :+1:

PR #1314, opened 2025-01-27
refactor(smart): add typechecking on smart

pdesoyres-cc:
## What this PR do?

This PR adds typechecking to the `onContextUpdate` callback of the `defineSmartComponent` function. It uses a generic on the component so that the `updateComponent` function is typechecked related to the given component.

After this move, few bugs have been raised by typescript. So there are several fix commits.

Finally, I took the opportunity to move all the smart related files into a `smart` directory. Let me know if it's a good idea or not.

## How to test?

* check the code
* test some random stories
* test some random smart stories
* test demo-smart/ endpoint
* test sandbox/ endpoint


hsablonniere (approved):
Well done @pdesoyres-cc, I looked at the code, I tried a few stories and `/demo-smart`, LGTM :+1:

src/lib/smart/smart-component.types.d.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export interface SmartDefinitionParam<TypeHint = unknown> {
+  type: TypeHint;
+  optional?: boolean;
+}
+
+export type SmartContext = Record<string, any>;
+
+export type OnEventCallback = (type: string, listener: (detail: any) => void) => void;
```
florian-sanders-cc, line 44: **nitpick**: we could go for a stricter type based on our custom event naming convention to discriminate against native events:
```suggestion
export type OnEventCallback = (type: `${string}:${string}`, listener: (detail: any) => void) => void;
```

But since we plan on changing the dispatch event names & models, I'm really fine if we stick with the current implementation :+1:

florian-sanders-cc (approved):
Such a great PR @pdesoyres-cc thanks for that!

I've played with a few components + checked the code and it works fine.

Having a type checked & type safe `updateComponent` is a game changer, thanks a lot for that :wink: 

There are 2 optimizations that we could do later on:
- I think we could go one step further thanks to `HTMLElementTagNameMap` to match the `selector` against the component. It could offer a little bit more type safety (and maybe make it unnecessary to pass the generic in the typedef but not sure in this context). 
  - You can see some examples in the [makeStory refactoring branch (WIP)](https://github.com/CleverCloud/clever-components/commit/2042c171b29dc403b048d7560afaeee701eb78cc#diff-c4faae3014beca412141b080f4640c787901b3e0d7f1a38a8878f4b86af4bf3cR9), the component type is inferred from the component tag name.
  - But it's not that important and it requires that all components expose their tag name which they don't at the moment. I don't think it's worth postponing this PR so we should merge as is because it contains all we need at the moment :rocket:.
- Compared to what we currently have, we are losing types for the `context` properties. I don't think it's a real issue for most cases since we almost always pass ApiConfig or strings but we could have gone for a second generic to pass to `onContextUpdateArgs` with the context types :thinking:.
  - This is just an idea, I think we can merge as it is, the `any` types from context properties are not that problematic.

src/lib/smart/smart-component.types.d.ts
pdesoyres-cc: I didn't think about this, but anyway I didn't want to touch this part because of our plan to refactor the event handling deeper.

PR #1315, opened 2025-01-28
Story files tests / Accessibility tests: use a Web Test Runner Plugin instead of individual files

### LINKED ISSUE #731: test: discuss the best strategy to ignore some a11y tests
a11y unit tests are based on stories: each story produced for a component is tested against a11y rules.

Sometimes, we want to ignore some stories. As for now, we can do that within the test file like [here](https://github.com/CleverCloud/clever-components/blob/master/src/components/cc-icon/cc-icon.test.js#L7-L8).

An alternative would be to add a `skipTests` property for stories we want to be ignored. like below:

```
export const remixIcons = makeStory(conf, {
  css: ICONS_LAYOUT_CSS,
  dom: (container) => {
    ICONS_LAYOUT_BUILDER(container, remixIconsModule);
  },
  skipTests: true,
});
```

We all agree that whatever the adopted solution, we should enhance the documentation about this feature.

### LINKED ISSUE #854: accessibility test helper: remove the redundant `described` that wraps the tests
## Context

In the `tests/helpers/accessibility.js` file, we wrap mobile + desktop tests in a `describe('Accessibility',...).

This not necessary and this is not the right place to specify that these tests are about accessibility.
In the report, it generates this:
`Component: cc-input-text > Story: All form controls > Accessibility > Mobile: width = 360 height = 640`

The line should be as follows:
`cc-input-text > Accessibility > Story: All form controls > Mobile: width = 360 height = 640`

## Solution

- [ ] Remove the `describe('Accessibility')` from the helper,
- [ ] Change every component test file + example to make it look like this:
```js
describe('name of the component', () => {

  describe(`Accessibility`, function () {
    testAccessibility(storiesToTest);
  });

});
```

Thanks @pdesoyres-cc for getting it right in #845 :wink:

### LINKED ISSUE #963: Web Test Runner: improve developer experience
## Context

Currently, we rely on Web Test Runner to test components as well as libs and stories associated to them.

We have:
- a11y tests that loop through most of the stories of every component. These tests are set up from the `${componentName}.test.js` files,
- component controller tests (see `src/cc-logs/logs-controller.test.js`,
- unit tests related to libs (within the `test` folder).

## What we'd like to improve

- We would like to separate a11y tests from the rest so we can easily skip them (locally, not in CI). This means we would have to move and regroup them within a single file.
- We need to improve the output of a11y tests (see #854).
- We would like the reporter to show the list of tests that pass and fail (currently it only shows the list of tests that fail). A good example found by @roberttran-cc is https://modern-web.dev/docs/test-runner/reporters/summary/. Do we want this only locally or both for CI and local tests? :thinking:

florian-sanders-cc:
Fixes #731 
Fixes #963 
Fixes #854 

## What does this PR do?

- Removes all component test files that only contained a11y tests,
- Removes a11y tests from component test files that container other unit tests,
- Makes WTR run tests on every story file:
  - adds `src/components/**/*.stories.js` to the list of files to be handled by Web Test Runner. They are then transformed into test files by a plugin.
  - adds a `test-stories` plugin that:
    - tells the Web Dev Server used by WTR to import a virtual `.stories.test.js` file instead of the story,
    - produces the content of the `.stories.test.js` file when it is requested by the browser and serves it,
    - the content of the test file is fairly basic: import actual stories and pass them to the `testStories` helper that will loop over all stories to run tests (at the moment a11y tests to be specific).
- Merges `getStories` & `testStories` helpers into a single file because we never use them separately (`testStories` is the only one using `getStories` so they may as well be in the same file).
- Changes the logic to exclude stories from tests: 
  - we used to provide a list of stories to ignore because we manually created test files,
  - now all stories are tested by default and you can opt out of automated testing from the story options (docs have been adapted to explain that, and I'll add some comments to point out some examples). This is very close to what @pdesoyres-cc suggested in #731.
- Experiments with [groups](https://modern-web.dev/docs/test-runner/cli-and-configuration/#test-groups) to make it easier to focus on unit tests (outside of components) or component stories tests, or component unit tests. 
- Filters out `ResizeObserver` logs + adds export conditions so that we pull prod dependencies instead of dev (this way we test the components in the same conditions as when they are published).

## How to review?

- Check the commits and don't hesitate to weigh in about the architecture / naming,
  - there are many things that could be discussed, I'll try to leave some comments to add some context for the review + hesitations I faced.
- Run the tests locally, play with the groups.


hsablonniere (approved):
Impressive work @florian-sanders-cc, congrats! :clap:

I looked at the different commits and tried the commands.

* praise: nice touch on the plugin
* praise: I really like the group thing
* question: why tinyglobby and not globby? FYI, we already have globby as a transitive dep, I wanted to suggestion native node glob but it still triggers a warning and I'm not sure it's in v20
* note: the `window.process = {env: { NODE_ENV: "production" }}` is scary (event if required) :rofl:

florian-sanders-cc:
> question: why tinyglobby and not globby? FYI, we already have globby as a transitive dep, I wanted to suggestion native node glob but it still triggers a warning and I'm not sure it's in v20

@hsablonniere I meant to add a comment about that for the review but I forgot. I needed a sync glob library, not async + I needed it to accept several globs. I had the choice to update the `glob` dependency we currently have (it's an explicit dev dep at the moment, but it's really out of date), but then I would have had to test for breaking changes because it's used in our build process. Since I've created an issue to replace this lib with `tinyglobby` or move to Node.js 22 and use the native one, I decided to use `tinyglobby` and deal with that subject in another PR where we can test that the build process still works properly.

pdesoyres-cc (approved):
Really nice work @florian-sanders-cc .

I've just a naming nitpick: You named tests, plugin and test group with the "stories" word (`test-stories.js`, `test-stories.types.d.ts`, `test-story-plugin.js`, `npm run test:group stories:cc-example-component`). I would have used the "a11y" word instead because it is what tests check. The fact that those a11y tests are based on the stories is more an implementation detail.

HeleneAmouzou (approved):
Thanks and well done @florian-sanders-cc for all this work ! :clap: 

I did not enter in detail in the code but I went through the commits and read all the PR documentation you made, thanks a lot, it gave me  a good overview of the subject. :pray: 

I did not know the group option, I have tested while running the tests, very useful !

florian-sanders-cc:
> I've just a naming nitpick: You named tests, plugin and test group with the "stories" word (test-stories.js, test-stories.types.d.ts, test-story-plugin.js, npm run test:group stories:cc-example-component). I would have used the "a11y" word instead because it is what tests check. The fact that those a11y tests are based on the stories is more an implementation detail.

@pdesoyres-cc This is tricky but the `group` feature comes from Web Test Runner. It's heavily coupled to file names and there is no way to build a group based on something other than a file name pattern.
Even though the story test files only contain a11y tests at the moment, they will soon also include visual regression tests. At the moment I don't think I'll be able to make a group for a11y tests and another group for visual regression tests so I'd rather stick with `stories` until I find a solution to build two separate groups (I have some ideas with environment variables but I'm not sure it will work).

test/helpers/test-stories.js
```diff
@@ -0,0 +1,130 @@
...
+    height: 640,
+  },
+  mobile: {
+    width: 360,
+    height: 640,
+  },
+};
+
+/** @type {StoryContext} */
+// @ts-ignore we don't need to provide all options because the story won't be displayed in storybook
```
Galimede, line 31: question: couldn't we use a `Partial<StoryContext>`?

Galimede (approved):
Hey, nice work Florian! :muscle: 

That's actually pretty cool and impressive! :exploding_head:

roberttran-cc (approved):
Great PR for a great improvement to the stack, good job!

test/helpers/test-stories.js
florian-sanders-cc: No because `storyFunction` expects a full `StoryContext` (line 103 & 116).

This is something I'll try to improve when I type makeStory properly.

Issue #1200, opened 2024-10-10
Strange behavior in Chrome with a11y visually-hidden style


pdesoyres-cc:
When using the `visually-hidden` style, we get a strange behavior in Chrome:

The bug occurs when an element with this style is inside a grid area and with overflow set to auto.

```html
<div style="display: grid; grid-template-rows: 1fr auto">
  <div>CLI</div>
  <div style="overflow: auto; max-height=100px;">
    <div></div>
    <div></div>
    ...many <divs> here ...
    <div></div>
    <div></div>
    <label class=".visually-hidden" for="prompt">prompt</label>
    <input name="prompt">
  </div>
</div>
```

With this combination, Chrome displays a white gap below the grid container.
The more you have `<div>` in the overflow container, the higher the white gap is.

pdesoyres-cc:
This is due to the `position:absolute` of the `visually-hidden` css class.

The solution we have found is to force placing the visually hidden element at the top of the document using a `top: 0` css rule.

PR #1319, opened 2025-01-31
fix: revert `top` a11y visually-hidden style

Galimede:
## What does this PR do? 

* This PR reverts the changes that were initially fixing the issue #1200 
* Instead we're directly making the fix on `cc-kv-terminal`

## Why?

* While using `top: 0px` works and fixes the problem for `cc-kv-terminal`, it caused problems on other components using the `visually-hidden` styles, especially `cc-zone-picker` and `cc-plan-picker`
* It made these components scroll to top as the inputs were not correctly positioned anymore.
* That's why this PR reverts the fix and fixes `cc-kv-terminal` directly instead.

## How to review?

* Check the code
* Ideally check the sandbox of `cc-kv-explorer` to be sure that the problem doesn't appear anymore.

### DIFF
--- a/src/components/cc-kv-terminal/cc-kv-terminal.js
+++ b/src/components/cc-kv-terminal/cc-kv-terminal.js
@@ -384,6 +384,7 @@ export class CcKvTerminal extends LitElement {
           align-items: center;
           display: flex;
           gap: 0.2em;
+          position: relative;
         }
 
         .prompt input {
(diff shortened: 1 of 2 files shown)


florian-sanders-cc (approved):
Tested in sync in the console directly because I cannot test in storybook at the moment.

Thanks for the clear explanation, LGTM!

florian-sanders-cc (changes requested):
I would probably go for a refactor commit for the `kv-terminal` since it was working properly before, we're just swapping the fix for another one.

florian-sanders-cc (approved):

PR #1329, opened 2025-02-12
Fix/cc badge border radius

### LINKED ISSUE #990: cc-badge: set border radius on host instead of inner wrapper
## Context

The `cc-badge` component sets the `border-radius: 1em` on its inner wrapper instead of the `host`.

This means this `host` (the `<cc-badge>` element) does not have rounded borders.

When we use the `cc-badge` as a label for a hidden radio button (see `cc-product-list` component, category filters), we want the outline to be visible around the `cc-badge` and we want it to be rounded.

This means the rounded borders should be set on the host instead of the wrapper.
At the moment, the `cc-product-list` sets `border-radius: 1em` by itself but this is not robust (if we change to `1.5em` within the `cc-badge` one day, then it will look weird in `cc-product-list`.

clara-layus-cc:
Fixes #990 

# What does this PR do?

Set border radius on host instead of inner wrapper.

# How to review?

Check the commits,
Check the cc-badge stories in the preview.
1 reviewer should be enough.


pdesoyres-cc (changes requested):
LGTM.

Do not forget to add the Fixes #xxx footer in the commit message so that the generated changelog can display a direct link to the issue (see [conventionalcommits.org/en/v1.0.0](https://www.conventionalcommits.org/en/v1.0.0/))

pdesoyres-cc (approved):

PR #1332, opened 2025-02-13
build: add code format check

pdesoyres-cc:
Adds check of the code format during build process


florian-sanders-cc (approved):
LGTM, nice catch!

PR #1337, opened 2025-02-17
fix(cc-ssh-key-list): reduce spacing below the first section

### LINKED ISSUE #1262: cc-ssh-key-list: too much spacing below the first section
## Context

In the first section of the `cc-ssh-key-list` component, we show text to help people generate their SSH key.

This text is implemented using paragraphs but the last paragraph adds a spacing below that makes the spacing of the first section bigger than typical `cc-block-section`.

This should be easy to fix: 
- either remove all paragraph margins & use `display: grid` + `gap: 0.5em` on their parent,
- or only set `margin-bottom` for paragraphs that are not the last with the `:not(:last-of-type)` selector. 

image

clara-layus-cc:
Fixes #1262

# What does this PR do?

Reduce the spacing below the first section by rearranging the margins and gaps in the cc-block info slot

# How to review?

Check the commits,
Check the `cc-ssh-key-list` stories in the preview.
1 reviewer should be enough.

### DIFF
--- a/src/components/cc-ssh-key-list/cc-ssh-key-list.js
+++ b/src/components/cc-ssh-key-list/cc-ssh-key-list.js
@@ -455,6 +455,12 @@ export class CcSshKeyList extends LitElement {
         /* region misc */
         /* TODO tokenize border & border-color */
 
+        [slot='info'] {
+          display: grid;
+          gap: 1em;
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
Great job @clara-layus-cc, LGTM!

PR #1344, opened 2025-02-21
refactor(cem): silent doc generation

### LINKED ISSUE #1150: cem(typedef): introduce a `--silent` flag for the plugin
Every time the plugin is used, it logs every component and every cases.
However, there are times when we don't want the plugin to log, like when we're running tests, it feels a bit like spam.

For this purpose, we'd like to introduce a `--silent` flag in order to not log in some usage of the plugin.

Galimede:
Fixes #1150 

## What does this PR do?

* This PR removes the log that were produced when generating doc types of a component.
* Now, only the errors when writing a `typedef` will be displayed.
* Note: we couldn't add a `--silent` flag to the plugin as the CLI parameters are handled by CEM directly. 

## How to review?

* Run some tests, the logs shouldn't be there anymore.
* In a component try changing a `@typedef` to something like  `@tyedef` with a typo.
* Run the script `npm run component:docs`, it should show an error.
* One or two reviewer should be enough.

### DIFF
--- a/cem/support-typedef-jsdoc.js
+++ b/cem/support-typedef-jsdoc.js
@@ -114,12 +114,10 @@ export default function supportTypedefJsdoc() {
       moduleTypeCache.clear();
 
       const componentName = node.name.escapedText;
-      console.log(`\n⌛ Generating docs for ${componentName}`);
 
       const types = getTypesFromClass(node, ts);
 
       if (types.length === 0) {
-        console.info('✅ Successfully generated documentation.');
         return;
       }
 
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
LGTM @Galimede I wonder if we still need to warn for wrong typedef because it will trigger a TypeScript error anyway if the typedef syntax is wrong :thinking: 

One thing TypeScript doesn't handle properly is unused typedefs though. I'm not sure but didn't your plugin handle that because and I removed it? I can't find the commit where I did that :thinking:

pdesoyres-cc (approved):
LGTM

roberttran-cc (approved):

PR #1347, opened 2025-02-27
feat(cc-order-summary): improve configuration item implementation

### LINKED ISSUE #1346: cc-order-summary: improve skeleton state & accessibility
Future use case of the `cc-order-summary` (in the console) component will require:
- a more precise control on skeleton state
- proper accessibility management of data being updated (price for example)

roberttran-cc:
Fixes #1346

## What does this PR do?

For each line of the "configuration" part (aka the main part):
- Adds a `skeleton` property 
- Adds a `ariaLive` property

## How to review?

- Check the commits,
- Check the stories in [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-order-summary/improve-configuration-item/index.html?path=/story/%F0%9F%9A%A7-beta-%F0%9F%9B%A0-creation-tunnel-cc-order-summary-beta--default-story), especially the "Skeleton" and "With aria live" ones.

Two should be enough, with a mandatory one from @florian-sanders-cc. :hand_with_index_finger_and_thumb_crossed: 

## Thoughts

At some point, I considered making skeleton state on label & value separately but I'm not sure if it's worth the complexity. Maybe later if really needed.

### DIFF
--- a/src/components/cc-order-summary/cc-order-summary.js
+++ b/src/components/cc-order-summary/cc-order-summary.js
@@ -1,5 +1,6 @@
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { dispatchCustomEvent } from '../../lib/events.js';
 import { isStringBlank, isStringEmpty } from '../../lib/utils.js';
 import { skeletonStyles } from '../../styles/skeleton.js';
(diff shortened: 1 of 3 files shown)


pdesoyres-cc (approved):
LGTM, well done @roberttran-cc

florian-sanders-cc (approved):
Great job @roberttran-cc!

We may need to revisit the API later on because I'm wary about something:
- We let developers set the live region on the property/ies they want, making the component as agnostic as possible but we also open ourselves to more a11y issues: they could set a live region for each property or they could forget to set any live region at all. I'm not sure we should have a story with no live region (the default story) since it's an accessibility issue but for now let's stay like this and revisit the subject once we want to get out of beta? If we need to stay agnostic about the data that's provided to the component, then we'll add some guidance as comments within the types to help developers.
- We reflect the `aria-live` API in the `ConfigItem` property but here again I would be less open and expose a boolean with some comment that explains what it does internally within the component. I don't really like the idea that devs may set an `assertive` live region as it can be very detrimental if it's used in the wrong cases (for instance if it's used for the price here, it would be really bad). If we don't anticipate any case where we'd want an assertive live region, then our API should not expose it (this is debatable and very much my opinion).

TLDR; looks good to me for now, let's revisit at the end and see if we still need the API to be so open or if we can close some doors a little bit. If we do stay open, we'll add more guidance to mitigate possible issues :+1:

florian-sanders-cc (approved):
All good, great job @roberttran-cc, ship ship ship :rocket:

PR #1349, opened 2025-02-27
feat(cc-block-detail): init cc-block-detail component

### LINKED ISSUE #1348: cc-block-detail: init component
## Context

The purpose of this component is to add more information in the footer of a cc-block. This will allow for example to add a CLI command list going with the cc-block information as well as documentation links.

### Attributes/properties

| `isOpen`    | `Boolean` | `false`        | `false`    | Defines whether the content is visible or not |

### Slots

| Name  | Description |
|-------|-------------|
| `tab-title` | The title of the tab. Try to only use text.        |
| `link` | A zone dedicated to link, for example to documentation. |
| `content` | A zone dedicated to main content. |

### CSS custom property

No CSS custom property.

### CSS part

No CSS part.

## Outputs

No outputs.

clara-layus-cc:
Fix #1348 

# What does this PR do?

Introduce the `cc-block-detail` component. This component displays more information in the `cc-block` footer as command lines.

# How to review?
- Check the code
- Check the stories in the preview
- Check if the design is close enough to the [wireframe](https://www.figma.com/design/rC5NuHHEZ1E7HEEAbQyvuq/CC-Block-new?node-id=8-771&t=SkZ82Je83zrQ47OH-1) made by Marion

### DIFF
--- a/src/components/cc-block-detail/cc-block-details.js
+++ b/src/components/cc-block-detail/cc-block-details.js
@@ -1,0 +1,131 @@
+import { css, html, LitElement } from 'lit';
+import { iconRemixArrowDownSLine as iconArrowDown } from '../../assets/cc-remix.icons.js';
+import { dispatchCustomEvent } from '../../lib/events.js';
+import '../cc-icon/cc-icon.js';
+
+/**
+ * A display component with mostly HTML+CSS and an open/close toggle feature organized with slots to display information such as CLI commands.
+ * The main purpose is to be used with a cc-block.
+ *
+ * @cssdisplay block
+ *
+ * @fires {CustomEvent<boolean>} cc-block-details:is-open-change - Fires is-open state whenever it changes.
+ *
+ * @slot tab-title - The title of the tab. Try to only use text.
+ * @slot link - A zone dedicated to link, for example to documentation.
+ * @slot content - A zone dedicated to main content.
+ */
+
+export class CcBlockDetails extends LitElement {
+  static get properties() {
+    return {
+      isOpen: { type: Boolean, reflect: true, attribute: 'is-open' },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {boolean} Sets the state of the toggle. */
+    this.isOpen = false;
+  }
+
+  _onClickToggle() {
+    this.isOpen = !this.isOpen;
+    dispatchCustomEvent(this, 'is-open-change', !this.isOpen);
+  }
+
+  render() {
+    return html`
+      <div class="wrapper">
+        <button class="button" aria-expanded="${this.isOpen}" aria-controls="content" @click=${this._onClickToggle}>
+          <slot name="button-text"></slot>
+          <cc-icon .icon="${iconArrowDown}"></cc-icon>
+        </button>
+        <div id="content" class="content"><slot name="content"></slot></div>
+        <div class="links">
+          <slot name="link"></slot>
+        </div>
+      </div>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+
+        .wrapper {
+          column-gap: 1em;
+          display: grid;
+          grid-template-areas: 'button links' 'content content';
+          justify-items: start;
+        }
+
+        .button {
+          align-items: center;
+          background-color: transparent;
+          border: solid 1px transparent;
+          display: flex;
+          font-family: inherit;
+          font-size: 1em;
+          grid-area: button;
+          padding: 0.25em 0.6em 0.35em;
+          transition: all 0.3s;
+        }
+
+        :host([is-open]) .button {
+          background-color: var(--cc-color-bg-default, #fff);
+          border-radius: var(--cc-border-radius-default, 0.25em) var(--cc-border-radius-default, 0.25em) 0 0;
+          margin-bottom: 0;
+        }
+
+        .button:hover {
+          border-radius: var(--cc-border-radius-default, 0.25em);
+        }
+
+        :host(:not([is-open])) .button:hover {
+          border: solid 1px var(--cc-color-border-neutral-strong, #8c8c8c);
+        }
+
+        .button cc-icon {
+          margin-left: 0.1em;
+          transition: all 0.3s;
+        }
+
+        :host([is-open]) .button cc-icon {
+          transform: rotate(180deg);
+        }
+
+        :host(:not([is-open])) .button:hover cc-icon {
+          transform: rotate(90deg);
+        }
+
+        .content {
+          background-color: var(--cc-color-bg-default, #fff);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: none;
+          grid-area: content;
+          padding: 1em;
+        }
+
+        :host([is-open]) .content {
+          display: block;
+        }
+
+        .links {
+          align-items: center;
+          display: flex;
+          gap: 0.5em;
+          grid-area: links;
+          justify-self: end;
+        }
+      `,
+    ];
+  }
+}
+
+window.customElements.define('cc-block-details', CcBlockDetails);
--- a/src/components/cc-block-detail/cc-block-details.stories.js
+++ b/src/components/cc-block-detail/cc-block-details.stories.js
@@ -1,0 +1,68 @@
+import { makeStory } from '../../stories/lib/make-story.js';
+import './cc-block-details.js';
+import '../cc-block/cc-block.js';
+
+export default {
+  tags: ['autodocs'],
+  title: '🧬 Molecules/<cc-block-details>',
+  component: 'cc-block-details',
+};
+
+const conf = {
+  component: 'cc-block',
+};
+
+const ccBlockHtmlExample = `
+  <div slot="header-title">This is my block</div>
+  <div slot="content">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque feugiat dui at leo porta dignissim. Etiam ut purus ultrices, pulvinar tellus quis, cursus massa. Mauris dignissim accumsan ex, at vestibulum lectus fermentum id. Quisque nec magna arcu. Quisque in metus sed erat sodales euismod eget id purus. Sed sagittis rhoncus mauris. Ut sit amet urna ac nunc semper porta. Nam ut felis eu velit luctus rutrum. Nam leo nisl, molestie a varius non, ullamcorper sit amet tortor. Donec in convallis ex. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Praesent hendrerit venenatis erat, eu malesuada nulla viverra eu. Curabitur porta risus augue, non rutrum lectus hendrerit a.</div>
+`;
+
+const defaultHtmlExample = `
+  ${ccBlockHtmlExample}
+  <cc-block-details slot="footer-left">
+    <div slot="button-text">Command line</div>
+    <a slot="link" href="">See documentation</a>
+    <div slot="content">Praesent tincidunt dapibus elit, sed posuere risus sodales quis. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Integer ultrices maximus ligula eu efficitur. Nulla luctus ipsum vitae dui gravida, sodales ornare dolor viverra. Vivamus pretium scelerisque ultricies. Nulla consectetur euismod euismod. Interdum et malesuada fames ac ante ipsum primis in faucibus. Cras dignissim id sapien non facilisis. Suspendisse vitae pretium orci. Nunc aliquam pellentesque nunc, at tincidunt nisi lobortis a. <a href="#">Fake link</a></div>
+  </cc-block-details>
+`;
+
+export const defaultStory = makeStory(conf, {
+  items: [
+    {
+      innerHTML: defaultHtmlExample,
+    },
+  ],
+});
+
+const openHtmlExample = `
+  ${ccBlockHtmlExample}
+  <cc-block-details is-open=true slot="footer-left">
+    <div slot="button-text">Command line</div>
+    <a slot="link" href="">See documentation</a>
+    <div slot="content">Praesent tincidunt dapibus elit, sed posuere risus sodales quis. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Integer ultrices maximus ligula eu efficitur. Nulla luctus ipsum vitae dui gravida, sodales ornare dolor viverra. Vivamus pretium scelerisque ultricies. Nulla consectetur euismod euismod. Interdum et malesuada fames ac ante ipsum primis in faucibus. Cras dignissim id sapien non facilisis. Suspendisse vitae pretium orci. Nunc aliquam pellentesque nunc, at tincidunt nisi lobortis a. <a href="#">Fake link</a></div>
+  </cc-block-details>
+`;
+
+export const openStory = makeStory(conf, {
+  items: [
+    {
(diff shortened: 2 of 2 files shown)


src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @slot tab-title - The title of the tab. Try to only use text.
+ * @slot link - A zone dedicated to link, for example to documentation.
+ * @slot content - A zone dedicated to main content.
+ */
+
+export class CcBlockCli extends LitElement {
+  static get properties() {
+    return {
+      toggle: { type: String, reflect: true },
```
pdesoyres-cc, line 23: suggestion: if the state can take only two values, why not using a `boolean` and rename the property to `isOpen` (`false` by default) ?

src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+ * A display component with mostly HTML+CSS and an open/close toggle feature organized with slots to display information such as CLI commands.
+ *
+ * @cssdisplay block
+ *
+ * @slot tab-title - The title of the tab. Try to only use text.
+ * @slot link - A zone dedicated to link, for example to documentation.
+ * @slot content - A zone dedicated to main content.
+ */
+
+export class CcBlockCli extends LitElement {
```
pdesoyres-cc, line 20: fix: Since this component seams to be very generic, I'm not sure we should name it with `cli` suffix. Suggestion:

* `CcBlockDetail`
* `CcBlockFooter`

src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+          align-items: center;
+          background-color: transparent;
+          border: none;
+          display: flex;
+          font-size: 1em;
+          padding: 0.25em 0.6em 0.35em;
+        }
+
+        .tab:hover {
+          border-radius: var(--cc-border-radius-default);
```
pdesoyres-cc, line 82: fix: we need a default value

src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        .tab.open {
+          background-color: var(--cc-color-bg-default);
```
pdesoyres-cc, line 86: fix: we need a default value

src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+          border-radius: var(--cc-border-radius-default) var(--cc-border-radius-default) 0 0;
```
pdesoyres-cc, line 87: fix: we need a default value

src/components/cc-block-cli/cc-block-cli.js
```diff
@@ -0,0 +1,122 @@
...
+        .tab.close:hover cc-icon {
+          transform: rotate(-90deg);
+        }
+
+        cc-icon {
+          margin-left: 0.1em;
+        }
+
+        .content {
+          background-color: var(--cc-color-bg-default);
```
pdesoyres-cc, line 108: fix: we need a default value

src/components/cc-block-cli/cc-block-cli.stories.js
```diff
@@ -0,0 +1,56 @@
...
+  <div slot="footer" style="background-color: var(--cc-color-bg-neutral); border-top: solid 2px #cdcdcd">
+    <cc-block-cli toggle="open">
+      <div slot="tab-title">Command line</div>
+      <a slot="link" href="">See documentation</a>
+      <div slot="content">Praesent tincidunt dapibus elit, sed posuere risus sodales quis. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Integer ultrices maximus ligula eu efficitur. Nulla luctus ipsum vitae dui gravida, sodales ornare dolor viverra. Vivamus pretium scelerisque ultricies. Nulla consectetur euismod euismod. Interdum et malesuada fames ac ante ipsum primis in faucibus. Cras dignissim id sapien non facilisis. Suspendisse vitae pretium orci. Nunc aliquam pellentesque nunc, at tincidunt nisi lobortis a. </div>
+    </cc-block-cli>
+  </div>
+`;
+
+export const openByDefaultStory = makeStory(conf, {
```
pdesoyres-cc, line 50: nitpick; maybe rename to `openStory`

pdesoyres-cc (changes requested):
Nice PR @clara-layus-cc , I left some comments and suggestions. My main concern is about the name of this component.

src/components/cc-block-cli/cc-block-cli.js
clara-layus-cc: It’s a good idea. Initially, the values were not the same but now that there are only two, passing with booleans will simplify the code.

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+  static get properties() {
+    return {
+      isOpen: { type: Boolean, reflect: true, attribute: 'is-open' },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {Boolean} Sets the state of the toggle behaviour. */
```
hsablonniere, line 26: nitpick: `boolean` should be lowercase

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  _onClickToggle() {
+    this.isOpen = !this.isOpen;
+  }
+
+  render() {
+    return html`
+      <div class="menu">
+        <button class="tab ${this.isOpen ? 'open' : 'close'}" @click=${this._onClickToggle}>
```
hsablonniere, line 37: note: not a real issue but since the component has the `is-open` attribute reflected, we can remove the `${this.isOpen ? 'open' : 'close'}` (see below for CSS).

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+          transform: rotate(-180deg);
+        }
+
+        .tab.close:hover cc-icon {
+          transform: rotate(-90deg);
+        }
+
+        cc-icon {
+          margin-left: 0.1em;
+        }
```
hsablonniere, line 97: issue: this should be grouped with the `.tab cc-icon` selector. Maybe both can be put in just `cc-ion` first.

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+
+        .tab:hover {
+          border-radius: var(--cc-border-radius-default, 0.25em);
+        }
+
+        .tab.open {
+          background-color: var(--cc-color-bg-default, #fff);
+          border-radius: var(--cc-border-radius-default, 0.25em) var(--cc-border-radius-default, 0.25em) 0 0;
+          margin-bottom: 0;
+        }
```
hsablonniere, line 81: note: we can rely directly on the reflected `is-open` attribute with this selector instead:

```
:host([is-open="true"]) .tab
```

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+
+        .tab cc-icon {
+          transition: all 0.3s;
+        }
+
+        .tab.open cc-icon {
+          transform: rotate(-180deg);
+        }
```
hsablonniere, line 89: note: we can rely directly on the reflected `is-open` attribute with this selector instead:

```
:host([is-open="true"]) cc-icon
```

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+
+        .tab.close:hover cc-icon {
+          transform: rotate(-90deg);
+        }
```
hsablonniere, line 93: note: we can rely directly on the reflected `is-open` attribute with this selector instead:

```
:host([is-open="true"]) .tab:hover cc-icon
```

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+          display: flex;
+          justify-content: space-between;
+        }
+
+        .tab {
+          align-items: center;
+          background-color: transparent;
+          border: none;
+          display: flex;
+          font-size: 1em;
```
hsablonniere, line 69: question: is this needed?

src/components/cc-block-detail/cc-block-detail.js
```diff
@@ -0,0 +1,114 @@
...
+          display: block;
+        }
+
+        .menu {
+          column-gap: 1em;
+          display: flex;
+          justify-content: space-between;
+        }
+
+        .tab {
```
hsablonniere, line 64: issue: we may need an explicit `font-family` (set to a token or unset), for now it's a different font than the rest because `button` has default styles.

src/components/cc-block-detail/cc-block-detail.stories.js
```diff
@@ -0,0 +1,56 @@
...
+  tags: ['autodocs'],
+  title: '🧬 Molecules/<cc-block-detail>',
+  component: 'cc-block-detail',
+};
+
+const conf = {
+  component: 'cc-block',
+};
+
+const ccBlockHtmlExemple = `
```
hsablonniere, line 15: nitpick: `ccBlockHtmlExemple` => `ccBlockHtmlExample`

src/components/cc-block-detail/cc-block-detail.stories.js
```diff
@@ -0,0 +1,56 @@
...
+
+const defaultHtmlExample = `
+  ${ccBlockHtmlExemple}
+  <div slot="footer" style="background-color: var(--cc-color-bg-neutral); border-top: solid 2px #cdcdcd">
+    <cc-block-detail>
+      <div slot="tab-title">Command line</div>
+      <a slot="link" href="">See documentation</a>
+      <div slot="content">Praesent tincidunt dapibus elit, sed posuere risus sodales quis. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Integer ultrices maximus ligula eu efficitur. Nulla luctus ipsum vitae dui gravida, sodales ornare dolor viverra. Vivamus pretium scelerisque ultricies. Nulla consectetur euismod euismod. Interdum et malesuada fames ac ante ipsum primis in faucibus. Cras dignissim id sapien non facilisis. Suspendisse vitae pretium orci. Nunc aliquam pellentesque nunc, at tincidunt nisi lobortis a. </div>
+    </cc-block-detail>
+  </div>
```
hsablonniere, line 28: suggestions:

* if we use the `slot="footer-left"` instead of `footer`, we'll get the gray background, the border and padding
  * we can remove the shadow in the cc-block in another commit if Marion agrees
* I we do this, we can remove the extra div and set the `slot` directly on the `cc-block-detail`

(same for the other story)

hsablonniere (changes requested):
Hey @clara-layus-cc, congrats for this first component!! :+1: :clap: 

I added a few comments here and there. Here are additional feedbacks:

* I want footers with CLI tab + documentation links to look the same as footers with just a documentation link
  * We discussed that this component should maybe support situations without the CLI tab but I'm not sure
  * Let's discuss it together
* We will need to make sure the link looks the same as other links (color, hover...), maybe it's not that important for the story of this component as it's just a slot
* The following border radius are not the same as the design (bottom left, bottom right, top right)
* I have some doubts about the `iconArrowUp`, I like the animation but I'm not sure about the directions
* I like the focus state but I think the hover is too suble

src/components/cc-block-detail/cc-block-detail.js
Galimede: I was thinking the same :+1:

Galimede (approved):
Hey Clara, nice work and congrats for your first component! :tada: 

I agree with Hubert feedbacks, don't have a lot to say on my side, GG! :muscle:

src/components/cc-block-detail/cc-block-detail.js
clara-layus-cc: Yes because all its properties have default values that do not match the design.

florian-sanders-cc (changes requested):
congrats on your first component @clara-layus-cc 

The code is very clean and well documented, this is great work!

As discussed the timing was difficult and I could not make a11y recommendations before the implementation so we need to modify some parts of the component to fix a few issues.

This is normal and absolutely not your fault :wink:  

Let's discuss the fixes together in sync when you are available (with @roberttran-cc if he wants / has the time to join).

Here are a few global notes:
- **suggestion:** for the icon animation + orientation, we need to agree on a standard and use the same everywhere. The `cc-block` component already implements an animation for the same usecase: [cc-block - ribbon with toggle story](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-molecules-cc-block--ribbon-with-toggle). I personally prefer the one from `cc-block` but it should probably be discussed with Marion and we'll adjust to use the same everywhere (we'll open a separate issue if `cc-block` needs to be modified),
- **nitpick:** I think we should rename the component to `cc-block-details` (plural) because the singular doesn't exactly mean the same in English (very subtle difference though).
- **suggestion:** add a story for cases without documentation links.

src/components/cc-block-detail/cc-block-details.test.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-env node, mocha */
```
florian-sanders-cc, line 1: **Todo:** you can remove this file, tests on stories are now handled automatically by a plugin :+1:

florian-sanders-cc (changes requested):
Only need to remove the `.test` file but apart from that, all good.

Excellent work @clara-layus-cc, you rock (yeah, I know, only old people use this phrase nowadays :older_man:)!!

src/components/cc-block-detail/cc-block-details.js
```diff
@@ -0,0 +1,124 @@
...
+      <div class="menu">
+        <div class="links">
+          <slot name="link"></slot>
+        </div>
+        <button class="button" aria-expanded="${this.isOpen}" aria-controls="content" @click=${this._onClickToggle}>
+          <slot name="button-text"></slot>
+          <cc-icon .icon="${iconArrowDown}"></cc-icon>
+        </button>
+      </div>
+      ${this.isOpen ? html` <div id="content" class="content"><slot name="content"></slot></div> ` : null}
```
florian-sanders-cc, line 50: **suggestion:** sorry I didn't notice this before
With components that show / hide content, it's better (not mandatory) to place the content in the DOM and hide / show it through CSS (or the `hidden` attribute).

Why is it better? Because when it's supported, we may use the [hidden="until-found" (only works in Chrome)](https://developer.chrome.com/docs/css-ui/hidden-until-found). 
Is it always better? No, sometimes it's better for perfs to only add content if it's needed but since we're dealing with simple / short content here, I think we're good :+1:

florian-sanders-cc (approved):
All good! Great work again @clara-layus-cc :wink:

florian-sanders-cc:
> @clara-layus-cc This seems good to me (code and SB preview). I only have one last thing. The tab order of the toggle button and the link is not natural. I wonder if we want to "fix" it. @florian-sanders-cc WDYT?

Damn my bad I didn't pay enough attention & wasn't specific enough in my guidelines. Yes, we need to place the link as the very last item of the component :+1: (feel free to ping me if it's unclear @clara-layus-cc, especially because these days I'm really absent minded :sweat_smile:)

hsablonniere (approved):

PR #1353, opened 2025-03-07
feat(cc-date-range-selector): move custom date range form inside popover body

pdesoyres-cc:
# Context

* When user selects a `custom` date range, a form containing the `since` and the `from` date inputs is displayed under the selector.
* This makes the component having a column based layout
* This layout fits well in the `cc-logs-app-runtime` component because it has got a left panel with the instances.
* But it doesn't fit with the future `cc-logs-addon-runtime` component which doesn't have a left column.
* It also wont fit the `cc-logs-app-access` component when we will integrate a date selection in it.

# What this PR do?

The main purpose of this PR is to change how the custom date range form is display inside the `cc-date-range-selector` component.

For the moment I left all the a11y considerations apart, I know there are a lot to do but this will be done in another PR.

So there are several commits:

* `cc-popover` needed two additional css props so that we can tweak the trigger button display without affecting the buttons inside the popover body
* Next commit is about changing the layout of the `cc-date-range-selector` component
* Next one is about adding a `left` slot on the `cc-logs-control` component so that we can put the instances panel
  * This was needed because we need to handle this layout:
image
* Last commit is about applying the change on `cc-logs-app-runtime` component

# How to review?

* check the code
* check the stories
* check the sandbox

### DIFF
--- a/sandbox/cc-logs-app-runtime/cc-logs-app-runtime-sandbox.js
+++ b/sandbox/cc-logs-app-runtime/cc-logs-app-runtime-sandbox.js
@@ -71,9 +71,11 @@ class CcLogsAppRuntimeSandbox extends LitElement {
         <cc-button type="submit">Apply</cc-button>
       </form>
 
-      <cc-smart-container ${ref(this._smartContainerRef)}>
-        <cc-logs-app-runtime-beta class="cc-logs-app-runtime main"></cc-logs-app-runtime-beta>
-      </cc-smart-container>
+      <div class="main">
+        <cc-smart-container ${ref(this._smartContainerRef)}>
+          <cc-logs-app-runtime-beta class="cc-logs-app-runtime"></cc-logs-app-runtime-beta>
+        </cc-smart-container>
+      </div>
     `;
   }
 
@@ -94,14 +96,15 @@ class CcLogsAppRuntimeSandbox extends LitElement {
           min-height: 0;
         }
 
-        cc-smart-container {
+        .cc-logs-app-runtime {
           display: flex;
           flex: 1;
           flex-direction: column;
           min-height: 0;
         }
 
-        .cc-logs-app-runtime {
+        .main {
+          display: grid;
           flex: 1;
           min-height: 0;
         }
--- a/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
+++ b/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
@@ -216,7 +216,6 @@ export class CcLogsAppRuntime extends LitElement {
     return html`
       <div class=${classMap(overlay)}>
         <div class=${classMap(wrapper)}>
-          ${this._renderDateRangeSelection()} ${this._renderInstances()}
           <div class="logs-wrapper">${this._renderLogs()}</div>
           ${this._renderLoadingProgress()}
         </div>
@@ -254,6 +253,7 @@ export class CcLogsAppRuntime extends LitElement {
 
     return html`<cc-logs-instances-beta
       class="instances"
+      slot="left"
       .state=${state}
       @cc-logs-instances:selection-change=${this._onInstanceSelectionChange}
     ></cc-logs-instances-beta>`;
@@ -324,6 +324,8 @@ export class CcLogsAppRuntime extends LitElement {
         @cc-logs-control:option-change=${this._onLogsOptionChange}
       >
         <div slot="header" class="logs-header">
+          ${this._renderDateRangeSelection()}
+
           <cc-logs-message-filter-beta
             class="logs-message-filter"
             .filter=${this._messageFilter}
@@ -340,6 +342,8 @@ export class CcLogsAppRuntime extends LitElement {
             @cc-button:click=${this._onFullscreenToggle}
           ></cc-button>
         </div>
+
+        ${this._renderInstances()}
       </cc-logs-control-beta>
 
       ${this.state.type === 'loaded' &&
@@ -388,6 +392,8 @@ export class CcLogsAppRuntime extends LitElement {
       css`
         /* stylelint-disable no-duplicate-selectors */
         :host {
+          --instances-width: 20em;
+
           display: block;
         }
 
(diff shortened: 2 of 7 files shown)


src/components/cc-logs-date-range-selector/cc-logs-date-range-selector.js
```diff
@@ -147,8 +150,23 @@ export class CcLogsDateRangeSelector extends LitElement {
...
+
+          return html`
+            <div class="options-popover-trigger--with-detail">
+              <span>${i18n('cc-logs-date-range-selector.option.custom')}</span>
+              <span class="date-range">${dateRange}</span>
+            </div>
+          `;
+        } else {
+          return i18n('cc-logs-date-range-selector.option.custom');
+        }
```
hsablonniere, line 168: nitpick: the else can be remove

src/components/cc-logs-date-range-selector/cc-logs-date-range-selector.js
```diff
@@ -131,9 +133,10 @@ export class CcLogsDateRangeSelector extends LitElement {
 
   /**
    * @param {LogsDateRangeSelectOption} option
-   * @return {string}
+   * @param {boolean} withDate
+   * @return {string|TemplateResult}
    */
-  _getOptionLabel(option) {
+  _getOptionLabel(option, withDate) {
```
hsablonniere, line 139: nitpick: should this be renamed to a `_renderOptionLabel` since it can return a `TemplateResult`?

src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
```diff
@@ -471,7 +469,7 @@ export class CcLogsAppRuntime extends LitElement {
         }
 
         .overlay-logs-wrapper {
-          left: 50%;
+          left: calc(50% + var(--instances-width) / 2);
```
florian-sanders-cc, line 472: **suggestion:** to me this is a smell that the `.overlay-logs-wrapper` should be handled by & be relative to the component it is supposed to be in the middle of (`cc-logs`).

It could be through a slot so that parent component can customize the content as they want.

I don't have a very good knowledge of the components and how they all work together so as usual feel free to dismiss this comment :wink:

sandbox/cc-logs-app-runtime/cc-logs-app-runtime-sandbox.js
```diff
@@ -94,15 +96,17 @@ class CcLogsAppRuntimeSandbox extends LitElement {
...
           flex: 1;
           flex-direction: column;
           min-height: 0;
         }
 
-        .cc-logs-app-runtime {
+        .main {
+          display: flex;
           flex: 1;
+          flex-direction: column;
```
florian-sanders-cc, line 109: **suggestion:**

```suggestion
          display: grid;
          flex: 1;
```

florian-sanders-cc (approved):
Well done @pdesoyres-cc the design looks good and it's a good way to make this component work for all contexts (with or without a left column) :clap: 

We have some work to do to make it accessible but we already had work to do about this before this refactoring so let's handle that separately. 
It only confirms that we need to stabilize the components before diving into complex a11y fixes and I guess these components should be stable enough very soon :+1:.

All my comments are suggestions, mostly about CSS. They are fairly opinionated so feel free to dismiss them, your code is good as it is :wink:

hsablonniere (changes requested):
Well done @pdesoyres-cc :clap: 

* I added some small comments :wink:
* Could you please add a story for the left slot in the 3rd commit?
  * It can be just a div with some placeholder text and a border or background

src/components/cc-popover/cc-popover.js
```diff
@@ -57,6 +57,8 @@ import '../cc-button/cc-button.js';
  * @cssprop {Size} --cc-popover-padding - Sets the padding of the floating area (default 0.5em).
  * @cssprop {Number} --cc-popover-z-index - Sets the z-index of the floating content (defaults: `999`).
  * @cssprop {Width} --cc-popover-trigger-button-width - Sets the width of the trigger button (defaults: `inherit`).
+ * @cssprop {FontWeight} --cc-popover-trigger-button-font-weight - Sets the font-weight CSS property of the trigger button (defaults: `bold`).
+ * @cssprop {TextTransform} --cc-popover-trigger-button-text-transform - Sets the text-transform CSS property of the trigger button (defaults: `uppercase`).
```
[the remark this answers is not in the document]
pdesoyres-cc: If you do that, every `cc-button` inside the `cc-popover` will be impacted. I want only the trigger button to be targeted by this custom prop.

src/components/cc-logs-control/cc-logs-control.js
```diff
@@ -437,32 +441,27 @@ export class CcLogsControl extends LitElement {
...
-          display: grid;
-          grid-template-areas:
-            'header scroll-button options-popover'
-            'logs   logs          logs';
-          grid-template-columns: 1fr auto auto;
-          grid-template-rows: max-content 1fr;
-          row-gap: 0.5em;
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
```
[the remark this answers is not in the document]
pdesoyres-cc: Just for my curiosity, why is it better to use grid in that case ?

src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
pdesoyres-cc: The right solution is to make the `cc-logs-control` have an `overlay` slot. I'll do that as soon as other logs panel need it.

src/components/cc-logs-control/cc-logs-control.js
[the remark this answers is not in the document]
florian-sanders-cc: It's not better and your code is perfectly fine as it is.
It just feels more straightforward for me: 
- the default behavior of grid is to place items below each other,
- the default behavior of flex is to place items next to each other.

So when I know I want to place things below each other + I don't want a wrapping system, I go for grid. This is an opinionated suggestion that only removes 1 CSS rule (`flex-direction`) :laughing: I won't suggest it in the next reviews but here I felt I could because the review was not big :smile:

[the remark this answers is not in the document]
pdesoyres-cc: I used that :star_struck: 

```css
 display: grid;
 gap: 0.5em;
 grid-template-rows:
     [header] auto
     [center] 1fr;
```

hsablonniere (approved):
LGTM :+1:

PR #1357, opened 2025-03-10
fix(cc-logs-date-range-selector): fix `since` and `until` dates edition

### LINKED ISSUE #1356: cc-logs-date-range-selector: bug when editing since or until dates
reproduce with:

* change since value
* change until value
* 💥 the since value is reset to its previous value
* change since value again
* 💥 the until value is reset to its previous value

pdesoyres-cc:
# What does this PR do?

This PR fixes #1356

# How to review?

* Just check that the scenario described in the issue is fixed
* One reviewer should be enough


florian-sanders-cc (approved):
Tested & approved!

Sorry it was missed during review :sweat_smile:

PR #1358, opened 2025-03-11
feat(cc-logs-app-access): add date range selector

pdesoyres-cc:
# What this PR do?

* This PR adds a date range selector to the `cc-logs-app-access`.
* It allows also to set the date range from the smart context, so that we will be able to store and restore date range selection using the Settings API.

# How to review?

* check the code
* check the stories
* check the sandbox


roberttran-cc (approved):
Perfect, LGTM!

PR #1361, opened 2025-03-12
ci(code-quality): add a check on the commit messages

pdesoyres-cc:
# What this PR do?

This PR adds a check on the commit(s) message(s), ensuring that they respect conventional commit format.

If one doesn't activate the local git hook that ensures conventional commit messages, the CI will be there to verify it anyway.

# How to test?

* pull this branch
* add a commit with an invalid message (use `--no-verify` option in the commit command to bypass the local git hook)
  ```
  git commit --no-verify -m "invalid message"
  ```
* push
* See the Tests and build job fail.
* Please remove your commit and force push afterward ;)

### DIFF
--- a/.github/workflows/test-and-build.yml
+++ b/.github/workflows/test-and-build.yml
@@ -24,6 +24,9 @@ jobs:
       - name: '[Prepare] Install dependencies'
         run: npm ci
 
+      - name: '[Code quality] Check commit message'
+        if: github.event_name == 'pull_request'
+        run: node ./node_modules/@commitlint/cli/cli.js --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
       - name: '[Code quality] Check ESLint'
         run: npm run lint
       - name: '[Code quality] Check Lit'


roberttran-cc (approved):

.github/workflows/test-and-build.yml
```diff
@@ -24,6 +24,9 @@ jobs:
       - name: '[Prepare] Install dependencies'
         run: npm ci
 
+      - name: '[Code quality] Check commit message'
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
```
florian-sanders-cc, line 29: **suggestion:** since we already have node + our dependencies, we could rely on the local commitlint instead of the latest remote one.
This a tiny bit safer (I'm nitpicking) since we control the upgrades of commitlint and if it's compromised, we're more likely to see it. With npx we may execute a comprised version before anyone has noticed any issue.

(haven't tested but this should work?)
```suggestion
        run: node ./node_modules/@commitlint/cli/cli.js --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
```

florian-sanders-cc (approved):
Great addition to our CI :+1: 
I suggested to move away from npx but it's non blocking so I'll approve as it is :wink:

PR #1367, opened 2025-03-25
fix(cc-logs-app-runtime): fix instance selection in live mode

### LINKED ISSUE #1366: cc-logs-app-runtime: in live mode, the instance selection is reset when new log arrives
The instances selection is reset to its initial value every time a new log arrives.

## Steps to reproduce

* use an app that emit logs permanently (no matter the emit speed)
* select live mode
* see some logs arriving
* change instance selection
* see some more logs arriving
* 💥 the instance selection is reset

pdesoyres-cc:
# What this PR do?

Fixes #1366 

There is also an additional commit that I wanted to tackle a long time ago. It is more like a workaround to something I don't understand with `lit-virtualizer`. Sometimes, `lit-virtualizer` asks for a `null` item to be rendered.

# How to review?

* Use the sandbox
* Select live mode
* Play with instance selection
* Instance selection should not be reset when new logs arrive

### DIFF
--- a/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
+++ b/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
@@ -147,24 +147,6 @@ export class CcLogsAppRuntime extends LitElement {
   }
 
   /**
-   * @param {CustomEvent<Array<string>>} event
-   */
-  _onInstanceSelectionChange(event) {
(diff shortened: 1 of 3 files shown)


florian-sanders-cc (approved):
LGTM!

Checked the code + Played with the sandbox as you suggested and everything seems fine!

PR #1374, opened 2025-03-31
chore(prettier): remove Lit property sorting plugin from Prettier config

Galimede:
## What does this PR do?

* This PR disables temporarily the plugin that sorts the Lit properties in our components.
* This has been done because Prettier can't handle multiple plugins that modify its AST.
* That's why we decided to remove the Lit properties sort plugin in favor for the import one for now.
* The PR also sorts the unsorted imports due to the plugin not functioning.
* Note: We plan to activate the Lit sort plugin again when we'll find a workaround. 

## How to review?

* Check the code.
* Do a `npm run format:check`, no warnings should come out.
* 2 reviewers should be enough.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,6 @@
 {
   "arrowParens": "always",
-  "plugins": ["prettier-plugin-organize-imports", "./prettier-rules/sort-lit-get-properties.js"],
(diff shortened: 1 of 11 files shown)


pdesoyres-cc (approved):
LGTM !

PR #1378, opened 2025-04-02
chore(prettier): switch config to allow multiple plugins

### LINKED ISSUE #1375: prettier: find a workaround to allow multiple Prettier plugins
A few weeks ago we noticed that our plugin that sorts imports was not working anymore. After investigation, we noticed that Prettier can't handle multiple plugins that modify the AST of the code. The one on the most righthand side of the config will always win and the others will not be triggered as they're probably overwritten.

It seems there are some workarounds available here and there, we'll need to test them and add them to the config.

For now we decided to deactivate our Lit properties sorting plugin as done in #1374.

Galimede:
Fixes #1375 

## What does this PR do?

* Removes the initial `.prettierrc` config
* Add a `prettier.config.js` config
* Add a hack to allow multiple plugins (c.f: the issue)

## How to review?

* Check the config
* Change an import and a prop in the `static get properties` of a component
* Everything should go back in place
* Two reviewers should be enough


prettier.config.js
```diff
@@ -0,0 +1,25 @@
+// @ts-expect-error
+// eslint-disable-next-line import/no-extraneous-dependencies
```
florian-sanders-cc, line 2: **suggestion:** I think we can add this file to the list of exemped files in the eslint config :+1: 
This way you can just keep the expect error?

prettier.config.js
```diff
@@ -0,0 +1,25 @@
...
+import pluginSortImports from 'prettier-plugin-organize-imports';
+import sortGetPropsPlugin from './prettier-rules/sort-lit-get-properties.js';
+
+/** @type {import("prettier").Parser}  */
+const myParser = {
+  ...pluginSortImports.parsers.babel,
+  parse: sortGetPropsPlugin.parsers.babel.parse,
+};
```
pdesoyres-cc, line 10: nitpick: can you add a explanation of the hack?

pdesoyres-cc (approved):
LGTM. Well done Mathieu. I've just a nitpick.

prettier.config.js
florian-sanders-cc, line 10: **question:** I don't know much about babel so my question may be stupid
How do we add a third plugin there?  
For instance, if we create another prettier plugin, it will probably also rely on babel. Will this work (I don't think so but I'm trying to understand :sweat_smile:)?

```js
```suggestion
const myParser = {
  ...pluginSortImports.parsers.babel,
  ...myCustomPlugin.parsers.babel,
  parse: sortGetPropsPlugin.parsers.babel.parse,
};
```

Anyway, whether it's future proof or not, it's a good solution for now!

florian-sanders-cc (approved):
Nice one @Galimede it works just fine!

I only have a suggestion and a question, nothing blocking ;) 

Gg

prettier.config.js
Galimede: Didn't think a bout it, will do :+1:

prettier.config.js
Galimede: I haven't tested but I guess it would just work like the example you provided

PR #1381, opened 2025-04-07
cc-addon-linked-apps: update zones API call in smart

### LINKED ISSUE #1380: cc-addon-linked-apps: update zones API call in smart
Based on last version of the clever-client, we should pass `ownerId` to the `/v4/products/zones` endpoint.

roberttran-cc:
Closes #1380.

# What to review

- one commit to bump `clever-client.js` dependency
- one commit to update the `cc-addon-linked-apps` smart

# How to review

- check the code
- checkout the branch and check with the smart demo that the component properly pass the `ownerId` query param to the `/v4/products/zones` endpoint

One person should be enough.

### DIFF
--- a/src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
+++ b/src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
@@ -1,7 +1,7 @@
 // @ts-expect-error FIXME: remove when clever-client exports types
 import { getLinkedApplications } from '@clevercloud/client/esm/api/v2/addon.js';
 // @ts-expect-error FIXME: remove when clever-client exports types
-import { getAllZones } from '@clevercloud/client/esm/api/v4/product.js';
+import { listZones } from '@clevercloud/client/esm/api/v4/product.js';
 // @ts-expect-error FIXME: remove when clever-client exports types
 import { ONE_DAY } from '@clevercloud/client/esm/with-cache.js';
 import { sendToApi } from '../../lib/send-to-api.js';
@@ -56,7 +56,7 @@ defineSmartComponent({
  */
 function fetchApplications({ apiConfig, signal, ownerId, addonId }) {
   return Promise.all([
-    fetchZones({ apiConfig, signal }),
+    fetchZones({ apiConfig, signal, ownerId }),
     fetchLinkedApplications({ apiConfig, signal, ownerId, addonId }),
   ]).then(([zones, applications]) => {
     return applications.map((app) => {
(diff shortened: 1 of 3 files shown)


src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
```diff
@@ -79,11 +79,12 @@ function fetchApplications({ apiConfig, signal, ownerId, addonId }) {
...
  * @param {Object} parameters
  * @param {ApiConfig} parameters.apiConfig
+ * @param {string} parameters.ownerId
  * @param {AbortSignal} parameters.signal
  * @returns {Promise<Zone[]>}
  */
-function fetchZones({ apiConfig, signal }) {
-  return getAllZones().then(sendToApi({ apiConfig, signal, cacheDelay: ONE_DAY }));
+function fetchZones({ apiConfig, signal, ownerId }) {
+  return listZones({ ownerId }).then(sendToApi({ apiConfig, signal, cacheDelay: ONE_DAY }));
```
Galimede, line 87: issue: we can't use cacheDelay anymore as we might not retrieve the zones for other orgas if you switched in the console.

Galimede (changes requested):
Hey Bob, just one thing to note otherwise LGTM! :muscle:

src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
roberttran-cc: Wow, thank you, I'll fix this. :brain:

Galimede (approved):

src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
florian-sanders-cc: @Galimede doesn't the cache system provided by `sendToApi` take into account the `queryParams`? I think they are part of the cache key and if they are.

If `queryParams` are part of the cache key, I think it's fine and we'll only rely on the client cache if we've made a call with this specific `ownerId` before :thinking: (not sure, that's just how I understand the `sendToApi` code)

florian-sanders-cc:
LGTM but I'm not sure about the cache removal here :thinking:

src/components/cc-addon-linked-apps/cc-addon-linked-apps.smart.js
Galimede: Yeah you're right, that's my bad, nice catch! :hugs:

roberttran-cc: TIL. Thanks to both of you, I'll fix this!

florian-sanders-cc (approved):
All good! Gg @roberttran-cc :sunglasses:

PR #1386, opened 2025-04-14
Feat/cc block detail integration

clara-layus-cc:
## What does this PR do?

- Integrates the cc-block-details component into other components.

## How to review?

- Check the default component story, where the cc-block-details component is integrated, within the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/cc-block-detail-integration/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story),

## Component list

- [x] cc-domain-management
- [x] cc-env-var-form
- [x] cc-tcp-redirection-form
- [x] cc-addon-backups

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -23,9 +23,11 @@ import {
 import { focusBySelector } from '../../lib/focus-helper.js';
 import { generateDocsHref } from '../../lib/utils.js';
 import { accessibilityStyles } from '../../styles/accessibility.js';
+import { cliCommandsStyles } from '../../styles/cli-commands.js';
 import { ccLink, linkStyles } from '../../templates/cc-link/cc-link.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-badge/cc-badge.js';
+import '../cc-block-details/cc-block-details.js';
 import '../cc-block-section/cc-block-section.js';
 import '../cc-block/cc-block.js';
 import '../cc-button/cc-button.js';
@@ -66,6 +68,7 @@ export class CcDomainManagement extends LitElement {
       dnsInfoState: { type: Object, attribute: 'dns-info-state' },
       domainFormState: { type: Object, attribute: 'domain-form-state' },
       domainListState: { type: Object, attribute: 'domain-list-state' },
+      resourceId: { type: String, attribute: 'resource-id' },
       _sortedDomains: { type: Array, state: true },
     };
   }
@@ -87,6 +90,9 @@ export class CcDomainManagement extends LitElement {
 
   constructor() {
     super();
+
+    /** @type {string} Sets the resource id for documentation */
+    this.resourceId = 'xxx_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx';
 
     /** @type {DomainManagementDnsInfoState} Sets the state of the DNS info section */
     this.dnsInfoState = { type: 'loading' };
@@ -350,12 +356,14 @@ export class CcDomainManagement extends LitElement {
             ${this.domainListState.type === 'loaded' ? this._renderDomains(this._sortedDomains) : ''}
           </cc-block-section>
 
-          <div slot="footer-right">
-            ${ccLink(
-              DOMAIN_NAMES_DOCUMENTATION,
-              html`<cc-icon .icon="${iconInfo}"></cc-icon> ${i18n('cc-domain-management.names.documentation.text')}`,
-            )}
-          </div>
+          <cc-block-details slot="footer-left">
+            <div slot="button-text">${i18n('cc-block-details.cli.text')}</div>
+            <div slot="link">
+              <cc-icon .icon="${iconInfo}"></cc-icon>
+              <a href="${DOMAIN_NAMES_DOCUMENTATION}">${i18n('cc-domain-management.names.documentation.text')}</a>
+            </div>
+            <div slot="content">${i18n('cc-domain-management.names.cli.content', { resourceId: this.resourceId })}</div>
+          </cc-block-details>
         </cc-block>
 
         <cc-block>
@@ -398,13 +406,14 @@ export class CcDomainManagement extends LitElement {
           ${this.dnsInfoState.type === 'loaded'
             ? this._renderDnsInfo(this.dnsInfoState.cnameRecord, this.dnsInfoState.aRecords)
             : ''}
-
-          <div slot="footer-right">
-            ${ccLink(
-              DNS_DOCUMENTATION,
-              html`<cc-icon .icon="${iconInfo}"></cc-icon> ${i18n('cc-domain-management.dns.documentation.text')}`,
-            )}
-          </div>
+          <cc-block-details slot="footer-left">
+            <div slot="button-text">${i18n('cc-block-details.cli.text')}</div>
+            <div slot="link">
+              <cc-icon .icon="${iconInfo}"></cc-icon>
+              <a href="${DNS_DOCUMENTATION}">${i18n('cc-domain-management.dns.documentation.text')}</a>
+            </div>
+            <div slot="content">${i18n('cc-domain-management.dns.cli.content', { resourceId: this.resourceId })}</div>
+          </cc-block-details>
         </cc-block>
       </div>
     `;
@@ -646,6 +655,7 @@ export class CcDomainManagement extends LitElement {
     return [
       accessibilityStyles,
       linkStyles,
+      cliCommandsStyles,
       css`
         :host {
           display: block;
--- a/src/components/cc-env-var-form/cc-env-var-form.js
+++ b/src/components/cc-env-var-form/cc-env-var-form.js
@@ -1,8 +1,10 @@
 import { css, html, LitElement } from 'lit';
 import { iconRemixInformationFill as iconInfo } from '../../assets/cc-remix.icons.js';
 import { generateDocsHref } from '../../lib/utils.js';
-import { ccLink, linkStyles } from '../../templates/cc-link/cc-link.js';
+import { cliCommandsStyles } from '../../styles/cli-commands.js';
+import { linkStyles } from '../../templates/cc-link/cc-link.js';
 import { i18n } from '../../translations/translation.js';
+import '../cc-block-details/cc-block-details.js';
 import '../cc-block/cc-block.js';
 import '../cc-button/cc-button.js';
 import '../cc-env-var-editor-expert/cc-env-var-editor-expert.js';
@@ -46,6 +48,7 @@ export class CcEnvVarForm extends LitElement {
       context: { type: String, reflect: true },
       heading: { type: String, reflect: true },
       readonly: { type: Boolean, reflect: true },
+      resourceId: { type: String, attribute: 'resource-id' },
       restartApp: { type: Boolean, attribute: 'restart-app' },
       state: { type: Object },
       _editorsState: { type: Object, state: true },
@@ -59,6 +62,9 @@ export class CcEnvVarForm extends LitElement {
 
     /** @type {string} Defines add-on name used in some heading/description (depending on context). */
     this.addonName = '?';
+
+    /** @type {string} Sets the resource id for documentation */
+    this.resourceId = 'xxx_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx';
 
     /** @type {string} Defines application name used in some heading/description (depending on context). */
     this.appName = '?';
@@ -372,12 +378,14 @@ export class CcEnvVarForm extends LitElement {
             `
           : ''}
 
-        <div slot="footer-right">
-          ${ccLink(
-            `${ENV_VAR_DOCUMENTATION}`,
-            html`<cc-icon .icon="${iconInfo}"></cc-icon>${i18n('cc-env-var-form.documentation.text')}`,
-          )}
-        </div>
+        <cc-block-details slot="footer-left">
+          <div slot="button-text">${i18n('cc-block-details.cli.text')}</div>
+          <div slot="link">
+            <cc-icon .icon="${iconInfo}"></cc-icon>
+            <a href="${ENV_VAR_DOCUMENTATION}">${i18n('cc-env-var-form.documentation.text')}</a>
+          </div>
+          <div slot="content">${i18n('cc-env-var-form.cli.content', { resourceId: this.resourceId })}</div>
+        </cc-block-details>
       </cc-block>
     `;
   }
@@ -385,6 +393,7 @@ export class CcEnvVarForm extends LitElement {
   static get styles() {
     return [
       linkStyles,
+      cliCommandsStyles,
       // language=CSS
       css`
         :host {
--- a/src/translations/translations.en.js
+++ b/src/translations/translations.en.js
@@ -470,6 +470,21 @@ export const translations = {
   'cc-env-var-editor-simple.empty-data': `There are no variables.`,
   //#endregion
   //#region cc-env-var-form
+  'cc-env-var-form.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p class="text">
+        You can manage environment variables directly from your terminal using the commands below.
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>List environment variables:</dt>
+        <dd><code>clever env --app ${resourceId}</code></dd>
+        <dt>Get a sourceable env file:</dt>
+        <dd><code>clever env --app ${resourceId} -F shell</code></dd>
+        <dt>Add or update an environment variable:</dt>
+        <dd><code>clever env set VAR_NAME VAR_VALUE --app ${resourceId}</code></dd>
+      </dl>
+    `,
   'cc-env-var-form.description.config-provider': /** @param {{addonName: string}} _ */ ({ addonName }) =>
     sanitize`Configuration exposed to dependent applications. <a href="${generateDocsHref('/deploy/addon/config-provider/')}">Learn more</a><br>These variables will be injected as environment variables in applications that have the add-on <strong>${addonName}</strong> in their service dependencies.<br>Every time you update your changes, all the dependent applications will be automatically restarted.`,
   'cc-env-var-form.description.env-var': /** @param {{appName: string}} _ */ ({ appName }) =>
--- a/src/translations/translations.en.js
+++ b/src/translations/translations.en.js
@@ -272,6 +272,16 @@ export const translations = {
     sanitize`<p>If you choose to use <code>A</code> records, for instance with a root domain (APEX), you'll need to update them yourself. Follow our <a href="${generateDevHubHref('/changelog')}">changelog</a> or check our <a href="${generateDevHubHref('/api/v4/#load-balancers')}">v4 API documentation</a> for this.</p>`,
   'cc-domain-management.dns.a.heading': `A records`,
   'cc-domain-management.dns.a.label': `A Record values`,
+  'cc-domain-management.dns.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p>
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>Diagnose the current configuration:</dt>
+        <dd><code>clever diag --app ${resourceId}</code></dd>
+      </dl>
+    `,
   'cc-domain-management.dns.cname.desc': () =>
     sanitize`<p>Using a <code>CNAME</code> record is recommended. This keeps your configuration up to date.</p>`,
   'cc-domain-management.dns.cname.heading': `CNAME record`,
@@ -338,6 +348,21 @@ export const translations = {
   'cc-domain-management.list.primary.success': /** @param {{domain: string}} _ */ ({ domain }) =>
     `"${domain}" has been successfully marked as primary domain`,
   'cc-domain-management.main-heading': `Manage your domain names`,
+  'cc-domain-management.names.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p class="text">
+        You can manage domains directly from your terminal using the commands below. 
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>List domains:</dt>
+        <dd><code>clever domain --app ${resourceId}</code></dd>
+        <dt>Diagnose DNS records:</dt>
+        <dd><code>clever domain diag --app ${resourceId}</code></dd>
+        <dt>Add a domain:</dt>
+        <dd><code>clever domain add myapp.example.com --app ${resourceId}</code></dd>
+      </dl>
+    `,
   'cc-domain-management.names.documentation.text': `Domain names - Documentation`,
   'cc-domain-management.new-window': `New Window`,
   'cc-domain-management.tls.certificates.documentation.text': `TLS certificates - Documentation`,
--- a/src/translations/translations.en.js
+++ b/src/translations/translations.en.js
@@ -1458,6 +1458,21 @@ export const translations = {
     sanitize`You can create a redirection in the <strong>${namespace}</strong> namespace.`,
   //#endregion
   //#region cc-tcp-redirection-form
+  'cc-tcp-redirection-form.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p class="text">
+        You can manage TCP redirections directly from your terminal using the commands below.
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>List TCP redirections:</dt>
+        <dd><code>clever tcp-redirs --app ${resourceId}</code></dd>
+        <dt>Add a TCP redirection:</dt>
+        <dd><code>clever tcp-redirs add --namespace my-namespace --app ${resourceId}</code></dd>
+        <dt>Remove a TCP redirection:</dt>
+        <dd><code>clever tcp-redirs remove --namespace my-namespace {portNumber} --app ${resourceId}</code></dd>
+      </dl>
+    `,
   'cc-tcp-redirection-form.create.error': /** @param {{namespace: string}} _ */ ({ namespace }) => {
     return sanitize`Something went wrong while creating a TCP redirection in the <strong>${namespace}</strong> namespace.`;
   },
--- a/src/translations/translations.en.js
+++ b/src/translations/translations.en.js
@@ -84,6 +84,19 @@ export const translations = {
   'cc-addon-admin.update': `Update name`,
   //#endregion
   //#region cc-addon-backups
+  'cc-addon-backups.cli.content': /** @param {{addonId: string}} _ */ ({ addonId }) =>
+    sanitize`
+      <p>
+        You can manage backups directly from your terminal using the commands below. 
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>List available database backups:</dt>
+        <dd><code>clever database backups ${addonId}</code></dd>
+        <dt>Download a database backup:</dt>
+        <dd><code>clever database backups download ${addonId} BACKUP_ID</code></dd>
+      </dl>
+    `,
   'cc-addon-backups.close-btn': `Close this panel`,
   'cc-addon-backups.command-password': `This command will ask for your password, here it is:`,
   'cc-addon-backups.delete': /** @param {{createdAt: string|number}} _ */ ({ createdAt }) =>
--- a/src/translations/translations.fr.js
+++ b/src/translations/translations.fr.js
@@ -283,6 +283,16 @@ export const translations = {
     sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, par exemple pour un domaine racine (APEX), vous devrez vous-même assurer leur mise à jour. Pensez à suivre notre <a href="${generateDevHubHref('/changelog')}" lang="en">changelog</a> ou à lire la documentation de notre <a href="${generateDevHubHref('/api/v4/#load-balancers')}" lang="en">API v4</a> pour cela.</p>`,
   'cc-domain-management.dns.a.heading': `Enregistrements A`,
   'cc-domain-management.dns.a.label': `Valeurs d'enregistrement A`,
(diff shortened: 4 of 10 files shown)


pdesoyres-cc (commented):
Nice work @clara-layus-cc. I left some nitpicks. And also I'm wondering if it's a good idea to put the example commands themselves inside the translations. We need to duplicate those commands in french and english even if commands are not things that we can translate.

florian-sanders-cc (changes requested):
I have a few nitpicks and fixes (low impact though) but overall this is great!

About the "visuals":

- I think we need to cancel the `p` margins,
- We could emphasize the command names with some font weight,
- We could add some more padding to the code blocks (even it means they take up more space),
- Should code blocks take up the whole width or fit to their content? I'm not sure about this one :thinking:.

Here's an example of what I mean
cc-block-details in cc-domain-management with all the command names in bold, no margin for paragraphs and more padding for code blocks

Also, I think there are some things to discuss:
- Ideally, we would place command examples in `<dl><dt>command name</dt><dd><code>the command example</code></dd>...</dl>` but we would need to authorize these tags in the sanitizer (I personally think it's fine to add simple HTML tags to the sanitizer but maybe this should be discussed),
- in any case, the styling of the cc-block-details content should be the same (`<code>` padding, spacing, font family, etc.) between all components where it's being used but since it's slotted, it means `cc-block-details` cannot really control that. I think we should create some reusable style that we import in each component (like linkStyles or accessibilityStyles).

If any of this is unclear, ping me so we can discuss it in sync (+ all of this should probably be discussed with other reviewers anyway :thinking:)

florian-sanders-cc:
Actually, I wonder if we shouldn't go for a component to display the commands.

For instance a component that accepts an array of objects like these:

```ts
{ 
  commandName: string; // comes from i18n, 1 command name = 1 i18n key)
  command: string // no need for i18n here as @pdesoyres-cc pointed out
}
```

Then the component would generate:

```html
<dl>
  <div>
    <dt>Translated Command Name 1</dt>
    <dd><code>clever ...</code></dd>
  <div>
  <div>
    <dt>Translated Command Name 1</dt>
    <dd><code>clever ...</code></dd>
  <div>
</dl>
```

The component would encapsulate the styles so we wouldn't have to import them, just use the component and feed date to it :thinking:

hsablonniere (changes requested):
Well done @clara-layus-cc :wink: 

## Texts

* I'm hesitating on the "Command line" on the toggle button
  * "CLI" could also be nice
* "How to copy and use commands"
  * I wonder if we need this title
* I don't think we should have "command to" for each example
* I think they should have colon at the end ":"
* "Classic command"
  * "List domains:"
  * "Lister les domaines :"
* "Command to see the status of domains"
  * "Diagnose DNS records:"
  * "Diagnostiquer les enregistrements DNS :"
* "Command to add a domain"
  * "Add a domain:"
  * "Ajouter un domaine :"
  * You can use an example instead of `domainName`, something like `myapp.example.com`
* "quickly install the CLI and Clever Tools"
  * I seems like we are referring to 2 things
  * Maybe something like this:
    * "quickly install the Clever Tools (our CLI)"
    * "installez rapidement les Clever Tools (notre CLI)"
* The french text don't have the app ID
* The first sentence lacks a link to the CLI documentation

## UI

* 5 commands seems a bit much
  * I think we can remove "remove" and "favourite"
* The documentation link does not have the right color
* Be careful, the space after the icon and documentation the link is underlined
* The code blocks don't have the border that was on the Figma design
* The code blocks don't have the right spacing (see Figma design)
* There is still a rounded corner on the details block at the top left corner, we need to remove it
* The focus ring of the toggle button should be above the main block

## About others' feedbacks

> I'm wondering if it's a good idea to put the example commands themselves inside the translations

I think we should keep them in the translations as in some cases we'll put french examples in the commands.

> We could emphasize the command names with some font weight

I like that.

> We could add some more padding to the code blocks (even it means they take up more space),

Agreed, just like it is on the Figma design.

> Should code blocks take up the whole width or fit to their content? I'm not sure about this one 🤔.

I think so. I wanted a "GitHub readme look" so maybe I'm biased.

> in any case, the styling of the cc-block-details content should be the same (<code> padding, spacing, font family, etc.) between all components where it's being used but since it's slotted, it means cc-block-details cannot really control that. I think we should create some reusable style that we import in each component (like linkStyles or accessibilityStyles).

Agreed.

> Actually, I wonder if we shouldn't go for a component to display the commands.

Let's try it without for now so we have more feedbacks.
We could also put some styles in the cc-block-details with `::slotted`

roberttran-cc (approved):
Apart from the feedback from my fellow teammates (which I agree with), code-wise, LGTM!

About the content, I'd put the instructions for the CLI after the commands: if you already know these instructions or after first read, you don't want them to be this much present (ideally, it should be a notice that could be dismiss).

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -649,14 +659,6 @@ export class CcDomainManagement extends LitElement {
           gap: 1.5em;
         }
 
-        code {
```
florian-sanders-cc, line 652: **fix**: the `cc-domain-management` component still relies on `<code>` elements for its own texts so this style needs to remain here (and it can differ from the one you use for `code` elements in `cliCommands` :+1:

src/styles/cli-commands.js
```diff
@@ -0,0 +1,21 @@
...
+  span {
+    font-weight: bold;
+  }
+  code {
+    background-color: var(--cc-color-bg-neutral);
+    border-radius: var(--cc-border-radius-default, 0.25em);
+    font-family: var(--cc-ff-monospace);
+    padding: 0.5em 1em;
+    word-break: break-all;
+  }
```
florian-sanders-cc, line 20: **fix:** you'll have to be more specific with the selector so that you don't impact the style of components that import this

For instance, if you look a the preview, you'll see that all `<span>` of the `cc-domain-management` are now in bold (like the domain names in the list).

I think something like this should be enough:
```suggestion
  cc-block-details a {
    color: var(--cc-color-text-primary-highlight, blue);
  }
  
  cc-block-details p {
    margin: 0;
  }
  
  cc-block-details span {
    font-weight: bold;
  }
  
  cc-block-details code {
    background-color: var(--cc-color-bg-neutral);
    border-radius: var(--cc-border-radius-default, 0.25em);
    font-family: var(--cc-ff-monospace);
    padding: 0.5em 1em;
    word-break: break-all;
  }
```

Either this or we force people to use specific classes like we do for `linkStyles` (must be `cc-link` class).

I'm still really leaning towards a component (in the long run) because it would solve all these issues + would make it a lot easier later on to add clipboard buttons (we would only have to add them in one place and it wouldn't impact i18n) but this should be enough in the meantime :wink:

src/translations/translations.en.js
```diff
@@ -334,6 +345,19 @@ export const translations = {
...
+      <p class="text">
+        You can manage domains directly from your terminal using the commands below. 
+        If you haven't already, quickly install the Clever Tools (our CLI) using our <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <span>List domains:</span>
+      <code>clever domain --app ${appId}</code>
+      <span>Diagnose DNS records:</span>
+      <code>clever domain diag --app ${appId}</code>
+      <span>Add a domain:</span>
+      <code>clever domain add myapp.example.com --app ${appId}</code>
```
florian-sanders-cc, line 359: **suggestion:** we still need a way to regroup the command name & the command itself semantically.
I think headings would be overkill and `dl / dt / dd` should be fine here :+1: 

```suggestion
      <dl>
        <dt>List domains:</dt>
        <dd><code>clever domain --app ${appId}</code></dd>
        <dt>Diagnose DNS records:</dt>
        <dd><code>clever domain diag --app ${appId}</code></dd>
        <dt>Add a domain:</dt>
        <dd><code>clever domain add myapp.example.com --app ${appId}</code></dd>
      </dl>
```

src/translations/translations.fr.js
```diff
@@ -279,6 +282,14 @@ export const translations = {
     sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, par exemple pour un domaine racine (APEX), vous devrez vous-même assurer leur mise à jour. Pensez à suivre notre <a href="https://developers.clever-cloud.com/changelog/" lang="en">changelog</a> ou à lire la documentation de notre <a href="https://developers.clever-cloud.com/api/v4/#load-balancers" lang="en">API v4</a> pour cela.</p>`,
   'cc-domain-management.dns.a.heading': `Enregistrements A`,
   'cc-domain-management.dns.a.label': `Valeurs d'enregistrement A`,
+  'cc-domain-management.dns.cli.content': () =>
+    sanitize`
+      <p>
+      Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <span>Commande pour diagnostiquer l'installation actuelle.</span>
+      <code>clever diag</code>
```
florian-sanders-cc, line 291: **suggestion:** same, I think I'd go for dl dt dd even if there's only one

src/translations/translations.fr.js
```diff
@@ -344,6 +355,19 @@ export const translations = {
...
+      <p>
+      Vous pouvez gérer les domaines directement depuis votre terminal grâce aux commandes ci-dessous.
+      Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <span>Lister les domaines :</span>
+      <code>clever domain --app ${appId}</code>
+      <span>Diagnostiquer les enregistrements DNS :</span>
+      <code>clever domain diag --app ${appId}</code>
+      <span>Ajouter un domaine :</span>
+      <code>clever domain add myapp.example.com --app ${appId}</code>
```
florian-sanders-cc, line 369: **suggestion:**

```suggestion
            <dl>
        <dt>Lister les domaines :</dt>
        <dd><code>clever domain --app ${appId}</code></dd>
        <dt>Diagnostiquer les enregistrements DNS :</dt>
        <dd><code>clever domain diag --app ${appId}</code></dd>
        <dt>Ajouter un domaine :</dt>
        <dd><code>clever domain add myapp.example.com --app ${appId}</code></dd>
      </dl>
```

florian-sanders-cc (changes requested):
@clara-layus-cc great job on the wording and moving to a shared style

We need to be cautious to be sure we don't impact the styles of components that rely on this shared style, feel free to ping me if you want to discuss or if you don't agree :wink:

src/translations/translations.en.js
```diff
@@ -268,6 +271,16 @@ export const translations = {
...
   'cc-domain-management.dns.a.heading': `A records`,
   'cc-domain-management.dns.a.label': `A Record values`,
+  'cc-domain-management.dns.cli.content': () =>
+    sanitize`
+      <p>
+        If you haven't already, quickly install the Clever Tools (our CLI) using our <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>Command to diagnose the current installation:</dt>
+        <dd><code>clever diag</code></dd>
```
florian-sanders-cc, line 281: **nitpick:** add the `appId` just like other commands

src/translations/translations.fr.js
```diff
@@ -279,6 +282,16 @@ export const translations = {
...
   'cc-domain-management.dns.a.heading': `Enregistrements A`,
   'cc-domain-management.dns.a.label': `Valeurs d'enregistrement A`,
+  'cc-domain-management.dns.cli.content': () =>
+    sanitize`
+      <p>
+      Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>Commande pour diagnostiquer l'installation actuelle:</dt>
+        <dd><code>clever diag</code></dd>
```
florian-sanders-cc, line 292: **nitpick:** add the `appId` just like other commands

florian-sanders-cc (approved):
All good, just a very quick suggestion but great job :wink:

roberttran-cc:
One last suggestion: in `cc-block-detail-integration`, I think titles should be closer to their respective commands and there should be more space between command blocks (meaning "title + command").

Currently, a title is closer to the previous command than its related command.

Galimede (approved):
Hey Clara, nice work, LGTM! :muscle: 

I just wonder if it wouldn't be better to have the part where you mention to install the CLI as a dedicated translation. It would make it easier to maintain in the future.

src/translations/translations.fr.js
```diff
@@ -450,6 +478,21 @@ export const translations = {
...
+  'cc-env-var-form.cli.content': /** @param {{appId: string}} _ */ ({ appId }) =>
+    sanitize`
+      <p class="text">
+        Vous pouvez gérer les variables d'environnement directement depuis votre terminal en utilisant les commandes ci-dessous.
+        Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>Lister les variables d'environnement:</dt>
+        <dd><code>clever env --app ${appId}</code></dd>
+        <dt>Récupérer un fichier de variables d’environnement executable:</dt>
```
florian-sanders-cc, line 490: **nitpick:** 
```suggestion
        <dt>Récupérer un fichier de variables d'environnement executable&nbsp;:</dt>
```

florian-sanders-cc (changes requested):
Well done, the integration looks clean!

I have a few nitpicks, small fixes and a few notes:
- I think the sentence order is weird ("you can manage .. through CLI" and then "Install the CLI"), I would first invite people to install the CLI and then say "... the commands below" but it's just me, feel free to ignore,
- In French, there's always a space before double punctuation like ":", ";"so in your French translations, you should have something like "Lister les variables d'environnement&nbsp;:" or "Lister les variables d'environnement :" (`nbsp;` is not really essential here),
- Note that the fixes requested about the port number and the missing `--app` apply to both French & English :+1:,
- Don't forget to rebase, this way the first commits will go away since they are already on master now :+1:

src/components/cc-env-var-form/cc-env-var-form.js
```diff
@@ -42,6 +44,7 @@ export class CcEnvVarForm extends LitElement {
   static get properties() {
     return {
       addonName: { type: String, attribute: 'addon-name' },
+      appId: { type: String, attribute: 'app-id' },
```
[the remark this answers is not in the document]
pdesoyres-cc: maybe, 'app-or-addon-id'?

src/translations/translations.fr.js
```diff
@@ -279,6 +282,16 @@ export const translations = {
     sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, par exemple pour un domaine racine (APEX), vous devrez vous-même assurer leur mise à jour. Pensez à suivre notre <a href="https://developers.clever-cloud.com/changelog/" lang="en">changelog</a> ou à lire la documentation de notre <a href="https://developers.clever-cloud.com/api/v4/#load-balancers" lang="en">API v4</a> pour cela.</p>`,
   'cc-domain-management.dns.a.heading': `Enregistrements A`,
   'cc-domain-management.dns.a.label': `Valeurs d'enregistrement A`,
+  'cc-domain-management.dns.cli.content': /** @param {{appId: string}} _ */ ({ appId }) =>
+    sanitize`
+      <p>
+      Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
```
pdesoyres-cc, line 288: fix: link title needs to be in french

src/translations/translations.fr.js
```diff
@@ -344,6 +357,21 @@ export const translations = {
   'cc-domain-management.list.primary.success': /** @param {{domain: string}} _ */ ({ domain }) =>
     `"${domain}" a bien été défini comme nom de domaine principal`,
   'cc-domain-management.main-heading': `Gérez vos noms de domaine`,
+  'cc-domain-management.names.cli.content': /** @param {{appId: string}} _ */ ({ appId }) =>
+    sanitize`
+      <p>
+      Vous pouvez gérer les domaines directement depuis votre terminal grâce aux commandes ci-dessous.
+      Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
```
pdesoyres-cc, line 364: fix: link title needs to be in french

src/translations/translations.fr.js
```diff
@@ -450,6 +478,21 @@ export const translations = {
   'cc-env-var-editor-simple.empty-data': `Il n'y a pas de variable.`,
   //#endregion
   //#region cc-env-var-form
+  'cc-env-var-form.cli.content': /** @param {{appId: string}} _ */ ({ appId }) =>
+    sanitize`
+      <p class="text">
+        Vous pouvez gérer les variables d'environnement directement depuis votre terminal en utilisant les commandes ci-dessous.
+        Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
```
pdesoyres-cc, line 485: fix: link title needs to be in french

src/translations/translations.fr.js
```diff
@@ -1386,6 +1429,20 @@ export const translations = {
     sanitize`Vous pouvez créer une redirection dans l'espace de nommage <strong>${namespace}</strong>.`,
   //#endregion
   //#region cc-tcp-redirection-form
+  'cc-tcp-redirection-form.cli.content': /** @param {{appId: string}} _ */ ({ appId }) =>
+    sanitize`
+      <p class="text">
+        Vous pouvez gérer les redirections TCP directement depuis votre terminal grâce aux commandes ci-dessous.
+        Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.      </p>
```
pdesoyres-cc, line 1436: fix: link title needs to be in french

src/translations/translations.en.js
```diff
@@ -268,6 +271,16 @@ export const translations = {
     sanitize`<p>If you choose to use <code>A</code> records, for instance with a root domain (APEX), you'll need to update them yourself. Follow our <a href="https://developers.clever-cloud.com/changelog/">changelog</a> or check our <a href="https://developers.clever-cloud.com/api/v4/#load-balancers">v4 API documentation</a> for this.</p>`,
   'cc-domain-management.dns.a.heading': `A records`,
   'cc-domain-management.dns.a.label': `A Record values`,
+  'cc-domain-management.dns.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p>
+        If you haven't already, quickly install the Clever Tools (our CLI) using our <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installing Clever Tools - new window">documentation</a>.
+      </p>
+      <dl>
+        <dt>Command to diagnose the current installation:</dt>
```
florian-sanders-cc, line 280: **Question:** should we go for a common "UX writing style" for commands? 
Like do we go for verbs?
`Diagnose the current configuration` or `Diagnosing the current configuration?`

pdesoyres-cc (changes requested):
Well done Clara. I left a suggestion about translating the documentation link title into french.

Globally, I find that there are already a lot of repetitions:
- header and link to documentation
- the need to import styles
- commands are the same in french and in english, only the command labels are different.

Maybe it's ok for the moment, but I think that we will soon need a dedicated component.

HeleneAmouzou (approved):
GG, it looks very nice ! :clap: 
I have nothing more to add to what my colleagues have already said !

src/components/cc-env-var-form/cc-env-var-form.js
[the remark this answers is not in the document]
roberttran-cc: (5 years later) I like the `resourceId` proposal!

src/components/cc-block-details-command/cc-block-details-command.js
```diff
@@ -0,0 +1,106 @@
...
+  render() {
+    return html`
+      <div class="description">${this.description}</div>
+      <div class="command-container">
+        <code class="command">${this.command}</code>
+        <button class="copy-button" @click=${this._copyToClipboard} title="Copy command">
+          ${this.copied
+            ? html`<cc-icon .icon="${iconSuccess}" size="lg"></cc-icon>`
+            : html`<cc-icon .icon="${iconCopy}" size="lg"></cc-icon>`}
+        </button>
```
florian-sanders-cc: **fix:** the i18n is missing and the button state (copy / copied) is not conveyed to assistive technologies.

It's a tricky one because we may not want to change the state of the button (copy vs copied) but dispatch a notification instead? 
Would we want this notification to be visible? I think the icon change you went for is enough for the visual notification :thinking: 

I think we should go for a hidden live region but I'm not sure yet.

Anyway, these questions make me think this subject should be handled separately, in a PR dedicated to the clipboard button component. I would remove the clipboard feature for now because it will be easier to reason about it with a dedicated component & PR.

src/components/cc-block-details-command/cc-block-details-command.js
```diff
@@ -0,0 +1,106 @@
...
+    super();
+
+    /** @type {String} Sets the command. */
+    this.command = '';
+
+    /** @type {String} Sets the command description. */
+    this.description = '';
+
+    /** @type {Boolean} Feedback state for copy action. */
+    this.copied = false;
```
florian-sanders-cc: **fix:** this should be an internal prop because parents shouldn't be able to mess with this one

```suggestion
    /** @type {Boolean} Feedback state for copy action. */
    this._copied = false;
```

src/components/cc-block-details-command/cc-block-details-command.js
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @cssdisplay block
+ *
+ */
+
+export class CcBlockDetailsCommand extends LitElement {
+  static get properties() {
+    return {
+      command: { type: String, reflect: true },
+      copied: { type: Boolean },
```
florian-sanders-cc: fix: this should be an internal prop because parents shouldn't be able to mess with this one
```suggestion
      _copied: { type: Boolean, state: true },
```

src/components/cc-block-details-command/cc-block-details-command.js
```diff
@@ -0,0 +1,106 @@
...
+ * The main purpose is to be used with a cc-block-details.
+ *
+ * @cssdisplay block
+ *
+ */
+
+export class CcBlockDetailsCommand extends LitElement {
+  static get properties() {
+    return {
+      command: { type: String, reflect: true },
```
florian-sanders-cc: **fix:** the `reflect` option is mainly used for styling purposes so I don't think you need it here (same for description). 

The goal is usually to do something like this from outside the parent component:
```
cc-input-text[multi] { ... stuff to apply when the cc-input-text has the multi attribute }
```

So in your context, I don't see a usecase for that

src/components/cc-block-details-command/cc-block-details-command.js
florian-sanders-cc, line 1: **fix:** we would need to add the same semantics as we already discussed in our last round of review: `dl`, `dt` & `dd`.

The issue is that this component targets each command and not the list itself so devs would have to do something like this:
```html
<dl>
  <cc-block-details-command ...></cc-block-details-command> <!-- which would contain the <dt> & <dd> tags -->
  <cc-block-details-command ...></cc-block-details-command> <!-- which would contain the <dt> & <dd> tags -->
</dl>
```

This is doable but not great because:
- the child component uses `dt` & `dd` but devs shouldn't know about this and more importantly, they shouldn't have to adapt to it? 
- What if we changed the semantics of the inner HTML and used `li` instead of `dt` and `dd`, then the parent components would have to be updated to use `ul` instead of `dl` :thinking: (this is theoritical, it's very unlikely we change the semantics)
- This makes me think the subject is trickier than it looks if we want to get it right. Right now I'm not sure about which way to go, I would have tried a list component with dynamic slots but I'm not sure it's a good idea (some people consider this an anti-pattern),
- Another option would be to handle this like `select` and `option` components are usually handled in a design system: one component for the list (`sl-select` in Shoelace for instance) that expects list items components (`sl-option` in Shoelace for instance) to be slotted in it :thinking: 
- I do think slots are interesting for this usecase and we should explore solutions around them :eyes: 
- Or maybe we just document that it's meant to be used in a `dl` tag :shrug: 

Anyway, this should be discussed and I think this means the component requires a separate, dedicated PR

florian-sanders-cc (changes requested):
Well done @clara-layus-cc the component raises some very interesting questions that make me want to experiment a little bit but it's not my subject so I'll refrain myself :smile: 

I do think it's trickier than it looks and it should be handled separately (and I do think it's worth spending time on a dedicated component because I don't like having that much HTML inside our translations + such a big translation node).

Don't try to fix my comments, these are mostly questions / opinions and I might be wrong so it's better to wait for other opinions and decide what to do next :+1:

src/components/cc-block-details-command/cc-block-details-command.js
HeleneAmouzou: I think it would be interesting to test this option : a component for the list with list items components

HeleneAmouzou (commented):
I have just catch-up with your last updates, GG @clara-layus-cc ! 👏
I have nothing much to say, agree with @florian-sanders-cc on the quantity of HTML in the translations and I well understand now the necessity of having separate components. I am curious to see the feedback of my collegues and the next iterations of this component.

roberttran-cc (approved):
LGTM!

src/lib/i18n/i18n-sanitize.js
florian-sanders-cc, line 1: **nitpick:** @clara-layus-cc we could remove this change (I had to add the tags to the list when you were away because I needed it for API token components :innocent: )

src/translations/translations.fr.js
```diff
@@ -1442,6 +1495,20 @@ export const translations = {
     sanitize`Vous pouvez créer une redirection dans l'espace de nommage <strong>${namespace}</strong>.`,
   //#endregion
   //#region cc-tcp-redirection-form
+  'cc-tcp-redirection-form.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p class="text">
+        Vous pouvez gérer les redirections TCP directement depuis votre terminal grâce aux commandes ci-dessous.
+        Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installation des Clever Tools - nouvelle fenêtre">documentation</a>.      </p>
```
florian-sanders-cc: **fix:** for all doc lines, use the `generateDocs` helper that we've introduced last week :innocent:

florian-sanders-cc (changes requested):
@clara-layus-cc almost there, should be easy to fix but ping me if it's unclear!

src/translations/translations.fr.js
```diff
@@ -454,6 +492,21 @@ export const translations = {
...
+  'cc-env-var-form.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p class="text">
+        Vous pouvez gérer les variables d'environnement directement depuis votre terminal en utilisant les commandes ci-dessous.
+        Si ce n'est pas déjà fait, installez rapidement les Clever Tools (notre CLI) à l'aide de notre <a href="https://github.com/CleverCloud/clever-tools/blob/master/docs/setup-systems.md" title="documentation - Installation des Clever Tools - nouvelle fenêtre">documentation</a>.
+      </p>
+      <dl>
+        <dt>Lister les variables d'environnement :</dt>
+        <dd><code>clever env --app ${resourceId}</code></dd>
+        <dt>Récupérer un fichier de variables d'environnement executable :</dt>
```
Galimede: typo: `executable` => `exécutable`

src/translations/translations.en.js
```diff
@@ -272,6 +285,16 @@ export const translations = {
     sanitize`<p>If you choose to use <code>A</code> records, for instance with a root domain (APEX), you'll need to update them yourself. Follow our <a href="${generateDevHubHref('/changelog')}">changelog</a> or check our <a href="${generateDevHubHref('/api/v4/#load-balancers')}">v4 API documentation</a> for this.</p>`,
   'cc-domain-management.dns.a.heading': `A records`,
   'cc-domain-management.dns.a.label': `A Record values`,
+  'cc-domain-management.dns.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p>
+        To install Clever Tools CLI, follow the instructions from the <a href="${generateDocsHref('/cli/install/')}" title="documentation - Install Clever Tools - new window">documentation</a>.
```
pdesoyres-cc, line 291: nitpick: here, the header `You can manage XXX directly from your terminal using the commands below.` is not present, is this intended?

src/translations/translations.fr.js
```diff
@@ -283,6 +296,16 @@ export const translations = {
     sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, par exemple pour un domaine racine (APEX), vous devrez vous-même assurer leur mise à jour. Pensez à suivre notre <a href="${generateDevHubHref('/changelog')}" lang="en">changelog</a> ou à lire la documentation de notre <a href="${generateDevHubHref('/api/v4/#load-balancers')}" lang="en">API v4</a> pour cela.</p>`,
   'cc-domain-management.dns.a.heading': `Enregistrements A`,
   'cc-domain-management.dns.a.label': `Valeurs d'enregistrement A`,
+  'cc-domain-management.dns.cli.content': /** @param {{resourceId: string}} _ */ ({ resourceId }) =>
+    sanitize`
+      <p>
+      Pour installer les Clever Tools (CLI), suivez les instructions de la <a href="${generateDocsHref('/cli/install/')}" title="documentation - Installer les Clever Tools - nouvelle fenêtre - en Anglais">documentation</a>.
```
pdesoyres-cc, line 302: nitpick: here the `Vous pouvez gérer XXX directement depuis votre terminal en utilisant les commandes ci-dessous.` sentence is not present. Is this intended?

src/translations/translations.en.js
clara-layus-cc: The component in question is present on the same screen as another one which already has the introductory sentence so I think it is not necessary to repeat it again.

src/translations/translations.fr.js
clara-layus-cc: The component in question is present on the same screen as another one which already has the introductory sentence so I think it is not necessary to repeat it again.

PR #1388, opened 2025-04-15
feat(cc-zone): use diplayName when it's defined

hsablonniere:
## What does this PR do?

- Show the `displayName` of a zone when it's defined (even if it does not have a `scope:private` tag)
- Rework the stories to reflect better this modified behavior
- Rework the stories to reflect better the inline display of the cc-zone in headers

## How to review?

- look at the cc-zone stories
- look at the cc-zone-input stories

The zone with a `displayName` should show this instead of "City, Country".


Galimede (approved):
LGTM! :muscle:

PR #1405, opened 2025-05-15
refactor: add snakeCase to change-case.js

### LINKED ISSUE #1294: cc-oauth-consumer: init component
## Context

This component is dedicated to manage an oauth consumer.

## To be discussed

- [x] naming (component, prop, etc.)
  - [ ] access => read ? manage => write ? 
- [x] naming oauth-consumer ou oauth-consumer-info ? 
- [ ] keep or remove placeholder in inputs ?
- [ ] Textes et traductions à (voir avec David
- [x] Options order
- [ ] Mutualiser les traductions ou ok d'avoir des doublons ?

## Technical details
This component will be split into 2 components:
- `cc-oauth-consumer-form` : a component displaying a form to create or edit of an oauth consumer,
- `cc-oauth-consumer-info` : a component displaying oauth consumer information.

### Attributes/properties

| `state`    | `Object` | `{ type:  'loading' }`        | `true`    | Sets the state of the component.        |

### CSS custom property

| `xxx` | `Yyyy` | `null`        | TODO        |

### CSS part

| Name  | Description |
|-------|-------------|
| `xxx` | TODO        |

## Outputs

| `cc-oauth-consumer-form:create` | `Yyyy`    | Fires when clicking the creation form submit button        |
| `cc-oauth-consumer-form:update` | `Yyyy`    | Fires when clicking the update form submit button.        |
| `cc-oauth-consumer-form:delete` | `Yyyy`    | Fires whenever the delete button is clicked.        |

## Type Definitions

```typescript
export type OAuthConsumerFormState =
  | OAuthConsumerFormStateIdleCreate
  | OAuthConsumerFormStateCreating
  | OAuthConsumerFormStateIdleUpdate
  | OAuthConsumerFormStateUpdating
  | OAuthConsumerFormStateDeleting
  | OAuthConsumerFormLoading
  | OAuthConsumerFormStateError;

export interface OAuthConsumerFormStateIdleCreate {
  type: 'idle-create';
}

export interface OAuthConsumerFormStateCreating {
  type: 'creating';
}

export interface OAuthConsumerFormStateIdleUpdate {
  type: 'idle-update';
  values?: {
    name: string;
    homePageUrl: string;
    appBaseUrl: string;
    description: string;
    image: string;
    rights: Array<Right>;
  };
}

export interface OAuthConsumerFormStateUpdating {
  type: 'updating';
  values?: {
    name: string;
    homePageUrl: string;
    appBaseUrl: string;
    description: string;
    image: string;
    rights: Array<Right>;
  };
}

export interface OAuthConsumerFormStateDeleting {
  type: 'deleting';
  values?: {
    name: string;
    homePageUrl: string;
    appBaseUrl: string;
    description: string;
    image: string;
    rights: Array<Right>;
  };
}

export interface OAuthConsumerFormLoading {
  type: 'loading';
}

export interface OAuthConsumerFormStateError {
  type: 'error';
}

export interface NewOauthConsumer {
  name: string;
  homePageUrl: string;
  appBaseUrl: string;
  description: string;
  image: string;
  rights: Array<Right>;
}

interface Right {
  value: string;
  isEnable: boolean;
}

```

## Data/APIs

* `GET api.clever-cloud.com/v2/organisations/{orga_id}/consumers/{key}` 
* `CREATE api.clever-cloud.com/v2/organisations/{orga_id}/consumers`
* `PUT api.clever-cloud.com/v2/organisations/{orga_id}/consumers/{key}` 
* `DELETE api.clever-cloud.com/v2/organisations/{orga_id}/consumers/{key}`

## Stories & states

* Idle create
* Idle update
* Creating
* Updating
* Skeleton/loading
* Error
* Deleting
* Simulation

## TO DO

- [x] design meetings with Marion
- [x] Listing API
- [x] Onboarding formulaires/smart
- [ ] Create component (create and edit form)
  - [ ]  Create
    - [x] Integration (HTML/CSS)
    - [x] i18n
    - [x]  Inputs (props, attributs, slots, etc.)
    - [x] Form management
    - [x] Events
    - [x] Smart, API calls
  - [ ] Edit
    - [x] Integration (HTML/CSS)
    - [x] i18n
    - [x]  Inputs (props, attributs, slots, etc.)
    - [x] Form management
    - [x] Events
    - [x] Smart, API calls
- Create component (information screen)
   - [x] Integration (HTML/CSS)
    - [x] i18n
    - [x]  Inputs (props, attributs, slots, etc.)
    - [x] Smart, API calls
- Accessibility
- Q/A
- Tests

HeleneAmouzou:
Fix #1294

## What does this PR do?
- Introduce `cc-oauth-consumer-form` component, this component display two forms, one for the creation of an OAuth consumer and one for the update with a danger zone to delete the consumer,
- Refactor of `change-case` helper to add snakeCase,
- Add the `cc-oauth-consumer-forms` events to `event-map`.

## How to review?
- Check the code and the commits,
- Check the stories in the [preview](),
- Check the responsive,
- Play with demo-smart,
- QA is available on [notion](url)

## TODO
- [ ] Add preview link
- [ ] Merge the two screens of QA
- [ ] Validate texts

## QUESTIONS
- Do we need a working version on demo-smart ? Is is risky regarding the security to commit a OAuth consumer key ?


hsablonniere (dismissed):
Hey @HeleneAmouzou, well done on this one :wink: 

I haven't looked much at the code, I'll let the others review. I focused on the UI, the i18n and the docs.

## UI

* Why do we have fixed length buttons?
  * It creates a double line button in french
  * I wonder if we should remove the `changes/les changements` from the button label
* We need a link to the docs at the bottom of the cc-block
  * https://www.clever-cloud.com/developers/api/howto/#oauth1
* Can we move the description after the name in the form?
  * This way we ask for all the URLs at once
* Can we add some details for the inputs?
  * A but like GitHub does here https://github.com/settings/applications/new

## i18n

I proposed my changes in this branch: https://github.com/CleverCloud/clever-components/tree/cc-oauth-consumer-form/init-component--hubert

Let me know what you think.

### Docs

* The smart docs contain "CREATE", "REMOVE" or "UPDATE", this is not a valid HTTP method
* Can you add the `smart-mode="create"` and `smart-mode="update"` in the smart examples
* You can remove the leading `/` from the URL in the API endpoints docs for create

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.js
```diff
@@ -0,0 +1,177 @@
...
+
+    onEvent('cc-oauth-consumer-form-delete', () => {
+      updateComponent('state', (state) => {
+        state.type = 'deleting';
+      });
+      api
+        .deleteOauthConsumer()
+        .then(() => {
+          notifySuccess(i18n('cc-oauth-consumer-form.delete.success'));
+          window.dispatchEvent(new Event('oauth-consumer-delete'));
```
florian-sanders-cc, line 118: **question/fix:** 

@HeleneAmouzou you would need to create 1 class per event you want to dispatch and dispatch them from the component instead of window (we used to dispatch these from window just like you did but we changed this with the new events PR):


```suggestion
            component.dispatchEvent(new CcOauthConsumerFormDeletedEvent());
```

You will probably need to adapt your code within the console to match the new event type + don't forget the Fired events section in the smart md doc :wink:

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.js
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  /** @return {Promise<RawOauthConsumer>} */
+  getOauthConsumer() {
+    return getOauthConsumer({ id: this._ownerId, key: this._key }).then(sendToApi({ apiConfig: this._apiConfig }));
+  }
+
+  /**
+   * @param {OauthConsumerWithoutKeyAndSecret} data
+   * @return {Promise<any>}
```
florian-sanders-cc, line 150: **nitpick:** if your promise doesn't return anything you can use `void`


```suggestion
   * @return {Promise<void>}
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.js
```diff
@@ -0,0 +1,177 @@
...
+      description: data.description,
+      picture: data.picture,
+      rights: rights,
+    };
+    return update({ id: this._ownerId, key: this._key }, updatedOauthConsumer).then(
+      sendToApi({ apiConfig: this._apiConfig }),
+    );
+  }
+
+  /** @return {Promise<any>} */
```
florian-sanders-cc, line 173: **nitpick:** if your promise doesn't return anything, you can use `void`


```suggestion
  /** @return {Promise<void>} */
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.md
```diff
@@ -0,0 +1,61 @@
...
+      API_HOST: "",
+      API_OAUTH_TOKEN: "",
+      API_OAUTH_TOKEN_SECRET: "",
+      OAUTH_CONSUMER_KEY: "",
+      OAUTH_CONSUMER_SECRET: "",
+    },
+    "ownerId": "",
+    "key": "",
+}'>
+  <cc-oauth-consumer-form></cc-oauth-consumer-form>
```
florian-sanders-cc, line 58: **fix:** 
```suggestion
  <cc-oauth-consumer-form smart-mode="update"></cc-oauth-consumer-form>
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+## 🌐 API endpoints
+
+| Method   | URL                                       | Cache?  |
+|----------|-------------------------------------------|---------|
+| `GET`    | `/v2/organisations/${ownerId}/consumers/${key}`          | Default |
+| `REMOVE` | `/v2/organisations/${ownerId}/consumers/${key}`          | Default |
+| `UPDATE` | `/v2/organisations/${ownerId}/consumers/${key}`          | Default |
```
florian-sanders-cc, line 40: **fix:**

(you could also format the md table so that rows have the same length)
```suggestion
| `DELETE` | `/v2/organisations/${ownerId}/consumers/${key}`          | Default |
| `PUT` | `/v2/organisations/${ownerId}/consumers/${key}`          | Default |
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-update.md
```diff
@@ -0,0 +1,61 @@
...
+
+  ```ts
+interface ApiConfig {
+  API_HOST: string,
+  API_OAUTH_TOKEN: string,
+  API_OAUTH_TOKEN_SECRET: string,
+  OAUTH_CONSUMER_KEY: string,
+  OAUTH_CONSUMER_SECRET: string,
+}
+```
```
florian-sanders-cc, line 32: **fix:** you should add a `Fired events` section (see `cc-domain-management` or `cc-orga-member-list` smart docs)

src/lib/change-case.js
```diff
@@ -93,3 +93,18 @@ export function camelCase(input) {
...
 }
+
+/**
+ * snakeCase('string'); //=> 'string'
+ * snakeCase('dot.case'); //=> 'dot_case'
+ * snakeCase('PascalCase'); //=> 'pascal_case'
+ * snakeCase('camelCase'); //=> 'camel_case'
+ * snakeCase('version 1.2.10'); //=> 'version_1_2_10'
+ * @param {String} input
+ * @returns {String}
```
florian-sanders-cc, line 104: **nitpick:**

```suggestion
 * @param {string} input
 * @returns {string}
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,731 @@
...
+    this._formRef.value.reset();
+  }
+
+  updated() {
+    this._validateCheckboxGroup();
+  }
+
+  render() {
+    if (this.state.type === 'error') {
+      return html` <cc-notice slot="content" intent="warning" message="error"></cc-notice> `;
```
florian-sanders-cc, line 313: **fix:** missing i18n here

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.stories.js
```diff
@@ -0,0 +1,448 @@
...
+    urlInputElement.reportInlineValidity();
+    /** @type {CcInputText} */
+    const baseUrlInputElement = component._formRef.value.querySelector('[name="baseUrl"]');
+    baseUrlInputElement.reportInlineValidity();
+    /** @type {CcInputText} */
+    const descriptionInputElement = component._formRef.value.querySelector('[name="description"]');
+    descriptionInputElement.reportInlineValidity();
+    /** @type {CcInputText} */
+    const pictureInputElement = component._formRef.value.querySelector('[name="picture"]');
+    pictureInputElement.reportInlineValidity();
```
florian-sanders-cc, line 227: **suggestion:** you can remove all of this and query the form + use `requestSubmit()`, this way it will call `reportInlineValidity()` + set the focus by itself

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.stories.js
```diff
@@ -0,0 +1,448 @@
...
+    component._formRef.value.picture.validate();
+    /** @type {CcInputText} */
+    const urlInputElement = component._formRef.value.querySelector('[name="url"]');
+    urlInputElement.reportInlineValidity();
+    /** @type {CcInputText} */
+    const baseUrlInputElement = component._formRef.value.querySelector('[name="baseUrl"]');
+    baseUrlInputElement.reportInlineValidity();
+    /** @type {CcInputText} */
+    const pictureInputElement = component._formRef.value.querySelector('[name="picture"]');
+    pictureInputElement.reportInlineValidity();
```
florian-sanders-cc, line 257: **suggestion:** same, you don't need to call `validate` because it's done automatically when you change a value + you don't have to call `reportInlineValidity` on each form control, you can call `requestSubmit` on the form directly and it will handle all of this by itself


```suggestion
    component._formRef.value.requestSubmit();
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.stories.js
```diff
@@ -0,0 +1,448 @@
...
+    urlInput.value = oAuthConsumerData.url;
+    /** @type {HTMLInputElement} */
+    const baseUrlInput = component.shadowRoot.querySelector('form [name="baseUrl"]');
+    baseUrlInput.value = oAuthConsumerData.baseUrl;
+    /** @type {HTMLInputElement} */
+    const descriptionInput = component.shadowRoot.querySelector('form [name="description"]');
+    descriptionInput.value = oAuthConsumerData.description;
+    /** @type {HTMLInputElement} */
+    const pictureInput = component.shadowRoot.querySelector('form [name="picture"]');
+    pictureInput.value = oAuthConsumerData.picture;
```
florian-sanders-cc, line 147: **suggestion:** create some helper function to do the `querySelector` for you.
The helper could accept 2 params for instance, `form` and `formControlName`, and inside the function you would do something like this

```js
return form.querySelector(`[name=${formControlName}]`);
```

You could type the return type of the helper with something like this
```
@return {CcInputText | HTMLInputElement}
```

florian-sanders-cc (changes requested):
Great and not easy work @HeleneAmouzou :exploding_head: 

This is probably the most complex form component we've done (native checkbox group validation, custom validator for URLs, prefilled but only when the context is `update` etc.).
It's also one of the very few components that has 2 smart components so considering this was your first experience with smart components, really well done :clap: 

Most of my comments are nitpicks as usual and all of them are mostly things you couldn't know or stuff you forgot (and considering the size of the PR, 1819 lines added, it's perfectly normal).
I have a few global notes as well:
- **fix:** Every input that controls the format must specify the expected format,
  - You could use the `help` slot to specify `https://example.com` for every URL input.
- **nitpick:** In your stories, you could set `orga-href` to `https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-overview-cc-overview--orga-mode` (use the complete prod URL because relative URLs won't work on Storybook)
- **nitpick:** When loading, I think I'd put the `cc-button` elements to `disabled` instead of `skeleton` since we already know what value they will have and how they look like once loaded.
- **nitpick:** We should reset the `link:visited` style (same for my components :eyes:)
- **fix:** error empty, the name of the OAuth consumer is not set so it looks weird in the danger zone `Delete oAuth consumer ()`
  - I think this can only happen inside stories because the name will always be set in reality, it can be emptied afterwards but it will be there in the state.
  - This is more of a design thing and it's only my opinion but I think we could remove the ({consumer name}) anyway
- **nitpick:** text heading danger zone `oAuth` => `OAuth`
- **praise:** simulations & error stories are really cool, the stories are thorough and the types are great,
- **suggestion:** we should standardize how to handle the `go back` link when the viewport is narrow (maybe sync with Marion) because we both have this design but we don't really do the same thing (yours is centered horizontally for instance)
- **question/suggestion:** I wonder if it's worth mentioning "application" in all of the form labels

Haven't done the QA yet

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.events.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerWithoutKeyAndSecret} OauthConsumerWithoutKeyAndSecret
+ */
+
+/**
+ * Dispatched when an OAuth consumer creation is requested.
+ * @extends {CcEvent<OauthConsumerWithoutKeyAndSecret>}
+ */
+export class CcOauthConsumerFormCreateEvent extends CcEvent {
```
pdesoyres-cc, line 11: The naming of events should talk about the manipulated entity and not the thing that initiate the event. So I would have used the following names:

- cc-oauth-consumer-create
- cc-oauth-consumer-update
- cc-oauth-consumer-delete
- cc-oauth-consumer-created

This way, if, for example, another component deletes an oauth consumer, it will make more sense to reuse `cc-oauth-consumer-delete` event than `cc-oauth-consumer-form-delete`.

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,773 @@
...
+  'manageOrganisationsServices',
+  'managePersonalInformation',
+  'manageSshKeys',
+];
+
+/**
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormState} OAuthConsumerFormState
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormStateIdleUpdate} OauthConsumerFormStateIdleUpdate
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormStateCreating} OauthConsumerFormStateCreating
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormStateUpdating} OauthConsumerFormStateUpdating
```
pdesoyres-cc, line 89: nitpick: import not used

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,773 @@
...
+  'manageOrganisationsMembers',
+  'manageOrganisationsServices',
+  'managePersonalInformation',
+  'manageSshKeys',
+];
+
+/**
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormState} OAuthConsumerFormState
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormStateIdleUpdate} OauthConsumerFormStateIdleUpdate
+ * @typedef {import('./cc-oauth-consumer-form.types.js').OauthConsumerFormStateCreating} OauthConsumerFormStateCreating
```
pdesoyres-cc, line 88: nitpick: import not used

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.smart-create.js
```diff
@@ -0,0 +1,95 @@
...
+      });
+      api
+        .createOauthConsumer(data)
+        .then(
+          /** @param {OauthConsumer} response */ (response) => {
+            notifySuccess(i18n('cc-oauth-consumer-form.create.success', { oauthConsumerName }));
+            component.resetOauthConsumerForm();
+            updateComponent('state', { type: 'idle-create' });
+            const key = response.key;
+            window.dispatchEvent(new CcOauthConsumerFormDispatchKeyEvent(key));
```
pdesoyres-cc, line 46: fix: I think it's better to dispatch this event from the component instead of window. The event will bubble from component and can be catch more easily

```suggestion
            component.dispatchEvent(new CcOauthConsumerFormDispatchKeyEvent(key));
```

pdesoyres-cc (dismissed):
What a nice PR! Well done Hélène!

I left a few comments and I have one question:
* why do we need a `Go back` link in the creation form?
* Is it something we would like to generalize in our creation forms?

Galimede (commented):
Hi Hélène, nice work! :muscle: 

Apart from what my colleagues already said in the review, I don't have much to say. 

I just wonder if we should let the user be able to use the link to go back if wer'e in waiting mode. :thinking:

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,731 @@
...
+    }
+    const creatingContext = this.state.type === 'idle-create' || this.state.type === 'creating';
+    const isWaiting =
+      this.state.type === 'creating' || this.state.type === 'updating' || this.state.type === 'deleting';
+    const skeleton = this.state.type === 'loading';
+    let formValues = SKELETON_OAUTH_CONSUMER_FORM_DATA;
+
+    if (this.state.type === 'idle-update' || this.state.type === 'updating' || this.state.type === 'deleting') {
+      formValues = this.state.values;
+    }
```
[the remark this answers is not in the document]
florian-sanders-cc: Since nobody commented that one, I think you can resolve and decide what you prefer @HeleneAmouzou, both are fine :wink:

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.events.js
```diff
@@ -0,0 +1,83 @@
...
+    super(CcOauthConsumerCreatedEvent.TYPE, key);
+  }
+}
+
+/**
+ * Dispatched when an OAuth consumer update is requested.
+ * @extends {CcEvent<OauthConsumerWithoutKeyAndSecret>}
+ */
+export class CcOauthConsumerUpdateEvent extends CcEvent {
+  static TYPE = 'cc-oauth-consumer-update';
```
florian-sanders-cc, line 39: **nitpick:** I'm only thinking about it now because I've faced a similar case in my update component. It seems events like these are typically named `...-change` instead of `...-update` in other components.
Maybe we should stick `change` to match other components? (you would also need to rename the class :wink: 

```suggestion
  static TYPE = 'cc-oauth-consumer-change';
```

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.events.js
florian-sanders-cc, line 1: **nitpick:** I'm only thinking about it now because I've faced a similar case in my update component. The difference between `...-update` & `...-updated` / `...-create` & `...-created` is only one letter, which makes it very easy to mixup :see_no_evil: 

In my component I went for `...-has-changed`. It wouldn't work for `created`, maybe `...-has-been-created` or `...-was-created`?

Maybe @pdesoyres-cc has an opinion about that :eyes: 

Note: if you modify this, don't forget to update the listeners within the console :smile:

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,750 @@
...
+
+/**
+ * A component displaying a form to create an OAuth Consumer and a form to update an OAuth Consumer with a delete zone.
+ *
+ * @cssdisplay block
+ */
+export class CcOauthConsumerForm extends LitElement {
+  static get properties() {
+    return {
+      orgaHref: { type: String, attribute: 'orga-href' },
```
florian-sanders-cc, line 103: **fix:** not used anymore since you removed the go back link :+1:

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.js
```diff
@@ -0,0 +1,750 @@
...
+      state: { type: Object, attribute: false },
+      _shouldDisplayCheckboxGroupError: { type: Boolean, state: true },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {string|null} Sets the link to navigate back to the organisation information screen. */
+    this.orgaHref = null;
```
florian-sanders-cc, line 113: fix: not used anymore since you removed the go back link 👍

florian-sanders-cc (changes requested):
Great job @HeleneAmouzou, only a few details left :+1:

marion-izd:
We saw your questions during our little synchro this week. I've just done another pass on Storybook, and it's OK for me! Thanks for all your work :)

src/components/cc-oauth-consumer-form/cc-oauth-consumer-form.events.js
pdesoyres-cc: My opinion is that we should try to use the same policy everywhere. Maybe we can go for:

- `...-was-updated`
- `...-was-deleted`
- `...-was-created`

florian-sanders-cc (approved):

PR #1411, opened 2025-05-16
`cc-token-api-list`: support expired tokens

### LINKED ISSUE #1407: `cc-token-api-list`: show expired tokens and allow users to explicitely delete them
## Context

When we created the `cc-token-api-list`, the API was not ready to return expired tokens, only the active ones.

Now the API can return expired tokens as well as active tokens so we need to display them within the list, probably with a specific badge.

We also need a button that will delete an expired token.

Currently we're designing what this should look like, we'll start implementation right after that.

### LINKED ISSUE #1410: `cc-token-api-list`: import link styles
The link in the intro text doesn't have the proper link color because I forgot to import `linkStyles`

florian-sanders-cc:
Fixes #1410 
Fixes #1407 

## What does this PR do?

- Adapts the `cc-token-api-list` component to display expired tokens,
  - These token cards have a different style (background, opacity & button icon).
- Imports the missing styles for links. 

## How to review?

- Check the commits,
- Check the preview,
- Note that you cannot check the smart behavior because the Auth backend hasn't been released and the API doesn't return the `EXPIRED` | `ACTIVE` info for now.


florian-sanders-cc:
@marion-izd I chose not to go for the remove all button as I said in figma, is it okay with you? We can reconsider it if we get feedback asking for this feature?

src/components/cc-token-api-list/cc-token-api-list.js
```diff
@@ -216,215 +217,218 @@ export class CcTokenApiList extends LitElement {
   }
 
   static get styles() {
-    return css`
-      :host {
-        --list-item-icon-size: 1.2em;
+    return [
+      linkStyles,
+      css`
```
florian-sanders-cc, line 222: the rest is just prettier formatting / indenting

pdesoyres-cc (approved):
LGTM. I was a bit surprised by the deletion versus revoke icons, but it's ok like that :+1:

roberttran-cc (approved):
LGTM! Did not test the demo-smart though.

marion-izd:
test

florian-sanders-cc:
test

florian-sanders-cc:
@marion-izd test ping

PR #1427, opened 2025-06-02
Fix/opacity when disabled var

### LINKED ISSUE #617: Use the new `--cc-opacity-when-disabled` variable when relevant
# Context

With `cc-ssh-key-list` we introduced a new `--cc-opacity-when-disabled` variable to be used for instance when UI is in waiting mode.
The value of this variable is `0.65` because selecting a value below this may break a11y tests (color contrasts).

# Issue

In `cc-button` the opacity is set to `0.5` in waiting mode.
The `cc-button` should use the new `--cc-opacity-when-disabled`.

There might be other components to update like `cc-button`.

# Todo

- [x] Check which components have some kind of waiting mode
- [x] Use the variable for all these components.

clara-layus-cc:
Fixes #617 

# What does this PR do?

- Changes `opacity: 0.5` to `opacity: --cc-opacity-when-disabled`  in waiting or disabled mode.

# How to review?

- Check the commits,
- Check the preview,

### DIFF
--- a/src/components/cc-button/cc-button.js
+++ b/src/components/cc-button/cc-button.js
@@ -438,7 +438,7 @@ export class CcButton extends LitElement {
 
         button[aria-disabled='true'] {
           cursor: inherit;
-          opacity: 0.5;
+          opacity: var(--cc-opacity-when-disabled);
         }
 
         .skeleton {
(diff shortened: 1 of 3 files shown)


florian-sanders-cc (approved):
LGTM! Thanks for handling this @clara-layus-cc :sunglasses:

PR #1428, opened 2025-06-02
fix(cc-input-text): add the possibility to use command key on implicit submit

### LINKED ISSUE #1227: cc-input-text (multi):  add the possibility to use `command` key on implicit submit
Currently, we're only checking for the `Ctrl` key for a `Crtl+Enter` implicit submit on multi mode. However, common usage for mac OS is to use `cmd+Enter`.

clara-layus-cc:
Fixes #1227

# What does this PR do?

- Add the possibility to use `cmd+Enter` on mac OS

# How to review?

- Check the commits,


roberttran-cc (approved):
Tested on a Macbook Air & now working as expected: good job!

PR #1433, opened 2025-06-04
refactor(cc-input-date): improve help and error messages

### LINKED ISSUE #1391: `cc-input-date`: improve help text & error message
## Context

The `cc-input-date` component expects a specific format but by default, the help text only says that the value must be a date without specifying the format.

The error message in case the format is invalid says that it should be a date but doesn't provide any example.

HeleneAmouzou:
Fixes #1391

## What does this PR do?

- Updates help and errors message in the `cc-input-date`.

## How to review?

- Check the code,
- Check on the preview the `cc-input-date` and  dependant components, for instance `cc-token-api-creation-form`.


src/translations/translations.en.js
```diff
@@ -591,8 +591,8 @@ export const translations = {
   'cc-heptapod-info.storage-description': `Storage size`,
   //#endregion
   //#region cc-input-date
-  'cc-input-date.error.bad-input': `You must enter a date.`,
-  'cc-input-date.error.empty': `You must enter a value`,
+  'cc-input-date.error.bad-input': () => sanitize`You must enter a date. <br> For instance: 2025-06-03 17:25:50.`,
+  'cc-input-date.error.empty': () => sanitize`You must enter a value. <br> Format: YYYY-MM-DD HH:MM:SS.`,
```
roberttran-cc: **❔ question:** Is there a reason we have `<br>` here and not in the french translations? :thinking:

HeleneAmouzou: My bad, I just forgot to add them in the french translations !

roberttran-cc (approved):

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -94,25 +94,28 @@ export const helpMessage = makeStory(conf, {
   items: baseItems.map((p) => ({
     ...p,
     required: true,
-    innerHTML: '<p slot="help">Must be a date</p>',
+    innerHTML: '<p slot="help">Must be a date. Format: YYYY-MM-DD HH:MM:SS.</p>',
```
florian-sanders-cc: **nitpick/suggestion:** 

```suggestion
    innerHTML: '<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>',
```

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -94,25 +94,28 @@ export const helpMessage = makeStory(conf, {
...
+    component.reportInlineValidity();
+  },
 });
 
 export const errorMessageWithHelpMessage = makeStory(conf, {
   items: baseItems.map((p) => ({
     ...p,
     required: true,
-    errorMessage: 'You must enter a value',
+    errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix**: the `errorMessage` is supposed to be used for stuff the `cc-input-date` cannot handle on its own. 
The goal of your story is to showcase the error message that's automatically handled by `cc-input-date` so you should do the same as in the previous story (`onUpdateComplete` + `reportInlineValidity`) and keep the innerHTML below to showcase the help message :+1:

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -137,9 +140,9 @@ export const inlineWithErrorAndHelpMessages = makeStory(conf, {
     ...p,
     inline: true,
     required: true,
-    errorMessage: 'You must enter a value',
+    errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix:** same, should use `onUpdateComplete` + `reportInlineValidity` :+1:

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -201,16 +204,16 @@ export const customLabelStyle = makeStory(
       ...customBaseItems,
       ...customBaseItems.map((item) => ({
         ...item,
-        innerHTML: `<p slot="help">Must be a date</p>`,
+        innerHTML: `<p slot="help">Must be a date. Format: YYYY-MM-DD HH:MM:SS.</p>`,
```
florian-sanders-cc: **nitpick/suggestion:** 
```suggestion
        innerHTML: `<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>`,
```

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -201,16 +204,16 @@ export const customLabelStyle = makeStory(
...
       })),
       ...customBaseItems.map((item) => ({
         ...item,
-        errorMessage: 'You must enter a value',
+        errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix:** should be removed (use `onUpdateComplete` + `reportInlineValidity` instead)

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -201,16 +204,16 @@ export const customLabelStyle = makeStory(
...
       })),
       ...customBaseItems.map((item) => ({
         ...item,
-        errorMessage: 'You must enter a value',
-        innerHTML: `<p slot="help">Must be a date</p>`,
+        errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix:** should be removed (use `onUpdateComplete` + `reportInlineValidity` instead)

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -201,16 +204,16 @@ export const customLabelStyle = makeStory(
...
+        innerHTML: `<p slot="help">Must be a date. Format: YYYY-MM-DD HH:MM:SS.</p>`,
```
florian-sanders-cc: **nitpick/suggestion:** 
```suggestion
        innerHTML: `<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>`,
```

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -219,18 +222,18 @@ export const customLabelStyle = makeStory(
       ...customBaseItems.map((item) => ({
         ...item,
         inline: true,
-        innerHTML: `<p slot="help">Must be a date</p>`,
+        innerHTML: `<p slot="help">Must be a date. Format: YYYY-MM-DD HH:MM:SS.</p>`,
```
florian-sanders-cc: **nitpick/suggestion:**

```suggestion
        innerHTML: `<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>`,
```

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -219,18 +222,18 @@ export const customLabelStyle = makeStory(
...
       })),
       ...customBaseItems.map((item) => ({
         ...item,
         inline: true,
-        errorMessage: 'You must enter a value',
+        errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix:** should be removed (use `onUpdateComplete` + `reportInlineValidity` instead)

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -219,18 +222,18 @@ export const customLabelStyle = makeStory(
...
       })),
       ...customBaseItems.map((item) => ({
         ...item,
         inline: true,
-        errorMessage: 'You must enter a value',
-        innerHTML: `<p slot="help">Must be a date</p>`,
+        errorMessage: 'You must enter a value. Format: YYYY-MM-DD HH:MM:SS.',
```
florian-sanders-cc: **fix:** should be removed (use `onUpdateComplete` + `reportInlineValidity` instead)

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -219,18 +222,18 @@ export const customLabelStyle = makeStory(
...
+        innerHTML: `<p slot="help">Must be a date. Format: YYYY-MM-DD HH:MM:SS.</p>`,
```
florian-sanders-cc: **nitpick/suggestion:**

```suggestion
        innerHTML: `<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>`,
```

src/translations/translations.en.js
```diff
@@ -592,8 +592,8 @@ export const translations = {
   'cc-heptapod-info.storage-description': `Storage size`,
   //#endregion
   //#region cc-input-date
-  'cc-input-date.error.bad-input': `You must enter a date.`,
-  'cc-input-date.error.empty': `You must enter a value`,
+  'cc-input-date.error.bad-input': () => sanitize`You must enter a date. <br> For instance: 2025-06-03 17:25:50.`,
```
florian-sanders-cc: **question/suggestion:** I think the `cc-input-date` should send a date representing now + 15 mins for instance (or maybe 1 day, 1 week, not sure).
Hardcoded values are guaranteed to be obsolete and irrelevant at some point.

The choice of using a date in the future is not perfect: if we use the date for something in the past, it doesn't make sense. But this error message can be overridden by the components using the `cc-input-date` so we just have to find a default that works in most cases and let people customize if / when they need (this is why I chose to use a date in the future).

You can see an example in the `cc-token-api-creation-form` component (we can discuss it in sync if it's unclear)

src/translations/translations.en.js
florian-sanders-cc: **question:** on one hand I like the idea of specifying the format in the error message but on the other hand, it's mandatory that the format remain visible even when there are no errors so people will have to use the help slot for that (in the long run, we should probably also make the component display the help message by default and let people override / hide it if they need, but that's another subject).

TLDR; I would remove the expected format here because it must already be visible elsewhere

florian-sanders-cc (changes requested):
Well done @HeleneAmouzou, reviewing this PR made me think of several questions / ideas.

Don't worry about the number of comments: they are all related to the 2 only issues I found (using the `errorMessage` when it's not necessary & the help message content).

florian-sanders-cc (approved):
LGTM! Thanks for handling that @HeleneAmouzou :+1:

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -137,11 +161,14 @@ export const inlineWithErrorAndHelpMessages = makeStory(conf, {
...
-    errorMessage: 'You must enter a value',
     innerHTML: `
-      <p slot="help">Must be a date</p>
+      <p slot="help">Format: YYYY-MM-DD HH:MM:SS.</p>
     `,
   })),
+  /** @param {CcInputDate} component */
+  onUpdateComplete: (component) => {
+    component.reportInlineValidity();
+  },
```
roberttran-cc, line 172: **🔨 issue:** In the story, this call triggers the first component of the list, thus only showing the error message of the first `cc-input-date` and not the others.

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -219,27 +244,29 @@ export const customLabelStyle = makeStory(
...
         inline: true,
-        errorMessage: 'You must enter a value',
-        innerHTML: `<p slot="help">Must be a date</p>`,
+        innerHTML: `<p slot="help">Format: YYYY-MM-DD HH:MM:SS.</p>`,
       })),
     ],
+    /** @param {CcInputDate} component */
+    onUpdateComplete: (component) => {
+      component.reportInlineValidity();
+    },
```
roberttran-cc: **🔨 issue:** Same issue (check the [production Storybook](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-atoms-cc-input-date--custom-label-style) for reference).

pdesoyres-cc (approved):
Thanks for handling that @HeleneAmouzou

florian-sanders-cc (approved):
LGTM! Thanks @HeleneAmouzou :sunglasses:

Galimede (approved):
LGTM!

PR #1448, opened 2025-06-10
fix(cc-input-date, date-displayer,...): remove obsolete eslint comments that set mocha / node env

### LINKED ISSUE #1339: test files: remove obsolete eslint comments that set mocha / node env
## Context

Since we've moved to ESLint 9 & flat config, we no longer need comments to set the eslint environment.
The `/* eslint-env node, mocha */` comments should be removed from all test files.

If new errors arise after removing these comments, then it means the related file should be added in the proper section of the eslint.config.js file.

clara-layus-cc:
Fixes #1339 

# What does this PR do?

- Remove obsolete eslint comments that set mocha / node env

# How to review?

- Check the commits,


florian-sanders-cc (approved):

pdesoyres-cc (approved):
Thank you @clara-layus-cc

PR #1451, opened 2025-06-11
refactor(translations): remove the extra "s" on "information" word

HeleneAmouzou:
## What does this PR do?

Remove the extra "s" on the word "information".

## How to review?
- check the code,
- check the preview,


florian-sanders-cc (approved):
Nice catch! LGTM!

florian-sanders-cc:
Nitpick about the commit scope: do we consider typos a refactor or a fix? It depends but I think I would go for a "fix" (if someone was using the previous version and had noticed the typo, then they would want to see that we've fixed it in this version and upgrade)

HeleneAmouzou:
@florian-sanders-cc Thanks, I was hesitating between refactor and fix, let's go for fix then !

marion-izd (approved):

PR #1453, opened 2025-06-13
fix(form elements): use reportInlineValidity in error stories

### LINKED ISSUE #1447: Fix error stories of form inputs
## Issue 

While fixing an accessibility issue on the `cc-input-date`, we realize that error stories of form inputs currently display error messages on every input examples. The error messages should only be triggered when the input is required and empty.
We can also benefit from this fix to add different stories for different kinds of error (empty, badEmail, badType, etc.)

For example, the Error Message story of the `cc-input-text` : 

Image

## Form input components
 
- [x] `cc-input-date` 
- [ ] `cc-input-text`
- [ ] `cc-input-number`

HeleneAmouzou:
Fixes #1447

## What does this PR do?

- Reworks error stories of form elements to use `reportInlineValidity` on inputs,
- Simplifies the error stories to only display one case,
- Adds error stories with invalid format,
- Removes impossible states (ex : required inputs with a valid value and empty error message),
- Removes "." at the end of simple sentence.

## How to review ?

- Check to commits,
- Check the preview stories and compare with prod.

## Questions / Reflexions

- On the `cc-input-text`, I figured out that a component with `multi` prop can not be visually seen with expended input and an error message (as the input expends regarding the text added inside and the empty error message is triggered on empty input). I have left the example of a `cc-input-text` multilines on the `customLabelStyle` story, but we can not see it anymore on error mode as it is in prod.

### DIFF
--- a/src/components/cc-input-number/cc-input-number.stories.js
+++ b/src/components/cc-input-number/cc-input-number.stories.js
@@ -2,6 +2,10 @@ import { allFormControlsStory } from '../../stories/all-form-controls.js';
 import { makeStory, storyWait } from '../../stories/lib/make-story.js';
 import './cc-input-number.js';
 
+/**
+ * @typedef {import('./cc-input-number.js').CcInputNumber} CcInputNumber
+ */
+
 const baseItems = [
   { label: 'The Label' },
   { label: 'The Label', value: 100 },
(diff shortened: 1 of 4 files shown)


pdesoyres-cc (approved):
LGTM

Galimede (approved):
Nice work Hélène, LGTM!

src/components/cc-input-number/cc-input-number.stories.js
```diff
@@ -50,22 +54,43 @@ export const helpMessage = makeStory(conf, {
...
-  })),
+  items: [
+    {
+      ...baseItems[0],
+      required: true,
+    },
+  ],
+  onUpdateComplete: (component) => {
+    component.reportInlineValidity();
+  },
```
florian-sanders-cc, line 66: **nitpick:** you could type `component` here to avoid the `any` (same for all cases where `onUpdatedComplete` is used)


```suggestion
  /** @param {CcInputNumber} component */
  onUpdateComplete: (component) => {
    component.reportInlineValidity();
  },
```

src/translations/translations.en.js
```diff
@@ -656,7 +656,7 @@ export const translations = {
   //#endregion
   //#region cc-input-number
   'cc-input-number.decrease': `decrease`,
-  'cc-input-number.error.bad-type': `You must enter a number.`,
+  'cc-input-number.error.bad-type': `You must enter a number`,
```
florian-sanders-cc, line 660: **praise:** nice catch, I missed it when I was supposed to remove trailing dots everywhere :see_no_evil:

florian-sanders-cc (changes requested):
Nice work Hélène! I only have a nitpick and I think your commits should be `refactor` not `fix` because changes related to stories shouldn't be part of the changelog (they are not part of what we ship to users).

PR #1454, opened 2025-06-16
fix(cc-env-var-create): prefilled with a value the validationWithExistingNames story input

### LINKED ISSUE #1442: `cc-env-var-create` stories are broken
## Context

This story https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%9B%A0-environment-variables-cc-env-var-create--validation-with-existing-names is broken because the form controls are supposed to be prefilled with a value and they are not.

It's probably related to the fact that we've moved to ElementInternals instead of tracking values within the state but I haven't checked.

clara-layus-cc:
Fixes #1442 

# What does this PR do?

- Fix cc-env-var-create validationWithExistingNames story by prefilling with a value

# How to review?

- Check the commits,
- Check the cc-env-var-create validationWithExistingNames story, within the [preview](),


florian-sanders-cc (changes requested):
@clara-layus-cc I always forget to check the commit names: the commit shouldn't be a fix because stories are not part of what we ship to users, they are not part of the components so they should not be part of the changelog.

You can go for `refactor` instead of fix fox the commit scope :+1:

HeleneAmouzou (approved):
LGTM ! :clap: 
I was just wondering if the story `validationWithStrictMode` isn't broken as well ? Shouln't it be displaying something ? :thinking:

florian-sanders-cc (approved):
Apart from @HeleneAmouzou 's question (which can be dealt with a separate issue & PR if needed), all good to me!

Thanks @clara-layus-cc :sunglasses:

PR #1457, opened 2025-06-18
refactor(cc-input-date): improve format displayed

### LINKED ISSUE #1455: `cc-input-date`: range overflow / underflow example formats may be confusing
## Context

The `cc-input-date` component error messages contain examples when the given date is before the min date or after the max date.

These examples use the ISO format because the `cc-input-date` supports both the ISO and short format but it's confusing for users because the displayed value always uses the short format.

It would be better to format the min & max examples so that they match what's displayed in the field.

### LINKED ISSUE #1456: `cc-input-date`: always show the expected format
## Context

The `cc-input-date` currently behaves like all cc form controls: it accepts an optional help message as slotted content.

However, this component differs from others because it always expects a given format (ISO or short ISO).

The `cc-input-date` component should enforce the format help message so that it's always visible.

We discussed it with @HeleneAmouzou and we decided it was better to place the help message about the expected format within the render method, below the `help` slot.
This way the text is always visible, and people may slot content on top of the expected format if they want to.

Alternatively, we could have provided the expected format as a default slot value (present if nothing is slotted, but erased if anything is slotted) but:
- it would mean we allow people to remove the help message,
- it would force people to add the expected format within their slotted content if they want to add something more 🤔.

HeleneAmouzou:
Fixes #1455
Fixes #1456

## What does this PR do?
- Refactors the `cc-input-date` render to always display the help message,
- Removes the unnecessary `helpMessage` and `errorMessageWithHelpMessage` stories,
- Formats the underflow and overflow date ranges,
- Adds a `errorMessageInvalidRangeOverflowAndUnderflow`.

## How to review?
- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/refactor/cc-input-date-improve-format-displayed/index.html?path=/docs/%F0%9F%A7%AC-atoms-cc-input-date--docs) and compare with [prod](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-atoms-cc-input-date--default-story),
- Look for any visual regression on different screen sizes, especially on impacted components (`cc-logs-date-range-selector` and `cc-token-api-creation-form`),

### DIFF
--- a/src/components/cc-input-date/cc-input-date.stories.js
+++ b/src/components/cc-input-date/cc-input-date.stories.js
@@ -124,6 +124,29 @@ export const errorMessageInvalidFormat = makeStory(conf, {
   },
 });
 
+export const errorMessageInvalidRangeOverflowAndUnderflow = makeStory(conf, {
+  items: [
+    {
+      ...baseItems[0],
+      value: '2023-07-23T00:00:00.000Z',
+      min: '2023-07-23T00:00:00.000Z',
+      max: '2023-05-22T00:00:00.000Z',
+      label: 'Since',
+    },
+    {
+      ...baseItems[0],
+      value: '2023-05-22T00:00:00.000Z',
+      min: '2023-07-23T00:00:00.000Z',
+      max: '2023-09-23T00:00:00.000Z',
+      label: 'Until',
+    },
+  ],
+  /** @param {CcInputDate} component */
+  onUpdateComplete: (component) => {
+    component.reportInlineValidity();
+  },
+});
+
 export const inline = makeStory(conf, {
   items: baseItems.map((p) => ({
     ...p,
--- a/src/components/cc-input-date/cc-input-date.stories.js
+++ b/src/components/cc-input-date/cc-input-date.stories.js
@@ -89,11 +89,11 @@ export const required = makeStory(conf, {
   items: baseItems.map((p) => ({ ...p, required: true })),
 });
 
-export const helpMessage = makeStory(conf, {
+export const helpMessageSlotted = makeStory(conf, {
   items: baseItems.map((p) => ({
     ...p,
     required: true,
-    innerHTML: '<p slot="help">Format: YYYY-MM-DD HH:MM:SS</p>',
+    innerHTML: '<p slot="help">Specify the date using this field</p>',
   })),
 });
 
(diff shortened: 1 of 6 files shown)


src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -89,14 +89,6 @@ export const required = makeStory(conf, {
   items: baseItems.map((p) => ({ ...p, required: true })),
 });
 
-export const helpMessage = makeStory(conf, {
```
florian-sanders-cc, line 92: **suggestion:** we could keep this story to show that you can still use the help slot to provide additional info

florian-sanders-cc (changes requested):
Thanks for handling that @HeleneAmouzou !! 

I only have a suggestion about the current content of the PR, but 2 requests for changes overall:
- the commits should be fixes / or even feats because we're actually impacting the component this time, not just the stories. 
- this change requires other changes in components that currently rely on the `cc-input-date`. We need to check all components that use it (there aren't so many) and remove duplicate help messages (for instance in `cc-token-api-creation-form`) + check if it looks good (for instance in logs components, format was not specified, which was an accessibility issue that you fixed with this PR but it changes the design so we need to check if it works on desktop + mobile).
- all of this makes me wonder if this change could / should be considered breaking: if you used the component before and you used the help slot to provide the format, you shouldn't anymore, which seems breaking (a good change, but breaking nonetheless) :thinking: WDYT?

src/translations/translations.fr.js
```diff
@@ -664,6 +664,7 @@ export const translations = {
     `Saisissez une date inférieure à ${max}`,
   'cc-input-date.error.range-underflow': /** @param {{min: string}} _ */ ({ min }) =>
     `Saisissez une date supérieure à ${min}`,
+  'cc-input-date.help': `Format: AAAA-MM-JJ HH:MM:SS`,
```
Galimede: nitpick: there should be a space between `format` and  the colon `:`

florian-sanders-cc:
LGTM! I vote for a feat and no breaking change (although this is technically one, I don't think we have anyone out there using this component except us so should be fine)

src/components/cc-input-date/cc-input-date.stories.js
```diff
@@ -89,11 +89,11 @@ export const required = makeStory(conf, {
...
+export const helpMessageSloted = makeStory(conf, {
```
roberttran-cc: **🤓 nitpick:**

```suggestion
export const helpMessageSlotted = makeStory(conf, {
```

roberttran-cc (approved):
Nice PR!

pdesoyres-cc (approved):
Well done @HeleneAmouzou.

Note that the ISO date format is also accepted. it will be automatically reformatted using the formatter.

Could we take the opportunity to fix the english i18n label for:
* `You must enter a date lower that 2023-05-22 02:00:00.` to `You must enter a date lower than 2023-05-22 02:00:00.`
* `You must enter a date greater that 2023-05-22 02:00:00.` to `You must enter a date greater than 2023-05-22 02:00:00.`

HeleneAmouzou:
@pdesoyres-cc Do you think we should inform more explicitly the users that they can also enter a date on ISO format in the help message ?

PR #1458, opened 2025-06-19
feat(cc-clipboard): init component

Galimede:
## What does this PR do?

* This PR introduces a new `cc-clipboard` component,
* The old clipboard has been replaced with the component in `cc-input-text`,
* Please note, this PR doesn't replace the copy button present in other components with a cc-clipboard for now.

## How to review?

* Check the commits,
* Check the preview (cc-clipboard & cc-input-text).


HeleneAmouzou (approved):
Very nice !
I was just wondering if we should add a loading story as we mentioned that the component supports skeleton loading state ? Other LGTM ! :clap:

marion-izd:
Well done Mathieu for this component!  💪
I've nothing to say about what's been done. 

I have a few comments/questions to complete or take further:
- Even if the use cases will be rare, perhaps a Disabled and Error state should be included?
- For popover on the default Hover state (and potentially also on the error state during hover), the number of characters rule is : there are 20 characters in ‘copy to clipboard - ’ (including spaces), you should avoid going beyond a total of 35 characters. This leaves a maximum of 15 characters for the copied ‘content’.
- In storybook, is this the right place to potentially add a preview of the component's behaviour with the text? (particularly with long text). On Figma we had a debate between fixing it on the right or fixing it after the text, even if it's underneath.
- Having a view with the text would also make it possible to set the gap rule between the two elements: 8px

I've also had a look at the preview of the text input with the clipboard: I've got nothing to say, it's perfect. 👍
Don't hesitate to get back to me if you have any questions or need to synchronise my comments or integration with other components.
GG again !

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,118 @@
...
+    setTimeout(() => {
+      this._copied = false;
+    }, 2000);
+  }
+
+  render() {
+    return html`
+      <button
+        type="button"
+        class="copy-button"
```
roberttran-cc: **🤓 nitpick:** Ultra nitpick but you could remove this class name and target the element using `button` selector in the CSS.

roberttran-cc (approved):
Clean component. Me gusto.

src/translations/translations.fr.js
```diff
@@ -274,6 +274,11 @@ export const translations = {
   //#region cc-button
   'cc-button.cancel': `Cliquez pour annuler`,
   //#endregion
+  //#region cc-clipboard
+  'cc-clipboard.copied': `Le texte a été copié`,
+  'cc-clipboard.copy': /** @param {{text: string}} _ */ ({ text }) =>
```
hsablonniere, line 279: nitpick: use a "real" ellipsis `…` instead of `...`

src/translations/translations.en.js
```diff
@@ -263,6 +263,10 @@ export const translations = {
   //#region cc-button
   'cc-button.cancel': `Click to cancel`,
   //#endregion
+  //#region cc-clipboard
+  'cc-clipboard.copied': `The text has been copied`,
+  'cc-clipboard.copy': /** @param {{text: string}} _ */ ({ text }) => `Copy to clipboard - ${text.substring(0, 5)}...`,
```
hsablonniere: nitpick: use a "real" ellipsis `…` instead of `...`

pdesoyres-cc: nitpick: if text length is lower than 5 chars, you don't need to add the ellipsis

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
+import { css, html, LitElement } from 'lit';
+import { iconRemixFileCopyLine as iconCopy, iconRemixCheckLine as iconSuccess } from '../../assets/cc-remix.icons.js';
+import { copyToClipboard } from '../../lib/clipboard.js';
+import { isStringEmpty } from '../../lib/utils.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-icon/cc-icon.js';
+
+/**
+ * @typedef {import('../common.types.js').IconModel} IconModel
```
florian-sanders-cc: **nitpick:** not used

```suggestion
```

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
...
+ * @typedef {import('lit').PropertyValues<CcClipboard>} CcClipboardPropertyValues
+ */
+
+/**
+ * A component to copy text to clipboard with visual feedback.
+ *
+ * ## Details
+ *
+ * * Provides a button to copy text to the clipboard
+ * * Shows visual feedback when copy operation succeeds or fails
```
florian-sanders-cc: **fix:** since we're not going for an error mode, we should modify this desc

```suggestion
 * * Shows visual feedback when copy operation succeeds
```

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
...
+ *
```
florian-sanders-cc, line 20: **fix:** since we don't support skeleton / loading, this desc should be removed


```suggestion
```

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
...
+          padding: 0;
+        }
+
+        button:focus {
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: var(--cc-focus-outline-offset, 2px);
+        }
+
+        button:active,
+        button:hover {
```
florian-sanders-cc: **question:** why do we need that? :thinking:

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
...
+          box-shadow: none;
+          outline: 0;
+        }
+
+        button:hover {
+          background-color: var(--cc-color-bg-neutral-hovered);
+        }
+
+        .copy-icon {
+          --cc-icon-color: var(--cc-input-btn-icons-color, #595959);
```
florian-sanders-cc: **question:** why not a typical em value? :eyes:

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,121 @@
...
+
+        button:active,
+        button:hover {
+          box-shadow: none;
+          outline: 0;
+        }
+
+        button:hover {
+          background-color: var(--cc-color-bg-neutral-hovered);
+        }
```
florian-sanders-cc, line 100: **suggestion:** Since this component is not a form control, I'm not sure it should rely on a custom prop that's dedicated to form controls. As a user, I wouldn't expect `cc-clipboard` not in the context of an input to change color with this variable.

I suggest we create an intermediate CSS prop that should be used from the `cc-input-*` components:

```suggestion
          --cc-icon-color: var(--cc-clipboard-icon-color, #595959);
```

And from the `cc-input-*` component, you would do this:

```
// could be a parent selector, whatever
cc-clipboard {
    --cc-clipboard-icon-color: var(--cc-input-btn-icons-color, ...);
}
```

Or maybe we could even skip the intermediate CSS prop and directly use the `--cc-icon-color` (as long as it's scoped to the proper selector).

@roberttran-cc do you have opinions about that? :eyes:

src/components/cc-input-text/cc-input-text.js
```diff
@@ -722,14 +696,19 @@ export class CcInputText extends CcFormControlElement {
           padding: 0;
         }
 
+        .btn,
+        .btn-copy {
+          flex-shrink: 0;
+          margin: 0.2em 0.2em 0.2em 0;
+          z-index: 2;
+        }
```
florian-sanders-cc: **nitpick:** I feel like the alignment is slightly off (the clipboard button is slightly closer to the top than the bottom)

I tested quickly but the component is tricky so feel free to dismiss my suggestion if it's not a good idea:
we could remove the margins which are not very robust anyway and use `align-items: center` on `.meta-input` instead?
As for the spacing on the right and between buttons, we can still use margins, or padding, and gap (I think I would go a for a margin-right since it would deal with the trailing spacing for the last element + the spacing between the two buttons).

src/components/cc-clipboard/cc-clipboard.js
```diff
@@ -0,0 +1,110 @@
...
+ * @typedef {import('lit').PropertyValues<CcClipboard>} CcClipboardPropertyValues
+ */
+
+/**
+ * A component to copy text to clipboard with visual feedback.
+ *
+ * ## Details
+ *
+ * * Provides a button to copy text to the clipboard
+ * * Shows visual feedback when copy operation succeeds
```
florian-sanders-cc, line 19: **nitpick:** @Galimede we should document the new CSS prop here :innocent:

florian-sanders-cc (changes requested):
A last tiny nitpick about the doc and then we're good to go! Great job again @Galimede :raised_hands:

florian-sanders-cc (approved):
LGTM!

PR #1471, opened 2025-07-08
chore(eslint): use common esm configuration

Galimede:
## What does this PR do?

* This PR makes use of the new Clever Cloud ESLint ESM common configuration package.
* _Note: this PR will only be merged when the packages will be released on NPM._

## How to review?

* Run `npm install`
* Check the files
* Run `npm run lint`
* You can test the root package, and the ESLint specific package by:
 * installing the package with the command available within the `package.json` (either `npm run install-local:config` (for the ESLint specific one) or `npm run install-local:config-global` for the root package)
 * run `npm run lint` afterwards


florian-sanders-cc (approved):
Great job @Galimede this works great! :raised_hands:

src/stories/lib/make-story.js
```diff
@@ -1,5 +1,5 @@
 // @ts-ignore not worth helping TS understand this is valid since we don't need to type check this module
-// eslint-disable-next-line import/no-unresolved
+// eslint-disable-next-line
```
florian-sanders-cc: I think it's an unwanted replacement, @Galimede it would be better to go for


```suggestion
// eslint-disable-next-line import-x/no-unresolved
```

Or maybe even customize the rule in our local project config to add `custom-elements.json`? I had not thought about that but it would be simpler since it would cover all cases 🤔

florian-sanders-cc (changes requested):
Small suggestions but apart from that it's all good @Galimede 🙌

PR #1475, opened 2025-07-09
fix(cc-input-text): align back button correctly on multi

Galimede:
## What does this PR do?

* This PR reverts and set the initial margin for the buttons instead of doing `align-items: center` in the `cc-input-text`
* This caused the `cc-clipboard` to be aligned vertically in the center of the `cc-input-text` on mutli mode.
image

## How to review?

* Check the commit
* Check the preview
* Two reviewers should be enough


pdesoyres-cc (approved):
LGTM

PR #1482, opened 2025-07-17
Initialize visual testing

### LINKED ISSUE #1069: Web Test Runner: add visual regression tests
## Context

Modern Web has developed a [`test-runner-visual-regression` package](https://github.com/modernweb-dev/web/tree/master/packages/test-runner-visual-regression) that may be used with Web Test Runner to take screenshots of stuff and compare to the previous screenshot of the same stuff.

It is fairly agnostic (which is great) and can easily be used to compare stories before & after modifications.

I've played with it within the components repo and it worked + it's fairly easy to integrate.

The only drawback is that you have to commit the screenshots so it's able to compare them. 
I believe they are fairly low weight but I need to check.

florian-sanders-cc:
Fixes #1069 

## What does this PR do?

- Refactors a11y tests and its helpers to separate a11y tests from helpers used to run tests on stories,
- Adds support for long lists of keys and retrieving images to the `CellarClient`,
- Adds 4 new components used to generate visual tests reports,
- Makes `cc-expand` respect `prefers-reduced-motion`,
- Renames the `wds` directory to `web-test-runner` since we no longer use `wds` on its own and we only use `wds` plugins for Web Test Runner now,
- Adds visual tests tasks and GitHub workflow.

## How to review?

> [!NOTE]
> Considering the size of this PR and the fact it almost doesn't impact end-users, you may **focus mostly on the workflow file** and the **related tasks** to check there are **no security issues** and review the rest based on the time you have (if you don't have much time, you may easily skip the review of the components I think?).

- As a first step, it's probably a good idea to read these:
  - the [Setting up visual tests ADR](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tests/visual-changes/index.html?path=/docs/%F0%9F%93%8C-architecture-decision-records-adr-0030-setting-up-visual-tests--docs) (this has been AI generated from my frontend friday notes and then reworked a little bit),
  - the [Contributing - Test](https://clever-components-preview.cellar-c2.services.clever-cloud.com/tests/visual-changes/index.html?path=/docs/%F0%9F%91%8B-contributing-test--docs) docs,
  - the [workflow explanation on TLDRaw](https://www.tldraw.com/f/mpntH67u32GSmSSKSDCDC?d=v1520.-191.3545.1759.page) (namings have changed since then but the process remains the same).
- I've left a few comments below to point out specific / weird stuff,
- Check the commits, 
- Check the new components in storybook,
- Check a few examples of PRs:
  - [Example with no changes detected](https://github.com/florian-sanders-cc/clever-components--gh-actions-playground/pull/10#issuecomment-3073300331),
  - [Example with a lot of changes (`cc-button` primary color)](https://github.com/florian-sanders-cc/clever-components--gh-actions-playground/pull/9#issuecomment-3072446522).

## TODO Flo

- [ ] Change the CDN URL to fetch components from the prod CDN once components have been released (probably in a separate PR afterwards with no reviews necessary).


.github/workflows/manage-visual-tests.yml
```diff
@@ -0,0 +1,297 @@
...
+on:
+  pull_request:
+    types: [synchronize, opened, reopened, closed, labeled, unlabeled]
+
+env:
+  VISUAL_TESTS_REPORTS_DIR: 'visual-tests-reports'
+  VISUAL_TESTS_RAW_REPORT_NAME: 'visual-tests-results.json'
+  VISUAL_TESTS_FINAL_REPORT_NAME: 'visual-tests-report.json'
+  VISUAL_TESTS_REPORT_COMMENT_ID: 'visual-changes-report-comment-${{ github.event.pull_request.number }}'
+  VISUAL_TESTS_CELLAR_BUCKET_NAME: ${{ vars.VISUAL_TESTS_CELLAR_BUCKET_NAME }}
```
florian-sanders-cc, line 16: Using a repository variable here allowed me to run the action on my other repository, on another bucket from the ACME FOO org without having to modify the source code of the workflow :innocent:

.github/workflows/manage-visual-tests.yml
```diff
@@ -0,0 +1,297 @@
...
+  BRANCH_NAME: ${{ github.event.pull_request.head.ref }}
```
florian-sanders-cc, line 17: The `BRANCH_NAME` is used extensively in the workflow and its dependencies. Sometimes, we cannot rely on our local `getCurrentBranch()` helper because we're in detached head mode, for instance when we checkout the base commit.

.github/workflows/manage-visual-tests.yml
```diff
@@ -0,0 +1,297 @@
...
+      # Expected outputs:
+      # - should-update-expectation: boolean - true if expectation should be updated (base commit has changed or no previous report was found)
+      # - last-expectation-update-utc: string - date as string representing the last update of the expectation screenshots (UTC)
+      # - last-expectation-update-local: string - date as string representing the last update of the expectation screenshots (Local)
+      - name: '[Run] Check for last expectation update'
+        id: check-for-expectation-update
+        run: node ./tasks/visual-tests/manage-visual-tests-data.js check-for-expectation-update
+        env:
+          VISUAL_TESTS_CELLAR_KEY_ID: ${{ secrets.VISUAL_TESTS_CELLAR_KEY_ID }}
+          VISUAL_TESTS_CELLAR_SECRET_KEY: ${{ secrets.VISUAL_TESTS_CELLAR_SECRET_KEY }}
```
florian-sanders-cc, line 81: it's fairly easy to make a node script output stuff for other steps / jobs in the workflow but the drawback is that these outputs are undocumented.
The only way I found to mitigate that was to add comments :shrug:

.github/workflows/manage-visual-tests.yml
```diff
@@ -0,0 +1,297 @@
...
+          cache: 'npm'
+
+      - name: '[Prepare] Install dependencies'
+        run: npm ci
+
+      # Expected outputs
+      # - stories-groups=Array<`stories-group-{number}`>
+      - name: '[Run] Get stories groups'
+        id: get-stories-groups
+        run: node ./.github/scripts/get-stories-groups.js
```
florian-sanders-cc, line 49: it's fairly easy to make a node script output stuff for other steps / jobs in the workflow but the drawback is that these outputs are undocumented.
The only way I found to mitigate that was to add comments :shrug:

.github/workflows/manage-visual-tests.yml
```diff
@@ -0,0 +1,297 @@
...
+          VISUAL_TESTS_CELLAR_KEY_ID: ${{ secrets.VISUAL_TESTS_CELLAR_KEY_ID }}
+          VISUAL_TESTS_CELLAR_SECRET_KEY: ${{ secrets.VISUAL_TESTS_CELLAR_SECRET_KEY }}
+
+  update-expectation:
+    if: needs.check-for-expectation-update.outputs.should-update-expectation == 'true' && !contains(github.event.pull_request.labels.*.name, 'skip-visual-changes-tests')
+    name: Update expectation ${{ matrix.stories-group }}
+    needs: [split-into-groups, check-for-expectation-update]
+    runs-on: ubuntu-latest
+    container:
+      image: mcr.microsoft.com/playwright:v1.53.2-noble
```
florian-sanders-cc: Using the playwright docker image speeds up the install but it comes with a price: when we update playwright (`@web/test-runner-playwright`), we'll have to update this as well...
I'm not sure where to document this actually :shrug: 
We could "just" install playwright using `npx` but it can easily take 1min30s so this is why I went this way.

src/components/cc-icon/cc-icon.stories.js
```diff
@@ -133,6 +136,9 @@ export const cleverIcons = makeStory(conf, {
     accessibility: {
       enable: false,
     },
+    visual: {
+      enable: false,
```
florian-sanders-cc, line 140: we disable visual testing for Remix & Clever icons because it's irrelevant and very slow (sometimes it even timeouts) since there are a lot of icons.

src/lib/change-case.js
```diff
@@ -108,3 +108,39 @@ export function snakeCase(input) {
...
+ * camelCaseToSpacedCapitalized('string'); //=> 'String'
+ * camelCaseToSpacedCapitalized('dotCase'); //=> 'Dot Case'
+ * camelCaseToSpacedCapitalized('PascalCase'); //=> 'Pascal Case'
+ * camelCaseToSpacedCapitalized('snakeCase'); //=> 'Snake Case'
+ * camelCaseToSpacedCapitalized('version_1_2_10'); //=> 'Version_1_2_10'
+ * camelCaseToSpacedCapitalized('version1_2_10'); //=> 'Version1_2_10'
+ * @param {string} input
+ * @returns {string}
+ */
+export function camelCaseToSpacedCapitalized(input) {
```
florian-sanders-cc, line 141: Clearly AI generated :innocent:

src/stories/assets/cc-addon-backups-default-story-desktop-actual.png
florian-sanders-cc, line 1: ignore screenshot assets, it's only needed for stories

src/stories/fixtures/visual-tests-results.js
florian-sanders-cc, line 1: Can be safely ignored, this is just data seeds for stories :shrug:

test/helpers/global-mock-date.js
florian-sanders-cc, line 1: @pdesoyres-cc I couldn't find a good way to use your helper even though it's way better than what I did :see_no_evil: 

The issue is that I need to mock the Date before the stories are even imported and the stories imports are static, not dynamic.
We can discuss it in sync when you have time!

web-test-runner/get-story-files-groups.js
florian-sanders-cc, line 1: This is mostly generated by AI (especially the smart parts of the code to distribute stuff evenly :sweat_smile:)  and then reworked.

tasks/cellar-client.js
```diff
@@ -40,14 +40,21 @@ export class CellarClient {
...
-    const listObjectsResponse = await this._client.send(
-      new AWS.ListObjectsV2Command({
-        Bucket: this._bucket,
-        Prefix: prefix,
-      }),
-    );
-    const files = listObjectsResponse.Contents ?? [];
-    return files.map((file) => file.Key);
+    let keys = [];
+    let ContinuationToken = undefined;
```
pdesoyres-cc: nitpick: rename variable `continuationToken`

.github/scripts/delete-comment.js
pdesoyres-cc, line 1: question: Can we refactor to avoid code duplication with `create-or-edit-comment.js`?

.github/workflows/manage-visual-tests.yml
pdesoyres-cc: At least, you could add a comment here.

Maybe, at the beginning of the workflow, we can add a step that auto detects the installed version of playwright and check if it matches with the image version. If not, make the workflow fail with a proper message telling that we need to change the image version.

src/components/cc-img-comparator/cc-img-comparator.js
```diff
@@ -0,0 +1,176 @@
...
+    /** @type {string|null} */
+    this.baseImgText = null;
+
+    /** @type {string|null} */
+    this.comparisonImgSrc = null;
+
+    /** @type {string|null} */
+    this.comparisonImgText = null;
+
+    this._revealPercentage = '50';
```
pdesoyres-cc: nitpick: I would have make this property a `Number` and do the number casting inside `the _onSlide` method instead of the `render`

src/components/cc-img-comparator/cc-img-comparator.stories.js
pdesoyres-cc, line 1: fix: There is a problem with the height of the images:

Image

src/components/cc-visual-tests-report-entry/cc-visual-tests-report-entry.stories.js
pdesoyres-cc, line 1: There is a problem with images:

Image

Image

src/components/cc-visual-tests-report-menu/cc-visual-tests-menu.stories.js
pdesoyres-cc, line 1: fix: rename file to `cc-visual-tests-report-menu.stories.js`

src/components/cc-visual-tests-report-menu/cc-visual-tests-report-menu.js
pdesoyres-cc, line 1: remark: I find the direction of the arrows un-intuitive. I'm not sure we need to change the direction of the arrow when mouse hover.

pdesoyres-cc: There is a problem when selecting the third item:

https://github.com/user-attachments/assets/c6bbc9ba-2a33-40bc-a8d6-93a6cf24af00

src/components/cc-visual-tests-report-menu/cc-visual-tests-menu.stories.js
pdesoyres-cc, line 1: fix: nothing is displayed in the stories

src/components/cc-visual-tests-report/cc-visual-tests-report.js
```diff
@@ -0,0 +1,436 @@
...
+      return '';
+    }
+
+    const { repositoryOwner, repositoryName, prNumber, workflowId, branchName, expectationMetadata, actualMetadata } =
+      this.report;
+    const activeTestResult =
+      this._sortedTestResults.find((result) => result.id === this.activeTestResultId) ?? this._sortedTestResults[0];
+
+    return html`
+      <a class="skip-link" href="#main-content" @click="${this._skipToMain}">Skip to content</a>
```
pdesoyres-cc, line 121: question: what is that for?

pdesoyres-cc (changes requested):
what a great PR! Congrats @florian-sanders-cc.

I could not review all details, but I left few comments and suggestions.

I found several issues around the stories.

Also, for some commits, you can consider using the `test` type instead of `refactor`.

.github/workflows/manage-visual-tests.yml
florian-sanders-cc: I forgot to mention that when the playwright version doesn't match the container version, you already get an error telling you to bump the container version @pdesoyres-cc so I think what you are suggesting is already covered (but it was a good idea :+1:)

docs/adr/adr-0030-setting-up-visual-tests.md
```diff
@@ -0,0 +1,99 @@
...
+
+2.**Expectation/baseline strategy:** The baseline for comparison will be the **base commit of the PR**.
+    * **Advantages:** This gives developers control over their baseline (it updates logically on a rebase), avoids network flakiness from fetching a production baseline, and is much more lightweight.
+    * **Rejected Alternative (Production Baseline):** Using the production deployment as the baseline was rejected. It's slow, and the baseline can change unexpectedly if a new deployment occurs during a test run, leading to unstable tests.
+
+3.**Screenshot Storage:** Screenshots (both baseline and actual) will be stored on **Clever Cloud Cellar (Object Storage)**, not in the Git repository.
+    * **Reasoning:** Storing images in the repository is heavy, pollutes PRs with binary files, and creates complex commit/push workflows from the CI.
+
+4.**Tooling:** **Web Test Runner** with Playwright was chosen as the primary tool. This provides a good balance of performance, modern features, and control without the full complexity of a standalone Playwright setup.
```
roberttran-cc: **🔨 issue:** add a space before the text, so that Storybook renders the markdown correctly

src/components/cc-img-comparator/cc-img-comparator.stories.js
florian-sanders-cc: Was a issue with storybook, it worked fine locally but when storybook is built, the stories assets are supposed to be processed by vite.
Since I was relying on a helper to factorize the URL construction, vite was not picking up these assets and not updating their path accordingly (from `src/stories/assets/` to `dist/assets/` when Storybook is built).

Anyway, went back to inlined URLs and everything should work fine now :wink:

src/components/cc-visual-tests-report-entry/cc-visual-tests-report-entry.stories.js
florian-sanders-cc: Was a issue with storybook, it worked fine locally but when storybook is built, the stories assets are supposed to be processed by vite.
Since I was relying on a helper to factorize the URL construction, vite was not picking up these assets and not updating their path accordingly (from src/stories/assets/ to dist/assets/ when Storybook is built).

Anyway, went back to inlined URLs and everything should work fine now 😉

src/components/cc-visual-tests-report-menu/cc-visual-tests-report-menu.js
florian-sanders-cc: > remark: I find the direction of the arrows un-intuitive. I'm not sure we need to change the direction of the arrow when mouse hover.

This is the way it's done in all other components but I agree and I've expressed my opinion before the change was made. We can probably change this in all related components `cc-block` & `cc-block-details` at once in a separate PR after discussing it with Marion.

src/components/cc-visual-tests-report/cc-visual-tests-report.js
florian-sanders-cc: This is a skip link: https://webaim.org/techniques/skipnav/

TLDR; it's a mandatory link, supposed to be placed at the very beginning of a website / app to allow users to by-pass at least the menu. This way keyboard users may reach the main content of a page without having to tab through the whole menu. Note that this is not really aimed at blind users who already have ways to skip the menu thanks to screen reader shortcuts, this is mostly for users who only navigate with keyboard but no pointer device.

src/components/cc-visual-tests-report-menu/cc-visual-tests-report-menu.js
florian-sanders-cc: Create an issue for the arrow animation: https://github.com/CleverCloud/clever-components/issues/1546

PR #1449, opened 2025-06-10
Feat/cc block details command

clara-layus-cc:
# What does this PR do?

- Creates and integrates the cc-block-details-command component into other components.

# How to review?

- Check the commits,
- Check the default component story, where the cc-block-details-command component is integrated, within the [preview](),

## Component list

- [x] cc-domain-management
- [x] cc-env-var-form
- [x] cc-tcp-redirection-form
- [x] cc-addon-backups

PR #1483, opened 2025-07-17
fix: adjust typography for french translations

hsablonniere:
## What does this PR do?

In french translations:

* replace `...` with ellipsis character
* use NNBSP before `;`, `!` and `?`
* use NBSP before `:`
* use unicode escaped character instead of HTML entity
  * remove useless `sanitize` calls

## TODO

* [ ] remove temporary commit
* [ ] rebase on #1449


src/translations/translations.fr.js
```diff
@@ -251,8 +249,8 @@ export const translations = {
   //#region cc-ansi-palette
   'cc-ansi-palette.compliant': `Couleur qui respecte le RGAA`,
   'cc-ansi-palette.fg-bg': /** @param {{foreground: string, background: string}} _ */ ({ foreground, background }) =>
-    `Texte : ${foreground}, Fond: ${background}`,
-  'cc-ansi-palette.hover': /** @param {{color: string}} _ */ ({ color }) => `Survol : ${color}`,
+    `Texte\u00A0: ${foreground}, Fond: ${background}`,
```
florian-sanders-cc: **nitpick:**


```suggestion
    `Texte\u00A0: ${foreground}, Fond\u00A0: ${background}`,
```

florian-sanders-cc: **nitpick:** also missing one in `cc-ansi-palette.selected`

src/translations/translations.fr.js
```diff
@@ -322,7 +319,7 @@ export const translations = {
   'cc-domain-management.form.domain.error.empty': `Saisissez un nom de domaine`,
   'cc-domain-management.form.domain.error.format': `Saisissez un domaine valide, par exemple "example.com"`,
   'cc-domain-management.form.domain.error.wildcard': () =>
-    sanitize`Saisissez un domaine valide.<br>Les <span lang="en">wildcard</span> "*" ne peuvent être utilisés qu'en sous-domaine, par exemple&nbsp;: "*.example.com"`,
+    sanitize`Saisissez un domaine valide.<br>Les <span lang="en">wildcard</span> "*" ne peuvent être utilisés qu'en sous-domaine, par exemple\u00A0: "*.example.com"`,
   'cc-domain-management.form.domain.help': () =>
     sanitize`Par exemple: <code>example.com</code>, <code>*.example.com</code> ou <code>example.cleverapps.io</code>`,
```
florian-sanders-cc: **nitpick:**


```suggestion
'cc-domain-management.form.domain.help': () =>
sanitize`Par exemple\u00A0: <code>example.com</code>, <code>*.example.com</code> ou <code>example.cleverapps.io</code>`,
```

florian-sanders-cc: **nitpick:** also missing one in `cc-domain-management.form.path.help`.

src/translations/translations.fr.js
```diff
@@ -69,10 +69,10 @@ function getUnit(value) {
  * @return {string}
  */
 function formatFlavor(flavor) {
-  const cpu = `CPUs : ${flavor.cpus}`;
+  const cpu = `CPUs\u00A0: ${flavor.cpus}`;
   const shared = flavor.microservice ? ` (partagé)` : '';
   const gpu = flavor.gpus > 0 ? `GPUs : ${flavor.gpus}` : '';
```
florian-sanders-cc: **nitpick:** 

```suggestion
  const gpu = flavor.gpus > 0 ? `GPUs\u00A0: ${flavor.gpus}` : '';
```

florian-sanders-cc (approved):
I compared with the result I got using Zed and added a few missed occurrences.

There might be others because I did find a few in your changes that had not been detected in my branch but were properly handled in yours 😄 Anyway, I don't think it's that big of a deal 🤷 

I'm wondering if we should also use French quotation marks if we really want to stick with the French typo rules 🤔 (I'm not in favor or against that, just pointing it out for consistency).

As for using the unicode char instead of the HTML entities, I really like it 👍 Should we remove the `&nbsp;` from accepted entities in our sanitizer? It could be a breaking change for some of our usage in the console so I'm in favor of keeping it for the moment.

hsablonniere:
> I compared with the result I got using Zed and added a few missed occurrences.
Nice, thanks.

> There might be others because I did find a few in your changes that had not been detected in my branch but were properly handled in yours 😄 Anyway, I don't think it's that big of a deal 🤷
Hum, interesting.

> I'm wondering if we should also use French quotation marks if we really want to stick with the French typo rules 🤔 (I'm not in favor or against that, just pointing it out for consistency).
For a CSP dashboard interface, I would say no (at least for now).

> Should we remove the &nbsp; from accepted entities in our sanitizer? It could be a breaking change for some of our usage in the console so I'm in favor of keeping it for the moment.
Yeah, let's do that later.

PR #1487, opened 2025-07-22
fix(cc-code): fix component refresh

pdesoyres-cc:
## What this PR do ?

This PR fixes the refresh of the component when the slotted text changes.

The problem was that the slot change event is not fired when text nodes inside a slot changes.

This PR adds a `MutationObserver` and watch for changes inside the slot. Note that slots are "distribution points" (they don't actually contain the content), MutationObserver will miss mutation so we need to listen to assigned nodes individually.


florian-sanders-cc (approved):
Nice catch, I always forget about this limitation :/

I was a little bit wary about adding a mutation observer to every `cc-code` instance considering it's such a low level component that will be used several times inside many components.
After reading a little bit about the subject, it seems that as long as we don't mutate the DOM very often (which is clearly the case here), it's almost as lightweight as adding an event listener 🤷.

TLDR; LGTM!

PR #1490, opened 2025-07-24
feat(cc-picker): init component

### LINKED ISSUE #1489: cc-picker: init component
## Context

- factorize some common code between `cc-plan-picker` and `cc-zone-picker`
- externalize an UI part from one of our project
- add a sub component

## To be discussed

- the component name

### Attributes/properties
| Property     | Type      | Default value | Mandatory | Description                                                    |
|--------------|-----------|---------------|-----------|----------------------------------------------------------------|
| `disabled`   | `Boolean` | `false`       | `false`   | Whether the component should be disabled                       |
| `inline`     | `Boolean` | `false`       | `false`   | Sets the `<label>` on the left of the tiles                    |
| `label`      | `String`  | `''`          | `false`   | The label describing the picker, displayed on before the tiles |
| `mode`       | `String`  | `'none'`      | `false`   | How tiles selected state is rendered                           |
| `options`    | `Array`   | `null`        | `false`   | The list of options                                            |
| `readonly`   | `Boolean` | `false`       | `false`   | Whether the component should be readonly                       |
| `required`   | `Boolean` | `false`       | `false`   | Sets the "required" text inside the label                      |
| `resetValue` | `String`  | `null`        | `false`   | Sets the `value` to set when parent `<form>` element is reset  |
| `value`      | `String`  | `null`        | `false`   | Current selected value                                         |

### Slots

| Name   | Description                                                   |
|--------|---------------------------------------------------------------|
| `help` | The help message to be displayed right below the tiles        |

### CSS custom property

| Name                           | Type         | Default value | Description                                                       |
|--------------------------------|--------------|---------------|-------------------------------------------------------------------|
| `--cc-form-label-gap`          | `Size`       | `0.35em`      | The space between the label and the control                       |
| `--cc-form-label-gap-inline`   | `Size`       | `0.75em`      | The space between the label and the control when layout is inline |
| `--cc-input-label-color`       | `Color`      | `inherit`     | The color for the input's label                                   |
| `--cc-input-label-font-size`   | `FontSize`   | `inherit`     | The font-size for the input's label                               |
| `--cc-input-label-font-weight` | `FontWeight` | `normal`      | The font-weight for the input's label                             |
| `--cc-tile-content-width`      | `Width`      | `fit-content` | Sets the width of the form control content                        |

### CSS part

| Name    | Description                                                      |
|---------|------------------------------------------------------------------|
| `tiles` | Styles the tiles container, mainly to modify their layout        |

## Outputs

| Event                            | Data type | Description |
|----------------------------------|-----------|-------------|
| `cc-select` from `CcSelectEvent` | `string`  |             |

## Type Definitions

```typescript
export interface PickerOption {
  body: string | Node;
  disabled?: boolean;
  footer?: string | Node;
  value?: string;
}
```

## Data/APIs

n/a

## Stories & states

- default
- required
- with help message

roberttran-cc:
Fixes #1489 

# What does this PR do?

- add a `cc-picker` component
- add a `cc-tile` sub component
- add some design tokens
- fix some design token usage

# How to review?

- check the commits
- check the stories in the preview

# What's to discuss

Feel free to challenge/comment on the following points:
- the components name
- the components properties naming: thinking of `mode` especially
- where the components are located in the Storybook menu
- for `cc-picker`, the error state: should we work on a more elaborate UI in this case
  - there's only the error message for now
  - should we work on a more explicit design
- the `cc-tile` disabled state is based on weaker color and not on opacity
- the `mode="none"` looks not explicit enough


docs/getting-started/design-tokens.md
florian-sanders-cc, line 1: **praise**: thanks for maintaining this doc up to date (and creating it in the first place) :raised_hands:

src/styles/default-theme.css
```diff
@@ -245,7 +249,28 @@
...
+  /* Usage: for borders on hover
+  * For instance: input / select borders on hover */
+  --cc-color-border-primary-hovered: var(--color-blue-80);
+
+  /* Usage: for contrasted borders.
+  * For instance: click targets like input fields */
+  --cc-color-border-primary-strong: var(--color-blue-70);
+
+  /* Usage: for very subtle primary decorative borders
+  * For instance: row separators within a table. */
```
florian-sanders-cc, line 276: **question**: will we need to update the D & T console themes to add these? I feel like it's ok if we don't since it's only used in the tunnel + they are subtle colors but I'm raising the point just in case.

src/styles/default-theme.css
florian-sanders-cc, line 1: **praise**: I think I like the idea of using a lighter color instead of opacity so thanks for that! :raised_hands:

src/components/cc-tile/cc-tile.js
florian-sanders-cc, line 1: **suggestion**: We could name the component `cc-picker-option` to mimic a fairly common HTML naming (select option, dropdown option, menu option, etc.).
It would make it clear that this component is supposed to be used within a `cc-picker` (but maybe you don't agree that it's supposed to only be used in `cc-picker`? :thinking:)

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+    super();
+
+    /** @type {boolean} Whether the component should be disabled (default: 'false') */
+    this.disabled = false;
+
+    /** @type {boolean} Whether the component should be in error mode when not disabled (default: 'false') */
+    this.error = false;
+
+    /** @type {TileMode} How the component selected state is rendered (default: 'none') */
+    this.mode = 'none';
```
florian-sanders-cc: **suggestion**:  (not convinced but it could lead to better ideas)
`selectedDecoration` 
`selectionMarker`
`selectionIndicator`

:thinking:

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        .wrapper ::slotted([slot='footer']) {
+          flex: 0 0 auto;
+        }
+        /* endregion */
+
+        /* region body section */
+        ::slotted([slot='body']) {
+          font-size: 1em;
```
florian-sanders-cc: **question**: why do we need to force the font-size here?

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+          padding: 0.5em;
+        }
+
+        .icon-mode--radio {
+          --cc-icon-color: var(--cc-color-text-weak, #404040);
+
+          /* if text is multiline, allows the icon staying centered with the first line */
+          align-items: center;
+          display: flex;
+          height: 1lh;
```
florian-sanders-cc: **praise**: nice one! :raised_hands:

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        /* region global */
+        :host {
+          align-items: stretch;
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: inline-flex;
+          justify-content: stretch;
```
florian-sanders-cc: **question**: why do we need `align-items` & `justify-content` on the host since it only contains the `.wrapper`?

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        .wrapper {
+          border: 2px solid var(--cc-color-border-neutral, #bfbfbf);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: flex;
+          flex-direction: column;
+          line-height: 1.5;
+          overflow: hidden;
+          width: 100%;
```
florian-sanders-cc: **question**: I'm not sure you still need the `width: 100%` but it's very easy to think you don't need something until it breaks in CSS so I'll leave it up to you :sweat_smile:

src/components/cc-tile/cc-tile.js
```diff
@@ -0,0 +1,263 @@
...
+  constructor() {
+    super();
+
+    /** @type {boolean} Whether the component should be disabled (default: 'false') */
+    this.disabled = false;
+
+    /** @type {boolean} Whether the component should be in error mode when not disabled (default: 'false') */
+    this.error = false;
+
+    /** @type {TileMode} How the component selected state is rendered (default: 'none') */
```
florian-sanders-cc: **fix (but not sure)**: I'm not sure the `none` mode should be allowed at all because it creates a11y issues: we provide information through the use of color only.

Since we're only changing the border color, it's probably too subtle to be acceptable: to rely on contrast difference only, we need:
- a strong contrast diff (> 3 if I remember correctly but it might be more, I need to check),
- a good surface area (large text, large background area, etc.).

I think border color change doesn't meet these requirements and we should force either some background area or a shape (check vs radio) :thinking:

We can discuss it in sync, I'm mostly thinking and writing at the same time here :sweat_smile:

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * A generic picker component that allows users to select from a list of options displayed as tiles.
+ *
+ * Once a value is selected, it cannot be unselected without script.
+ *
+ * It extends `CcFormControlElement`, so it can be used in a form and its value can be submitted.
+ *
+ * @cssdisplay inline-block
+ *
+ * @csspart tiles - Styles the tiles container, mainly to modify their layout.
```
florian-sanders-cc, line 35: **praise:** I like the idea of exposing CSS parts, especially for form controls :raised_hands:

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+    this.inline = false;
+
+    /** @type {string} The label describing the picker, displayed on before the tiles (default: '') */
+    this.label = '';
+
+    /** @type {TileMode} How tiles selected state is rendered (default: 'none') */
+    this.mode = 'none';
+
+    /** @type {PickerOption[]} The list of options */
+    this.options = null;
```
florian-sanders-cc, line 83: **note**: looking back I think we could do better in the future design system for components that are "composable" like the `cc-select`, `cc-toggle`, and this one.

I think slots are a lot more convenient (for instance the `sl-select` component expects to be slotted with a series of `sl-option` components: https://shoelace.style/components/option).
This is a little more technically challenging but it feels better to compose the options directly in the template rather than in JS (very subjective though).

Anyway, not a point for this component as it is right now but more like "let's discuss this subject for the new design system" :eyes:

src/components/cc-picker/cc-picker.js
florian-sanders-cc, line 1: I'm not sure we should allow `disabled` and selected state (as shown in the `disabled (all options)` story): 
- disabled elements don't participate in form (their values shouldn't be part of the formData unless I'm wrong which happen fairly often),
- disabled elements are not part of the accessibility tree, meaning they don't even exist for screen readers. So it's weird that something which doesn't exist may still be "selected" :thinking: 
- visually, I wasn't sure what the blue border meant at first in the `disabled (all options)` (not a big deal though, I'm tired and I understood it like 1 min after).

florian-sanders-cc: When disabled, several of our components have a different background (well, mostly input components). I tried to do the same for the `cc-tile` and I think I like it but it's very subjective. The difference between disabled / not disabled didn't feel obvious enough for me with the border color only.
As it is now:
image
With a different bg color (probably contrast issues that we should tackle):
image

florian-sanders-cc: As for the `readonly` mode, there's is nothing that indicates that the option cannot be changed, people may be wondering if it's normal or not.
I wonder if we should `disable` all options except the selected one? But `disabling` options is not great for a11y so maybe we would need a soft disable (`aria-disabled="true"` on radio inputs + bg color or border to highlight this behavior)
As it is now:
image
With readonly bg variable:
image

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+          display: inline-block;
+        }
+        /* endregion */
+
+        /* region fieldset */
+        fieldset {
+          border: none;
+          display: inline-block;
+          margin: 0;
+          outline: none;
```
florian-sanders-cc: **nitpick:** I think we could go for something like 

```css
fieldset:focus-visible {
  // our typical outline variables
}
```

This way mouse users won't get the outline but keyboard users may still get it

florian-sanders-cc (changes requested):
Awesome PR @roberttran-cc, there are plenty of details that I really like, the stories are exhaustive, the different states (colors, etc.) of the component are well thought :raised_hands: 

I do have nitpicks (as usual :innocent:) and some fixes / things to discuss. 
I guess you can merge as it is if you need it and we can fix stuff when I come back from holidays if you want, I don't want to block you :wink:

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+    this.value = e.target.value;
+    this.dispatchEvent(new CcSelectEvent(this.value));
+  }
+
+  render() {
+    if (this.options?.length === 0) {
+      return '';
+    }
+
+    const hasErrorMessage = this.errorMessage != null && this.errorMessage !== '';
```
Galimede, line 173: nitpick: we could use our `isStringEmpty` helper

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+
+    return html`
+      <fieldset class="fieldset" @input=${this._onTileSelect} ${ref(this._pickerRef)} tabindex="-1">
+        <div class="fieldset-content">
+          <legend class="legend">
+            <span class="legend-text">${this.label}</span>
+            ${this.required ? html` <span class="required">${i18n('cc-select.required')}</span> ` : ''}
```
Galimede: question: shouldn't there be a dedicated `cc-picker.required` translation instead?

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+    /** @type {boolean} Sets the `<label>` on the left of the tiles.
+     * Only use this if your form contains 1 or 2 fields and your labels are short.
+     */
+    this.inline = false;
+
+    /** @type {string} The label describing the picker, displayed on before the tiles (default: '') */
+    this.label = '';
+
+    /** @type {TileMode} How tiles selected state is rendered (default: 'none') */
+    this.mode = 'none';
```
pdesoyres-cc: suggestion: maybe rename into `tileSelectionStyle`

src/components/cc-picker/cc-picker.js
pdesoyres-cc: I agree, but let's do that in our future design system.

src/components/cc-picker/cc-picker.js
pdesoyres-cc: Since `errorMessage` can also be a `Node`, no you cannot use the `isStringEmpty` helper.

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+    return html`
+      <fieldset class="fieldset" @input=${this._onTileSelect} ${ref(this._pickerRef)} tabindex="-1">
+        <div class="fieldset-content">
+          <legend class="legend">
+            <span class="legend-text">${this.label}</span>
+            ${this.required ? html` <span class="required">${i18n('cc-select.required')}</span> ` : ''}
+          </legend>
+
+          <div class="tiles" part="tiles">
+            ${this.options?.map((option) => this._renderOption(option, hasErrorMessage))}
```
pdesoyres-cc: nitpick: useless `?` after `this.options`. Check for options nullity has already been done before.

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,349 @@
...
+      </fieldset>
+    `;
+  }
+
+  /**
+   * @param {PickerOption} tileOption
+   * @param {boolean} isError
+   * @private
+   */
+  _renderOption(tileOption, isError) {
```
pdesoyres-cc: nitpick: rename `tileOption` to `pickerOption` or just `option`

src/components/cc-tile/cc-tile.js
pdesoyres-cc: maybe `selectionStyle` ?

src/components/cc-tile/cc-tile.js
pdesoyres-cc: +1 for `cc-picker-option`

HeleneAmouzou (approved):
GG @roberttran-cc ! :clap: 
I have catch up all the comments and have nothing further to add !

src/styles/default-theme.css
roberttran-cc: Good point! Noted and I'll take care of it after this PR being merged.

src/components/cc-tile/cc-tile.js
roberttran-cc: Hmm, this seems unnecessary, probably a temporary rule that I forgot to remove.

src/components/cc-tile/cc-tile.js
roberttran-cc: Hmm, this seems unnecessary too, probably older rules that I forgot to remove.

src/components/cc-tile/cc-tile.js
roberttran-cc: This one has a real purpose :sweat_smile: for when the component is used inside a grid: I'll leave a comment to explain.

src/components/cc-tile/cc-tile.js
roberttran-cc: I had doubt myself, I'll remove it.

src/components/cc-picker/cc-picker.js
roberttran-cc: Not sure what you mean: those styles are to reset the `fieldset` browser style. :thinking:

src/components/cc-tile/cc-tile.js
roberttran-cc: I went with `selectionStyle`.

For the record, Claude Code made two interesting suggestions: `variant` and `appearance`.

src/components/cc-picker/cc-picker.js
roberttran-cc: I went with `selectionStyle`, the same name as in the tile component.

src/components/cc-picker/cc-picker.js
florian-sanders-cc: Yes sorry I didn't provide any context :sweat_smile:

We can discuss it in sync but the idea is `outline` should never be set to `none` / `0` or anything that makes it invisible unless we have a good fallback for keyboard users.

When the component is focused (for instance if it's the first input with an error within a form), then the focus should be visible. See this story for a practical example: https://www.clever.cloud/developers/clever-components/?path=/story/%F0%9F%9B%A0-oauth-consumer-cc-oauth-consumer-form--error-with-empty-checkboxes

If you only want keyboard users to see this focus, then your best bet is to use a `fieldset:focus-visible {}` selector

src/components/cc-picker/cc-picker.js
roberttran-cc: Agree for the `readonly` & `disabled` styles and implemented.

@florian-sanders-cc About `selected + disabled`, I finally kept it and went for the same grey style as disabled (instead of the blue-ish one I showed you), as we always have a visual indicator of the `selected` state.

src/components/cc-picker/cc-picker.js
roberttran-cc: Understood & implemented!

src/components/cc-picker/cc-picker.js
```diff
@@ -0,0 +1,363 @@
...
+    return html`
+      <input
+        class="visually-hidden"
+        type="radio"
+        id="${id}"
+        name="${this.name}"
+        .value=${value}
+        .checked=${isChecked}
+        ?disabled=${isDisabled}
+        aria-disabled=${isReadonly}
```
florian-sanders-cc: **fix:** as discussed in sync, we'd better rely on the native `disabled` attribute to avoid having DOM elements in an impossible state (`disabled` & `selected`)
```suggestion
        ?disabled=${isDisabled || (isReadonly && !isChecked)}
```
Same for the `cc-zone-picker` component implementation :+1:

pdesoyres-cc (approved):
LGTM

PR #1495, opened 2025-07-30
feat(cc-env-var-form): add smart part to env-var context

clara-layus-cc:
# What does this PR do?

- Adds a smart part to the component `cc-env-var-form` to env-var context

# How to review?

- Check the commit,
- Check the demo-smart page on the [preview]()


src/components/cc-env-var-form/cc-env-var-form.smart-env-var.js
```diff
@@ -0,0 +1,108 @@
...
+ * @typedef {import('./cc-env-var-form.types.js').EnvVarFormState} EnvVarFormState
+ * @typedef {import('./cc-env-var-form.types.js').EnvVarFormStateLoaded} EnvVarFormStateLoaded
+ * @typedef {import('./cc-env-var-form.types.js').EnvVarFormStateSaving} EnvVarFormStateSaving
+ * @typedef {import('../common.types.js').EnvVar} EnvVar
+ * @typedef {import('../../lib/send-to-api.js').ApiConfig} ApiConfig
+ * @typedef {import('../../lib/smart/smart-component.types.js').OnContextUpdateArgs<CcEnvVarForm>} OnContextUpdateArgs
+ */
+
+defineSmartComponent({
+  selector: 'cc-env-var-form[context="env-var"]',
```
florian-sanders-cc: **suggestion:** I would have gone for `env-var-app` to make it very obvious in which cases the component is used (unless it's not just for apps?). I would apply the same logic to the smart component name / files


```suggestion
  selector: 'cc-env-var-form[context="env-var-app"]',
```

src/components/cc-env-var-form/cc-env-var-form.smart-env-var.md
```diff
@@ -0,0 +1,59 @@
...
+
+```js
+interface ApiConfig {
+  API_HOST: String,
+  API_OAUTH_TOKEN: String,
+  API_OAUTH_TOKEN_SECRET: String,
+  OAUTH_CONSUMER_KEY: String,
+  OAUTH_CONSUMER_SECRET: String,
+}
+```
```
HeleneAmouzou:
```suggestion
```ts
interface ApiConfig {
  API_HOST: string,
  API_OAUTH_TOKEN: string,
  API_OAUTH_TOKEN_SECRET: string,
  OAUTH_CONSUMER_KEY: string,
  OAUTH_CONSUMER_SECRET: string,
}
```
```

File: src/components/cc-env-var-form/cc-env-var-form.smart-env-var.md [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+
+## 🌐 API endpoints
+
+<!-- List API endpoints used by the component here with the details. -->
```
HeleneAmouzou:
```suggestion
```

src/components/cc-env-var-form/cc-env-var-form.js
florian-sanders-cc, line 1: @clara-layus-cc you also need to change the name of your smart component (and its doc) so that it matches the context value (for instance if you check the demo-smart now it doesn't work anymore)

I'm also realizing it's technically a "breaking change" because we re changing the value of a documented prop...
If you think it's too annoying, just remove the commit and stick with `env-var`, or modify file names and add the "!" to your commit to flag the breaking change, it's up to you! :+1:

PR #1498, opened 2025-07-31
Fix: pass ressourceId from smart to component

clara-layus-cc:
# What does this PR do?

- Change param name `resourceId` to `applicationId` in `cc-domain-management` and `cc-tcp-redirection-form`
- Pass the param `applicationId` or `addonId` from the smart to the component

# How to review?

- Check the commits,
- Check the demo-smart page on the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/fix/add-ressourceid-to-smart/index.html)

### DIFF
--- a/src/components/cc-env-var-form/cc-env-var-form.js
+++ b/src/components/cc-env-var-form/cc-env-var-form.js
@@ -65,7 +65,7 @@ export class CcEnvVarForm extends LitElement {
     this.addonName = '?';
 
     /** @type {string} Sets the resource id for documentation */
-    this.resourceId = '<APPLICATION_ID>';
+    this.resourceId = '<RESOURCE_ID>';
 
     /** @type {string} Defines application name used in some heading/description (depending on context). */
     this.appName = '?';


src/components/cc-tcp-redirection-form/cc-tcp-redirection-form.js
```diff
@@ -43,7 +43,7 @@ export class CcTcpRedirectionForm extends LitElement {
     super();
 
     /** @type {string} Sets the resource id for documentation */
```
Galimede: nitpick/suggestion: `resource` => `application`

florian-sanders-cc (changes requested):
Well done @clara-layus-cc!

Just like the `cc-tcp-redirection-form`, note that the `cc-domain-management` is a breaking change since you are renaming the prop so you should add the "!" right after the scope in your commit :+1:

> On another note, I just wonder if we should harmonize the appId/applicationId in cc-tcp-redirection-form and cc-domain-management as we have applicationId for the component and appId for the smart. Personally, I would put appId on the component and in the smart.

@galimede I agree with you but I wonder if it's worth it: we would have to make this change in every other component that has this kind of prop and it would be a breaking change. If we choose to go this way, we probably should update all the components in this release to group all the breaking changes together.

Let's see what Hélène thinks about that and if you and Clara think it's worth it?

@clara-layus-cc I only have a small nitpick apart from the breaking change thing: the `cc-env-var-form` may be used in differents contexts so I think the placeholder should be `<RESOURCE_ID>` instead of `<APPLICATION_ID>`. People will only see that placeholder when loading (very unlikely) so this is clearly a nitpick :innocent:

image

HeleneAmouzou (approved):
LGTM ! :clap: 

About the appId/applicationId harmonization, I'm ok with going with applicationId everywhere to match the new client, but if updates are needed, I wonder if it could not be done in another PR ?

HeleneAmouzou:
Sorry for the confusion :see_no_evil: 
Nothing to change indeed in the UI component, `applicationId` is right :+1: 
For the the smart, as `appId` / `addonId` matche the console router I think it is also right to keep it that way :+1: 
And as @Galimede mentionned, we can come back to it in the future to match the new client

florian-sanders-cc (approved):

PR #1502, opened 2025-08-01
`cc-addon-credentials-beta`: init with keycloak smart component

florian-sanders-cc:
## What does this PR do?

- Adds two new components:
  - `cc-addon-credentials-beta` which shows credentials of an add-on. It may contain tabs to show different kind of credentials when needed. Note that the current `cc-addon-credentials` will be deprecated once it's no longer used by any product dashboard.
  - `cc-addon-credentials-content` which is just a low level component to display the actual content of each tab.
  - When the component only has one tab, we don't show the toggle on the top right and we display the content directly.
- Adds a smart component for keycloak credentials.

## How to review?

- Check the commits,
- Play with the stories in [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-addon-credentials/init/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-credentials-beta--default-story),
- Review the [QA](https://www.notion.so/cc-addon-credentials-beta-246fe1a33d38806c9489cc30142acc9f),
- **Vote for the [low level component](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-addon-credentials/init/index.html?path=/story/%F0%9F%9B%A0-addon-cc-addon-credentials-content--default-story) name**: 
  - :tada: `cc-addon-credentials-content`,
  -  :+1: `cc-addon-credentials-tabpanel`, 
  - :heart: `cc-addon-credentials-panel`,
  - :rocket: `cc-addon-credentials-tab`.

### DIFF
--- a/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
+++ b/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
@@ -1,0 +1,161 @@
+import { LitElement, css, html } from 'lit';
+import { iconRemixInformationFill as iconInfo } from '../../assets/cc-remix.icons.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-addon-credentials-content/cc-addon-credentials-content.js';
+import '../cc-block-details/cc-block-details.js';
+import '../cc-block/cc-block.js';
+import '../cc-code/cc-code.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-link/cc-link.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-notice/cc-notice.js';
+import '../cc-toggle/cc-toggle.js';
+
+/**
+ * @typedef {import('./cc-addon-credentials-beta.types.js').AddonCredentialsBetaState} AddonCredentialsBetaState
+ * @typedef {import('./cc-addon-credentials-beta.types.js').TabName} TabName
+ * @typedef {import('../cc-toggle/cc-toggle.types.js').Choice} Choice
+ * @typedef {import('lit').PropertyValues<CcAddonCredentialsBeta>} CcAddonCredentialsBetaPropertyValues
+ */
+
+/**
+ * A component to display credentials for an add-on.
+ *
+ * ## Details
+ *
+ * This component displays various credentials for an add-on, such as username, password, direct access information, or other relevant connection details.
+ * It supports different views through tabs, allowing users to switch between various access methods or categories.
+ *
+ * The content displayed within each tab is managed by the `cc-addon-credentials-content` component. It also provides a link to the official documentation for more information.
+ *
+ * @cssdisplay block
+ */
+
+export class CcAddonCredentialsBeta extends LitElement {
+  static get properties() {
+    return {
+      docLink: { type: Object, attribute: 'doc-link' },
+      state: { type: Object },
+      _selectedTabName: { type: String, state: true },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {{ text: string; href: string; }} */
+    this.docLink = null;
+
+    /** @type {AddonCredentialsBetaState} */
+    this.state = { type: 'loading', tabs: { default: [] } };
+
+    /** @type {TabName} */
+    this._selectedTabName = null;
+  }
+
+  /**
+   * @param {TabName} toggleChoiceValue
+   */
+  _getToggleChoiceLabel(toggleChoiceValue) {
+    switch (toggleChoiceValue) {
+      case 'api':
+        return i18n('cc-addon-credentials-beta.choice.api');
+      case 'apm':
+        return i18n('cc-addon-credentials-beta.choice.apm');
+      case 'default':
+        return i18n('cc-addon-credentials-beta.choice.default');
+      case 'direct':
+        return i18n('cc-addon-credentials-beta.choice.direct');
+      case 'elastic':
+        return i18n('cc-addon-credentials-beta.choice.elastic');
+      case 'kibana':
+        return i18n('cc-addon-credentials-beta.choice.kibana');
+      default:
+        return '';
+    }
+  }
+
+  /**
+   *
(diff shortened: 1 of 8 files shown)


src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.types.d.ts
```diff
@@ -0,0 +1,76 @@
...
+    CC_JAVA_VERSION: string;
+    CC_PRE_RUN_HOOK: string;
+    CC_METRICS_PROMETHEUS_PORT: string;
+    CC_POST_BUILD_HOOK: string;
+    CC_KEYCLOAK_REALMS: string;
+    CC_KEYCLOAK_HOSTNAME: string;
+    CC_FS_BUCKET: string;
+    CC_KEYCLOAK_ADMIN_DEFAULT_PASSWORD: string;
+    CC_KEYCLOAK_VERSION: string;
+    CC_KEYCLOAK_DB_POOL_MAX_SIZE: string;
```
pdesoyres-cc, line 74: nitpick: I'm not sure we need to be so precise! A `Record<string, string>` may be enough.

src/translations/translations.fr.js
```diff
@@ -178,6 +178,44 @@ export const translations = {
...
+  'cc-addon-credentials-content.code.initial-password': `Mot de passe initial`,
+  'cc-addon-credentials-content.code.network-group': `Network Group`,
+  'cc-addon-credentials-content.code.password': `Mot de passe`,
+  'cc-addon-credentials-content.code.port': `Port`,
+  'cc-addon-credentials-content.code.tenant': `Tenant`,
+  'cc-addon-credentials-content.code.token': `Jeton`,
+  'cc-addon-credentials-content.code.uri': `URI`,
+  'cc-addon-credentials-content.code.user': `Utilisateur`,
+  'cc-addon-credentials-content.ng.disable': `Désactiver`,
+  'cc-addon-credentials-content.ng.enable': `Enable Network Groups`,
```
pdesoyres-cc: fix: translate to french

src/translations/translations.en.js
```diff
@@ -165,6 +165,44 @@ export const translations = {
...
+  'cc-addon-credentials-content.code.initial-password': `Initial password`,
+  'cc-addon-credentials-content.code.network-group': `Network group`,
+  'cc-addon-credentials-content.code.password': `Password`,
+  'cc-addon-credentials-content.code.port': `Port`,
+  'cc-addon-credentials-content.code.tenant': `Tenant`,
+  'cc-addon-credentials-content.code.token': `Token`,
+  'cc-addon-credentials-content.code.uri': `URI`,
+  'cc-addon-credentials-content.code.user': `User`,
+  'cc-addon-credentials-content.ng.disable': `Disable`,
+  'cc-addon-credentials-content.ng.enable': `Enable Network Groups`,
```
pdesoyres-cc: fix: use singular


```suggestion
  'cc-addon-credentials-content.ng.enable': `Enable Network Group`,
```

src/components/cc-addon-credentials-content/cc-addon-credentials-content.js
```diff
@@ -0,0 +1,267 @@
...
+          flex-wrap: wrap;
+          gap: 0.5em;
+        }
+
+        cc-input-text {
+          width: 100%;
+        }
+
+        .skeleton {
+          background-color: #bbb !important;
```
HeleneAmouzou: **question**: why do we need `!important` here ?

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
```diff
@@ -0,0 +1,162 @@
...
+import '../cc-code/cc-code.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-link/cc-link.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-notice/cc-notice.js';
+import '../cc-toggle/cc-toggle.js';
+
+/**
+ * @typedef {import('./cc-addon-credentials-beta.types.js').CcAddonCredentialsBetaState} CcAddonCredentialsBetaState
+ * @typedef {import('./cc-addon-credentials-beta.types.js').Tabs} Tabs
```
HeleneAmouzou: **nitpick:** unused import

florian-sanders-cc:
> Really nice work @florian-sanders-cc . I like the design which is not too generic, but enough to handle many use cases.
> 
> I'm not sure if you want to handle that, but the component header hasn't got the same height when there are tabs and when there are no tabs.
> 
>     * So if, for some reason, the state switches from one tab (not tab selector) to multiple tabs, there will be a kind of blinking effect.
> 
>     * Or if there are multiple components in the same page, the space between the top-left corner and the header won't be exactly the same.

@pdesoyres-cc yes this is definitely an issue, which we already have in prod with `cc-token-*` components that contain a `cc-link` or a `cc-button` on the right side of their header. 
This is more of a `cc-block` issue so we can deal with it later on (but soon) and we'll fix all occurrences at once :+1: 
I've created an issue because @Galimede had already raised this when reviewing the `cc-token-*` component and it was still in my TODO :see_no_evil:

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.types.d.ts
florian-sanders-cc: We don't need to be so specific but this type is mostly a copy paste from the API payload. It's convenient because I don't have to go and look back to my browser to check for the name of the env variables (the best would be for the API to expose a proper field for me to retrieve info without picking from environment variables but well :shrug:)

Since it doesn't hurt and it's not costly to maintain I'll keep it :+1:

src/components/cc-addon-credentials-content/cc-addon-credentials-content.js
florian-sanders-cc: I think we don't :see_no_evil: I probably copied & pasted this one to get the exact color, good catch!

PR #1503, opened 2025-08-06
fix sorted-list in smart-component.md

jbnoblot:
Hello, I fix in smart-component.md a Numbered List markdown

ps: all other pages in markdown is broken with "Importing a module script failed."


florian-sanders-cc (changes requested):
:wave: @jbnoblot nice catch and thanks for this PR! 

The content of the PR looks good to me but the commit message doesn't match our conventions.
We rely on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to generate (or not) changelog entries and determine the type of release the changes will lead to.

TLDR; if you can edit your commit message to something like this:
```
docs(getting-started/smart-component): fix numbered list syntax
```

It would be perfect! 
If you can't or don't want to, I can do it myself later 😉

PR #1505, opened 2025-08-06
chore: move documentation links to clever.cloud

davlgd:


florian-sanders-cc (approved):
I've reworked the commits just to split things up (not essential but I had time 😄), thanks your PR @davlgd, made me realize we still had some cases where we didn't rely on the helper 🙈 .

PR #1510, opened 2025-08-08
Components build: fix container queries minify issue

### LINKED ISSUE #1387: `babel-plugin-template-html-minifier`: project is unmaintained and incompatible with container queries
## Context

When building our components before releasing (`npm` & `CDN`), we rely `babel-plugin-template-html-minifier` to minify the `html` & `css` template strings content.

It seems that this babel plugin hasn't been updated in 5 years and does not support container queries.
There is an open issue about this subject but no answer since 2023 (https://github.com/cfware/babel-plugin-template-html-minifier/issues/62).

We may need to replace it with another solution.

florian-sanders-cc:
Fixes ##524
Fixes #1387

## What does this PR do?

The main goal of this PR is to allow us to use container queries.
The issue was that the minify process was not parsing `@container` rules properly because it relied on an outdated dependency.

- Bumps `clean-css` to latest (`5.3.3`),
  - this package is used to minify our `default-theme.css` file,
- Replaces `babel-plugin-template-html-minifier` (unmaintained and old) with `@lit-labs/rollup-plugin-minify-html-literals` (freshly added to lit-labs),
  - both these packages actually rely on the same packages for the heavy work so the options are roughly the same,
  - had to force `@lit-labs/rollup-plugin-minify-html-literals` to use `clean-css@5.3.3` so that container queries are properly parsed,
  - short term: we'll create a PR on lit labs to bump the dependency there,
  - mid term: we need to decide on whether we want to stop minifying or not and what we want to minify (CSS, HTML, JS?).

## How to review?

- Check the commits,
- Locally, run `npm i`, then `npm ls clean-css` => you should see `5.3.3` everywhere,
- If you want to check that it works (@HeleneAmouzou it would be nice of you to confirm this so that others can skip this step), you can:
  1. add this to the styles of a component (for instance `cc-badge`): 
  ```css
        @container (max-width: 34em) {
          .wrapper {
            grid-template-areas:
              'button'
              'content'
              'links';
            grid-template-columns: 1fr;
          }
        }
  ```
  2. run `npm run components:build`,
  3. get the `dist/cc-badge.js` content,
  4. copy the css part (at the end of the line usually) and unminify it => you should get rougly the same result as the source and the `@container` stuff should be the same.


roberttran-cc (approved):
👍 🙏 💪

PR #1511, opened 2025-08-08
fix(cc-block-details): improve responsive

### LINKED ISSUE #1450: `cc-block-details`: improve responsive
## Context

The `cc-block-details` component has been designed to always display:
- the command line button on the left, 
- the doc link on the right,
- the command line content below the two.

in desktop command line button on the left, doc link on the right, content of the command line spanning below the two elements

For accessibility, we had to make sure the command line button & the command line content remain right next to each other within the DOM. The doc link is the last element of the cc-block-details in the DOM.

We had to rely on `grid-template-areas` to move stuff around visually and match the expected display described above.

The issue is that the `grid-template-areas` is not flexible and when the component width is narrow, the display is not optimal:

in mobile, command line button on the left, doc link on the right, content spanning below the two but the doc link and command button span on several rows

## How to solve the issue?

With @clara-layus-cc, we tried every possible solution we could think of to make the doc link wrap below the command content when there's not enough space with no `ResizeController` and no `Container queries` but we failed to find a good solution. 
We got close, by using a grid with `grid-template-columns: repeat(auto-fit, minmax(400px, 1fr))` and specifying the command content `grid-column: 1 / -1` `grid-row: 2 / 3` but we couldn't find a solution to align the doc link on the right on desktop and on the left in mobile.

In the end, we had to go for a `@container (max-width...)` solution.

This solution is very clean & easy to maintain but we cannot use it right now because of #1387.
Once it's solved, we can merge this branch: https://github.com/CleverCloud/clever-components/tree/cc-block-details/fix-responsive

florian-sanders-cc:
Fixes #1450

## What does this PR do?

- Improves the responsive aspect of `cc-block-details` by using container queries.

## How to review?

- Check the commit,
- Check the preview vs prod.

## TODO

- [x] Merge #1510 before (or at least make sure both these PRs are part of the same release) to allow using container queries

### DIFF
--- a/src/components/cc-block-details/cc-block-details.js
+++ b/src/components/cc-block-details/cc-block-details.js
@@ -55,14 +55,26 @@ export class CcBlockDetails extends LitElement {
       // language=CSS
       css`
         :host {
+          container-type: inline-size;
           display: block;
         }
 
         .wrapper {
+          align-items: center;
           display: grid;
           grid-template-areas: 'button links' 'content content';
           grid-template-columns: 1fr auto;
           justify-items: start;
+        }
+
+        @container (max-width: 34em) {
+          .wrapper {
+            grid-template-areas:
+              'button'
+              'content'
+              'links';
+            grid-template-columns: 1fr;
+          }
         }
 
         .button {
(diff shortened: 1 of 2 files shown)


roberttran-cc (approved):

HeleneAmouzou (approved):
LGTM ! :clap:

PR #1516, opened 2025-08-25
fix(cc-notice): align icon to top when no heading is present

### LINKED ISSUE #828: cc-notice: icon not aligned in some case
When the message is displayed on multiple lines, the icon is vertically centered when there are no heading (instead of begin top aligned).

Screen Shot 2023-09-11 at 17 48 39

Screen Shot 2023-09-11 at 17 49 42

roberttran-cc:
Fixes #828

# What to review

Icon alignment when there are no heading.

# How to review

Compare the `With long text and no heading` story in both [production](https://www.clever-cloud.com/doc/clever-components/?path=/story/%F0%9F%A7%AC-atoms-cc-notice--with-long-text-and-no-heading) & preview Storybook .

# How many reviewers

One should be enough.

### DIFF
--- a/src/components/cc-notice/cc-notice.js
+++ b/src/components/cc-notice/cc-notice.js
@@ -135,7 +135,7 @@ export class CcNotice extends LitElement {
         }
 
         .wrapper {
-          align-items: center;
+          align-items: start;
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (approved):
Perfect

PR #1518, opened 2025-08-27
fix(cc-env-var-form): display restart button only if app already been deployed

### LINKED ISSUE #1513: cc-env-var-form: hides the restart application button if the application has never been started
# Context

When a user updates the environment variables of their application, the "restart application" button is always displayed. The problem is in the case where the application has never had a deployment, using the button returns an error message. 

The smart component should therefore be modified in order to check if an application deployment has already taken place; otherwise, the restart button should not be displayed.

clara-layus-cc:
Fixes #1513 

# What to review

- Changes the smart part to display the restart button only if the app already been deployed

# How to review

- Check the commit

### DIFF
--- a/src/components/cc-env-var-form/cc-env-var-form.smart-env-var-app.js
+++ b/src/components/cc-env-var-form/cc-env-var-form.smart-env-var-app.js
@@ -1,5 +1,6 @@
+//prettier-ignore
 // @ts-expect-error FIXME: remove when clever-client exports types
-import { getAllEnvVars, redeploy, updateAllEnvVars } from '@clevercloud/client/esm/api/v2/application.js';
+import { getAllDeployments,getAllEnvVars,redeploy,updateAllEnvVars } from '@clevercloud/client/esm/api/v2/application.js';
 // @ts-expect-error FIXME: remove when clever-client exports types
 import { toNameValueObject } from '@clevercloud/client/esm/utils/env-vars.js';
 import { notify, notifyError, notifySuccess } from '../../lib/notifications.js';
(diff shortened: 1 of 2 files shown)


src/components/cc-env-var-form/cc-env-var-form.smart-env-var-app.js
```diff
@@ -151,6 +156,13 @@ class Api {
     );
   }
 
+  async isAlreadyBeenDeployed() {
```
roberttran-cc: **🎯 suggestion:** non-blocking

```suggestion
  async hasAlreadyBeenDeployed() {
```

Or maybe simply:
```suggestion
  async hasDeployments() {
```

src/components/cc-env-var-form/cc-env-var-form.smart-env-var-app.js
```diff
@@ -78,7 +79,11 @@ defineSmartComponent({
             },
           );
           notifySuccess(i18n('cc-env-var-form.update.success'));
-          updateComponent('restartApp', true);
+
+          api.isAlreadyBeenDeployed().then((alreadyBeenDeployed) => {
+            alreadyBeenDeployed ? updateComponent('restartApp', true) : null;
```
roberttran-cc: **💭 thought:** non-blocking

I think ternary are more clear & suitable for assignment statement rather than function call (though it works perfectly). But that's some nitpick & personal preference!

```suggestion
            if (alreadyBeenDeployed) {
              updateComponent('restartApp', true);
            }
```

PR #1519, opened 2025-08-29
feat(cc-addon-info): init component with keycloak smart 

HeleneAmouzou:
## What does this PR do?

- Adds a new component:
  - `cc-addon-info` which display information of an add-on.
- Adds a smart component for keycloak info.

## How to review?

- Check the commits,
- Play with the stories in [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-addon-info/init/index.html?path=/docs/%F0%9F%9B%A0-addon-cc-addon-info--docs).
- QA is available [here](https://www.notion.so/cc-addon-info-261fe1a33d388061bb7deb9c298ac139)
- The component is available on [staging-six](https://console-staging-six.cleverapps.io/organisations/orga_3547a882-d464-4c34-8168-add4b3e0c135/addons/addon_b0156348-3825-4026-895a-bcaacafc4746).

## TODO

- [x] Add QA

### DIFF
--- a/src/components/cc-addon-info/cc-addon-info.js
+++ b/src/components/cc-addon-info/cc-addon-info.js
@@ -1,0 +1,645 @@
+import { css, html, LitElement } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { createRef, ref } from 'lit/directives/ref.js';
+import {
+  iconRemixCloseLine as iconClose,
+  iconRemixInformationFill as iconInfo,
+  iconRemixSettings_3Line as iconUpdate,
+} from '../../assets/cc-remix.icons.js';
+import { LostFocusController } from '../../controllers/lost-focus-controller.js';
+import { hasSlottedChildren } from '../../directives/has-slotted-children.js';
+import { getAssetUrl } from '../../lib/assets-url.js';
+import { fakeString } from '../../lib/fake-strings.js';
+import { formSubmit } from '../../lib/form/form-submit-directive.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { skeletonStyles } from '../../styles/skeleton.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-block/cc-block.js';
+import '../cc-link/cc-link.js';
+import '../cc-notice/cc-notice.js';
+import '../cc-select/cc-select.js';
+import { CcAddonVersionChangeEvent } from './cc-addon-info.events.js';
+
+/** @type {Record<FormattedFeature['code'], () => string>} */
+const FEATURES_I18N = {
+  'connection-limit': () => i18n('cc-addon-info.feature.connection-limit'),
+  cpu: () => i18n('cc-addon-info.feature.cpu'),
+  databases: () => i18n('cc-addon-info.feature.databases'),
+  dedicated: () => i18n('cc-addon-info.feature.dedicated'),
+  'disk-size': () => i18n('cc-addon-info.feature.disk-size'),
+  gpu: () => i18n('cc-addon-info.feature.gpu'),
+  'has-logs': () => i18n('cc-addon-info.feature.has-logs'),
+  'has-metrics': () => i18n('cc-addon-info.feature.has-metrics'),
+  'is-migratable': () => i18n('cc-addon-info.feature.is-migratable'),
+  'max-db-size': () => i18n('cc-addon-info.feature.max-db-size'),
+  memory: () => i18n('cc-addon-info.feature.memory'),
+  version: () => i18n('cc-addon-info.feature.version'),
+  'encryption-at-rest': () => i18n('cc-addon-info.feature.encryption-at-rest'),
+  users: () => i18n('cc-addon-info.feature.users'),
+  'data-exploration': () => i18n('cc-addon-info.feature.data-exploration'),
+  'db-analysis': () => i18n('cc-addon-info.feature.db-analysis'),
+};
+
+const GRAFANA_LOGO_URL = getAssetUrl('/logos/grafana.svg');
+
+/**
+ * @typedef {import('./cc-addon-info.types.js').AddonInfoState} AddonInfoState
+ * @typedef {import('./cc-addon-info.types.js').AddonVersionStateUpdateAvailable} AddonVersionStateUpdateAvailable
+ * @typedef {import('./cc-addon-info.types.js').AddonVersionStateRequestingUpdate} AddonVersionStateRequestingUpdate
+ * @typedef {import('./cc-addon-info.types.js').LinkedService} LinkedService
+ * @typedef {import('../cc-select/cc-select.types.js').Option} Option
+ * @typedef {import('../common.types.js').FormattedFeature} FormattedFeature
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLDialogElement>} HTMLDialogElementRef
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLElement>} HTMLElementRef
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLFormElement>} HTMLFormElementRef
+ * @typedef {import('lit').PropertyValues<CcAddonInfo>} CcAddonInfoPropertyValues
+ */
+
+/**
+ * A component to display detailed info about an add-on.
+ *
+ * @cssdisplay block
+ *
+ * @slot billing - A customised text regarding the billing.
+ * @slot linked-services - A customised text regarding the linked services.
+ */
+export class CcAddonInfo extends LitElement {
+  static get properties() {
+    return {
+      docLink: { type: Object, attribute: 'doc-link' },
+      state: { type: Object },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {{ text: string; href: string; }} */
+    this.docLink = null;
+
+    /** @type {AddonInfoState} Sets the state of the component*/
+    this.state = { type: 'loading', creationDate: '2025-08-04 15:03:02' };
+
+    /** @type {HTMLDialogElementRef} */
+    this._versionDialogRef = createRef();
+
+    /** @type {HTMLElementRef} */
+    this._versionTextRef = createRef();
+
+    /** @type {HTMLFormElementRef} */
+    this._versionFormRef = createRef();
+
+    new LostFocusController(this, 'dialog', () => {
+      this._versionTextRef.value?.focus();
+    });
+  }
+
+  /**
+   * Returns the formatted value corresponding to a feature
+   *
+   * @param {FormattedFeature} feature - the feature to get the formatted value from
+   * @return {string|Node|void} the formatted value for the given feature or the feature value itself if it does not require any formatting
+   */
+  _getFeatureValue(feature) {
+    if (feature == null) {
+      return '';
+    }
+
+    switch (feature.type) {
+      case 'boolean':
+        return i18n('cc-addon-info.type.boolean', { boolean: feature.value === 'true' });
+      case 'boolean-shared':
+        return i18n('cc-addon-info.type.boolean-shared', { shared: feature.value === 'shared' });
+      case 'bytes':
+        return feature.code === 'memory' && feature.value === '0'
+          ? i18n('cc-addon-info.type.boolean-shared', { shared: true })
+          : i18n('cc-addon-info.type.bytes', { bytes: Number(feature.value) });
+      case 'number':
+        return feature.code === 'cpu' && feature.value === '0'
+          ? i18n('cc-addon-info.type.boolean-shared', { shared: true })
+          : i18n('cc-addon-info.type.number', { number: Number(feature.value) });
+      case 'number-cpu-runtime':
+        return i18n('cc-addon-info.type.number-cpu-runtime', {
+          /**
+           * Narrowing the type would make the code less readable for no gain, improving the type to separate
+           * `number-cpu-runtime` from other types makes the code a lot more complex for almost no type safety gain
+           */
+          // @ts-ignore
+          cpu: feature.value.cpu,
+          // @ts-ignore
+          shared: feature.value.shared,
+        });
+      case 'string':
+        return feature.value.toString();
+    }
+  }
+
+  /**
+   * Returns the complete linked service name with the 'app' or 'addon' mention
+   *
+   * @param {LinkedService} service
+   * @param {boolean} skeleton
+   * @return {string|Node}
+   * @private
+   */
+  _getServiceType(service, skeleton) {
+    if (skeleton) {
+      return fakeString(15);
+    }
+
+    switch (service.type) {
+      case 'app':
+        return i18n('cc-addon-info.service.name.app', { name: service.name });
+      case 'addon':
+        return i18n('cc-addon-info.service.name.addon', { name: service.name });
+      default:
+        return fakeString(15);
+    }
+  }
+
+  /** @param {{ version: string }} formData */
+  _onUpdateVersionRequested({ version }) {
+    this.dispatchEvent(new CcAddonVersionChangeEvent(version));
+  }
+
+  _onVersionDialogOpen() {
+    this._versionDialogRef.value.showModal();
+  }
+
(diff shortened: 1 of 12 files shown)


src/components/cc-addon-info/cc-addon-info.types.d.ts
```diff
@@ -0,0 +1,162 @@
...
+  ];
+  features: [
+    {
+      name: string;
+      enabled: boolean;
+    },
+  ];
+}
+
+export interface ServiceConfig {
```
pdesoyres-cc: fix: Unused interface

src/components/cc-addon-info/cc-addon-info.types.d.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export interface CcAddonInfoStateLoading extends BaseProperties {
+  type: 'loading';
+}
+
+export interface CcAddonInfoStateError {
+  type: 'error';
+}
+
+export interface BaseProperties {
```
pdesoyres-cc: fix: this exported type name seems a bit too generic. Can you rename it in something more specific like `CcAddonInfoBaseProperties`

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+          // @ts-ignore
+          shared: feature.value.shared,
+        });
+      case 'string':
+        return feature.value.toString();
+    }
+  }
+
+  /**
+   * Returns the complete the linked service name with the 'app' or 'addon' mention
```
pdesoyres-cc: nitpick: typo in this sentence

src/components/cc-addon-info/cc-addon-info.smart-keycloak.js
```diff
@@ -0,0 +1,349 @@
...
+      .then(({ operatorVersionInfo, addonInfo, grafanaAppLink, operator }) => {
+        const javaAppId = operator.resources.entrypoint;
+        logsUrl = context.logsUrlPattern.replace(':id', javaAppId);
+
+        updateComponent('state', {
+          type: 'loaded',
+          version: formatVersionState(operatorVersionInfo),
+          creationDate: addonInfo.creationDate,
+          openGrafanaLink: grafanaAppLink,
+          openScalabilityLink: scalabilityUrlPattern.replace(':id', operator.resources.entrypoint),
```
pdesoyres-cc: nitpick: use `javaAppId` var instead of `operator.resources.entrypoint`

src/components/cc-addon-info/cc-addon-info.smart-keycloak.js
```diff
@@ -0,0 +1,349 @@
...
+          linkedServices: [
+            {
+              type: 'app',
+              name: 'Java',
+              logoUrl: getAssetUrl('/logos/java-jar.svg'),
+              link: appOverviewUrlPattern.replace(':id', operator.resources.entrypoint),
```
pdesoyres-cc: nitpick: use `javaAppId` var instead of `operator.resources.entrypoint`

src/components/cc-addon-info/cc-addon-info.smart-keycloak.js
```diff
@@ -0,0 +1,349 @@
...
+      api
+        .updateOperatorVersion(targetVersion)
+        .then(() => {
+          notifySuccess(
+            i18n('cc-addon-info.version.update.success.content', { logsUrl }),
+            i18n('cc-addon-info.version.update.success.heading', { version: targetVersion }),
+          );
+          // Once update has been requested, we need to fetch up to date version info to refresh the UI
+          // The API is optimistic, when a version update is requested, it becomes the add-on current version even if the deployment is still running
+          // We could update the version number ourselves without fetching again but we need to know if new updates are available or not (users may update to a version that is not the latest)
```
pdesoyres-cc: suggestion: you say that we cannot know if the user upgraded to the last version or not. But I think we can :

- `api.getOperatorVersionInfo()` returns an `OperatorVersionInfo`
- This object contains the list of `available` versions but also the `latest` version
- When you `formatVersionState`, you could also indicate which one of the available version is the latest
- This way, you can know if the user updated to the latest version or not
- Then you can get rid of this additional API call.

Note that, we could ensure that this latest version is the preselected version in the upgrade dialog so that the "upgrade to latest" is the favourite action. I can see that it's already the case, because it seems that the available versions returned by the API are in the right order, where the last one is the latest version)

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+```ts
+interface ApiConfig {
+  API_HOST: string;
+  API_OAUTH_TOKEN: string;
+  API_OAUTH_TOKEN_SECRET: string;
+  OAUTH_CONSUMER_KEY: string;
+  OAUTH_CONSUMER_SECRET: string;
+}
+
+interface GrafanaLink {
```
pdesoyres-cc, line 37: doc: can you add explanation about what are `base` and `console` properties (their purpose, their expected format)

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+    "apiConfig": {
+      "API_HOST": "",
+      "API_OAUTH_TOKEN": "",
+      "API_OAUTH_TOKEN_SECRET": "",
+      "OAUTH_CONSUMER_KEY": "",
+      "OAUTH_CONSUMER_SECRET": ""
+    },
+    "ownerId": "",
+    "addonId": "",
+    "appOverviewUrlPattern": "",
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+    "addonDashboardUrlPattern": "",
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+      "API_OAUTH_TOKEN": "",
+      "API_OAUTH_TOKEN_SECRET": "",
+      "OAUTH_CONSUMER_KEY": "",
+      "OAUTH_CONSUMER_SECRET": ""
+    },
+    "ownerId": "",
+    "addonId": "",
+    "appOverviewUrlPattern": "",
+    "addonDashboardUrlPattern": "", 
+    "scalabilityUrlPattern": "",
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+    "grafanaLink": {
+      "base": "",
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+      "console": ""
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/components/cc-addon-info/cc-addon-info.smart-keycloak.md
```diff
@@ -0,0 +1,77 @@
...
+    },
+    "logsUrlPattern": ""
```
pdesoyres-cc: doc: can you set an example instead of a blank string ?

src/translations/translations.fr.js
```diff
@@ -275,6 +275,68 @@ export const translations = {
   'cc-addon-header.state-msg.deployment-is-deploying': `L'add-on est en cours de déploiement…`,
   'cc-addon-header.state-msg.unknown-state': `État inconnu, essayez de redémarrer l'add-on ou de contacter notre support si vous avez des questions.`,
   //#endregion
+  //#region cc-addon-info
+  'cc-addon-info.billing.heading': `Facturation`,
+  'cc-addon-info.creation-date.heading': `Date de création`,
+  'cc-addon-info.creation-date.human-friendly-date': /** @param {{ date: string | number }} _ */ ({ date }) =>
+    formatDatetime(date),
+  'cc-addon-info.documentation.text': `Voir la  documentation`,
+  'cc-addon-info.error': `Une erreur est survenue lors du chargement des informations de l'add-on.`,
```
pdesoyres-cc: fix: align the message with other errors

```suggestion
  'cc-addon-info.error': `Une erreur est survenue pendant le chargement des informations de l'add-on.`,
```

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+        }
+
+        @supports (backdrop-filter: blur(5px)) {
+          ::backdrop {
+            backdrop-filter: blur(5px);
+            background: rgb(30 30 30 / 35%);
+          }
+        }
+
+        .dialog-close {
```
pdesoyres-cc, line 559: suggestion: maybe add `cursor: pointer` ?

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+    this._versionDialogRef = createRef();
+
+    /** @type {HTMLElementRef} */
+    this._versionTextRef = createRef();
+
+    /** @type {HTMLFormElementRef} */
+    this._versionFormRef = createRef();
+
+    new LostFocusController(this, 'dialog', () => {
+      this._versionTextRef.value?.focus();
```
pdesoyres-cc, line 94: question: the focus is set but it is not visible (not focus ring). Is it ok?

pdesoyres-cc (changes requested):
Huge and impressive work!
And, you also made it ready for a lot of addons, I can't wait to have all this in the Console.

 :tada Well done :tada

src/lib/product.js
```diff
@@ -176,7 +176,7 @@ export function formatProductConsumptionIntervals(priceSystem, serviceName) {
  * @param {Array<FormattedFeature['code']>} [selectedFeatures] - Array of selected feature codes.
  * @returns {Array<FormattedFeature>} Formatted addon features.
  */
-function formatAddonFeatures(providerFeatures, selectedFeatures) {
+export function formatAddonFeatures(providerFeatures, selectedFeatures) {
```
roberttran-cc: **❔ question:** there's no external use of this function so I'm not sure if this `export` is relevant.

src/components/cc-addon-info/cc-addon-info.stories.js
```diff
@@ -0,0 +1,543 @@
...
+        ],
+        creationDate: '2023-01-15T10:30:00Z',
+        openGrafanaLink: 'https://grafana.example.com',
+        openScalabilityLink: 'https://scalability.example.com',
+      },
+      innerHTML: `
+        <p slot="billing"><strong>This add-on is free but its dependencies, mentioned above, are billed based on their consumptions,</strong> just like other applications and add-ons.</p>`,
+    },
+  ],
+  simulations: [storyWait(2000)],
```
roberttran-cc: **💣 chore:** Not needed anymore I guess (if yes, should remove the import as well).


```suggestion
```

src/translations/translations.fr.js
```diff
@@ -282,6 +282,68 @@ export const translations = {
...
+  'cc-addon-info.feature.connection-limit': `Limite de connexions`,
+  'cc-addon-info.feature.cpu': `vCPUs`,
+  'cc-addon-info.feature.data-exploration': `Exploration des données`,
+  'cc-addon-info.feature.databases': `Bases de données`,
+  'cc-addon-info.feature.db-analysis': `Analyse DB`,
```
roberttran-cc: **🎯 suggestion:** To align with `cc-addon-info.feature.max-db-size`.


```suggestion
  'cc-addon-info.feature.db-analysis': `Analyse BDD`,
```

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+
+        p,
+        dl,
+        dt,
+        dd {
+          margin: 0;
+        }
+
+        .section {
+          align-items: center;
```
roberttran-cc: **🎯 suggestion:** I'd go with `baseline` but that's more a personal preference.


```suggestion
          align-items: baseline;
```

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+        .linked-service__li cc-img {
+          border-radius: 0.19em;
+          height: 1.5em;
+          width: 1.5em;
+        }
+
+        .features__content {
+          display: flex;
+          flex-wrap: wrap;
+          gap: 1.5em;
```
roberttran-cc: **🎯 suggestion:** Consider a lower value for gap between lines when features list is wrapping.


```suggestion
          gap: 0.75em 1.5em;
```

src/components/cc-addon-info/cc-addon-info.js
```diff
@@ -0,0 +1,625 @@
...
+        .section:not(:first-child) {
+          border-top: solid 1px var(--cc-color-border-neutral-weak);
+        }
+
+        .section.section--billing:not([billing-is-slotted]) {
+          display: none;
+        }
+
+        .heading {
+          flex: 0 1 21em;
```
roberttran-cc: **🎯 suggestion:** When the width is 750px, the heading and the content have more or less the same width, which I find imbalance. Maybe consider:


```suggestion
          flex: 0 1 auto;
          max-width: 21em;
          width: 33%;
```

But I may have missed some use cases.

src/components/cc-addon-info/cc-addon-info.js
florian-sanders-cc: good point, I always forget that it's not the default for button :sweat_smile:

src/components/cc-addon-info/cc-addon-info.js
florian-sanders-cc: It's okayish. It's not mandatory to make the focus visible in such cases because it's the default browser behavior & it's not a result of a tab navigation.
We don't want the focus to be visible for mouse users but only for keyboard users, which used to be a tricky subject and the compromise was usually to just not set a focus ring in such cases.

But now that `:focus-visible` is widely available, we have no reason not to show the focus, so great point :+1:

src/lib/product.js
florian-sanders-cc: Ah yes, it's because we only have keycloak right now and we don't display features for it, but we need it for other add-ons (elastic, etc.). 

I'll remove it, the PR for elastic will include this change instead, it makes more sense :+1:

src/components/cc-addon-info/cc-addon-info.js
florian-sanders-cc: You're absolutely right, it was also part of Marion's feedback. We didn't pay attention to cases with multiple lines :see_no_evil: 

I went for `flex-start` since it `baseline` was a bit off for some cases:

with baseline:
image

with flex-start:
image

src/components/cc-addon-info/cc-addon-info.js
florian-sanders-cc: Works like a charm, thanks a lot!
We may need to do the same in other components to make all of them shift roughly at the same point :thinking:

florian-sanders-cc: actually it doesn't work when the viewport is reduced to a certain point:
image

Since we're setting the width to `33%`, at some point, it ends up being too small to show the whole heading on one line.

I have some ideas to solve this but I need to test

florian-sanders-cc: Got it with that:

```css
        .heading {
          flex: 1 1 33%;
          font-weight: bold;
          max-width: 21em;
        }
```

thanks again for this idea :muscle:

PR #1524, opened 2025-09-03
feat(cc-link): add `download` attribute to handle download links

### LINKED ISSUE #1461: cc-link: add a download icon possibility
for some cases where the link is a download link we could replace the external link with a download icon

HeleneAmouzou:
## What does this PR do?

- Fixes #1461
- adds a `download` attribute on the `cc-link` to handle download links cases

## How to review?

- check the commit,
- check the stories (especially the new `Download link` story)  in the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-link/handle-download/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-link--download-link),

### DIFF
--- a/src/components/cc-link/cc-link.js
+++ b/src/components/cc-link/cc-link.js
@@ -1,7 +1,7 @@
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { iconRemixExternalLinkLine as externalLinkIcon } from '../../assets/cc-remix.icons.js';
+import { iconRemixExternalLinkLine as externalLinkIcon, iconRemixDownloadLine } from '../../assets/cc-remix.icons.js';
 import { isStringEmpty } from '../../lib/utils.js';
 import { skeletonStyles } from '../../styles/skeleton.js';
 import { i18n } from '../../translations/translation.js';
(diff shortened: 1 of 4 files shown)


src/components/cc-link/cc-link.js
```diff
@@ -53,6 +54,9 @@ export class CcLink extends LitElement {
     /** @type {boolean} Disables the external link icon. */
     this.disableExternalLinkIcon = false;
 
+    /** @type {string|null} If set, enables `download` attribute value on the `<cc-icon>` tag. */
```
Galimede: nitpick: I find this comment rather confusing, it's not only about enabling the download icon, because if it was the case we could put a boolean, it's also about putting the download attribute on the `a`. 
Maybe we could change the comment to reflect that?

src/components/cc-link/cc-link.js
```diff
@@ -161,6 +171,12 @@ export class CcLink extends LitElement {
                 a11y-name=${i18n('cc-link.new-window.name')}
               ></cc-icon>`
             : ''}
+          ${this.download != null && isStringEmpty(this.innerHTML)
```
Galimede: question: why don't we display the icon if there's some HTML/text?

Galimede (commented):
Hey Hélène, great work! 💪 

I'm wondering if there's not a title or something missing when in download for the link, but I'll let @florian-sanders-cc handle the subject.

src/components/cc-link/cc-link.js
HeleneAmouzou: Absolutely !

src/components/cc-link/cc-link.js
HeleneAmouzou: It's a mistake, my bad !

PR #1534, opened 2025-09-09
fix(cc-addon-header): make documentation link dynamic

HeleneAmouzou:
## What does this PR do?

- Sets a `docName` in the smart and pass it to the success translation messages,
- Adds `docName` param in the success translation messages.

## How to review?

- Check the commit,
- Check on `demo-smart` ; select the `cc-addon-header.smart-keycloak`, restart the addon and click on the documentation link on the success message.

1 reviewer should be enough.

### DIFF
--- a/src/components/cc-addon-header/cc-addon-header.smart-keycloak.js
+++ b/src/components/cc-addon-header/cc-addon-header.smart-keycloak.js
@@ -6,9 +6,12 @@ import { fakeString } from '../../lib/fake-strings.js';
 import { notify, notifyError } from '../../lib/notifications.js';
 import { sendToApi } from '../../lib/send-to-api.js';
 import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
+import { generateDocsHref } from '../../lib/utils.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-smart-container/cc-smart-container.js';
 import './cc-addon-header.js';
+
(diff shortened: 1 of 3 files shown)


florian-sanders-cc (commented):
@HeleneAmouzou LGTM but I'm wondering if we shouldn't make the translation accept a `docsUrl` param and let the smart do the concat with `generateDocsHref(...)`?
This way if somehow one day the path is not the same for all products, we don't have any problem 🤔 (tell me if it's unclear we can discuss it in sync)

src/components/cc-addon-header/cc-addon-header.smart-keycloak.js
```diff
@@ -34,6 +35,8 @@ defineSmartComponent({
     const { apiConfig, ownerId, addonId, productStatus } = context;
     const api = new Api(apiConfig, ownerId, addonId, signal);
     let logsUrl = '';
+    const docsName = 'keycloak';
+    const docsUrl = generateDocsHref(`/addons/${docsName}/`);
```
florian-sanders-cc: **question**: @HeleneAmouzou you don't use `docsName` anywhere else so I'm wondering if we can go for this directly 🤔 
Note that this const could also be placed outside `defineSmartComponent` since it doesn't depend on the context (if you move it, you can rename it to `DOCS_URL`)

```suggestion
    const docsUrl = generateDocsHref(`/addons/keycloak/`);
```

PR #1536, opened 2025-09-10
logs: some tweaks

pdesoyres-cc:
## What this PR do ?

This PR adds some minor tweaks around logs.
- Uses ellipsis char when truncating metadata (and do it at cc-logs level)
- truncates access log status to 3 characters
- replace `??` by empty strings for access log city and country
- hide non-http access logs

## How to review

- check the commit
- check the stories


src/lib/logs/logs-stream.js
```diff
@@ -258,6 +258,10 @@ export class LogsStream {
   async #onStreamLogEvent(rawLog) {
     this.#waitingTimer.stop();
     const convertedLog = await this._convertLog(rawLog);
+    if (convertedLog == null) {
+      console.warn('received a log that could not be converted', rawLog);
+      return;
+    }
```
hsablonniere: question: does this mean we will have a `console.warn` for every non HTTP access log?

src/lib/utils.js
```diff
@@ -229,6 +229,22 @@ export function isStringBlank(string) {
...
+ * @param {string} [suffix]
+ */
+export function truncateString(string, maxLength, suffix = '\u2026') {
+  if (maxLength <= 0) {
+    throw new Error('maxLength must be greater than 0');
+  }
+  if (string.length <= maxLength) {
+    return string;
+  }
+  return string.slice(0, maxLength - suffix.length) + suffix;
```
hsablonniere, line 245: praise: well done :+1:

src/lib/logs/logs-stream.js
pdesoyres-cc: yes. Is this a problem? Maybe we should not print to console as it is valid to have non-http access log.

PR #1537, opened 2025-09-10
Allow customizing assets base URL

### LINKED ISSUE #1531: Add env variable for assets
## Context

Our components have hard coded references to assets.clever-cloud.com.

This doesn't really work in AirGap situations so the base URL should be set through a variable.

The system could be very similar to what we currently do for doc links at the moment.

florian-sanders-cc:
Fixes #1531 

## What does this PR do?

- Adds a `setAssetsBaseUrl` helper to override the default assets base URL (`https://assets.clever-cloud.com/`),
  - The goal is that the console may override the value based on an environment variable for contexts where the default URL does not work / exist.
- Adds a `getAssetsUrl` helper that concatenated a path with the base URL so that no component and no stories contain direct references to `https://assets.clever-cloud.com/`, only the paths to resources.
  - This means if we need to change the domain of assets, we can do so only from 1 file, just like for docs URLs,
  - This means the console may override the value as needed.

Thanks to @hsablonniere for the idea and @pdesoyres-cc for the peer coding session :wink: 

## How to review?

- Check the impacted components still work correctly:
  - cc-addon-elasticsearch-options,
  - cc-elasticsearch-info,
  - cc-grafana-info,
  - cc-heptapod-info,
  - cc-jenkins-info,
  - cc-matomo-info,
  - cc-pricing-page,
  - cc-zone,
  - cc-zone-picker,
  - cc-zone-input.
- If you want, you may use `setAssetsBaseUrl('https://example.com/toto')` in `.storybook/previews.js` and see the result in the impacted component stories,


src/lib/assets-url.js
```diff
@@ -0,0 +1,14 @@
...
+/** @param {string} value */
+export function setAssetsBaseUrl(value) {
+  assetsUrl = value.replace(/\/$/, '');
+}
+
+/**
+ * @param {string} path
+ * @returns {string}
+ */
+export function getAssetsUrl(path) {
```
hsablonniere: question: should this be `getAssetUrl` singular?

PR #1549, opened 2025-09-24
refactor(cc-addon-credentials-beta): use helper for docs URL

florian-sanders-cc:
## What does this PR do?

- Relies on the `generateDocsHref` helper instead of hardcoded value,
- Removes the docLink set within the state within the smart (it was irrelevant since this is not part of the state anyway).

## How to review?

- Check the commit,
- Check demo-smart locally, the doc link should still lead to the keycloak docs :+1:,


HeleneAmouzou (approved):
LGTM ! 👏

PR #1550, opened 2025-09-24
refactor(cc-addon-credentials-beta): align error translation

florian-sanders-cc:
## What does this PR do?

- Removes `the` from the error message when loading fails. This way the error message is exactly the same as other dashboard components.

## How to review?

- Check the commit,
- Check the error story => it should match the `cc-addon-info` & `cc-addon-header` error messages (in English).


florian-sanders-cc:
Automerging because there is really 0 impact :wink:

PR #1562, opened 2025-10-01
feat(cc-addon-*): add Matomo

HeleneAmouzou:
## What does this PR do?

- Adds Matomo dashboard to `index.html`,
- Declines `cc-addon-header` and `cc-addon-info` for Matomo product.

## How to review?

- Check the commits,
- Check the code,
- Play with `demo-smart`.


src/components/cc-addon-header/cc-addon-header.smart-matomo.js
```diff
@@ -0,0 +1,142 @@
...
+        {
+          url: fakeString(15),
+          name: fakeString(5),
+        },
+      ],
+      actions: {
+        restart: true,
+        rebuildAndRestart: true,
+      },
+      productStatus: fakeString(4),
```
florian-sanders-cc: **nitpick:** since Matomo is not in beta or anything, we could remove the product status from the skeleton

florian-sanders-cc: even better, I'm only realizing it now: we should not ever use fake strings for `productStatus`, we should just use the `productStatus` we get from `context`.
It's not something we load from the API anyway, it's given from the start by the console so we may as well use it right away.

This is something we should do for all add-ons, this way we only show a skeleton productStatus when needed.


```suggestion
      productStatus,
```

src/components/cc-addon-header/cc-addon-header.smart-matomo.md
```diff
@@ -0,0 +1,65 @@
...
+
+## ⚙️ Params
+
+| Name             | Type        | Details                                                                                       | Default |
+|------------------|-------------|-----------------------------------------------------------------------------------------------|---------|
+| `apiConfig`      | `ApiConfig` | Object with API configuration (target host, tokens...)                                        |         |
+| `ownerId`        | `string`    | UUID prefixed with orga_                                                                      |         |
+| `addonId`        | `string`    | ID of the add-on                                                                              |         |
+| `logsUrlPattern` | `string`    | Pattern for the logs url (Example : `/organisations/${ownerId}/applications/${appId}/logs`)   |         |
+| `productStatus`  | `string`    | Maturity status of the product                                                                |         |
```
florian-sanders-cc: **nitpick:** should be flagged as optional

src/components/cc-addon-info/cc-addon-info.smart-matomo.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /**
+   * @return {Promise<{addonInfo: RawAddon, operator: MatomoOperatorInfo, operatorVersion: string, grafanaAppLink: string}>}
+   */
+  async getMatomoAddonInfo() {
+    // Fetch addon to get realId,
+    const rawAddon = await this._getAddon();
+    this._realId = rawAddon.realId;
+    this._providerId = /** @type {import('../common.types.js').RawAddonProvider} */ rawAddon.provider.id;
```
florian-sanders-cc: **nitpick:** you probably need parenthesis when you use inline types like this


```suggestion
    this._providerId = /** @type {import('../common.types.js').RawAddonProvider} */ (rawAddon.provider.id);
```

florian-sanders-cc (changes requested):
A tiny request for change (which is related to a mistake I made actually 😄) but apart from that, all good! 😎

src/components/cc-addon-info/cc-addon-info.smart-matomo.js
HeleneAmouzou: It triggers TS errors when I had parenthesis :thinking:

florian-sanders-cc: I'm only noticing this now :sweat_smile:, you don't need this inline type at all, actually `rawAddon.provider.id` is a string (or a union if we want to be specific) so it's clearly not a `RawAddonProvider`.

```suggestion
    this._providerId = rawAddon.provider.id;
```

PR #1565, opened 2025-10-02
`cc-addon-credentials-*`: support multi-instances Network groups for keycloak

florian-sanders-cc:
## What does this PR do?

- Splits the network groups into two types:
  - `standard`, the regular ng used by most porducts,
  - `multi-instances`, ng + 2 instances, only for keycloak.
- Updates the `cc-addon-credentials-beta.smart.js` to use the `multi-instance` ng kind.

## How to review?

- Check the commits,
- Check the relevant stories,
- Check `demo-smart` (success & error cases for NG activation & deactivation).


src/translations/translations.en.js
```diff
@@ -191,10 +191,10 @@ export const translations = {
   'cc-addon-credentials-beta.doc-link.keycloak': `Keycloak - Documentation`,
   'cc-addon-credentials-beta.error': `Something went wrong while loading add-on information.`,
   'cc-addon-credentials-beta.heading': `Access`,
-  'cc-addon-credentials-beta.ng.disabling.error': `Something went wrong while trying to disable the Network Group`,
-  'cc-addon-credentials-beta.ng.disabling.success': `The Network Group has been successfully disabled`,
-  'cc-addon-credentials-beta.ng.enabling.error': `Something went wrong while trying to enable the Network Group`,
-  'cc-addon-credentials-beta.ng.enabling.success': `The Network Group has been successfully enabled`,
```
florian-sanders-cc, line 197: @HeleneAmouzou note that I had to remove these translations even though you need them for other components. If we kept them here, they would make the tests fails because they would be unused 😓 

In the PRs where you need them (Otoroshi I guess), you should just rename these to `cc-addon-credentials-beta.ng-standard.*` and they will look like a new addition related to your PR, WDYT?

HeleneAmouzou: I'll do that !

HeleneAmouzou (approved):
LGTM !

PR #1570, opened 2025-10-02
refactor(cc-addon-info): remove openScalabilityLink prop for Keycloak

HeleneAmouzou:
## What does this PR do?

- Removes `openScalability` prop for Keycloak,
- Rename `linked services` to `ressources`,
- Use `initial-user` and `initial-password` as labels.

## How to review?

- Check the commits,
- Check the code,
- Play with `demo-smart`.


florian-sanders-cc (approved):
Great job Hélène! :sunglasses:

florian-sanders-cc (changes requested):
Forgot to check the commit texts: 
I think we should go for `feat` commits instead of refactor.
These changes should be part of the changelog since the components have already been released. They change the component behavior & look so we should warn people about that (of course nobody uses them except us but still 😄)

For the `linked services` -> `resources` I would mention that it's only an i18n / text change but that the API is unchanged (still `linked services`). This could be specified in the commit note right after the one line message?

florian-sanders-cc (approved):

PR #1589, opened 2025-10-24
fix(cc-article-list): update article feed URLs to `clever.cloud` domain

florian-sanders-cc:
## What does this PR do?

- Currently, the `cc-article-list` fails to load because of CORS issues. The website has moved to `clever.cloud` so CORS rules for `clever-cloud.com` no longer apply.
- Corrects the feed URLs from `clever-cloud.com` to `clever.cloud` for both French and default language feeds.

## How to review?

- Check the commit,
- Check demo-smart,


HeleneAmouzou (approved):
LGTM, thanks for handling that ! 👏

PR #1591, opened 2025-10-24
feat(cc-range-selector): init component

roberttran-cc:
# What does this PR do?

- add a `cc-range-selector` component
- add a `cc-range-selector-option` sub component
- add a `range-selector-dragging-controller` to manage drag selection state
- add a `trimArray` utility function to `src/lib/utils.js` with comprehensive tests
- add `CcRangeSelectEvent` to `common.events.js`
- add `CcRangeSelectorSelectCustom` event for custom option selection
- add i18n translations (EN/FR) for error messages and UI labels
- integrate `cc-range-selector` in the form demo sandbox

# How to review?

- check the commits
- check the stories in the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-range-selector/init/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-range-selector--default-story)
  - test both modes: 'single' for radio-like selection and 'range' for drag selection
  - test the drag interaction: mousedown → drag → mouseup (range mode)
  - test keyboard navigation
  - test the custom option feature with show-custom attribute
- play with the sandbox

# What's to discuss

Feel free to challenge/comment on the following points:
- the component name (`cc-range-selector`)
- the property naming: `mode`, `selection`, `showCustom`
- the boundary-based selection approach (`{ startValue, endValue }`) vs storing full array
  - this optimizes memory but requires deriving the full array when needed
- the drag interaction pattern: clear → preview → apply (with rollback on outside click)
  - single-click without drag doesn't create a selection in range mode
- the handling of disabled options within a selected range:
  - they're visually included in the range
  - but excluded from form submission and the values array
  - disabled options at range boundaries are automatically trimmed on initialization
- the custom option button behavior:
  - clears current selection
  - dispatches custom event for parent handling
  - prevents multiple activations

### DIFF
--- a/src/components/cc-range-selector/cc-range-selector.js
+++ b/src/components/cc-range-selector/cc-range-selector.js
@@ -1,0 +1,1314 @@
+import { css, html, render } from 'lit';
+import { isTemplateResult } from 'lit/directive-helpers.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { createRef, ref } from 'lit/directives/ref.js';
+import { iconRemixArrowRightDoubleFill as iconArrow } from '../../assets/cc-remix.icons.js';
+import { EventHandler } from '../../lib/events.js';
+import { CcFormControlElement } from '../../lib/form/cc-form-control-element.abstract.js';
+import { RequiredValidator } from '../../lib/form/validation.js';
+import { findLastIndex, isStringBlank, trimArray } from '../../lib/utils.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-range-selector-option/cc-range-selector-option.js';
+import { CcSelectEvent } from '../common.events.js';
+import { CcRangeSelectEvent, CcRangeSelectorSelectCustom } from './cc-range-selector.events.js';
+import { RangeSelectorDraggingController } from './range-selector-dragging-controller.js';
+
+/**
+ * @import { RangeSelectorMode, RangeSelectorOption, RangeSelectorSelection, RenderOptionContext } from './cc-range-selector.types.js'
+ * @import { GenericEventWithTarget } from '../../lib/events.types.js'
+ * @import { Validator, ErrorMessageMap } from '../../lib/form/validation.types.js'
+ * @import { FormControlData } from '../../lib/form/form.types.js'
+ * @import { PropertyValues } from 'lit'
+ * @import { ClassInfo } from 'lit/directives/class-map.js'
+ * @import { Ref } from 'lit/directives/ref.js'
+ */
+
+/**
+ * A range selector component that allows users to select ranges from a list of options.
+ *
+ * It extends `CcFormControlElement`, so it can be used in a form and its value can be submitted.
+ *
+ * ## Important Caveats
+ *
+ * **Disabled Options in Selections:**
+ * For visualization purposes, disabled options can appear within a selected range in range mode.
+ * However, these disabled options are automatically excluded from form submission and the values array.
+ * Additionally, on component updates (willUpdate), any disabled options at the start or end
+ * of the values array are automatically trimmed to ensure clean selection boundaries.
+ *
+ * **Range Mode Interaction:**
+ * In range mode, the drag interaction uses a clear/preview/apply pattern:
+ * - On mousedown: Current selection is cleared and stored for potential rollback
+ * - During drag: "dragging" visual state shows preview, "selected" state is hidden
+ * - On mouseup: Selection is applied if range is valid (>1 option)
+ * - On outside click: Previous selection is restored (rollback)
+ *
+ * **Single-Click Limitation:**
+ * In range mode, clicking a single option without dragging will not create a selection.
+ * This is intentional to distinguish between single clicks and drag operations.
+ *
+ * **Hidden Range Inputs for Keyboard Navigation:**
+ * Both modes use hidden `<input type="range">` elements for keyboard navigation.
+ * These native inputs require a valid value within their min-max bounds.
+ * When no initial selection exists, default indices ensure the inputs are valid
+ * and accessible for keyboard interaction.
+ *
+ * @cssdisplay inline-block
+ *
+ * @csspart options - Styles the options container, mainly to modify their layout.
+ *
+ * @cssprop {Size} --cc-form-label-gap - The space between the label and the control (defaults: `0.35em`).
+ * @cssprop {Size} --cc-form-label-gap-inline - The space between the label and the control when layout is inline (defaults: `0.75em`).
+ * @cssprop {Color} --cc-input-label-color - The color for the input's label (defaults: `inherit`).
+ * @cssprop {FontSize} --cc-input-label-font-size - The font-size for the input's label (defaults: `inherit`).
+ * @cssprop {FontWeight} --cc-input-label-font-weight - The font-weight for the input's label (defaults: `normal`).
+ * @cssprop {Display} --cc-range-selector-trailing-arrow-display - Controls the display of the trailing arrow after the final option, mostly needed when options default layout is overridden to `grid` (defaults: `none`).
+ * @cssprop {Width} --cc-range-selector-options-width - Sets the width of the form control content (defaults: `fit-content`).
+ * @cssprop {Size} --cc-range-selector-options-indent - Horizontal space between the start of the line and the options (defaults: `0.25em`).
+ *
+ * @slot help - The help message to be displayed right below the options. Please use a `<p>` tag.
+ */
+export class CcRangeSelector extends CcFormControlElement {
+  static get properties() {
+    return {
+      ...super.properties,
+      customA11yDesc: { type: String, attribute: 'custom-a11y-desc' },
+      disabled: { type: Boolean, reflect: true },
+      inline: { type: Boolean, reflect: true },
+      label: { type: String },
+      mode: { type: String, reflect: true },
+      options: { type: Array },
+      readonly: { type: Boolean, reflect: true },
+      required: { type: Boolean, reflect: true },
+      selection: { type: Object },
+      showCustom: { type: Boolean, attribute: 'show-custom' },
+      value: { type: String },
+      _isCustomOptionActive: { type: Boolean, state: true },
+    };
+  }
+
+  static reactiveValidationProperties = ['required', 'options'];
+
+  constructor() {
+    super();
+
+    /** @type {string} Optional accessibility description appended to the custom option button's title attribute for screen readers (default: null) */
+    this.customA11yDesc = null;
+
+    /** @type {boolean} Whether the component should be disabled (default: 'false') */
+    this.disabled = false;
+
+    /**
+     * @type {boolean} Sets the `<label>` on the left of the options.
+     * Only use this if your form contains 1 or 2 fields and your labels are short (default: 'false').
+     */
+    this.inline = false;
+
+    /** @type {string} The label describing the range selector, displayed before the options (default: '') */
+    this.label = '';
+
+    /** @type {RangeSelectorMode} The selection mode: 'single' for single selection or 'range' for range selection (default: 'range') */
+    this.mode = 'range';
+
+    /** @type {RangeSelectorOption[]} The list of options */
+    this.options = null;
+
+    /** @type {boolean} Whether the component should be readonly when not disabled (default: 'false') */
+    this.readonly = false;
+
+    /** @type {boolean} Sets the "required" text inside the label (default: 'false'). */
+    this.required = false;
+
+    /** @type {boolean} Shows a custom option button at the end of the options list, allowing users to trigger custom selection logic (default: 'false'). */
+    this.showCustom = false;
+
+    /** @type {string} The current value of the selector for single mode. */
+    this.value = null;
+
+    /** @type {RangeSelectorSelection | null} The current boundaries of the selector for range mode - stores boundary values (start and end) of the selection. */
+    this.selection = null;
+
+    /** @type {Ref<HTMLElement>} */
+    this._inputRangeStartRef = createRef();
+
+    /** @type {Ref<HTMLElement>} */
+    this._inputRangeEndRef = createRef();
+
+    /** @type {Ref<HTMLElement>} */
+    this._errorRef = createRef();
+
+    /** @type {Ref<HTMLElement>} */
+    this._selectorRef = createRef();
+
+    /** @type {RangeSelectorDraggingController} */
+    this._dragCtrl = new RangeSelectorDraggingController(this);
+
+    /** @type {ErrorMessageMap} */
+    this._errorMessages = {
+      empty: () => i18n('cc-range-selector.error.empty'),
+    };
+
+    const onOutsideClick = (/** @type {Event} */ event) => {
+      if (!this._dragCtrl.isDragging()) {
+        return;
+      }
+
+      const contentElement = this._selectorRef.value;
+      if (contentElement != null && !event.composedPath().includes(contentElement)) {
+        if (this._isModeRange()) {
+          this._dragCtrl.stop();
+          this.selection = this._dragCtrl.getPreviousSelection();
+        }
+      }
+    };
+    this._onOutsideClickHandler = new EventHandler(window, 'click', onOutsideClick);
+    this._onCcButtonClickHandler = new EventHandler(window, 'cc-click', onOutsideClick);
+
+    this._isCustomOptionActive = false;
+
+    /** @type {string[]} */
+    this._optionsInnerText = null;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this._onOutsideClickHandler.connect();
+    this._onCcButtonClickHandler.connect();
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this._onOutsideClickHandler.disconnect();
+    this._onCcButtonClickHandler.disconnect();
+  }
+
+  /**
+   * Validates selection and trims disabled options from the start and end of the selected range.
+   * This ensures that the selection doesn't visually include leading or trailing disabled options.
+   * @param {PropertyValues<CcRangeSelector>} changedProperties
+   */
+  willUpdate(changedProperties) {
+    if (changedProperties.has('options')) {
+      this._optionsInnerText = this.options.map(this._getOptionText);
+    }
+
+    if (!changedProperties.has('selection') && !changedProperties.has('options')) {
+      return;
+    }
+
+    if (this.disabled || this.selection == null) {
+      return;
+    }
+
+    // Validate selection in range mode - this can only happen if set programmatically with invalid values
+    if (this._isModeRange()) {
+      const indexes = this._getSelectionIndexes();
+
+      if (indexes.start === -1) {
+        throw new Error(
+          `Invalid selection: startValue "${this.selection.startValue}" not found in options. Available values: ${this.options?.map((o) => o.value).join(', ')}`,
+        );
+      }
+      if (indexes.end === -1) {
+        throw new Error(
+          `Invalid selection: endValue "${this.selection.endValue}" not found in options. Available values: ${this.options?.map((o) => o.value).join(', ')}`,
+        );
+      }
+      if (indexes.start > indexes.end) {
+        throw new Error(
+          `Invalid selection: startValue "${this.selection.startValue}" comes after endValue "${this.selection.endValue}" in options array`,
+        );
+      }
+      if (indexes.start === indexes.end) {
+        throw new Error(
+          `Invalid selection: startValue and endValue are the same ("${this.selection.startValue}"). Use mode="single" with value="${this.selection.startValue}" for single selection`,
+        );
+      }
+    }
+
+    const valuesArray = this._getValuesArray(/* includeDisabled */ true);
+    const trimmedArray = trimArray(valuesArray, (/** @type {string} */ value) => {
+      return this.options?.find((option) => option.value === value)?.disabled;
+    });
+
+    // Only trim if not already clean (prevent loops)
+    if (trimmedArray.length !== valuesArray.length) {
+      this.selection =
+        trimmedArray.length === 0
+          ? null
+          : {
+              startValue: trimmedArray.at(0),
+              endValue: trimmedArray.at(-1),
+            };
+    }
+  }
+
+  /* region CcFormControlElement implementation */
+
+  /**
+   * @return {HTMLElement}
+   * @protected
+   */
+  _getFormControlElement() {
+    return this._selectorRef.value;
+  }
+
+  /**
+   * @return {HTMLElement}
+   * @protected
+   */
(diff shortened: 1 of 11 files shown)


src/components/cc-range-selector-option/cc-range-selector-option.js
```diff
@@ -0,0 +1,193 @@
...
+      disabled: this.disabled,
+      readonly: !this.disabled && this.readonly,
+      error: !this.disabled && !this.readonly && this.error,
+      dragging: this.dragging,
+      selected: !this.dragging && this.selected, // Selected styling is hidden during dragging
+    };
+
+    return html`
+      <div class="wrapper ${classMap(classes)}">
+        <slot name="body"></slot>
```
pdesoyres-cc: question: do we really need a named slot when there is only one slot? Why not using the default slot instead?

src/components/common.events.js
```diff
@@ -31,6 +31,21 @@ export class CcSelectEvent extends CcEvent {
...
+  static TYPE = 'cc-range-select';
+
+  /**
+   * @param {string[]} detail
+   */
+  constructor(detail) {
+    super(CcRangeSelectEvent.TYPE, detail);
+  }
+}
+
```
pdesoyres-cc: question: I'm wondering why you put this one in the common events. Do you fill it will be used in some other components?

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,998 @@
...
+    /** @type {boolean} Sets the "required" text inside the label (default: 'false'). */
+    this.required = false;
+
+    /** @type {boolean} Shows a custom option button at the end of the options list, allowing users to trigger custom selection logic (default: 'false'). */
+    this.showCustom = false;
+
+    /** @type {string} The current value of the selector for single mode. */
+    this.value = null;
+
+    /** @type {RangeSelectorSelection | null} The current boudaries of the selector for range mode - stores boundary values (start and end) of the selection. */
```
pdesoyres-cc: fix: typo

```suggestion
    /** @type {RangeSelectorSelection | null} The current boundaries of the selector for range mode - stores boundary values (start and end) of the selection. */
```

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,998 @@
...
+   */
+  _getValidator() {
+    return combineValidators([
+      this.required ? new RequiredValidator() : null,
+      createValidator(() => {
+        const indexes = this._getSelectionIndexes();
+        if (!this.required && this.selection == null) {
+          return Validation.VALID;
+        }
+        if (indexes.first == null || indexes.last == null || indexes.first >= indexes.last) {
```
pdesoyres-cc: question: about condition `indexes.first >= indexes.last`. It's impossible with user interaction to come to this situation right? Why do we need this test?

Maybe we could make the `_getSelectionIndexes()` method handle that case by switching start and end boundaries?

src/components/cc-range-selector-option/cc-range-selector-option.js
roberttran-cc: You're right, it's a relic from old iterations: I'll remove it, it will make the code simpler. :+1: 

_edit:_ as it's a minor update, will be merged in the initial commits to facilitate second round of review.

src/components/common.events.js
roberttran-cc: I did think that theoretically but I don't see it anytime soon: I'll move it to component-specific events file, it will be clearer. :+1: 

_edit:_ as it's a minor update, will be merged in the initial commits to facilitate second round of review.

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: It's intended for when selection data is set programmatically (as in the second example in the `errorWithInvalidSelection` story).

Let me know if I overthought it.

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+      const contentElement = this._selectorRef.value;
+      if (contentElement != null && !event.composedPath().includes(contentElement)) {
+        if (this._isModeRange()) {
+          this._dragCtrl.stop();
+          this.selection = this._dragCtrl.getPreviousSelection();
+        }
+      }
+    };
+    this._onOutsideClickHandler = new EventHandler(window, 'click', onOutsideClick);
+    this._onCcButtonClickHandler = new EventHandler(window, 'cc-click', onOutsideClick);
```
florian-sanders-cc, line 167: **question:** why do we need to listen to `cc-click` specifically? I think the `cc-click` event is dispatched in addition to regular `click` events so `click` events should cover all cases anyway (unless we catch & prevent default them I guess?)

src/components/cc-range-selector/cc-range-selector.types.d.ts
```diff
@@ -0,0 +1,12 @@
+export type RangeSelectorMode = 'single' | 'range';
+
+export interface RangeSelectorOption {
+  body: string | Node;
```
florian-sanders-cc: **nitpick/suggestion:** I kinda like `label` for cases like these. I think it emphasizes the fact that each option will be attached to a form control + that the content should not be too rich / long. For instance it should never contain any interactive element whatsoever.
It's just a suggestion but it would match what we do for `cc-toggle` & `cc-select` options.

```suggestion
  label: string | Node;
```

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+ */
+
+/**
+ * @typedef {object} RenderOptionContext
+ * @property {{ first: number, current: number, last: number }} indexes - Selection index information
+ * @property {boolean} isModeSingle - Whether the selector is in single selection mode
+ * @property {boolean} isModeRange - Whether the selector is in range selection mode
+ * @property {boolean} isLastOption - Whether this is the last option in the list (used to control arrow visibility)
+ * @property {RangeSelectorOption} nextOption - The next option in the list, used to determine arrow highlighting state
+ */
```
florian-sanders-cc: **suggestion:** move this to the `d.ts` file

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+    this._onOutsideClickHandler.disconnect();
+    this._onCcButtonClickHandler.disconnect();
+  }
+
+  /**
+   * Lifecycle method called after the first render.
+   * Trims disabled options from the start and end of the selected range.
+   * This ensures that the selection doesn't visually include leading or trailing disabled options.
+   */
+  firstUpdated() {
```
florian-sanders-cc: **question:** why only for the first render and not for others? I feel like the component could be "reused" / "reset" with new values and you would need to handle such cases but I'm pretty sure it's me who's missing the point of this 🤔

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+      }),
+    ]);
+  }
+
+  /**
+   * Returns the form control data for submission.
+   * Note: Return type differs based on mode:
+   * - Single mode: returns a string value (this.value)
+   * - Range mode: returns FormData with multiple entries under the same name
+   * @return {string|FormControlData} FormData for range mode, string for single mode
```
florian-sanders-cc: **nitpick:** `FormControlData` already is `string | FormData | File`

```suggestion
   * @return {string|FormData} FormData for range mode, string for single mode
```

or 


```suggestion
   * @return {FormControlData} FormData for range mode, string for single mode
```

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+        @input=${this._onControlInput}
+        ${ref(this._selectorRef)}
+        role="${isModeRange ? 'group' : 'radiogroup'}"
+        tabindex="-1"
+      >
+        <div class="fieldset-content">
+          <legend class="legend">
+            <span class="legend-text">${this.label}</span>
+            ${this.required ? html` <span class="required">${i18n('cc-range-selector.required')}</span> ` : ''}
+          </legend>
```
florian-sanders-cc: **fix:** the `legend` element must immediately follow the `fieldset` element. The `div` element in the middle breaks the association (the fieldset has no accessible name, the legend won't be announced upon entering the fieldset).

Since `fieldset` styling can be tricky, you can apply what is described in the `solutions` section from this ADR: https://www.clever-cloud.com/doc/clever-components/?path=/docs/%F0%9F%93%8C-architecture-decision-records-adr-0025-moving-away-from-fieldset-and-legend--docs

We can discuss it in sync if needed 😉

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+    // Calculate mode constants once for all options
+    const isModeSingle = this._isModeSingle();
+    const isModeRange = this._isModeRange();
+
+    return html`
+      <fieldset
+        class="fieldset ${classMap({ 'is-error': hasErrorMessage })}"
+        @input=${this._onControlInput}
+        ${ref(this._selectorRef)}
+        role="${isModeRange ? 'group' : 'radiogroup'}"
```
florian-sanders-cc: **note/praise:** last time I checked (more than 3 years ago), `radiogroup` was not really supported by most of browsers and ended up being computed as `group`.

Still I like the idea. Worst case scenario it ends up as `group` and it's perfectly fine, it's a nice bonus when it works 👍

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1002 @@
...
+        }
+        /* endregion */
+
+        /* region option - custom button */
+        .btn-custom {
+          align-items: stretch;
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: inline-flex;
+          flex: 1 1 auto;
+          outline-offset: var(--cc-focus-outline-offset, 2px);
```
florian-sanders-cc: **nitpick:** we usually set this right after the `outline` property to have them both together but it's not a big deal if you want to stick with that

src/translations/translations.en.js
```diff
@@ -1557,6 +1557,12 @@ export const translations = {
   'cc-product-list.search-empty': `No products matching your search criteria were found.`,
   'cc-product-list.search-label': `Search for a product`,
   //#endregion
+  //#region cc-range-selector
+  'cc-range-selector.custom': `Customize`,
+  'cc-range-selector.error.empty': `This field is required.`,
+  'cc-range-selector.error.invalid-selection': `Invalid selection.`,
```
florian-sanders-cc: **nitpick:** we try to avoid trailing dots for short messages (though I'm sure we still have exceptions)

```suggestion
  'cc-range-selector.error.empty': `This field is required`,
  'cc-range-selector.error.invalid-selection': `Invalid selection`,
```

src/translations/translations.fr.js
```diff
@@ -1581,6 +1581,12 @@ export const translations = {
   'cc-product-list.search-empty': `Aucun produit ne correspond à vos critères de recherche.`,
   'cc-product-list.search-label': `Chercher un produit`,
   //#endregion
+  //#region cc-range-selector
+  'cc-range-selector.custom': `Personnaliser`,
+  'cc-range-selector.error.empty': `Ce champ est obligatoire.`,
+  'cc-range-selector.error.invalid-selection': `Sélection invalide.`,
```
florian-sanders-cc: **nitpick:** we try to avoid trailing dots for short messages (though I'm sure we still have exceptions)

```suggestion
  'cc-range-selector.error.empty': `Ce champ est obligatoire`,
  'cc-range-selector.error.invalid-selection': `Sélection invalide`,
```

src/translations/translations.en.js
```diff
@@ -1557,6 +1557,12 @@ export const translations = {
   'cc-product-list.search-empty': `No products matching your search criteria were found.`,
   'cc-product-list.search-label': `Search for a product`,
   //#endregion
+  //#region cc-range-selector
+  'cc-range-selector.custom': `Customize`,
+  'cc-range-selector.error.empty': `This field is required.`,
```
florian-sanders-cc: **nitpick:** we usually go for `You must select a value` (maybe we could be even most specific for the range mode like you must select a range of two values)

src/translations/translations.en.js
florian-sanders-cc: **question (not sure):** when a selection is invalid, we must tell users what to do to correct their mistake.
Typically if the start value is higher than the end value, we should say it. 
But if I understand the component, this can only happen programmatically, so it's more of a developer issue than a user facing error 🤔 

This makes me wonder if the component should throw an error instead and break?

src/components/cc-range-selector/cc-range-selector.types.d.ts
roberttran-cc: Hard decision: unlike `cc-toggle` & `cc-select`, this component attribute accepts `string` but `Node` as well. And it's consistent with the other tile component (`cc-picker`).

Decision: leave it as is and discuss this topic separately.

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: You're right, it's a valid concern, I was too focused on a specific use case on this one.

Decision: will update.

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: Nice catch.

Decision: will update (minor update, will be merged in the initial commits).

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: Thanks for pointing this out.

Decision: will update.

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: Agree.

Decision: will update (minor update, will be merged in the initial commits).

src/translations/translations.en.js
roberttran-cc: Nice catch.

Decision: will update (minor update, will be merged in the initial commits).

src/translations/translations.fr.js
roberttran-cc: Nice catch.

Decision: will update (minor update, will be merged in the initial commits).

src/translations/translations.en.js
roberttran-cc: Nice suggestion.

Decision: will update (minor update, will be merged in the initial commits).

src/translations/translations.en.js
roberttran-cc: Another valid concern.

Decision: will update.

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: Actually, there's a `e.stopPropagation();` [here](https://github.com/CleverCloud/clever-components/blob/master/src/components/cc-button/cc-button.js#L192) that prevents that (it may no longer be needed but that's a different topic). And my actual use case was testing the component in the sandbox (see details in [this commit](https://github.com/CleverCloud/clever-components/pull/1591/changes/33e2e97f6bafd5720dc05bd5804fd2798d3827ae)).

Decision: leave it as is and discuss separately the `stopPropagation`.

roberttran-cc:
Answers and questions for @florian-sanders-cc.

### Disabled state styling

**Observation:** Disabled = blank/white, Active = grey. Usually opposite (disabled = grey, active = white).

**Answer:** Actually agree, I probably got lost when styling the component.

**Decision:** update to match other components (will be merged in the initial commits).

### Dragging Behavior Bug

**Issue:** Ending drag on non-option element (arrow) leaves selection in pending/preview state without applying.

**Root cause:** mouseup only handled on cc-range-selector-option elements. If mouseup fires on arrow or gap, selection never applied.

**Decision:** will be fixed to improve UX, although this state matches the behaviour when you start a selection with a simple click.

### Custom Option Usecase

**Question:** What's the usecase? Seems tunnel-specific. Should it be external button?

**Answer:** Use case: Provide "escape hatch" for users needing values outside preset range.

**Example scenarios:**
- Range offers 1-30 days, user needs 90 days → click custom → open modal
- Range offers S/M/L sizes, user needs custom dimensions → click custom → input form
- Integrates visually within selector while triggering parent-controlled flow

**Why inside component:**
- Visual consistency (looks like one cohesive control)
- Clear alternative option (not hidden external button)
- Form flow stays inline (no hunting for "other" button elsewhere)

**Alternative:** Could be external, but breaks visual grouping.

### Misc

For the record, in single mode, we keep a radio input.

Also, the 'same visual feedback when using the keyboard as we get when using the mouse (dotted blue borders)' feedback is not relevant anymore I guess.

### Question

With the input range implementation, should we keep the error story in range mode? I disabled the a11y test on this one to prevent error.

Galimede (changes requested):
I will continue my review later on, however here are two related bugs that comes from the `custom` option. 

## Bug #1: Customize Button Stays Active After Drag

When you click "Customize" and then drag to select options, the customize button remains visually active if you release the mouse outside of an option (on an arrow, gap, or elsewhere in the fieldset).

###   Reproduction Steps

  1. Open the range selector with custom option enabled
  2. Make a selection (e.g., select options 2-4)
  3. Click the "Customize" button
    - Button becomes visually active
    - Selection is cleared
  4. Start dragging from an option and drag across several options
  5. Release the mouse over an arrow icon or gap (not directly on an option)
    - ❌ Bug: The customize button remains highlighted
    - ❌ Bug: _isCustomOptionActive stays true even though new selection is applied

### Expected Behavior

  When a new selection is made via drag (regardless of where mouseup occurs), the customize button should become inactive.

##  Bug #2: Arrow Icons and Border-Radius Artifacts

 When _isCustomOptionActive = true, there's a state mismatch between where the rendermethod sources its index data vs where selection state is determined.

 The Mismatch:

  1. In render() (line 810):
  // When custom is active, uses _lastKnowSelection for rendering indexes
  const { start, end } = !this._isCustomOptionActive
    ? this._getSelectionIndexes()      // ← Uses this.selection
    : this._lastKnowSelection;         // ← Uses cached old selection
  2. In _isOptionSelected() (line 704):
  // ALWAYS uses this._getSelectionIndexes() which reads this.selection
  const { start, end } = this._getSelectionIndexes();  // ← Always uses this.selection
  3. In _renderOption() (lines 892-894):
  const isFirst = indexes.current === indexes.start;  // ← From render() - uses _lastKnowSelection when custom active
  const isLast = indexes.current === indexes.end;     // ← From render() - uses _lastKnowSelection when custom active
  const isSelected = this._isOptionSelected(value);   // ← ALWAYS uses this.selection

  Result: When custom is active:
  - isFirst and isLast are calculated based on the old selection (_lastKnowSelection)
  - isSelected is false because this.selection = null
  - CSS classes like 'within-selection', 'range-continues-left', 'range-continues-right' are applied based on the old selection
  - This removes border-radius on options that aren't actually selected
  - Arrows might be positioned incorrectly

### Reproduction Steps

  1. Open the range selector with custom option enabled
  2. Select a range (e.g., options 2-5)
  3. Click "Customize"
    - Selection is cleared
    - _isCustomOptionActive = true
    - _lastKnowSelection stores {start: 2, end: 5}
    - this.selection = null
  4. Look at the UI carefully
    - ❌ Bug: Options 2-5 might have modified border-radius (flattened on inner sides)
    - ❌ Bug: The visual state doesn't match the actual selection state

  This becomes more visible if you interact further or if the customize button stays active (Bug #1).

--- 

One of the fix would be to clear custom state in _onFieldsetMouseUp:
```
  File: cc-range-selector.js:453-457

  _onFieldsetMouseUp() {
    if (this._dragCtrl.isDragging() && this._dragCtrl.getSize() > 0) {
      this._applyRangeSelection();
     // -> this._isCustomOptionActive = false;  
    }
  }
```
This would ensure the custom button is deactivated when a drag ends anywhere in the fieldset, not just on options.

pdesoyres-cc (changes requested):
Code is easy to read, there are unit tests, lots of comments! Well done @roberttran-cc for this tricky component.

I've got nothing more to say apart from a bug which may be for the same reason as the one raised above by @Galimede. It may be just another way to end with this illegal state.

We can end up with a range selection AND the `Customize` button on:

Image

https://github.com/user-attachments/assets/0e3aff68-2fa6-481b-bc13-22794eba03fc

**Reproduce with:**

- go to [With custom option story](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-range-selector/init/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-range-selector--with-custom-option)
- click on `Customize` button
- click on second option
- click on `Customize` button

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1241 @@
...
+
+  static reactiveValidationProperties = ['required', 'options'];
+
+  constructor() {
+    super();
+
+    /** @type {boolean} Whether the component should be disabled (default: 'false') */
+    this.disabled = false;
+
+    /** @type {boolean} Sets the `<label>` on the left of the options.
```
Galimede: nitpick: could you put the type on a separate line? Type is not highlighted in VSCode if put on same line as `/**` if there are multiple lines

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -0,0 +1,1241 @@
...
+    }
+
+    // If body is Node, extract textContent
+    if (body instanceof Node) {
+      return body.textContent || value;
+    }
+
+    // If body is TemplateResult, render to temp container and extract text
+    // TemplateResult has .strings property (Lit-specific marker)
+    if (body != null && typeof body === 'object' && 'strings' in body) {
```
Galimede: suggestion: I don't know if you have tried it but there's a `isTemplateResult` helper function provided by Lit

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: No problem.

Decision: will update (minor update, will be merged in the initial commits).

src/components/cc-range-selector/cc-range-selector.js
roberttran-cc: Hoooo that's so nice and it works! :star_struck:

Decision: will update (minor update, will be merged in the initial commits).

src/components/cc-range-selector/cc-range-selector.js
```diff
@@ -1054,9 +1058,17 @@ export class CcRangeSelector extends CcFormControlElement {
     const inRangeWhileDragging = false;
     const hasPointer = !isDisabled && !this.readonly && !this._isCustomOptionActive;
 
+    const a11yDesc = this.customA11yDesc?.trim() ?? '';
+    const title = i18n('cc-range-selector.custom') + (a11yDesc !== '' ? ` - ${a11yDesc}` : '');
```
florian-sanders-cc: **nitpick/suggestion:** could use `!isStringEmpty()`

PR #1592, opened 2025-10-27
feat(cc-logs): add log inspect feature

### LINKED ISSUE #1017: As a user, I want to see the filtered logs "in context"
As of now, when filtering by log message, we show only the logs that correspond to the current filter.

When looking for an issue somewhere in a log stream, we first try to identify a log, but then, we also need to see what happened just before, or just after. Meaning: we need to see the log "in context".

Ideas:

- add a  "show in context" button somewhere along the log entry which
  - clears the filter
  - centers the view on the log entry
  - adds some kind of focus on the log line
  - forces a logs stream pause

- add a "show/hide surrounding logs" toggle button somewhere along the log entry which
  - shows/hides the 10 logs before and after while keeping the current filter

pdesoyres-cc:
Fixes #1017

## What this PR does?

This PR adds a log inspect feature across multiple logging components:

- cc-logs: base log inspect functionality
- cc-logs-app-runtime: for application runtime logs
- cc-logs-addon-runtime: for addon runtime logs
- cc-logs-app-access: for application access logs

After finding a specific log (with filter) the feature allows users to pause the stream, clear the filter and center the log on the screen.

##  How to review?

- check the code
- use sandbox to check the behavior

### DIFF
--- a/src/components/cc-logs-app-access/cc-logs-app-access.js
+++ b/src/components/cc-logs-app-access/cc-logs-app-access.js
@@ -196,6 +196,10 @@ export class CcLogsAppAccess extends LitElement {
     this._messageFilter = detail;
   }
 
+  _onLogInspect() {
+    this._messageFilter = { value: '', mode: this._messageFilter?.mode ?? 'loose' };
+  }
+
   /* endregion */
 
   render() {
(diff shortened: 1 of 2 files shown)


florian-sanders-cc (approved):
Works well and the code looks clean to me!

Well done @pdesoyres-cc 😉

PR #1599, opened 2025-10-30
Feat/cc feature list

clara-layus-cc:
# What does this PR do?

- add a `cc-feature-list` component

# How to review?

- check the commit
- check the stories in the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/cc-feature-list/index.html?path=/docs/%F0%9F%9B%A0-profile-cc-feature-list--docs)


src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+        detail: {
+          featureId: feature.featureId,
+          newValue,
+        },
+        bubbles: true,
+        composed: true,
+      }),
+    );
+
+    console.log(newValue);
```
pdesoyres-cc: fix: remove console.log

src/components/cc-feature-list/cc-feature-list.events.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Dispatched when a feature setting changes.
+ * @extends {CcEvent<string>}
+ */
+export class CcFeatureValueChangeEvent extends CcEvent {
+  static TYPE = 'cc-feature-setting-change';
+
+  /**
+   * @param {string} detail
```
pdesoyres-cc: fix: detail should contain the `featureId` and  the `newValue`

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+   * @param {Feature} feature
+   * @param {String} newValue
+   */
+  _onToggleChange(feature, newValue) {
+    if (this.state.type === 'loading') {
+      return;
+    }
+
+    this.dispatchEvent(
+      new CustomEvent('cc-feature-setting-change', {
```
pdesoyres-cc: fix: use `CcFeatureValueChangeEvent` instead

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+                  ${i18n('cc-feature-list.feedback-link')}
+                </cc-link>
+              `
+            : ''}
+        </div>
+        <div slot="header-right">
+          <cc-toggle
+            .choices="${feature.toggleOptions}"
+            value="${feature.toggleValue}"
+            @cc-select=${(/** @type {CcFeatureValueChangeEvent} */ event) =>
```
pdesoyres-cc: fix: wrong type `CcFeatureValueChangeEvent`. Here you receive the event dispatched by `cc-toggle` component which is a `CcSelectEvent`

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+   * @param {Feature} feature
+   */
+  renderFeature(feature) {
+    return html`
+      <cc-block class="feature">
+        <div slot="header-title">
+          ${feature.featureName}
+          ${feature.status
+            ? html`
+                <cc-badge intent="${feature.status === 'alpha' ? 'warning' : feature.status === 'beta' ? 'info' : ''}">
```
pdesoyres-cc: fix: this double ternary is hard to read. We try to avoid that. In that case, we prefer create a intermediate variable with a `if-then-else` or `switch` pattern.

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+                  ${feature.status}
+                </cc-badge>
+              `
+            : ''}
+        </div>
+        <div slot="content">${feature.featureDescription}</div>
+        <div slot="footer">
```
pdesoyres-cc: fix: maybe we should not have this div at all if `feature.documentationLink` and `feature.feedbackLink` are both null. I'm not sure if the cc-block handles properly when slot is empty (will we have an extra margin?)

src/components/cc-feature-list/cc-feature-list.stories.js
pdesoyres-cc, line 1: fix: can you add the following cases:

- feature with no `documentationLink` 
- feature with no `feedbackLink`
- feature with no `documentationLink` nor `feedbackLink`

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * @typedef {import('./cc-feature-list.types.d.ts').Feature} Feature
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListState} FeatureListState
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoading} FeatureListStateLoading
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoaded} FeatureListStateLoaded
+ * @typedef {import('./cc-feature-list.events.js').CcFeatureValueChangeEvent} CcFeatureValueChangeEvent
+ */
+
+/**
+ * A component to display a list of feature flag and to manage their activation.
```
pdesoyres-cc: typo:
```suggestion
 * A component to display a list of feature flags and to manage their activation.
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,162 @@
...
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
+          margin-top: 1em;
+        }
+
+        .feature [slot='footer'],
+        .feature [slot='header-title'] {
+          display: flex;
+          flex-direction: row;
```
pdesoyres-cc: nitpick: useless since it's the default

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,97 @@
...
+              'This feature flag enables the new Access Logs view, allowing users to explore detailed connection and activity records for their resources.  It provides improved filtering, search capabilities, and clearer data visualization.  Activating this flag will replace the legacy logs interface with the new version currently in final version.',
+            toggleOptions: [
+              { label: 'Disable', value: 'disable' },
+              { label: 'Version A', value: 'versionA' },
+              { label: 'Version B', value: 'versionB' },
+            ],
+            toggleValue: 'versionA',
+            status: 'alpha',
+            documentationLink: 'google.com',
+            feedbackLink: 'clever-cloud.com',
```
pdesoyres-cc: nitpick: prefer using agnostic link like `https://example.com/feedback`

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,97 @@
...
+            featureDescription:
+              'This feature flag enables the new Access Logs view, allowing users to explore detailed connection and activity records for their resources.  It provides improved filtering, search capabilities, and clearer data visualization.  Activating this flag will replace the legacy logs interface with the new version currently in final version.',
+            toggleOptions: [
+              { label: 'Disable', value: 'disable' },
+              { label: 'Version A', value: 'versionA' },
+              { label: 'Version B', value: 'versionB' },
+            ],
+            toggleValue: 'versionA',
+            status: 'alpha',
+            documentationLink: 'google.com',
```
pdesoyres-cc: nitpick: prefer using agnostic link like `https://example.com/doc`

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,97 @@
...
+            featureName: 'New creation tunnel',
+            featureDescription:
+              'This feature flag enables the new creation flow currently in testing.  It introduces an improved step-by-step experience designed to simplify setup and enhance clarity throughout the process.  Activating this flag will replace the current creation tunnel with the experimental version.  Please note that this feature is still in beta and may undergo changes before its official release.',
+            toggleOptions: [
+              { label: 'Disable', value: 'disable' },
+              { label: 'Enable', value: 'enable' },
+            ],
+            toggleValue: 'enable',
+            status: 'beta',
+            documentationLink: 'google.com',
```
pdesoyres-cc: nitpick: prefer using agnostic link like `https://example.com/doc`

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,97 @@
...
+            feedbackLink: 'clever-cloud.com',
```
pdesoyres-cc: nitpick: prefer using agnostic link like `https://example.com/feedback`

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface Feature {
+  featureId: string;
+  featureName: string;
+  featureDescription: string;
+  toggleOptions: { label: string; value: string }[];
+  toggleValue: string;
+  status: FeatureStatus;
+  documentationLink: string;
+  feedbackLink: string;
+}
```
pdesoyres-cc, line 14: nitpick: renaming propositions:

```suggestion
export interface Feature {
  id: string;
  name: string;
  description: string;
  options: { label: string; value: string }[];
  value: string;
  status: FeatureStatus;
  documentationLink: string;
  feedbackLink: string;
}
```

src/translations/translations.en.js
```diff
@@ -696,6 +696,14 @@ export const translations = {
   'cc-env-var-linked-services.loading.app': /** @param {{appName: string}} _ */ ({ appName }) =>
     sanitize`Loading configuration exposed by applications linked to <strong>${appName}</strong>...`,
   //#endregion
+  //#region cc-feature-list
+  'cc-feature-list.documentation-link': `See documentation`,
+  'cc-feature-list.error': `Something went wrong while loading available experimental features.`,
```
pdesoyres-cc: nitpick: we can remove the trailing dot:

```suggestion
  'cc-feature-list.error': `Something went wrong while loading available experimental features`,
```

src/translations/translations.en.js
```diff
@@ -696,6 +696,14 @@ export const translations = {
...
+  'cc-feature-list.feedback-link': `Give your feedback`,
+  'cc-feature-list.intro': `Enable or disable features currently under development and give us your feedback`,
+  'cc-feature-list.no-data': `No experimental features are available at this time.`,
```
pdesoyres-cc: nitpick: we can remove the trailing dot:


```suggestion
  'cc-feature-list.no-data': `No experimental features are available at this time`,
```

src/translations/translations.en.js
```diff
@@ -696,6 +696,14 @@ export const translations = {
   'cc-env-var-linked-services.loading.app': /** @param {{appName: string}} _ */ ({ appName }) =>
     sanitize`Loading configuration exposed by applications linked to <strong>${appName}</strong>...`,
   //#endregion
+  //#region cc-feature-list
+  'cc-feature-list.documentation-link': `See documentation`,
+  'cc-feature-list.error': `Something went wrong while loading available experimental features.`,
+  'cc-feature-list.feedback-link': `Give your feedback`,
+  'cc-feature-list.intro': `Enable or disable features currently under development and give us your feedback`,
```
pdesoyres-cc: nitpick: we can add a trailing dot since there may have some text after:

```suggestion
  'cc-feature-list.intro': `Enable or disable features currently under development and give us your feedback.`,
```

src/translations/translations.fr.js
```diff
@@ -705,6 +705,14 @@ export const translations = {
   'cc-env-var-linked-services.loading.app': /** @param {{appName: string}} _ */ ({ appName }) =>
     sanitize`Chargement de la configuration publiée par les applications liées à <strong>${appName}</strong>…`,
   //#endregion
+  //#region cc-feature-list
+  'cc-feature-list.documentation-link': `Voir la documentation`,
+  'cc-feature-list.error': `Une erreur s'est produite lors du chargement des fonctionnalités expérimentales disponibles.`,
```
pdesoyres-cc: nitpick: we can remove the trailing dot:

```suggestion
  'cc-feature-list.error': `Une erreur s'est produite lors du chargement des fonctionnalités expérimentales disponibles`,
```

src/translations/translations.fr.js
```diff
@@ -705,6 +705,14 @@ export const translations = {
...
+  'cc-feature-list.feedback-link': `Donnez votre avis`,
+  'cc-feature-list.intro': `Activez ou désactivez les fonctionnalités actuellement en cours de développement et donnez-nous votre avis.`,
+  'cc-feature-list.no-data': `Aucune fonctionnalité expérimentale n'est disponible pour le moment.`,
```
pdesoyres-cc: nitpick: we can remove the trailing dot:


```suggestion
  'cc-feature-list.no-data': `Aucune fonctionnalité expérimentale n'est disponible pour le moment`,
```

pdesoyres-cc (changes requested):
Nice component. Well done Clara. I left some nitpicks and suggestions

src/components/cc-feature-list/cc-feature-list.js
clara-layus-cc: This is not the default because there is no gap between the elements.

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
+export type FeatureListState = FeatureListStateLoaded | FeatureListStateLoading | FeatureListStateError;
```
florian-sanders-cc, line 1: **question (need more context to provide feedback):** we have typical `loaded` vs `loading` vs `error` state but no smart component to control these 🤔 
- Do you plan on fetching data from the console codebase? 
- What is the source of your data? Settings API or local storage only?
- If data comes from the settings API, do you plan on also updating it when settings change (for instance if the user enables a feature that was initially disabled)?

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type FeatureStatus = 'alpha' | 'beta' | 'preview' | '';
```
florian-sanders-cc: **question:** the empty string in a union seems weird to me. Why do you need it / What did you mean?

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  options: { label: string; value: string }[];
+  value: string;
+  status: FeatureStatus;
+  documentationLink: string;
+  feedbackLink: string;
+}
+
+export interface FeatureListStateLoaded {
+  type: 'loaded';
+  features: Feature[];
```
florian-sanders-cc: **nitpick:** I think we try to avoid plurals and go for less confusing `list` prefix in such cases (though I'm pretty sure you can easily find counter examples where we didn't)


```suggestion
  featureList: Feature[];
```

florian-sanders-cc: (personnally i'm fine with both)

src/translations/translations.en.js
```diff
@@ -696,6 +696,14 @@ export const translations = {
   'cc-env-var-linked-services.loading.app': /** @param {{appName: string}} _ */ ({ appName }) =>
     sanitize`Loading configuration exposed by applications linked to <strong>${appName}</strong>...`,
   //#endregion
+  //#region cc-feature-list
+  'cc-feature-list.documentation-link': `See documentation`,
+  'cc-feature-list.error': `Something went wrong while loading available experimental features`,
+  'cc-feature-list.feedback-link': `Give your feedback`,
```
florian-sanders-cc: **suggestion:** (order of preferences)

```suggestion
  'cc-feature-list.feedback-link': `Share feedback`,
```

```suggestion
  'cc-feature-list.feedback-link': `Give feedback`,
```

```suggestion
  'cc-feature-list.feedback-link': `Send feedback`,
```

src/translations/translations.en.js
```diff
@@ -696,6 +696,14 @@ export const translations = {
...
+  'cc-feature-list.intro': `Enable or disable features currently under development and give us your feedback`,
```
florian-sanders-cc: **suggestion:**

```suggestion
  'cc-feature-list.intro': `Enable or disable features currently under development and share feedback`,
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+      default:
+        featureStatus = '';
+        break;
+    }
+    return html`
+      <cc-block class="feature">
+        <div slot="header-title">
+          ${feature.name}
+          ${feature.status ? html` <cc-badge intent="${featureStatus}"> ${feature.status} </cc-badge> ` : ''}
+        </div>
```
florian-sanders-cc: **fix:** we try to avoid truthy / falsy comparisons even though JS allows it (the `feature.status ? ...`) + you could move the switch out of the render function inside a private method and call it within your render when it's relevant

```suggestion
  renderFeature(feature) {
    return html`
      <cc-block class="feature">
        <div slot="header-title">
          ${feature.name}
          ${feature.status != null ? html` <cc-badge intent="${this._getStatusIntent(feature.status)}"> ${feature.status} </cc-badge> ` : ''}
        </div>
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+        <div slot="content">${feature.description}</div>
+        ${feature.documentationLink || feature.feedbackLink
```
florian-sanders-cc: **fix:** (avoid fragile truthy / falsy comparisons)


```suggestion
        ${feature.documentationLink != null || feature.feedbackLink != null
```

Or if you're afraid of empty strings


```suggestion
        ${!isStringEmpty(feature.documentationLink) || !isStringEmpty(feature.feedbackLink)
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+          ? html`
+              <div slot="footer">
+                ${feature.documentationLink
+                  ? html`
+                      <cc-link
+                        .href=${feature.documentationLink}
+                        a11y-desc="Go to documentation page"
```
florian-sanders-cc: **fix:** should be translated and should convey more info compared to the actual link text, for instance

```suggestion
                        a11y-desc="${i18n('cc-feature-list.feedback-link')} - ${feature.name}"
```

This way we'll end up with a title like: `See documentation - Access Logs - new window`

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+                        .icon=${iconInfo}
+                      >
+                        ${i18n('cc-feature-list.documentation-link')}
+                      </cc-link>
+                    `
+                  : ''}
+                ${feature.feedbackLink
```
florian-sanders-cc: **fix:**

```suggestion
                ${feature.feedbackLink  != null
```

or


```suggestion
                ${!isStringEmpty(feature.feedbackLink)
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+          </div>
+        </div>
+      </cc-block>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
```
florian-sanders-cc, line 157: **fix:** missing the host display (by default web component hosts are `display: inline` and this will cause issues when / if you component is not used as a flex / grid child).

```suggestion
      css`
        :host {
          display: block;
        }
```

src/translations/translations.en.js
florian-sanders-cc: Or


```suggestion
  'cc-feature-list.intro': `Enable or disable features currently under development and tell us what you think`,
```

(I think I prefer this one :smile: )

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+        break;
+      default:
+        featureStatus = '';
+        break;
+    }
+    return html`
+      <cc-block class="feature">
+        <div slot="header-title">
+          ${feature.name}
+          ${feature.status ? html` <cc-badge intent="${featureStatus}"> ${feature.status} </cc-badge> ` : ''}
```
florian-sanders-cc: **fix:** it may seem silly but `feature.status` needs to come from `i18n` here:
- we need it to be capitalized (could do with CSS but meh),
- maybe at some point the status could be two words separated with space,
- maybe we'll add status (like "Preview") that are not the same in English & French,
- we cannot be sure it will be the same word in languages other than English & French (can discuss if this one is unclear).

Create a translation key for each status and a `_getFeatureStatus(status)` method with a switch that returns each translation value based on the status value.

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+          gap: 0.5em;
+          margin-top: 1em;
+        }
+
+        .feature [slot='footer'],
+        .feature [slot='header-title'] {
+          display: flex;
+          flex-direction: row;
+          gap: 1em;
+        }
```
florian-sanders-cc, line 176: **suggestion:** I think you need a `flex-wrap: wrap;` (+ `row-gap: 0.5em`) to avoid this

Image

With the suggestions

Image

```suggestion
        }
        
        .feature [slot='header-title'] {
          flex-wrap: wrap;
          row-gap: 0.5em;
        }
```

florian-sanders-cc (changes requested):
Well done @clara-layus-cc the component looks good and the code is really easy to read! 😎 

I'll let you tackle my comments (mostly small fixes & nitpicks as usual 😇) but I also have a more general question:
I'm not sure we should be using `cc-block` for the features themselves. I feel like the border is a less subtle than what we have in the wireframes + the headings should not be that big / bold.

We could customize the headings (font-size & weight from your component) but for the borders I'm not sure. It depends if we think it's important to have more subtle borders or not (I do prefer with more subtle ones but let's see what others think).

The overall toast behaviour is unclear to me, I feel like toasting everytime you change the value of a setting is a lot (bear in mind that keyboard users will have to go through each option if they want to reach the last one, so toasting & updating should probably be done on blur and not on input, or we need a submit input somewhere to validate choices 🤔 )
=> The comment above might be unclear, we can discuss it in sync and it could be nice to have other opinions about it.

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+const conf = {
+  component: 'cc-feature-list',
+};
+
+/**
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoaded} FeatureListStateLoaded
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoading} FeatureListStateLoading
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateError} FeatureListStateError
```
HeleneAmouzou: **chore**: unused import

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+import '../cc-link/cc-link.js';
+import '../cc-loader/cc-loader.js';
+import '../cc-notice/cc-notice.js';
+import '../cc-toggle/cc-toggle.js';
+import { CcFeatureSettingChangeEvent } from './cc-feature-list.events.js';
+
+/**
+ * @typedef {import('./cc-feature-list.types.d.ts').Feature} Feature
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListState} FeatureListState
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoading} FeatureListStateLoading
```
HeleneAmouzou: **chore**: unused import

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+ * @typedef {import('./cc-feature-list.types.d.ts').FeatureListStateLoaded} FeatureListStateLoaded
```
HeleneAmouzou: **chore**: unused import

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+                        a11y-desc="Go to documentation page"
+                        .icon=${iconInfo}
+                      >
+                        ${i18n('cc-feature-list.documentation-link')}
+                      </cc-link>
+                    `
+                  : ''}
+                ${feature.feedbackLink
+                  ? html`
+                      <cc-link .href=${feature.feedbackLink} a11y-desc="Go to feedback page" .icon=${iconMessage}>
```
HeleneAmouzou: **fix**: `a11y-desc` should be translated (see Florian's comment above)

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,153 @@
...
+            documentationLink: 'https://example.com/doc',
+            feedbackLink: 'https://example.com/feedback',
+          },
+        ],
+      },
+    },
+  ],
+});
+
+export const links = makeStory(conf, {
```
HeleneAmouzou: **question**: I am not sure to understand the purpose of this story ? 🤔

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+      return;
+    }
+
+    this.dispatchEvent(new CcFeatureSettingChangeEvent({ featureId: feature.id, newValue }));
+  }
+
+  /**
+   * @param {Feature} feature
+   */
+  renderFeature(feature) {
```
roberttran-cc: **💣 chore:** private function such as sub-render should be prefixed with `_render` (you can check `docs/cc-example-component.js` occasionally when in doubt).

src/components/cc-feature-list/cc-feature-list.js
roberttran-cc: **💭 thought:** Probably same feedback here as for the documentation link.

src/components/cc-feature-list/cc-feature-list.js
roberttran-cc: I think Pierre is saying that `flex-direction` default value is `row`, so removing this single rule line 151 should not change anything and should not have side effect unless I'm missing something.

src/components/cc-feature-list/cc-feature-list.types.d.ts
roberttran-cc: Agree. Without additional information, I'd go with:

```suggestion
// as `preview` value seems not used
type FeatureStatus = 'alpha' | 'beta';
```

And mark optional value in the `Feature` interface:

```ts
export interface Feature {
  ...
  status?: FeatureStatus;
  ...
}
```

And finally, I would be dealing with empty string directly in the component, as it should behave the same as when a property is missing.

Let me (us) know if clarification is needed or if I (we) missed context.

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+  constructor() {
+    super();
+
+    /** @type {FeatureListState} Set the state of the component */
+    this.state = { type: 'loading' };
+  }
+
+  /**
+   * @param {Feature} feature
+   * @param {String} newValue
```
roberttran-cc: **🤓 nitpick:** [Source](https://jsdoc.app/tags-param)


```suggestion
   * @param {string} newValue
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+                    `
+                  : ''}
+              </div>
+            `
+          : ''}
+        <div slot="header-right">
+          <cc-toggle
+            .choices="${feature.options}"
+            value="${feature.value}"
+            @cc-select=${(/** @type {CcSelectEvent} */ event) => this._onToggleChange(feature, event.detail)}
```
roberttran-cc: **💣 chore:** Missing `CcSelectEvent` type import.

```js
/** @typedef {import('../common.events.js').CcSelectEvent} CcSelectEvent */
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,164 @@
...
+    if (this.state.type === 'loading') {
+      return;
+    }
+
+    this.dispatchEvent(new CcFeatureSettingChangeEvent({ featureId: feature.id, newValue }));
+  }
+
+  /**
+   * @param {Feature} feature
+   */
```
roberttran-cc, line 57: **💣 chore:** Missing return type


```suggestion
  /**
   * @param {Feature} feature
   * @return {TemplateResult}
   */
```

Note: needs type definition import as well:

```js
/** @typedef {import('lit').TemplateResult<1>} TemplateResult */
```

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface Feature {
+  id: string;
+  name: string;
+  description: string;
+  options: { label: string; value: string }[];
+  value: string;
+  status: FeatureStatus;
+  documentationLink: string;
+  feedbackLink: string;
```
roberttran-cc: **🎯 suggestion:** Should be marked as optional.


```suggestion
  documentationLink?: string;
  feedbackLink?: string;
```

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,153 @@
+import { makeStory } from '../../stories/lib/make-story.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-link/cc-link.js';
+import '../cc-toggle/cc-toggle.js';
```
roberttran-cc: **🤓 nitpick:** Imports not needed (the main component already import them).


```suggestion
```

roberttran-cc (commented):
Nice component, great job!

I've got several feedback that are quite minor and mostly details that we all tended to overlook when learning this codebase (and still today occasionally): beside that, you followed pretty neatly the way we code our component. 💪 

A note on stories: I like the default story showing the most simple case (no link & no status) and subsequent ones showing more complex cases ("With link", "With status" and "With link and status" stories for example).\
But sometimes, the simplest case is not always the most common so that should be the one exception.\
But that's really a personal preference, showing all options in the default story makes sense too.

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+type FeatureStatus = 'alpha' | 'beta' | 'preview' | '';
+
+export interface Feature {
+  id: string;
+  name: string;
+  description: string;
+  options: { label: string; value: string }[];
+  value: string;
+  status: FeatureStatus;
+  documentationLink: string;
```
Galimede: nitpick: having a documentation link seems optional, can we reflect that in the type?

src/components/cc-feature-list/cc-feature-list.types.d.ts
Galimede: nitpick: having a feedback link seems optional, can we reflect that in the type?

src/components/cc-feature-list/cc-feature-list.stories.js
clara-layus-cc: The purpose of the story is to show the different possible combinations in relation to documentation and feedback links. That is, one component with only the documentation link, another with only the feedback link, and a third with no link at all. After that, I'm not sure if the name of the story is obvious.

src/components/cc-feature-list/cc-feature-list.types.d.ts
clara-layus-cc: As part of this topic, the list of features will be listed in a file in the console and the user's values will be stored in their local storage. That's why, even though there is no smart part in this component, I created different states to cover all possibilities: a problem retrieving the list from the console file, latency in retrieving information from local storage, etc.

src/components/cc-feature-list/cc-feature-list.stories.js
HeleneAmouzou: Thanks, it makes more sense now. If you follow @roberttran-cc 's advice (which I agree with), you could then rename the story  "With link"/"With links".

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
+// @ts-nocheck
```
Galimede: question: why is there a `ts-nocheck`? 🤔

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
...
+  _getFeatureStatus(status) {
+    switch (status) {
+      case 'alpha':
+        return { text: i18n('cc-feature-list.status.alpha'), intent: 'warning' };
+      case 'beta':
+        return { text: i18n('cc-feature-list.status.beta'), intent: 'info' };
+      case 'preview':
+        return { text: i18n('cc-feature-list.status.preview'), intent: '' };
+    }
+    return undefined;
```
Galimede: nitpick: we usually tend to return `null` instead of `undefined`
note: 
you could also add a default return to the switch instead of the return at the end 
```
// ...
default: 
return null;
// ...
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * @param {Feature} feature
+   * @return {TemplateResult}
+   * @private
+   */
+  _renderFeature(feature) {
+    let featureStatus = this._getFeatureStatus(feature.status);
```
Galimede: nitpick: as we won't reassign it, you can use `const` instead

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
...
+          <div class="feature-list">
+            ${this.state.type === 'loading' ? html` <cc-loader></cc-loader>` : ''}
+            ${this.state.type === 'error'
+              ? html` <cc-notice intent="warning" message=${i18n('cc-feature-list.error')}></cc-notice> `
+              : ''}
+            ${this.state.type === 'loaded'
+              ? this.state.featureList.length > 0
+                ? this.state.featureList.map((feature) => this._renderFeature(feature))
+                : html` <cc-notice intent="info" message=${i18n('cc-feature-list.no-data')}></cc-notice> `
+              : ''}
```
Galimede: nitpick: we usually avoid to have mutliple nested conditional operators and divide the two conditions into two mustaches.
suggestion: 

```js
// We'd put this before the return of the render
// Note: maybe something like isFeatureListEmpty would be better, I let you decide the naming
const empty = this.state.featureList.length === 0;
```

```suggestion    
            ${this.state.type === 'loaded' && empty
                ? html` <cc-notice intent="info" message=${i18n('cc-feature-list.no-data')}></cc-notice> `
              : ''}
            ${this.state.type === 'loaded' && !empty
                ? this.state.featureList.map((feature) => this._renderFeature(feature))
             : ''}      
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
...
+      </div>
+    `;
+  }
+
+  render() {
+    return html`
+      <cc-block>
+        <div slot="header-title">${i18n('cc-feature-list.title')}</div>
+        <div slot="content">
+          ${this.state.type === 'loaded' && this.state.featureList.length > 0
```
Galimede: suggestion: with the suggestion below we could change this check to
```suggestion
          ${this.state.type === 'loaded' && !empty
```

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,212 @@
...
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+
+        .intro {
+          margin: 0;
+        }
```
Galimede: chore: this class seems unused 🤔

src/components/cc-feature-list/cc-feature-list.types.d.ts
```diff
@@ -0,0 +1,27 @@
+export type FeatureListState = FeatureListStateLoaded | FeatureListStateLoading | FeatureListStateError;
+
+type FeatureStatus = 'alpha' | 'beta' | 'preview';
```
roberttran-cc: **💣 chore:** Missing `export`

```suggestion
export type FeatureStatus = 'alpha' | 'beta' | 'preview';
```

src/components/cc-feature-list/cc-feature-list.stories.js
```diff
@@ -0,0 +1,160 @@
...
+            documentationLink: 'https://example.com/doc',
+            feedbackLink: 'https://example.com/feedback',
+          },
+        ],
+      },
+    },
+  ],
+});
+
+export const withoutLink = makeStory(conf, {
```
HeleneAmouzou, line 77: **question**: on this story, should not all the feature blocks have no link (no documentation link neither feedback link) ?

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -0,0 +1,209 @@
...
+                  : ''}
+              </div>
+            `
+          : ''}
+      </div>
+    `;
+  }
+
+  render() {
+    const isFeatureListEmpty = this.state.type === 'loaded' && this.state.featureList.length === 0;
```
Galimede, line 129: nitpick: you don't need the `this.state.type === 'loaded'` as you're checking this condition below

src/components/cc-feature-list/cc-feature-list.stories.js
clara-layus-cc: There should be one story without the documentation link, one story without the feedback link, and one story without any links. But I understand that the name of the story is confusing; I just can't think of a suitable name.

src/components/cc-feature-list/cc-feature-list.js
clara-layus-cc: It seems to me that because I tried a first version without it, but I have a problem with this.state.featureList because featureList only exists when the state is Loaded, so if I don't check before, there is a problem with a property that does not exist.

Galimede: Ah yes, I understand why you had to do that, looks fine to me then :wink:

PR #1600, opened 2025-10-30
fix(cc-addon-header.smart-kubernetes): set provider name in status message

florian-sanders-cc:
## What does this PR do?

When reworking the deployment status messages to remove the `add-on` mention, I accidentally used `productId` instead of `providerId` as i18n key param.

This results in the following message in French:

> Votre instance undefined est disponible !

This PR fixes that :see_no_evil: 

## How to review?

- No review necessary, this is a quick fix for the release


pdesoyres-cc (approved):
LGTM

PR #1605, opened 2025-11-13
fix(cc-block-details): improve focus outline and button transition

### LINKED ISSUE #1547: cc-block-details: missing focus outline style for the Command line button
## Context

The `Command line` button has no focus styling:

Black outline surrouding the Command line button

We need to add `:focus-visible` & use the `--cc-focus-outline` & `--cc-focus-outline-offset`.

Galimede:
Fixes #1547

## What does this PR do?                   
 
* Adds a focus-visible outline to the button,                                                                                                    
* Moves the transition from the button to its hover state as it was transitioning when in focus which we don't want.                                                                                                                                                                            
 
Question: transitioning the focus from square to rounded feels weird to me, should we round it too beforehand?

## How to review?

* Check the commit,
* Check the preview,
* One or two reviewers is enough.

### DIFF
--- a/src/components/cc-block-details/cc-block-details.js
+++ b/src/components/cc-block-details/cc-block-details.js
@@ -36,10 +36,12 @@ export class CcBlockDetails extends LitElement {
   render() {
     return html`
       <div class="wrapper">
-        <button class="button" aria-expanded="${this.isOpen}" aria-controls="content" @click=${this._onClickToggle}>
-          <slot name="button-text"></slot>
-          <cc-icon .icon="${iconArrowDown}"></cc-icon>
-        </button>
+        <div class="btn-wrapper">
+          <button class="button" aria-expanded="${this.isOpen}" aria-controls="content" @click=${this._onClickToggle}>
+            <slot name="button-text"></slot>
+            <cc-icon .icon="${iconArrowDown}"></cc-icon>
+          </button>
+        </div>
         <div id="content" class="content">
           <slot name="content"></slot>
         </div>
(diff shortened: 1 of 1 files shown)


clara-layus-cc (approved):
Well done !

As for the focus, I think it would indeed be a good idea to smooth things over from the outset to avoid any weird changes.

src/components/cc-block-details/cc-block-details.js
```diff
@@ -77,27 +79,42 @@ export class CcBlockDetails extends LitElement {
...
           margin-bottom: 0;
         }
 
         .button:hover {
           border: solid 1px var(--cc-color-border-neutral-strong, #8c8c8c);
-          border-radius: var(--cc-border-radius-default, 0.25em);
+        }
+
+        .button:focus-visible {
+          border: none;
```
florian-sanders-cc: **fix/nitpick:** setting `border: none` here triggers a very small layout shift, do we really need this rule?

src/components/cc-block-details/cc-block-details.js
```diff
@@ -77,27 +79,42 @@ export class CcBlockDetails extends LitElement {
...
-          transition: all 0.3s;
+          position: relative;
+          z-index: 1;
         }
 
         :host([is-open]) .button {
-          background-color: var(--cc-color-bg-default, #fff);
           border-radius: var(--cc-border-radius-default, 0.25em) var(--cc-border-radius-default, 0.25em) 0 0;
           margin-bottom: 0;
         }
```
florian-sanders-cc: **fix:** depending on what you've decided to do, you may remove this whole part


```suggestion
```

florian-sanders-cc (changes requested):

PR #1623, opened 2025-12-03
cc-dialog: create components and use them

### LINKED ISSUE #1577: feat: create cc-dialog component with confirmation template pattern
## Context

We have several components using native `<dialog>` elements with similar patterns:
- `cc-addon-admin`: deletion confirmation dialog with text input validation
- `cc-addon-info`: version selection dialog
- `cc-addon-backups`: custom overlay (should migrate to native dialog)

Currently each component reimplements dialog logic (open/close, form handling, waiting states, etc.).

Related: #427 (RFC: modal/dialog system)

## Proposal

Create a **two-part solution**:

### 1. Base `cc-dialog` Component

A flexible, low-level dialog wrapper around native `<dialog>` element:

**Features:**
- Native `<dialog>` element with `closedby="any"`
- Declarative (`open` prop) + Imperative API (`showModal()`, `close()`)
- Slots: `default`, `heading`, `content`, `actions`, maybe `form`
- Close button (top-right) with waiting state support (not sure, but we could disable it when in waiting mode)
- `waiting` prop that disables close button and dims content (not sure we need that, let's see)
- Events: `cc-dialog-close`
- Responsive design, backdrop with blur
- Accessibility: aria-labelledby, keyboard navigation

**API:**
```js
html`
  <cc-dialog ?waiting="${isLoading}">
    <div slot="heading">Delete Addon</div>
    
    <p slot="content">This action cannot be undone.</p>
    
    <div slot="actions">
      <cc-button outlined @cc-click=${close}>Cancel</cc-button>
      <cc-button danger>Delete</cc-button>
    </div>
  </cc-dialog>
`;
```

**Properties:**
- `open` (Boolean, reflected) - Declarative open/close control
- `waiting` (Boolean) - Disables close button and dims content during async operations

**Methods:**
- `show()` - Opens dialog as modal (recommended)
- `hide()` - Closes dialog

**Events:**
- `cc-dialog-hide` - Fired when dialog closes

### 2. Confirmation Template Function

A reusable template for the common "type text to confirm" pattern:

**Location:** `src/templates/dialog-confirm-input.js`

**Usage:**
```js
import { dialogConfirmInput } from '../../templates/dialog-confirm-input.js';

html`
  <cc-dialog ?waiting=${isDeleting} ${ref(this._dialogRef)}>
    <div slot="heading">Delete addon</div>
    
    <p>This action cannot be undone.</p>
    
    ${dialogConfirmInput({
      confirmText: addonName,
      inputLabel: 'Type addon name to confirm',
      mismatchErrorMessage: 'Name does not match',
      onConfirm: this._onDelete.bind(this),
      onCancel: this._onCancel.bind(this),
      waiting: isDeleting,
      submitLabel: 'Delete',
      cancelLabel: 'Cancel',
    })}
  </cc-dialog>
`;
```

**What the template factorizes:**
- Text confirmation validator (value must match `confirmText`)
- `formSubmit` directive wiring
- `cc-input-text` field with help text
- Button layout (cancel + submit) with waiting/disabled states
- Form structure slotted into `content` + `actions`

**What consumers still control:**
- Dialog heading and description content
- When to open/close the dialog
- What happens on confirm (via callback)
- State management (waiting, refs, reset)

## Benefits

✅ **Flexible base component** - Slot-based API for any dialog scenario  
✅ **Factorized common pattern** - Template function for text confirmation  
✅ **No component bloat** - Template is just a function, not a component  
✅ **Backward compatible** - Existing dialogs can migrate gradually  
✅ **Composable** - Mix custom content with template functions  
✅ **Clear API** - Slots make structure obvious  

## Implementation Plan

- [ ] Create `cc-dialog` base component
  - [ ] Component structure with slots (heading, default, actions)
  - [ ] Declarative (`open` prop) + imperative APIs (`show`, `showModal`, `close`)
  - [ ] Close button with `waiting` state support
  - [ ] Waiting state handling (disable close, dim content)
  - [ ] Events: `cc-dialog-close`, `cc-dialog-cancel`
  - [ ] Accessibility: aria-labelledby, keyboard navigation, focus management
  - [ ] Responsive styles, backdrop with blur
- [ ] Create `dialogConfirmInput` template function
  - [ ] Confirmation validator (text matching)
  - [ ] Form submit directive integration
  - [ ] Template implementation with config object
- [ ] Documentation
  - [ ] Storybook stories for base `cc-dialog` (various examples)
  - [ ] Storybook story for `dialogConfirmInput` template
  - [ ] Usage examples and patterns
  - [ ] Migration guide for existing dialogs
- [ ] Migrate existing usages
  - [ ] `cc-addon-admin` deletion dialog
  - [ ] `cc-addon-info` version selection dialog
  - [ ] Consider `cc-addon-backups` (migrate from custom overlay to native dialog)

## Design Decisions

**Why slot-based API instead of props?**
- More flexible: can pass complex HTML, not just strings
- Clearer component structure
- Consistent with Lit/web component best practices
- Easier to style and customize per use case

**Why template function vs separate component?**
- Provides factorization without loss of flexibility
- No component nesting complexity
- Easier to debug (just a function returning HTML)
- Allows mixing custom content with factorized patterns
- Consumer retains full control over dialog lifecycle

**Why keep base dialog simple?**
- Single responsibility: wrap native `<dialog>` with common UX patterns
- Consumers control content/forms/validation via slots
- Easier to test and maintain
- More flexible for future use cases we haven't thought of

## Open Questions (asked Claude to challenge me)

1. Should we create additional template variants beyond text confirmation?
   - `dialogConfirmSelect` - confirmation with dropdown selection => No
   - `dialogConfirmButtons` - simple yes/no without input => No
2. Should `LostFocusController` be integrated or documented as consumer responsibility? (Leaning towards: consumer responsibility, as it needs to be in the parent component) => No
3. Do we need a `size` prop for different dialog widths, or handle via CSS custom properties? => Maybe but that would be a max-width anyway?

## Notes

- `LostFocusController` should remain in consumer components (won't work with dialog as host)
- Form validation handled by consumers using `formSubmit` directive (or via template function)
- Dialog uses native `<dialog>` element (no polyfill needed, baseline browser support is good)

florian-sanders-cc:
Fixes #1577 

## What does this PR do?

- Adds the `autofocus` attr / prop to each form control component [EDIT: this feature only works in Chrome right now, I'll fix it after the first review phase is done],
- Adds a `cc-dialog` component: 
  - not opinionated, mostly here to factorize the native `dialog` lifecycle and offer better Developer Experience,
- Adds a `cc-dialog-confirm-actions` component:
  - fairly opinionated, factorize the use of `cancel` & confirm buttons at the bottom of a `cc-dialog` component,
  - apart from the (opinionated) layout & responsive aspect, it also handles the whole waiting dance (disabled cancel + waiting confirm).
- Adds a `cc-dialog-confirm-form` component:
  - relies on `cc-dialog-confirm-actions` for the buttons,
  - factorizes form related stuff,
  - fairly opinionated as well (was designed with specific usecases and layout in mind to avoid repeating code).
- Uses these components where relevant:
  - `cc-addon-admin`: replace the native `dialog` with a `cc-dialog` + `cc-dialog-confirm-form`,
  - `cc-addon-info`: replace the native `dialog` with a `cc-dialog` + `cc-dialog-confirm-actions`,
  - `cc-addon-backups`: replace the custom overlays with a proper `cc-dialog` component,
  - `cc-domain-management`: adds a `cc-dialog` + `cc-dialog-confirm-form`,
    - Note: this means the console should stop plugging its own dialog system into this component once released.
    - This is not officially a breaking change because the console was never supposed to do that in the first place, it was always meant to be a temporary solution / hack.
- Updates the `lost-focus-controller`:
  - The current logic of `lost-focus-controller` is: if an element from the list I'm watching had focus and was removed, then I need to react,
  - When you put a dialog to confirm the deletion, the element being focused when the removal happens is actually the button within the dialog and not the button within the list,
  - So we adapt the logic to make it work in these scenarios while keeping the same API overall,
  - Stories & docs should explain this properly but feel free to ping me if it's unclear or if it fails somehow.

## How to review?

- Check the [stories](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-dialog/init/index.html?path=/docs/%F0%9F%A7%AC-atoms-cc-dialog--docs) (3 new components + `cc-addon-admin`, `cc-addon-info`, `cc-domain-management`,
- Check the commits,
- Check the components where the `cc-dialog` is used in demo-smart and make sure they still work (+ make sure focus is handled correctly).
- There are a few subjects you could also focus on specifically:
  - `autofocus`: is it a good idea? Should we remove this or on the contrary offer more guidance?
  - CSS props and overall customization of the `cc-dialog`: is it enough or too much? Should stuff be renamed?
  - The `cc-dialog-confirm-form` component reacts to its parent `cc-dialog` being closed (resets the form): it's part of the contract that any `cc-dialog-confirm-form` must be used within a `cc-dialog` but is it enough to document it or should we throw errors if not? Should we design this another way (for instance from the `cc-dialog` instead)? Or should we let people using the confirm form reset it themselves everytime the dialog closes?
  - Margins & paddings inside the `cc-dialog` & other dialog components: I followed what we already had in the different components using the native `dialog` but maybe these spacings do not fit all cases (for instance, maybe the margin-top should be reduced when there is no form and only buttons :man_shrugging:).

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -31,6 +31,9 @@ import '../cc-block-section/cc-block-section.js';
 import '../cc-block/cc-block.js';
 import '../cc-button/cc-button.js';
 import '../cc-code/cc-code.js';
+import '../cc-dialog-confirm-form/cc-dialog-confirm-form.js';
+import '../cc-dialog/cc-dialog.js';
+import '../cc-icon/cc-icon.js';
 import '../cc-input-text/cc-input-text.js';
 import '../cc-link/cc-link.js';
 import '../cc-loader/cc-loader.js';
@@ -44,7 +47,7 @@ const TLS_CERTIFICATES_DOCUMENTATION = getDocUrl('/administrate/ssl');
 const DNS_DOCUMENTATION = getDocUrl('/administrate/domain-names');
 
 /**
- * @import { DomainManagementDnsInfoState, DomainManagementListState, DomainManagementFormState, DomainManagementFormStateIdle, FormattedDomainInfo, DomainInfo, FormError } from './cc-domain-management.types.js'
+ * @import { DomainManagementDnsInfoState, DomainManagementListState, DomainManagementFormState, DomainManagementFormStateIdle, FormattedDomainInfo, DomainInfo, FormError, DomainState } from './cc-domain-management.types.js'
  * @import { CcInputText } from '../cc-input-text/cc-input-text.js'
  * @import { TemplateResult, PropertyValues } from 'lit'
  * @import { Ref } from 'lit/directives/ref.js'
@@ -62,6 +65,7 @@ export class CcDomainManagement extends LitElement {
       dnsInfoState: { type: Object, attribute: 'dns-info-state' },
       domainFormState: { type: Object, attribute: 'domain-form-state' },
       domainListState: { type: Object, attribute: 'domain-list-state' },
+      _idOfDomainToDelete: { type: String, state: true },
       _sortedDomains: { type: Array, state: true },
     };
   }
@@ -98,6 +102,9 @@ export class CcDomainManagement extends LitElement {
 
     /** @type {Ref<CcInputText>} */
     this._domainInputRef = createRef();
+
+    /** @type {string|null} Used to display the deletion dialog when set with a value other than null. */
+    this._idOfDomainToDelete = null;
 
     /** @type {Ref<HTMLParagraphElement>} */
     this._emptyMessageRef = createRef();
@@ -248,17 +255,9 @@ export class CcDomainManagement extends LitElement {
     return () => this.dispatchEvent(new CcDomainMarkAsPrimaryEvent(domainInfo));
   }
 
-  /** @param {FormattedDomainInfo} formattedDomainInfo */
-  _onDelete(formattedDomainInfo) {
-    /** @type {DomainInfo} */
-    const domainInfo = {
-      id: formattedDomainInfo.id,
-      hostname: formattedDomainInfo.hostname,
-      pathPrefix: formattedDomainInfo.pathPrefix,
-      isWildcard: formattedDomainInfo.isWildcard,
-      isPrimary: formattedDomainInfo.isPrimary,
-    };
-    return () => this.dispatchEvent(new CcDomainDeleteEvent(domainInfo));
+  /** @param {string} domainId */
+  _onDeleteRequest(domainId) {
+    return () => (this._idOfDomainToDelete = domainId);
   }
 
   /** @param {CcInputEvent} event */
@@ -319,6 +318,11 @@ export class CcDomainManagement extends LitElement {
   }
 
   render() {
+    const domainToDelete =
+      this.domainListState.type === 'loaded' && this._idOfDomainToDelete != null
+        ? this.domainListState.domains.find((domain) => domain.id === this._idOfDomainToDelete)
+        : null;
+
     return html`
       <div class="wrapper">
         <cc-block>
@@ -378,6 +382,8 @@ export class CcDomainManagement extends LitElement {
             </div>
           </cc-block-details>
         </cc-block>
+
+        ${this._renderDeleteDomainDialog(domainToDelete)}
 
         <cc-block>
           <div slot="header-title">${i18n('cc-domain-management.certif.heading')}</div>
@@ -529,6 +535,7 @@ export class CcDomainManagement extends LitElement {
   _renderDomain(domainInfo, isOneRowMarkingPrimary) {
     const {
       type: domainItemStateType,
+      id,
       hostname,
       pathPrefix,
       isWildcard,
@@ -626,7 +633,7 @@ export class CcDomainManagement extends LitElement {
             .icon=${iconDelete}
             hide-text
             circle
-            @cc-click=${this._onDelete(domainInfo)}
+            @cc-click=${this._onDeleteRequest(id)}
           >
             ${i18n('cc-domain-management.list.btn.delete.text')}
           </cc-button>
@@ -667,6 +674,56 @@ export class CcDomainManagement extends LitElement {
           ></cc-input-text>
         </div>
       </cc-block-section>
+    `;
+  }
+
+  _onDeleteDialogClose() {
+    this._idOfDomainToDelete = null;
+  }
+
+  /** @param {DomainState} domainToDelete */
+  _onDeleteConfirm(domainToDelete) {
+    if (this.domainListState.type !== 'loaded' || domainToDelete == null) {
+      return;
+    }
+
+    const domainInfo = {
+      id: domainToDelete.id,
+      hostname: domainToDelete.hostname,
+      pathPrefix: domainToDelete.pathPrefix,
+      isWildcard: domainToDelete.isWildcard,
+      isPrimary: domainToDelete.isPrimary,
+    };
+
+    this.dispatchEvent(new CcDomainDeleteEvent(domainInfo));
+  }
+
+  /** @param {DomainState|null} domainToDelete */
+  _renderDeleteDomainDialog(domainToDelete) {
+    return html`
+      <cc-dialog
+        ?open="${domainToDelete != null}"
+        heading="${i18n('cc-domain-management.delete-dialog.heading')}"
+        @cc-close="${this._onDeleteDialogClose}"
+      >
+        ${domainToDelete != null
+          ? html`
+              <p>${i18n('cc-domain-management.delete-dialog.desc')}</p>
+              <cc-dialog-confirm-form
+                submit-label="${i18n('cc-domain-management.delete-dialog.confirm-button')}"
+                submit-intent="danger"
+                confirm-text-to-input="${getHostWithWildcard(
+                  domainToDelete.hostname + domainToDelete.pathPrefix,
+                  domainToDelete.isWildcard,
+                )}"
+                confirm-input-label="${i18n('cc-domain-management.delete-dialog.input-label')}"
+                ?waiting="${domainToDelete.type === 'deleting'}"
+                @cc-confirm="${() => this._onDeleteConfirm(domainToDelete)}"
+              >
+              </cc-dialog-confirm-form>
+            `
+          : ''}
+      </cc-dialog>
     `;
   }
 
--- a/src/translations/translations.en.js
+++ b/src/translations/translations.en.js
@@ -459,6 +459,10 @@ export const translations = {
   'cc-domain-management.certif.custom': () =>
     sanitize`You can provide your own certificate by using the <cc-link href="https://api.clever-cloud.com/v2/certificates/new">Clever Cloud Certificate Manager</cc-link>.`,
   'cc-domain-management.certif.heading': `Secure your application`,
+  'cc-domain-management.delete-dialog.confirm-button': `Delete`,
+  'cc-domain-management.delete-dialog.desc': `By deleting this domain name, your application will immediatly become unreachable from that specific domain name.`,
+  'cc-domain-management.delete-dialog.heading': `Remove the domain name`,
+  'cc-domain-management.delete-dialog.input-label': `Enter the domain name to confirm its deletion`,
   'cc-domain-management.dns.a.desc': () =>
     sanitize`<p>If you choose to use <code>A</code> records, for instance with a root domain (APEX), you'll need to update them yourself. Follow our <cc-link href="${getDevHubUrl('/changelog')}">changelog</cc-link> or check our <cc-link href="${getDevHubUrl('/api/v4/#load-balancers')}">v4 API documentation</cc-link> for this.</p>`,
   'cc-domain-management.dns.a.heading': `A records`,
--- a/src/translations/translations.fr.js
+++ b/src/translations/translations.fr.js
@@ -470,6 +470,10 @@ export const translations = {
   'cc-domain-management.certif.custom': () =>
     sanitize`Vous pouvez fournir votre propre certificat grâce au <cc-link href="https://api.clever-cloud.com/v2/certificates/new">gestionnaire de certificats Clever Cloud</cc-link>.`,
   'cc-domain-management.certif.heading': `Sécurisez votre application`,
+  'cc-domain-management.delete-dialog.confirm-button': `Supprimer`,
+  'cc-domain-management.delete-dialog.desc': `Supprimer ce nom de domain rendra immédiatement votre application inaccessible depuis ce dernier.`,
+  'cc-domain-management.delete-dialog.heading': `Supprimer le nom de domaine`,
+  'cc-domain-management.delete-dialog.input-label': `Saisissez le nom de domaine pour confirmer la suppression`,
   'cc-domain-management.dns.a.desc': () =>
     sanitize`<p>Si vous choisissez d'utiliser des enregistrements de type <code>A</code>, par exemple pour un domaine racine (APEX), vous devrez vous-même assurer leur mise à jour. Pensez à suivre notre <cc-link href="${getDevHubUrl('/changelog')}" lang="en">changelog</cc-link> ou à lire la documentation de notre <cc-link href="${getDevHubUrl('/api/v4/#load-balancers')}" lang="en">API v4</cc-link> pour cela.</p>`,
   'cc-domain-management.dns.a.heading': `Enregistrements A`,


src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+ * @cssprop {Width} --cc-dialog-width - Sets the value of the width CSS property (defaults: `38em`). Note that the dialog width will never exceed 80% of the viewport width. If you need to override this, use the `dialog` CSS part.
+ */
+export class CcDialog extends LitElement {
+  static get properties() {
+    return {
+      closedBy: { type: String, attribute: 'closed-by' },
+      heading: { type: String },
+      headingIcon: { type: Object, attribute: 'heading-icon' },
+      headingIconA11yName: { type: String, attribute: 'heading-icon-a11y-name' },
+      hiddenCloseButton: { type: Boolean, attribute: 'hidden-close-button' },
```
pdesoyres-cc, line 46: question: why not `hideCloseButton` like in `<cc-button>`: `hideText`?

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+      hiddenHeading: { type: Boolean, attribute: 'hidden-heading' },
```
pdesoyres-cc, line 47: question: why not `hideHeading` like in `<cc-button>`: `hideText`?

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+
+    if (isClosing) {
+      this._dialogRef.value?.close();
+      this._tryToFocusOpenerElement();
+      this.dispatchEvent(new CcDialogCloseEvent());
+    }
+
+    if (isOpening) {
+      this._openerElement = findActiveElement();
+      this._dialogRef.value?.showModal();
```
pdesoyres-cc, line 106: we could also dispatch an event when opening, it may be useful

src/components/cc-dialog/cc-dialog.events.js
```diff
@@ -0,0 +1,42 @@
+import { CcEvent } from '../../lib/events.js';
+
+/**
+ * Dispatched when the dialog is closed.
+ *
+ * @extends {CcEvent}
+ */
+export class CcDialogCloseEvent extends CcEvent {
+  static TYPE = 'cc-dialog-close';
```
pdesoyres-cc: it seems to be a very generic event. We could move it to common events and rename it to `cc-close` (the future `cc-drawer` will need this kind of event)

src/components/cc-dialog/cc-dialog.events.js
```diff
@@ -0,0 +1,42 @@
...
+    super(CcDialogCloseRequestEvent.TYPE);
+  }
+}
+
+/**
+ * Dispatched when the focus is lost.
+ *
+ * @extends {CcEvent<Element>}
+ */
+export class CcDialogFocusRestorationFail extends CcEvent {
```
pdesoyres-cc: it could be a common event, named 'focus-lost' (the future `cc-grid` will need this kind of event)

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        ::backdrop {
+          background: rgb(30 30 30 / 55%);
+        }
+
+        @supports (backdrop-filter: blur(5px)) {
+          ::backdrop {
+            backdrop-filter: blur(5px);
+            background: rgb(30 30 30 / 35%);
```
pdesoyres-cc: question: do we need tokens for that (the future `cc-drawer` may need the same values)

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+
+          --cc-icon-size: 1.5em;
+        }
+
+        .dialog-heading-wrapper cc-icon,
+        .dialog-close cc-icon {
+          flex: 0 0 auto;
+        }
+
+        .dialog-close:focus-visible {
```
pdesoyres-cc, line 256: maybe we want the focus ring to be drawn when opening the dialog and when focus is set to the close button ? we may want to use :focus pseudo-class instead.

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+
+  static get styles() {
+    return [
+      accessibilityStyles,
+      css`
+        :host {
+          display: contents;
+
+          --cc-dialog-padding-xl: 4em;
+          --cc-dialog-padding-sm: 2em;
```
pdesoyres-cc, line 199: I find margins too big but this is only a matter of taste.

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+                <cc-icon
+                  class="dialog-heading-icon"
+                  .icon="${this.headingIcon}"
+                  a11y-name="${ifDefined(this.headingIconA11yName)}"
+                ></cc-icon>
+              `
+            : ''}
+          ${isStringEmpty(this.heading) ? html`<slot name="heading" id="dialog-heading"></slot>` : ''}
+          ${!isStringEmpty(this.heading) ? html` <span id="dialog-heading">${this.heading}</span>` : ''}
+        </div>
```
pdesoyres-cc, line 183: I would have put the heading and the close button on the same "line"

src/components/cc-dialog/cc-dialog.stories.js
```diff
@@ -0,0 +1,239 @@
...
+          </div>
+          <p>Content goes in the default slot.</p>
+        </cc-dialog>
+      `,
+      container,
+    );
+  },
+});
+
+export const longContentAndScroll = makeStory(conf, {
```
pdesoyres-cc, line 146: maybe we could have make only the content to scroll and the heading and close button to stay always visible

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
```diff
@@ -0,0 +1,184 @@
...
+  connectedCallback() {
+    this._parentCcDialog = this.closest('cc-dialog');
+    if (this._parentCcDialog instanceof CcDialog) {
+      this._dialogCloseHandler = new EventHandler(
+        this._parentCcDialog,
+        'cc-dialog-close',
+        this._onParentDialogClose.bind(this),
+      );
+      this._dialogCloseHandler.connect();
+    }
```
pdesoyres-cc, line 116: you can maybe add a warning in an else to warn developer of wrong usage

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
```diff
@@ -0,0 +1,184 @@
...
+
+    this._customErrorMessages = {
+      'no-match': () =>
+        this.confirmErrorMessage ?? i18n('cc-dialog-confirm-form.error', { name: this.confirmTextToInput }),
+    };
+
+    /** @type {HTMLFormElementRef} */
+    this._formRef = createRef();
+
+    this._dialogCloseHandler = null;
```
pdesoyres-cc, line 99: question: you don't need type here ?

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
```diff
@@ -0,0 +1,184 @@
...
+
+    this._parentCcDialog = null;
```
pdesoyres-cc, line 102: question: you don't need type here ?

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
```diff
@@ -0,0 +1,184 @@
...
+    return html`
+      <form ${formSubmit(this._onConfirm.bind(this))} ${ref(this._formRef)}>
+        <div class="dialog-content">
+          <cc-input-text
+            label="${this.confirmInputLabel}"
+            name="confirmation-input"
+            required
+            .customValidator="${this._confirmValidator}"
+            .customErrorMessages="${this._customErrorMessages}"
+            ?autofocus="${this.autofocusInput}"
```
pdesoyres-cc, line 158: autofocus doesn't work on firefox

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: Nice catch, we actually use both forms in the components:
- all form controls use `hiddenLabel` / `hidden-label`,
- `cc-toggle`, `cc-popover` and `cc-button` use `hideText`.

Personnally, for a component prop, I think `hidden...` is better because:
- props are usually names and not verbs, they describe the state of the component (while events are usually verbs) but there are probably exceptions to this,
- HTML already has a `hidden` attribute so it feels more familiar to me?

This is a personal opinion, I guess we could simply open a quick win issue (though it would lead to breaking changes) to discuss and standardize this naming?

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: I don't really understand your feedback, do you mean that as a sighted mouse user you want to see the focus ring on the close button when the dialog opens? :thinking:

pdesoyres-cc: yes, that's it

florian-sanders-cc: As a user or as a developer? Sorry for the pushback on this one but I've had the opposite feedback from many people for years so it kinda suprises me. 
I get that you wouldn't mind seeing the focus ring (I wouldn't mind either), but I don't really understand why you would need to see it?

I'll leave the comment open to see if we get other opinions about that but I anticipate opposite feedback from users so I'm leaning towards keeping it (in fact `:focus-visible` has been created exactly for this kind of usecase, focus set upon dialog opening is one the typical usecase that led to its creation).

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,268 @@
...
+import '../cc-button/cc-button.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-input-text/cc-input-text.js';
+import { CcDialogCloseEvent, CcDialogFocusRestorationFail } from './cc-dialog.events.js';
+
+/**
+ * @typedef {import('../common.types.d.ts').IconModel} IconModel
+ * @typedef {import('lit').PropertyValues<CcDialog>} CcDialogPropertyValues
+ * @typedef {import('lit/directives/ref.js').Ref<HTMLDialogElement>} HTMLDialogElementRef
+ * @typedef {import('../../lib/form/validation.js').Validator} Validator
```
HeleneAmouzou: *nitpick*: this import seem unused

src/components/cc-dialog/cc-dialog.stories.js
```diff
@@ -0,0 +1,239 @@
...
+
+export default {
+  tags: ['autodocs'],
+  title: '🧬 Atoms/<cc-dialog>',
+  component: 'cc-dialog',
+};
+
+/**
+ * @typedef {import('./cc-dialog.js').CcDialog} CcDialog
+ * @typedef {import('../cc-dialog-confirm-form/cc-dialog-confirm-form.js').CcDialogConfirmForm} CcDialogConfirmForm
```
HeleneAmouzou: **nitpick**: this import seem unused

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
florian-sanders-cc: it's a mistake! Thanks :+1:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
florian-sanders-cc: it's a mistake! Thanks :+1:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.stories.js
```diff
@@ -0,0 +1,266 @@
...
+          >
+          </cc-dialog-confirm-form>
+        </cc-dialog>
+      `,
+      container,
+    );
+  },
+});
+
+export const autofocusInput = makeStory(conf, {
```
HeleneAmouzou, line 73: **fix**:  I have checked this story (on firefox) and the input field is not  focused when the dialog opens, the focus is on the closing button

HeleneAmouzou (approved):
GG @florian-sanders-cc !! 👏 
I have only left nitpicks
I have checked the stories and it seems that the autofocus on `cc-dialog-confirm-form` might not work as planed on the `autofocusInput` story
I have check the impacted components on `demo-smart` and everything works perfectly

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: I did that at first, then I thought it didn't make sense because for the dialog to open, the parent has to switch the prop or call the `show` method so it felt redundant?

But your comment makes me reconsider it, it could still be useful in some cases and should be provided, I'll do it :+1:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
florian-sanders-cc: And not on WebKit either, it doesn't work with a shadow root in the middle which I had not tested properly :pensive: 
Thanks for the catch, I have a solution but I'll wait for other reviews first to avoid changing the code, I'll ping you once I've pushed it :+1:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.stories.js
florian-sanders-cc: Nice catch, it doesn't work on Firefox or WebKit when there's a shadow root in the middle. My bad for not testing that properly. I'll push a solution for that once others have had time to review and I'll ping you for confirmation :+1:

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,267 @@
...
+        part="dialog"
+        ${ref(this._dialogRef)}
+        @cancel="${this._onDialogClose}"
+        @cc-dialog-close-request="${this._onDialogClose}"
+      >
+        ${!this.hiddenCloseButton
+          ? html`
+              <button class="dialog-close" @click="${this._onDialogClose}">
+                <span class="visually-hidden">${i18n('cc-dialog.close')}</span>
+                <cc-icon .icon="${iconClose}"></cc-icon>
```
roberttran-cc: **🎯 suggestion:** consider using `size` property instead of CSS custom property L233.

```suggestion
                <cc-icon .icon="${iconClose}" size="lg"></cc-icon>
```

src/components/cc-dialog/cc-dialog.js
roberttran-cc: I had the same feedback from (sighted mouse) users as well in previous projects, not understanding why there was a focus ring (although it was with the default browser style, so not as nice).

So I'd be inclined to keep `:focus-visible` too.

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * A modal dialog component with optional heading, icon, and close button.
+ * Can be opened or closed programmatically or by user action.
+ *
+ * If both the `heading` prop and the `heading` slot are set, the prop takes precedence.
+ *
+ * @cssdisplay contents
+ *
+ * @slot Default Dialog content
```
Galimede: nitpick: there's no separating backtick here, also the `D` of `Default` starts with an uppercase whereas it's not the case for `heading` below. Or maybe am I missing something here?
```suggestion
 * @slot Default - Dialog content
```

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,267 @@
...
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {'any'|'closerequest'|'none'} Indicates which action closes the dialog (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy for more information) */
+    this.closedBy = 'any';
+
+    /** @type {string|null} Heading text displayed at the top of the dialog.
```
Galimede: nitpick: I'd break a line here, moreover it makes the type not highlighted on some IDEs by being here

src/components/cc-dialog/cc-dialog.js
```diff
@@ -0,0 +1,267 @@
...
+    /** @type {boolean} Hides the close button when true */
+    this.hiddenCloseButton = false;
+
+    /** @type {boolean} Hides the heading when true. Even when hidden, the heading is used as the dialog accessible name. */
+    this.hiddenHeading = false;
+
+    /** @type {IconModel|null} Sets the icon before the heading using a `<cc-icon>`. Icon is hidden if nullish. */
+    this.headingIcon = null;
+
+    /** @type {string|null} Sets the a11y name for the icon displayed next to the heading.
```
Galimede: nitpick: I'd break a line here, moreover it makes the type not highlighted on some IDEs by being here

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
```diff
@@ -0,0 +1,184 @@
...
+      submitIntent: { type: String, attribute: 'submit-intent' },
+      submitLabel: { type: String, attribute: 'submit-label' },
+      waiting: { type: Boolean, reflect: true },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {boolean} Automatically focuses the input when the dialog is opened.
```
Galimede: nitpick: I'd break a line here, moreover it makes the type not highlighted on some IDEs by being here

Galimede (approved):
Hey Florian, I don't have much to say except some nitpicks, GG. :muscle: 

I'm wondering, how we will proceed with the different closing options, as using them inconsistently might result in confusing users. :thinking: 

Also, I'm wondering what happens if we have a form modal, submit and close instantly the modal, should it be possible, should we allow it?

On another note, huge praise for the tests. :pray:

roberttran-cc (approved):
Great job, I have nothing much to say, you went up and beyond on this topic. 😍

About your questions:
- autofocus: I think it will please end users and the guidance is enough
- CSS props and overall customization of the cc-dialog: I like the CSS custom props and part combo, it offers good flexibility and it's worth the complexity tradeoff for a low-level dialog in my opinion
- cc-dialog-confirm-form contract: I think it's okay as is ; it works, the usage is clear with the documentation and the automatic reset does not leverage too much complexity (if a redesign is needed, maybe leave it for another PR, as I have no good idea right now)
- margins & paddings inside the cc-dialog & other dialog components: it's okay for now.

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: Tracked here: https://github.com/CleverCloud/clever-components/issues/1628

Will be handled separately if needed :+1:

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: Discussed with Marion & Bob, we reduced the margins in desktop a little bit.

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: I'll stick with `:focus-visible` for now :+1:

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: Discussed with Marion and Bob but we chose to stick with the current design for now :+1:

src/components/cc-dialog/cc-dialog.stories.js
florian-sanders-cc: Discussed with Bob & Marion, we decided to make the close button scroll with the content but not the heading :+1:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
florian-sanders-cc: Fixed it though the type doesn't bring much here, we'll probably need to improve the `EventHandler` type at some point.

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.js
florian-sanders-cc: Pushed the fix in a separate commit, it's a shame it's not handled natively :disappointed:

src/components/cc-dialog-confirm-form/cc-dialog-confirm-form.stories.js
florian-sanders-cc: Pushed the fix in a separate commit, it's a shame it's not handled natively 😞

src/components/cc-dialog/cc-dialog.js
florian-sanders-cc: Ah well, this is a `Lit Analyzer` issue, if I add a `-` then it doesn't recognize it as `default` slot and we end up with errors in CI. Same if I modify the named slot :-1:

marion-izd:
> @marion-izd
> 
> > Quick question: there's no story with two buttons, e.g. yes/no — is this a use case that's planned with the component or not?
> 
> This is covered by a separate component used inside `cc-dialog`, this: https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-dialog/init/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-dialog-confirm-actions--default-story
> 
> Or did you mean without a description above the buttons?


Oh, my mistake, I didn't see that you had created three components: dialog, dialog-confirm-form, and dialog-confirm-action. I only looked at dialog 🫣

src/styles/default-theme.css
```diff
@@ -185,6 +185,10 @@
   For instance: submit buttons */
   --cc-color-bg-success: var(--color-green-100);
 
+  /* Usage: backdrop overlay or dialog background.
+  For instance: modal backdrop. */
+  --cc-color-bg-backdrop: rgb(30 30 30 / 55%);
```
florian-sanders-cc: TODO Flo: not valid rgb value, crashes the console build

docs/getting-started/design-tokens.md
```diff
@@ -48,34 +48,35 @@ As our design tokens only exist (for now) as CSS values, keys will be listed as
...
+| `--cc-color-bg-strong`              | `--color-blue-100`       | `#012a51`             |
+| `--cc-color-bg-success`             | `--color-green-100`      | `#098846`             |
+| `--cc-color-bg-success-hovered`     | `--color-green-20`       | `#c3dab7`             |
+| `--cc-color-bg-success-weak`        | `--color-green-30`       | `#baf0be`             |
+| `--cc-color-bg-success-weaker`      | `--color-green-10`       | `#e3ffd6`             |
+| `--cc-color-bg-warning`             | `--color-orange-100`     | `#c15807`             |
+| `--cc-color-bg-warning-hovered`     | `--color-orange-20`      | `#d9d4ad`             |
+| `--cc-color-bg-warning-weak`        | `--color-orange-30`      | `#fcf3b5`             |
+| `--cc-color-bg-warning-weaker`      | `--color-orange-10`      | `#fff9cb`             |
+| `--cc-color-bg-backdrop`            | `rgb(30 30 30 / 55%)`    | `rgb(30 30 30 / 55%)` |
```
florian-sanders-cc: TODO Flo: not valid rgb value, crashes the console build + alphabetical order

pdesoyres-cc (approved):
LGTM. Thank you for that, and well done with the tests suites

PR #1625, opened 2025-12-12
feat(cc-cellar-explorer): init component

pdesoyres-cc:
This PR introduces a new `cc-cellar-explorer` component along with several supporting components (`cc-breadcrumbs`, `cc-grid`, `cc-drawer`) that enable browsing and managing Cellar storage. It also includes enhancements to existing components and utilities.

  Changes by Commit

- chore(dependencies): upgrade `baseline-browser-mapping` to version 2.9.11
  - because stylelint integration in webstorm was complaining about the old version we were using
- docs(resize-controller): fix JSDoc of the resize callback
- feat(cc-clipboard): add `skeleton` state
- feat(cc-clipboard): enable focus delegation
- feat(cc-button): enable focus delegation
  - Replaces the focus() implementation
 - feat(cc-grid): init component
  - Created new `cc-grid` component for displaying tabular data
  - Features include:
    - Column sorting (ascending/descending)
    - Keyboard navigation
    - Sticky header
- feat(cc-drawer): init component
  - Introduced new `cc-drawer` component for slide-out panels
  - First very simple implementation
- feat(cc-cellar-explorer): init component
  - This is the first iteration with buckets management only
  - Features:
    - Grid-based buckets listing
    - Filtering
    - Bucket creation
    - Bucket deletion
  - A bit of complexity because of the object listing coming in a future PR
    - a separate component: `cc-bucket-list`
    - a separate controller used in `cc-cellar-explorer` smart

### DIFF
--- a/src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
+++ b/src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
@@ -1,0 +1,675 @@
+import { css, html, LitElement } from 'lit';
+import { createRef, ref } from 'lit/directives/ref.js';
+import {
+  iconRemixAddFill as iconAdd,
+  iconRemixArchiveLine as iconBucket,
+  iconRemixDeleteBin_6Line as iconDelete,
+  iconRemixSearchLine as iconFilter,
+  iconRemixMoreFill as iconMore,
+} from '../../assets/cc-remix.icons.js';
+import { FormErrorFocusController } from '../../lib/form/form-error-focus-controller.js';
+import { formSubmit } from '../../lib/form/form-submit-directive.js';
+import { isStringEmpty, random, randomString } from '../../lib/utils.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-button/cc-button.js';
+import '../cc-clipboard/cc-clipboard.js';
+import '../cc-dialog-confirm-actions/cc-dialog-confirm-actions.js';
+import '../cc-dialog/cc-dialog.js';
+import '../cc-drawer/cc-drawer.js';
+import '../cc-grid/cc-grid.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-input-text/cc-input-text.js';
+import '../cc-notice/cc-notice.js';
+import {
+  CcCellarBucketCreateEvent,
+  CcCellarBucketDeleteEvent,
+  CcCellarBucketFilterEvent,
+  CcCellarBucketHideEvent,
+  CcCellarBucketShowEvent,
+  CcCellarBucketSortEvent,
+} from './cc-cellar-bucket-list.events.js';
+
+/**
+ * @import { CellarBucketListState, CellarBucketListStateLoading, CellarBucketListStateLoaded, CellarBucketState, CellarBucketCreateFormState, CellarBucketSortColumn } from './cc-cellar-bucket-list.types.js'
+ * @import { CellarBucketVersioning } from '../cc-cellar-explorer/cc-cellar-explorer.client.types.js'
+ * @import { CcGrid } from '../cc-grid/cc-grid.js'
+ * @import { CcGridColumnDefinition } from '../cc-grid/cc-grid.types.js'
+ * @import { CcGridSortEvent } from '../cc-grid/cc-grid.events.js'
+ * @import { TemplateResult } from 'lit'
+ * @import { Ref } from 'lit/directives/ref.js'
+ */
+
+/** @type {Array<CellarBucketState>} */
+const SKELETON_BUCKETS = [...Array(5)].map(() => ({
+  type: 'bucket',
+  state: 'idle',
+  name: randomString(random(10, 15)),
+  createdAt: new Date().toISOString(),
+  updatedAt: new Date().toISOString(),
+  sizeInBytes: random(150_000, 2_000_000),
+  objectsCount: random(1_000, 1_000_000),
+}));
+
+const NUMBER_OF_BUCKETS_ENABLING_FILTERING = 5;
+
+/**
+ * A component that allows to navigate through a Cellar addon.
+ *
+ * @cssdisplay block
+ */
+export class CcCellarBucketList extends LitElement {
+  static get properties() {
+    return {
+      state: { type: Object },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {CellarBucketListState} Sets state. */
+    this.state = { type: 'loading' };
+
+    /** @type {Ref<HTMLFormElement>} */
+    this._createBucketFormRef = createRef();
+
+    /** @type {Ref<HTMLElement>} */
+    this._createBucketButtonRef = createRef();
+
+    /** @type {Ref<CcGrid>} */
+    this._gridRef = createRef();
+
+    new FormErrorFocusController(this, this._createBucketFormRef, () => {
+      if (this.state.type === 'loaded') {
+        return this.state.createForm?.error;
+      }
+      return null;
+    });
+  }
+
+  //#region Public methods
+
+  /**
+   * @param {string} bucketName
+   */
+  scrollToBucket(bucketName) {
+    this.updateComplete.then(() => {
+      if (this.state.type === 'loaded') {
+        const index = this.state.buckets.findIndex((b) => b.name === bucketName);
+        this._gridRef?.value.scrollToIndex(index);
+      }
+    });
+  }
+
+  //#endregion
+
+  //#region Private methods
+
+  /**
+   * @param {CellarBucketCreateFormState} state
+   */
+  _getBucketCreationErrorMessage(state) {
+    if (state?.error == null) {
+      return null;
+    }
+    switch (state.error) {
+      case 'bucket-already-exists':
+        return i18n('cc-cellar-bucket-list.create.error.bucket-already-exists');
+      case 'bucket-name-invalid':
+        return i18n('cc-cellar-bucket-list.create.error.bucket-name-invalid');
+      case 'too-many-buckets':
+        return i18n('cc-cellar-bucket-list.create.error.too-many-buckets');
+    }
+  }
+
+  /**
+   * @param {CellarBucketVersioning} versioning
+   */
+  _getVersioningLabel(versioning) {
+    switch (versioning) {
+      case 'disabled':
+        return i18n('cc-cellar-bucket-list.details.overview.versioning.disabled');
+      case 'enabled':
+        return i18n('cc-cellar-bucket-list.details.overview.versioning.enabled');
+      case 'suspended':
+        return i18n('cc-cellar-bucket-list.details.overview.versioning.suspended');
+    }
+  }
+
+  /**
+   * @returns {{total: number|null, filtered: number|null}}
+   */
+  _getCountingState() {
+    const totalCount = this.state.type === 'loaded' ? this.state.total : null;
+    const filteredItemsCount =
+      this.state.type === 'loaded' && !isStringEmpty(this.state.filter) ? this.state.buckets.length : null;
+    return {
+      total: totalCount,
+      filtered: filteredItemsCount,
+    };
+  }
+
+  //#endregion
+
+  //#region Event handlers
+
+  /**
+   * @param {string} bucketName
+   */
+  _onDisplayBucketDetailRequested(bucketName) {
+    this.dispatchEvent(new CcCellarBucketShowEvent(bucketName));
+  }
+
+  _onCloseBucketDetails() {
+    this.dispatchEvent(new CcCellarBucketHideEvent());
+  }
+
+  /**
+   * @param {string} bucketName
+   */
+  _onDeleteBucket(bucketName) {
+    this.dispatchEvent(new CcCellarBucketDeleteEvent(bucketName));
+  }
+
+  _onDrawerFocusLost() {
+    if (this.state.type === 'loaded') {
+      if (this.state.buckets.length > 0) {
+        this._gridRef.value.focus();
+      } else {
+        this._createBucketButtonRef.value.focus();
+      }
+    }
+  }
+
+  /**
+   * @param {CcGridSortEvent} event
+   */
+  _onGridSort(event) {
+    const { columnIndex, direction } = event.detail;
+
+    /** @type {CellarBucketSortColumn} */
+    let column = 'name';
+    if (columnIndex === 1) {
+      column = 'updatedAt';
+    } else if (columnIndex === 2) {
+      column = 'sizeInBytes';
+    } else if (columnIndex === 3) {
+      column = 'objectsCount';
+    }
+
+    this.dispatchEvent(
+      new CcCellarBucketSortEvent({
+        column,
+        direction,
+      }),
+    );
+  }
+
+  _onCreateBucketButtonClick() {
+    if (this.state.type === 'loaded') {
+      this.state = {
+        ...this.state,
+        createForm: {
+          type: 'idle',
+          bucketName: '',
+          error: null,
+        },
+      };
+    }
+  }
+
+  _onConfirmBucketCreation() {
+    this._createBucketFormRef.value.requestSubmit();
+  }
+
+  _onCancelBucketCreation() {
+    if (this.state.type === 'loaded') {
+      this.state = {
+        ...this.state,
+        createForm: null,
+      };
+      this._createBucketFormRef.value.reset();
+    }
+  }
+
+  /**
+   * @param {{bucketName: string}} formData
+   */
+  _onCreateBucketFormSubmit({ bucketName }) {
+    this.dispatchEvent(new CcCellarBucketCreateEvent(bucketName));
+  }
+
+  /**
+   * @param {{filter: string}} formData
+   */
+  _onFilterFormSubmit({ filter }) {
+    this.dispatchEvent(new CcCellarBucketFilterEvent(filter));
+  }
+
+  //#endregion
+
+  //#region Rendering methods
+
+  render() {
+    if (this.state.type === 'error') {
+      return html`<cc-notice intent="warning" message=${i18n('cc-cellar-bucket-list.error')}></cc-notice>`;
+    }
+
+    return html`<div class="wrapper">
+      ${this._renderHeading(this.state)} ${this._renderGrid(this.state)}
+      ${this.state.type === 'loaded' ? this._renderBucketDetail(this.state) : ''}
+    </div>`;
+  }
+
+  /**
+   * @param {CellarBucketListStateLoading|CellarBucketListStateLoaded} state
+   * @returns {TemplateResult}
+   */
+  _renderHeading(state) {
+    const counts = this._getCountingState();
+    const enabledFiltering = counts.total != null && counts.total >= NUMBER_OF_BUCKETS_ENABLING_FILTERING;
+    const filter = state.type === 'loaded' ? state.filter : '';
+
+    return html`
+      <div class="list-heading">
+        <div class="list-heading--left">
+          <span class="list-heading--title">${i18n('cc-cellar-bucket-list.heading.title')}</span>
+          ${enabledFiltering
+            ? html`<cc-badge class="list-heading--count" weight="dimmed" intent="neutral">
+                ${counts.filtered != null ? `${counts.filtered}/` : ''}${counts.total}
+              </cc-badge>`
+            : ''}
+        </div>
+        <div class="list-heading--center">
+          ${enabledFiltering
+            ? html`<form ${formSubmit(this._onFilterFormSubmit.bind(this))}>
+                <cc-input-text
+                  name="filter"
+                  inline
+                  label=${i18n('cc-cellar-bucket-list.heading.filter.label')}
+                  ?disabled=${state.type !== 'loaded'}
+                  .value=${filter}
+                ></cc-input-text>
+                <cc-button type="submit" .icon=${iconFilter} hide-text outlined ?disabled=${state.type !== 'loaded'}>
+                  ${i18n('cc-cellar-bucket-list.heading.filter.button')}
+                </cc-button>
+              </form>`
+            : ''}
+        </div>
+        <cc-button
+          ${ref(this._createBucketButtonRef)}
+          class="create-bucket-button"
+          primary
+          type="submit"
+          .icon=${iconAdd}
+          ?disabled=${state.type !== 'loaded'}
+          @cc-click=${this._onCreateBucketButtonClick}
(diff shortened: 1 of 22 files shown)


src/components/cc-clipboard/cc-clipboard.js
florian-sanders-cc, line 1: **question:** for you and @marion-izd, we originally decided that the clipboard component would not have a skeleton state and that it would simply be hidden while loading.
I'm fine with adding a `skeleton` state (I was in favor of adding it in the first place) but:
1. Has it been discussed and is it a deliberate choice? 
2. Could we create an issue to adapt other components already using it to also use the `skeleton` prop when relevant? (for instance in add-on dashboard components) (I can create the issue if needed 👍)

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
+import '@lit-labs/virtualizer';
```
florian-sanders-cc: **fix:** not used
```suggestion
```

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+          table.style.gridTemplateColumns = this._getGridTemplateStyle(hiddenColumns);
+        }
+
+        this._hiddenColumns = hiddenColumns;
+
+        // adjust focus if necessary
+        if (this._hiddenColumns.includes(this._cursor.col)) {
+          this._moveFocus(-1, 0);
+        }
+      },
```
florian-sanders-cc: **suggestion:** this is fairly long and involves quite some logic, should we move this to a private method to declutter the constructor?

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+    this._tableRef.value.querySelector(`[data-row="${idx}"]`).scrollIntoView({ block: 'center' });
+  }
+
+  /**
+   * @param {number} columnIndex
+   */
+  _onSortButtonClick(columnIndex) {
+    const currentDirection = this.columns[columnIndex].sort;
+    const newDirection = currentDirection === 'asc' ? 'desc' : 'asc';
+    this.dispatchEvent(new CcGridSortEvent({ columnIndex, direction: newDirection }));
```
florian-sanders-cc, line 315: **question/suggestion:** should we avoid dispatching an event when `this.disabled` is `true`?

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+    // move the cursor to the right place if necessary
+    let newCursor;
+    if (changedProperties.has('columns')) {
+      if (this._cursor.col >= this.columns.length) {
+        newCursor = { col: 0, row: this._cursor.row };
+      }
+    }
+    if (newCursor != null) {
+      this._cursor = newCursor;
+      if (this._tableRef.value?.querySelector(':focus') != null) {
```
florian-sanders-cc, line 398: **suggestion:** I feel like this is more "efficient" (but let's be honest it doesn't matter at this point) and clearer but you can definitely keep it as it is


```suggestion
      if (this._tableRef.value?.contains(this.shadowRoot.activeElement)) {
```

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+    </div>`;
+  }
+
+  /**
+   * @returns {TemplateResult}
+   */
+  _renderHeader() {
+    return html`<tr>
+      ${this.columns.map((column, columnIndex) => {
+        const hasFocus = this._isHeaderSortable && this._cursor.row === -1 && this._cursor.col === columnIndex;
```
florian-sanders-cc, line 438: **question:** can't we design the code so that stuff handling the cursor position guards against cases where cursor is set to `-1` when it's not supposed to (`this._isHeaderSortable === false`)? 
I other words, can we handle the `this._isHeaderSortable` check when `this._cursor` is set, rather than checking it when it's read? This would make the impossible case `isHeaderSortable === false && this._cursor.row === -1` actually impossible?

I feel it's not the rendering responsibility to validate this (but I'm fine if you keep it this way, it works 😉)

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+        };
+
+        let ariaSort;
+        if (column.sort === 'asc') {
+          ariaSort = 'ascending';
+        } else if (column.sort === 'desc') {
+          ariaSort = 'descending';
+        } else if (column.sort === 'none') {
+          ariaSort = 'none';
+        }
```
florian-sanders-cc: **nitpick:** could be moved to a private method with a switch to match our usual patterns

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+
+    if (clampedCursor.col < 0 || clampedCursor.row < this._getFirstRowIndex()) {
+      return null;
+    }
+    return clampedCursor;
+  }
+
+  _getFirstRowIndex() {
+    return this._isHeaderSortable ? -1 : 0;
+  }
```
florian-sanders-cc: **nitpick:** should be moved before render

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+          ${this.a11yName}
+        </caption>
+        <thead>
+          ${this._renderHeader()}
+        </thead>
+        <tbody>
+          ${this.items.map((item, index) => this._renderRow(item, index))}
+        </tbody>
+      </table>
+      <div><!-- spacer to make the above <table> shrink properly --></div>
```
florian-sanders-cc: **fix:** if you remove the `.wrapper` + this spacer and set `height: 100%` to `table`, it seems to work, unless I'm missing something (feel free to ping if I'm wrong and you want to discuss this but I think we should avoid spacers as much as possible, they're usually a smell that something is wrong with the CSS).

src/components/cc-grid/cc-grid.js
```diff
@@ -0,0 +1,697 @@
...
+          overflow: hidden;
+          text-overflow: ellipsis;
+          white-space: nowrap;
+        }
+
+        button.sort {
+          align-items: center;
+          background: unset;
+          border: none;
+          border-radius: var(--cc-border-radius-default, 0.25em) !important;
```
florian-sanders-cc: **question:** why do we need the `!important` here?

src/components/cc-drawer/cc-drawer.js
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }
+
+  _tryToFocusOpenerElement() {
+    if (this._openerElement instanceof HTMLElement && this._openerElement.isConnected) {
+      this._openerElement.focus();
+    } else {
+      this.dispatchEvent(new CcFocusRestorationFailEvent(this._openerElement));
+    }
+  }
```
florian-sanders-cc: **nitpick:** move before render

src/components/cc-drawer/cc-drawer.js
```diff
@@ -0,0 +1,225 @@
...
+
+        @keyframes show {
+          from {
+            translate: 100%;
+          }
+
+          to {
+            translate: 0 0;
+          }
+        }
```
florian-sanders-cc, line 243: **praise:** I didn't know these were widly available, it's great that we don't have to use `transform` anymore 🙌

src/components/cc-drawer/cc-drawer.js
florian-sanders-cc, line 1: **note:** we should probably track somewhere that there are good refactoring opportunities here, should the dialog component handle this as an alternative style, or add a mixin to regroup common logic + shared styles. I don't know to be honest but would be nice to reflect on this at some point 👍

src/components/cc-grid/cc-grid.js
florian-sanders-cc, line 1: **suggestion:** I think it would be good to have tests that cover at least the keyboard navigation logic at some point. We could track this in a separate issue since this PR is already big enough?

src/components/cc-drawer/cc-drawer.stories.js
```diff
@@ -0,0 +1,47 @@
...
+      /** @type {CcDrawer} */
+      const drawer = container.querySelector('cc-drawer');
+      drawer.show();
+    }
+
+    function refresh() {
+      render(
+        html`
+          <cc-button @cc-click=${_onClick}>Open drawer</cc-button>
+          <cc-drawer heading="Hello world!">
```
florian-sanders-cc: **suggestion:** make it `open` by default in the story so that we benefit from visual tests in the long run (we'll have screenshots of the actual component instead of the button and we'll be able to detect if it changes) + it makes it even more obvious what this story is for

florian-sanders-cc: and if we want to showcase / emphasize the opening animation, we can rely on a simulation

src/translations/translations.en.js
```diff
@@ -425,6 +426,76 @@ export const translations = {
...
   //#endregion
+  //#region cc-cellar-bucket-list
+  'cc-cellar-bucket-list.count': /** @param {{count: number}} _ */ ({ count }) => formatNumber(lang, count),
+  'cc-cellar-bucket-list.create.bucket-name.help.case': `Bucket names must not contain uppercase characters or underscores.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.ip': `Bucket name cannot be formatted as IP address.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.labels': `Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.size': `Bucket names can be between 3 and 63 characters long.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.start': `Bucket names must start with a lowercase letter or number.`,
+  'cc-cellar-bucket-list.create.bucket-name.label': `Bucket Name`,
+  'cc-cellar-bucket-list.create.error.bucket-already-exists': `The bucket already exists`,
```
florian-sanders-cc: **suggestion:** we could be even more explicit to clarify that bucket names are shared with everybody like `This bucket name is already taken` (same goes for the FR version)

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+            <div class="details-actions">
+              <cc-button
+                danger
+                .icon=${iconDelete}
+                @cc-click=${() => this._onDeleteBucket(bucket.name)}
+                .waiting=${bucket.state === 'deleting'}
+                .disabled=${bucket.objectsCount > 0 && bucket.state !== 'deleting'}
+                >${i18n('cc-cellar-bucket-list.details.actions.delete.button')}</cc-button
+              >
+              <br />
```
florian-sanders-cc: **nitpick:** we can probably avoid the `<br />` with CSS

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+              ${bucket.objectsCount > 0
+                ? html`<cc-notice
+                    message=${i18n('cc-cellar-bucket-list.details.actions.delete.must-be-empty')}
+                    intent="info"
+                  ></cc-notice>`
```
florian-sanders-cc, line 452: **question:** do we plan on providing a way for users to delete the content AND the bucket through a single action after confirmation? I find the current pattern so frustrating as a user (even if I know that's how it works under the hood)

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+                : ''}
+            </div>
+
+            <div class="details-sub-title">${i18n('cc-cellar-bucket-list.details.overview.title')}</div>
+
+            <dl class="details-overview-list">
+              <dt><span>${i18n('cc-cellar-bucket-list.details.overview.objects-count')}</span></dt>
```
florian-sanders-cc: **nitpick:** do you need the `<span>`, couldn't the string be placed directly in `<dt>`?

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+  _renderBucketCreationForm(state) {
+    return html`<cc-dialog
+      class="create-bucket-dialog"
+      heading=${i18n('cc-cellar-bucket-list.create.title')}
+      ?open=${state != null}
+      @cc-close=${this._onCancelBucketCreation}
+    >
+      <form ${formSubmit(this._onCreateBucketFormSubmit.bind(this))} ${ref(this._createBucketFormRef)}>
+        <cc-input-text
+          ?autofocus=${true}
```
florian-sanders-cc, line 492: **question:** I don't remember which plugin was annoying and forced you to do that, could we track this subject in an issue? 👀

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+          label=${i18n('cc-cellar-bucket-list.create.bucket-name.label')}
+          name="bucketName"
+          required
+          .value=${state?.bucketName ?? ''}
+          .errorMessage=${this._getBucketCreationErrorMessage(state)}
```
florian-sanders-cc, line 498: **fix:** should add `readonly` when creating

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+   */
+  _getCountingState() {
+    const totalCount = this.state.type === 'loaded' ? this.state.total : null;
+    const filteredItemsCount =
+      this.state.type === 'loaded' && !isStringEmpty(this.state.filter) ? this.state.buckets.length : null;
+    return {
+      total: totalCount,
+      filtered: filteredItemsCount,
+    };
+  }
```
florian-sanders-cc: **nitpick:** move before render

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+        }
+
+        .wrapper {
+          background-color: var(--cc-color-bg-default, #fff);
+          border: 1px solid var(--cc-color-border-neutral, #aaa);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: flex;
+          flex-direction: column;
+          gap: 1em;
+          height: calc(100% - 2em);
```
florian-sanders-cc: **suggestion:** this is fragile, if the padding changes then this breaks. Switch to `100%` and add `box-sizing: border-box`

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+          border: 1px solid var(--cc-color-border-neutral-weak, #aaa);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          flex: 1;
+          min-height: 0;
+        }
+
+        .details-wrapper {
+          display: flex;
+          flex-direction: column;
+          width: 25em;
```
florian-sanders-cc, line 601: **fix:** this breaks when the screen is narrow, I think you can remove it unless I'm missing a case (if you do need to set some value, try `max-width` instead)

florian-sanders-cc (changes requested):
Well done @pdesoyres-cc I can't wait to see this in prod, I'm sure users will enjoy this a lot, and it's only the first iteration 🙌 

I found a few issues but you'll see they are pretty trivial.
Huge props for your awesome work on accessibility and the code quality in general 😉

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.types.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  | CellarBucketListStateError
+  | CellarBucketListStateLoaded;
+
+export interface CellarBucketListStateLoading {
+  type: 'loading';
+}
+
+export interface CellarBucketListStateError {
+  type: 'error';
+  filter?: string;
```
HeleneAmouzou: **question**: why do we need filter in error type ?

src/components/cc-cellar-explorer/cc-cellar-explorer.client.types.d.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface CellarBucket {
+  name: string;
+  createdAt: string;
+  updatedAt: string;
+  objectsCount: number;
+  sizeInBytes: number;
+  versioning?: CellarBucketVersioning;
+}
+
+export interface CellarBucketDetails extends CellarBucket {}
```
HeleneAmouzou, line 22: **question**: what is the benefit of extending without adding property ? why not having and using a unique type `CellarBucket` or `CellarBucketDetails` ?

src/translations/translations.en.js
```diff
@@ -425,6 +426,76 @@ export const translations = {
   //#region cc-button
   'cc-button.cancel': `Click to cancel`,
   //#endregion
+  //#region cc-cellar-bucket-list
+  'cc-cellar-bucket-list.count': /** @param {{count: number}} _ */ ({ count }) => formatNumber(lang, count),
+  'cc-cellar-bucket-list.create.bucket-name.help.case': `Bucket names must not contain uppercase characters or underscores.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.ip': `Bucket name cannot be formatted as IP address.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.labels': `Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.`,
+  'cc-cellar-bucket-list.create.bucket-name.help.size': `Bucket names can be between 3 and 63 characters long.`,
```
HeleneAmouzou: **suggestion:** `Bucket names must be between 3 and 63 characters long.`

HeleneAmouzou (approved):
I have nothing much to add to what Florian already noticed. I mostly have questions.
GG for this impressive first iteration ! 👏

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
```diff
@@ -0,0 +1,657 @@
...
+        .details-wrapper .details-icon-wrapper cc-icon {
+          height: 5em;
+          width: 5em;
+        }
+
+        .details-wrapper .details-icon-wrapper .details-name {
+          font-weight: bold;
+          white-space: wrap;
+          word-wrap: anywhere;
+        }
```
roberttran-cc, line 625: **🎯 suggestion:** `word-wrap` is the legacy name alias for `overflow-wrap` (but still compatible), consider using this one instead.

```suggestion
        .details-wrapper .details-icon-wrapper .details-name {
          font-weight: bold;
          overflow-wrap: anywhere;
          white-space: wrap;
        }
```

src/components/cc-drawer/cc-drawer.js
```diff
@@ -0,0 +1,225 @@
+import { css, html, LitElement } from 'lit';
+import { createRef, ref } from 'lit/directives/ref.js';
+import { iconRemixCloseLine as iconClose } from '../../assets/cc-remix.icons.js';
+import { i18n } from '../../lib/i18n/i18n.js';
```
roberttran-cc: **🎯 suggestion:** Use proper import (note: imports order will have to be updated).

```suggestion
import { i18n } from '../../translations/translation.js';
```

src/components/cc-cellar-explorer/cc-cellar-explorer.client.types.d.ts
pdesoyres-cc: This is just to prepare for the future. I prefer anticipating a different type when the bucket details will come with additional properties

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.types.d.ts
pdesoyres-cc: You're. this is not needed

src/components/cc-clipboard/cc-clipboard.js
pdesoyres-cc: This is a deliberated choice. I did not consult anyone about that. I thought it was just not useful before. Could you create the issue?

src/components/cc-grid/cc-grid.js
pdesoyres-cc: yes

src/components/cc-grid/cc-grid.js
pdesoyres-cc: yes + `align-content: start;` to avoid rows stretching to fill the container

src/components/cc-grid/cc-grid.js
pdesoyres-cc: i forgot to remove that

src/components/cc-grid/cc-grid.js
pdesoyres-cc: This part is a bit tricky. I prefer leaving it like that

src/components/cc-drawer/cc-drawer.js
pdesoyres-cc: https://github.com/CleverCloud/clever-components/issues/1650

src/components/cc-grid/cc-grid.js
pdesoyres-cc: #1651

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
pdesoyres-cc: deleting the content is about iterating over all objects and delete them one by one. This can take a long time and we need to figure out how to do that for bucket with lots of objects. This is quite tricky!

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
pdesoyres-cc: #1652

src/components/cc-clipboard/cc-clipboard.js
florian-sanders-cc: Claude did it for me because I was lazy :see_no_evil: 
#1653

roberttran-cc (approved):
Huge work, once again. The UX is really nice, a11y is great, and features are carefully though-out.

I approve the PR although I have two issues while testing the component in the console:
- consecutive errors in the bucket creation form makes it buggy,
- filter form does not appear/disappear immediately on bucket creation/deletion (must refresh page).

But really really great job.

PR #1626, opened 2025-12-15
`cc-addon-header.smart-*`: pass `ownerId` for Zone API calls

### LINKED ISSUE #1607: Use `ownerID` in zones API requests
In new dashboards, some customers get an error with the header which is calling zones API. It's because the response is built based on the more restrictive organisations the user has access to. The solution here is to add `ownerId=XXX` to the request, to get response for the current organisation.

florian-sanders-cc:
Fixes #1607 

## What does this PR do?

- Upgrades the `@clevercloud/client` to `11.4.0`,
- Removes all `@ts-expect-error` that were silencing the fact the client did not provide types since now it does,
- Fixes or silences errors related to the new types provided by the client (we'll need to tackle these once we move to the new client),
- Pass `ownerId` to the Zone API call in the `cc-addon-header` client (the actual goal of this PR :smile:).

## How to review?

- Check the commit about the ownerId stuff,
- Check the `demo-smart`, cc-addon-header => calls to the zone API should contain the `ownerId` as query param (can check from the network tab),
- Check the refactor commit that fixes or silences errors related to the client
  - most changes in there should have 0 impact on runtime code (either `@ts-expect-error` added, or type refactoring) => no need to spend time on these),
  - a few changes impact the runtime code (param removed because they were obsolete for instance. You may check the impacted component still works fine in the demo-smart if you want),
- 1 or 2 reviewers should be enough for this one.

### DIFF
--- a/src/components/cc-env-var-editor-json/cc-env-var-editor-json.js
+++ b/src/components/cc-env-var-editor-json/cc-env-var-editor-json.js
@@ -135,7 +135,8 @@ export class CcEnvVarEditorJson extends LitElement {
     if (changedProperties.has('state')) {
       this._skeleton = this.state.type === 'loading';
       const vars = this.state.type === 'loading' ? SKELETON_VARIABLES : this.state.variables;
-      const filteredVariables = vars.filter(({ isDeleted }) => !isDeleted);
+      // Filter deleted variables and only keep name/value (strip internal state properties like isNew, isEdited, isDeleted)
+      const filteredVariables = vars.filter(({ isDeleted }) => !isDeleted).map(({ name, value }) => ({ name, value }));
       this._variablesAsJson = toJson(filteredVariables);
       this._setErrors([]);
     }


src/components/cc-env-var-editor-json/cc-env-var-editor-json.js
```diff
@@ -136,7 +135,10 @@ export class CcEnvVarEditorJson extends LitElement {
     if (changedProperties.has('state')) {
       this._skeleton = this.state.type === 'loading';
       const vars = this.state.type === 'loading' ? SKELETON_VARIABLES : this.state.variables;
-      const filteredVariables = vars.filter(({ isDeleted }) => !isDeleted);
+      // Filter deleted variables and only keep name/value (strip internal state properties like isNew, isEdited, isDeleted)
+      const filteredVariables = vars
+        .filter(({ isDeleted }) => !isDeleted)
+        .map(({ name, value }) => ({ name, value }));
```
florian-sanders-cc: This fixes this issue:

Image

`isEdited` is not supposed to be displayed 🙈

src/components/cc-tcp-redirection-form/cc-tcp-redirection-form.smart.js
```diff
@@ -147,18 +145,20 @@ async function fetchTcpRedirectionsAndNamespaces({ apiConfig, signal, ownerId, a
  * @returns {Promise<{ port: number }>}
  */
 async function createTcpRedirection({ apiConfig, ownerId, appId, namespace }) {
-  return addTcpRedir({ id: ownerId, appId, payment: 'accepted' }, { namespace }).then(sendToApi({ apiConfig }));
+  return addTcpRedir({ id: ownerId, appId }, { namespace }).then(sendToApi({ apiConfig }));
```
florian-sanders-cc, line 148: I checked the client code and `payment` is not part of what the request body or params so it's clearly obsolete.

src/lib/api-helpers.js
```diff
@@ -107,13 +102,11 @@ function getDownloadUrl(apiConfig, ownerId, invoiceNumber) {
...
     .then(prefixUrl(apiConfig.API_HOST))
     .then(addOauthHeader(apiConfig))
-    .then(
-      /** @param {{url: string, headers: {authorization: string}}} requestParams */ (requestParams) => {
-        const url = new URL(requestParams.url);
-        url.searchParams.set('authorization', btoa(requestParams.headers.authorization));
-        return url.toString();
-      },
-    );
+    .then((requestParams) => {
```
florian-sanders-cc, line 105: the client now provides types so we don't need to type this ourselves + it didn't really match our local types so we might as well fully rely on the client.

PR #1635, opened 2025-12-19
fix(cc-addon-backups): prevent console error on dialog close

florian-sanders-cc:
## What does this MR do?

- Removes dead code that triggered an error upon closing dialogs

## How to review?

- Play with the add-on backups stories, upon closing the dialogs you should get no errors in the console.

### DIFF
--- a/src/components/cc-addon-backups/cc-addon-backups.js
+++ b/src/components/cc-addon-backups/cc-addon-backups.js
@@ -31,9 +31,7 @@ const SKELETON_BACKUPS = {
 
 /**
  * @import { Backup, OverlayType, AddonBackupsState, AddonBackupsStateLoaded, AddonBackupsStateLoading, ProviderId } from './cc-addon-backups.types.js'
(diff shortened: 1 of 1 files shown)


src/components/cc-addon-backups/cc-addon-backups.js
```diff
@@ -326,7 +309,6 @@ export class CcAddonBackups extends LitElement {
    */
   _renderBackups(state) {
     const skeleton = state.type === 'loading';
-    const areBtnsDisabled = state.type === 'loading' || this._overlayType != null;
```
florian-sanders-cc, line 329: we don't need to disable buttons when dialog is open (`overlayType` is `null`) because the `cc-dialog` component traps the focus anyway.
we don't need to disable buttons in skeleton since they don't do anything while in skeleton anything.

Galimede (approved):
Tested the stories and LGTM! 😛

PR #1640, opened 2026-01-06
fix(logs): refactor buffering to use time-based throttling only

### LINKED ISSUE #1627: investigate performance issue
When there are many logs, the browser starts to lag and the UI becomes really hard to use.
I can observe this on firefox (not in chromium).

reproduce with:

- go to an app that loads 10000 logs
- see firefox lagging

pdesoyres-cc:
# What this PR do?

Fixes #1627

**Context**: Logs UI lags and sometimes crashes when many logs are coming fast. This occurs on firefox only ( and also when dev tools are **not** opened)

The proposed solution is to removes that stress on the cc-logs refresh process by removing the size-based buffer limit in favor of time-based limit only.

# How to review

- Use sandbox
- Select live range
- See the UI not lagging anymore


Galimede (approved):
Hey Pierre, I tried to test it in the sandbox but whether with or without the fix my firefox wasn't crashing. So, I tested it locally in the console right away and it works fine. 😄 
Thanks for the fix and GG, LGTM! 💪

PR #1643, opened 2026-01-06
fix(cc-addon-credentials-content): update French token translation

davlgd:
This PR updates `cc-addon-credentials-content.code.token` translation


florian-sanders-cc (changes requested):
Technically this is a fix, I don't think we should have gone for `Jeton` in the first place so it's better to make this part of the changelog.

I'll just force push your commit to change the scope if it's fine for you 👍

PR #1648, opened 2026-01-16
fix: fix date formatting by handling inconsistent timezone output format

pdesoyres-cc:
# What this PR do?

This PR resolves this issue
`Intl.DateTimeFormat` with `timeZoneName: 'longOffset' `returns inconsistent formats across browsers/environments:
  - Some return "GMT"
  - Others return "GMT+00:00"


florian-sanders-cc (approved):
All good! Tested on webkit, firefox & chromium 😎

PR #1654, opened 2026-01-21
feat(pricing): update Pulsar plan intervals and pricing logic

florian-sanders-cc:
## What does this PR do?

- Updates `cc-pricing-product-consumption` (and its dependencies) to:
  - add `hot` vs `cold` storage sections,
  - make pricing progressive since that's how it's billed (since december),
  - update fixtures to stay up to date with pricing (not very important, mostly a copy / paste for stories).

## How to review?

- Review all files except Fixtures (the fixtures files are mostly copy / paste from API + formatting, and their result is visible in the stories so it's better to focus on the visible result than getting lost in the modified lines),
- Check the pricing stories in preview,
- Run storybook locally and check http://localhost:6006/demo-smart/cc-pricing-page/ at the very bottom of the page and play with the pulsar part,
- 2 reviewers should be enough (but I need @Galimede 's review :eyes:)


src/translations/translations.en.js
```diff
@@ -1547,7 +1547,11 @@ export const translations = {
   'cc-pricing-product-consumption.add': `Add`,
   'cc-pricing-product-consumption.bytes': /** @param {{bytes: number}} _ */ ({ bytes }) => formatBytesSi(bytes),
   'cc-pricing-product-consumption.bytes-unit': /** @param {{bytes: number}} _ */ ({ bytes }) => getUnit(bytes),
+  'cc-pricing-product-consumption.cold-storage.label': `storage`,
+  'cc-pricing-product-consumption.cold-storage.title': `Cold storage (HDD)`,
```
pdesoyres-cc: question: should we have colon char at the end?

src/translations/translations.en.js
```diff
@@ -1547,7 +1547,11 @@ export const translations = {
...
   'cc-pricing-product-consumption.error': `An error occurred while retrieving pricing details.`,
+  'cc-pricing-product-consumption.hot-storage.label': `storage`,
+  'cc-pricing-product-consumption.hot-storage.title': `Hot storage (SSD)`,
```
pdesoyres-cc: question: should we have colon char at the end?

src/components/cc-pricing-product-consumption/cc-pricing-product-consumption.js
```diff
@@ -493,8 +503,8 @@ export class CcPricingProductConsumption extends LitElement {
   _renderIntervalList({ type, progressive, intervals, maxInterval }) {
     return intervals.map((interval, intervalIndex) => {
       const maxIntervalIndex = intervals.indexOf(maxInterval);
-      const foo = progressive && intervalIndex <= maxIntervalIndex;
-      const highlighted = interval === maxInterval || foo;
+      const isInProgressiveRange = progressive && intervalIndex <= maxIntervalIndex;
```
pdesoyres-cc, line 506: nice enhancement 😆

florian-sanders-cc:
@pdesoyres-cc it's not a bug it's a feature I believe :laughing: :
Cellar pricing is not progressive so when you consume 15To, you don't benefit from the free tier and only the current price interval is used. 
This is why the price interval is "highlighted" in this mode and not in progressive because in progressive, all tiers add up.

To be honest, I've always been confused by that, I'm not a huge fan of this UX but I don't know how to "quick fix" this.

src/translations/translations.en.js
florian-sanders-cc: Yes we should! :+1:

src/translations/translations.en.js
florian-sanders-cc: Yes we should! 👍

PR #1661, opened 2026-01-27
feat(cellar-explorer): add objects management

pdesoyres-cc:
# Summary

- Add `cc-cellar-object-list` component to browse and manage objects (files/directories) within a Cellar bucket
- Add `cc-breadcrumbs` component for hierarchical path navigation
- Add `enableCopyToClipboard` option to cc-grid link cells
- Extract `Abortable` utility class to a generic lib for request cancellation
- Integrate object list navigation into `cc-cellar-explorer`

# Details

### `cc-cellar-object-list`

A full-featured component for browsing Cellar bucket contents with:

- Grid-based display of files and directories with type-specific icons (images, videos, audio, archives, PDFs, etc.)
- Breadcrumb navigation for directory traversal
- Filtering by object name prefix
- Pagination support (previous/next)
- Object actions: download, delete (with drawer)

### `cc-breadcrumbs`

A reusable breadcrumb navigation component that:

- Displays a hierarchical path of items
- Supports icons and custom labels
- Emits click events with the path to the selected item

## Enhancements

- `cc-grid`: Link cells can now display a copy-to-clipboard button via `enableCopyToClipboard: true`
- `cc-cellar-explorer`: Uses Abortable for smart request cancellation when navigating between buckets/paths

## Refactoring

- Moved Abortable class from kv-utils.js to src/lib/abortable.js for reuse across components

# How to review

- Verify `cc-breadcrumbs` stories render correctly and click events work
- Verify `cc-grid` stories
- Verify copy-to-clipboard button on grid link cells
- Verify `cc-cellar-object-list` stories
- Use sandbox or staging-five
  - Test navigation between directories using breadcrumbs
  - Test object filtering by prefix
  - Test pagination (previous/next page)
  - Test download and delete actions
  - Monkey test rapid switching buckets/paths using breadcrumb (you should never land with an inconsistent object lists)

### DIFF
--- a/src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
+++ b/src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.js
@@ -14,6 +14,7 @@ import { accessibilityStyles } from '../../styles/accessibility.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-badge/cc-badge.js';
 import '../cc-button/cc-button.js';
+import { CcCellarNavigateToBucketEvent } from '../cc-cellar-object-list/cc-cellar-object-list.events.js';
 import '../cc-clipboard/cc-clipboard.js';
 import '../cc-dialog-confirm-actions/cc-dialog-confirm-actions.js';
 import '../cc-dialog/cc-dialog.js';
@@ -98,8 +99,14 @@ export class CcCellarBucketList extends LitElement {
     this.updateComplete.then(() => {
       if (this.state.type === 'loaded') {
         const index = this.state.buckets.findIndex((b) => b.name === bucketName);
-        this._gridRef?.value.scrollToIndex(index);
+        this._gridRef.value?.scrollToIndex(index);
       }
+    });
+  }
+
+  focusFirstCell() {
+    this.updateComplete.then(() => {
+      this._gridRef.value?.focusFirstCell();
     });
   }
 
@@ -154,6 +161,13 @@ export class CcCellarBucketList extends LitElement {
   //#endregion
 
   //#region Event handlers
+
+  /**
+   * @param {string} bucketName
+   */
+  _onBucketClick(bucketName) {
+    this.dispatchEvent(new CcCellarNavigateToBucketEvent(bucketName));
+  }
 
   /**
    * @param {string} bucketName
@@ -339,10 +353,11 @@ export class CcCellarBucketList extends LitElement {
       {
         header: i18n('cc-cellar-bucket-list.grid.column.name'),
         cellAt: (bucketState) => ({
-          type: 'text',
+          type: 'link',
           value: bucketState.name,
           icon: iconBucket,
           enableCopyToClipboard: true,
+          onClick: () => this._onBucketClick(bucketState.name),
         }),
         width: 'minmax(max-content, 1fr)',
         sort: getSort('name'),
@@ -598,7 +613,7 @@ export class CcCellarBucketList extends LitElement {
         .details-wrapper {
           display: flex;
           flex-direction: column;
-          width: 25em;
+          max-width: 25em;
         }
 
         .details-wrapper .details-icon-wrapper {
--- a/src/components/cc-cellar-explorer/cc-cellar-explorer.js
+++ b/src/components/cc-cellar-explorer/cc-cellar-explorer.js
@@ -2,12 +2,14 @@ import { css, html, LitElement } from 'lit';
 import { createRef, ref } from 'lit/directives/ref.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-cellar-bucket-list/cc-cellar-bucket-list.js';
+import '../cc-cellar-object-list/cc-cellar-object-list.js';
 import '../cc-loader/cc-loader.js';
 import '../cc-notice/cc-notice.js';
 
 /**
  * @import { CellarExplorerState } from './cc-cellar-explorer.types.js'
  * @import { CcCellarBucketList } from '../cc-cellar-bucket-list/cc-cellar-bucket-list.js'
+ * @import { CcCellarObjectList } from '../cc-cellar-object-list/cc-cellar-object-list.js'
  * @import { Ref } from 'lit/directives/ref.js'
  */
 
@@ -31,6 +33,8 @@ export class CcCellarExplorer extends LitElement {
 
     /** @type {Ref<CcCellarBucketList>} */
     this._bucketListRef = createRef();
+    /** @type {Ref<CcCellarObjectList>} */
+    this._objectListRef = createRef();
   }
 
   /**
@@ -39,6 +43,13 @@ export class CcCellarExplorer extends LitElement {
   scrollToBucket(bucketName) {
     this.updateComplete.then(() => {
       this._bucketListRef.value?.scrollToBucket(bucketName);
+    });
+  }
+
+  focusFirstCell() {
+    this.updateComplete.then(() => {
+      this._bucketListRef.value?.focusFirstCell();
+      this._objectListRef.value?.focusFirstCell();
     });
   }
 
@@ -58,7 +69,11 @@ export class CcCellarExplorer extends LitElement {
       ></cc-cellar-bucket-list-beta>`;
     }
 
-    return html``;
+    return html`<cc-cellar-object-list-beta
+      ${ref(this._objectListRef)}
+      class="main"
+      .state=${this.state.level.state}
+    ></cc-cellar-object-list-beta>`;
   }
 
   static get styles() {
--- a/src/components/cc-cellar-explorer/cc-cellar-explorer.smart.js
+++ b/src/components/cc-cellar-explorer/cc-cellar-explorer.smart.js
@@ -2,6 +2,7 @@ import { getAllEnvVars } from '@clevercloud/client/esm/api/v2/addon.js';
 import { sendToApi } from '../../lib/send-to-api.js';
 import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
 import { BucketsListController } from '../cc-cellar-bucket-list/cc-cellar-bucket-list.ctrl.js';
+import { ObjectListController } from '../cc-cellar-object-list/cc-cellar-object-list.ctrl.js';
 import { CellarExplorerClient } from './cc-cellar-explorer.client.js';
 import './cc-cellar-explorer.js';
 
@@ -10,7 +11,9 @@ import './cc-cellar-explorer.js';
  * @import { CellarExplorerStateLoaded } from './cc-cellar-explorer.types.js'
  * @import { CellarEndpoint } from './cc-cellar-explorer.client.types.js'
  * @import { CcCellarBucketList } from '../cc-cellar-bucket-list/cc-cellar-bucket-list.js'
+ * @import { CcCellarObjectList } from '../cc-cellar-object-list/cc-cellar-object-list.js'
  * @import { CellarBucketListState } from '../cc-cellar-bucket-list/cc-cellar-bucket-list.types.js'
+ * @import { CellarObjectListState } from '../cc-cellar-object-list/cc-cellar-object-list.types.js'
  * @import { EnvVar } from '../common.types.js'
  * @import { UpdateCallback } from '../common.types.js'
  * @import { ApiConfig } from '../../lib/send-to-api.js'
@@ -42,7 +45,7 @@ defineSmartComponent({
         const cellarClient = new CellarExplorerClient(cellarProxyUrl, cellarEndpoint);
 
         /** @type {() => CcCellarBucketList} */
-        const getComponent = () => component.shadowRoot.querySelector('cc-cellar-bucket-list-beta');
+        const getBucketListComponent = () => component.shadowRoot.querySelector('cc-cellar-bucket-list-beta');
         /** @type {UpdateCallback<CellarBucketListState>} */
         const updateBucketComponent = (newState) => {
           updateComponent(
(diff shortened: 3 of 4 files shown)


src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+  'application/x-lz4',
+  'application/x-lzip',
+  'application/x-arj',
+  'application/x-lzh-compressed',
+  'application/x-ace-compressed',
+  'application/x-stuffit',
+];
+
+/**
+ * A component that allows to navigate through a Cellar addon.
```
HeleneAmouzou:
```suggestion
 * A component that allows to navigate through a Cellar addon's bucket.
```

src/components/cc-cellar-object-list/cc-cellar-object-list.stories.js
```diff
@@ -0,0 +1,345 @@
...
+        objects: [],
+        filter: '???',
+        hasPrevious: false,
+        hasNext: false,
+      },
+    },
+  ],
+});
+
+export const fetchingObject = makeStory(conf, {
```
HeleneAmouzou, line 172: **question :** I don't visually see the fetching state, is this story broken or am I missing something ? 🤔

src/components/cc-breadcrumbs/cc-breadcrumbs.events.js
```diff
@@ -0,0 +1,16 @@
+import { CcEvent } from '../../lib/events.js';
+
+/**
+ * Dispatched when a breadcrumb item is clicked.
+ * @extends {CcEvent<{path: Array<string>}>}
+ */
+export class CcBreadcrumbClickEvent extends CcEvent {
```
florian-sanders-cc, line 7: **question:** tough one but I wonder if removing the "s" here is a good idea, it makes it very easy to get the event name wrong since the component name contains a "s" (but I do understand why it's singlar here 🤷)

Or we could use a totally different event name like `cc-navigation-request` or something like that 🤔

src/components/cc-breadcrumbs/cc-breadcrumbs.js
```diff
@@ -0,0 +1,127 @@
...
+      // language=CSS
+      css`
+        :host {
+          display: block;
+
+          --gap: 0.5em;
+        }
+
+        ul {
+          align-items: end;
```
florian-sanders-cc: **fix:** this breaks alignment for breadcrumbs with no icons.

Image

To fix your issue, you can: 
1. set `display: inline-flex;` on `cc-link` (this will remove the blank space below the icon that led you to choose `align-items: end` in the first place),
2. remove this `align-items` (or replace `end` with `center` but I don't think you need it anyway 🤔)

Image

Image

src/components/cc-grid/cc-grid.js
```diff
@@ -558,16 +558,26 @@ export class CcGrid extends LitElement {
...
-          data-focusable=${skeleton ? 'false' : 'true'}
-          link
-          .icon=${cell.icon}
-          ?skeleton=${skeleton}
-          ?disabled=${this.disabled}
-          @cc-click=${() => this._onCellClick(rowIndex, columnIndex, cell.onClick)}
-          >${cell.value}</cc-button
-        >`,
+        template: html`<div class="icon-label">
+          ${cell.icon != null ? html`<cc-icon .icon=${cell.icon} ?skeleton=${skeleton}></cc-icon>` : ''}
```
florian-sanders-cc: **question:** are we sure we'll always have decorative icons or should we expose & use a `cell.iconA11yName`?

src/components/cc-cellar-object-list/cc-cellar-object-list.ctrl.js
```diff
@@ -0,0 +1,319 @@
...
+    );
+  }
+
+  /**
+   * @param {unknown} error
+   * @param {string} objectKey
+   * @param {function} orElse
+   * @param {boolean} [deleteMode]
+   */
+  #handleErrorOnObject(error, objectKey, orElse, deleteMode = false) {
```
florian-sanders-cc: **nitpick:** 4 positioned args feels a lot to me, especially with a boolean at the end, it makes the signature harder to read & manipulate

When I read this for instance:
```js
this.#handleErrorOnObject(
        error,
        objectKey,
        () => notifyError(i18n('cc-cellar-object-list.error.object-deletion-failed', { objectKey })),
        true,
      );
```

I have to check the actual function to understand what `true` means, what the callback is for.
It's a nitpick because when coding you have the LSP, which mitigates the issue but I feel like object + destructuring would be better.

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+  _onPathItemClick(event) {
+    const path = event.detail.path;
+    if (path.length === 1) {
+      this.dispatchEvent(new CcCellarNavigateToHomeEvent());
+    }
+    if (path.length === 2) {
+      this.dispatchEvent(new CcCellarNavigateToBucketEvent(this.state.bucketName));
+    }
+    this.dispatchEvent(new CcCellarNavigateToPathEvent(path.slice(2)));
+  }
```
florian-sanders-cc, line 167: **question/suggestion:** I bet this doesn't actually trigger an issue but it feels weird that we don't have any early returns in there. If `path.length === 1 || path.length === 2`, then we're dispatching both `CcCellarNavigateToHomeEvent` and `CcCellarNavigateToPathEvent` 🤔

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+        value: path,
+        icon: iconDirectory,
+      })),
+    ];
+
+    const copyValue = state.path.length > 0 ? state.path.join('/') : state.bucketName;
+
+    return html`
+      <div class="path-wrapper">
+        <cc-breadcrumbs .items=${items} @cc-breadcrumb-click=${this._onPathItemClick}></cc-breadcrumbs>
```
florian-sanders-cc: **fix:** the breadcrumb should be placed in a `<nav arial-label="{some translation for breadcrumb / fil d'Ariane}"`
Should the component itself wrap the list of links in `<nav>`? I think so 🤔 (I'm not even sure Shadow DOM surfaces region landmarks correctly but we should use a `<nav>` element anyway).

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+      ${object != null
+        ? html`<div class="details-wrapper">
+            <div class="details-icon-wrapper">
+              <cc-icon .icon=${this._getFileIcon(object.contentType)}></cc-icon>
+              <div class="details-name">${object.name}</div>
+            </div>
+
+            <div class="details-sub-title">${i18n('cc-cellar-object-list.details.actions.title')}</div>
+            <div class="details-actions">
+              <cc-button
```
florian-sanders-cc: **fix:** this should be a link unless we really don't have a choice. The current behavior does not actually trigger a download + opens a new window which is not the expected behavior. 
We can discuss it in sync but I think we should stick with the way we handled this for Kube: https://github.com/CleverCloud/clever-components/blob/1fbe222ab3744c63e7d170e7eb85c2a7fad3c1fe/src/components/cc-addon-header/cc-addon-header.smart-kubernetes.js#L46

It might be too complicated to do that in your case but it's worth trying because it would allow to provide the filename + extension so that it actually downloads the file when the user clicks (right now the user has to do all of this manually).

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+    return html`<cc-drawer
+      heading=${i18n('cc-cellar-object-list.details.heading')}
+      ?open=${object != null}
+      @cc-close=${this._onCloseObjectDetails}
+      @cc-focus-restoration-fail=${this._onDrawerFocusLost}
+    >
+      ${object != null
+        ? html`<div class="details-wrapper">
+            <div class="details-icon-wrapper">
+              <cc-icon .icon=${this._getFileIcon(object.contentType)}></cc-icon>
```
florian-sanders-cc: **fix:** the icon provides information about the type of file / media so we should provide an a11yName with this information

src/components/cc-breadcrumbs/cc-breadcrumbs.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * WARNING: When a breadcrumb item's `label` is an empty string, you MUST set the `iconA11yName` value to provide accessible text for the icon.
+ * Otherwise, the link will have no discernible text, which is a serious accessibility issue (WCAG: Links must have discernible text).
+ *
+ * @cssdisplay block
+ */
+export class CcBreadcrumbs extends LitElement {
+  static get properties() {
+    return {
+      disabled: { type: Boolean },
```
florian-sanders-cc: **fix:** this property is not implemented

src/translations/translations.fr.js
```diff
@@ -503,6 +503,56 @@ export const translations = {
   //#region cc-cellar-explorer
   'cc-cellar-explorer.error': `Une erreur est survenue pendant le chargement`,
   //#endregion
+  //#region cc-cellar-object-list
+  'cc-cellar-object-list.back-to-bucket-list': `Retour à la list des buckets`,
```
florian-sanders-cc: **fix:** 
```suggestion
  'cc-cellar-object-list.back-to-bucket-list': `Retour à la liste des buckets`,
```

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+
+        .details-wrapper .details-icon-wrapper cc-icon {
+          height: 5em;
+          width: 5em;
+        }
+
+        .details-wrapper .details-icon-wrapper .details-name {
+          font-weight: bold;
+          white-space: wrap;
+          word-wrap: anywhere;
```
florian-sanders-cc, line 588: **nitpick(not sure):** Claude tells me this, I'm not sure, I'll go and dig when I have time because I don't know enough about these 🙈 

```
    white-space: wrap;  // Invalid - should be 'normal'                                                                                                                                
    word-wrap: anywhere;  // This is also deprecated, prefer overflow-wrap: anywhere                                                                                                   
```

src/components/cc-cellar-object-list/cc-cellar-object-list.ctrl.js
```diff
@@ -0,0 +1,319 @@
...
+  /**
+   * @param {string} objectKey
+   * @returns {Promise<void>}
+   */
+  async download(objectKey) {
+    this.#updateDetails({ state: 'downloading' });
+    try {
+      const signedUrl = await this.#cellarClient.getObjectSignedUrl(this.#bucketName, objectKey);
+
+      const element = document.createElement('a');
```
florian-sanders-cc: **fix:** if we really want to go down that road, we should use the `download` attribute on the link so that it has the proper behavior. 
Still I think it's a smell that we might be doing this wrong: if what we need is the link behavior, we should use a link from the start because it will also bring the relevant semantics.

src/components/cc-cellar-object-list/cc-cellar-object-list.js
```diff
@@ -0,0 +1,617 @@
...
+  }
+
+  constructor() {
+    super();
+
+    /** @type {CellarObjectListState} Sets state. */
+    this.state = { type: 'loading', bucketName: '', path: [] };
+
+    /** @type {Ref<CcBreadcrumbs>} */
+    this._breadcrumbsRef = createRef();
```
florian-sanders-cc: **fix:** the breadcrumbsRef is not initialized on the element

src/components/cc-cellar-object-list/cc-cellar-object-list.stories.js
pdesoyres-cc: story was broken. thanks

src/components/cc-grid/cc-grid.js
pdesoyres-cc: I'll add a dedicated commit for that

src/components/cc-cellar-object-list/cc-cellar-object-list.js
pdesoyres-cc: I'll add that to the `<cc-breadcrumbs>` with a `label` property/attribute

src/components/cc-cellar-object-list/cc-cellar-object-list.js
pdesoyres-cc: I managed to do that with an auto refresh mechanism (not so hard to make it work)
I wrapped the `<cc-button>` around an `<a>`
Note that the `download` attribute only supports same origin URLs (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download). Here, we use direct URL to cellar so we cannot use this attribute. We have no choice to let the browser handle that. Note that the generated signed URL by Cellar keeps the file name intact so it should work naturally.

src/components/cc-cellar-object-list/cc-cellar-object-list.ctrl.js
pdesoyres-cc: I went for auto refresh download link instead. Note that the `download` attribute supports only same origin URLs

src/components/cc-cellar-object-list/cc-cellar-object-list.js
florian-sanders-cc: We can discuss in sync when you're available but we cannot wrap a button inside in a link, do you really need the `cc-button` except for its styles?

(thanks for the link about the `download` attribute, I didn't know about that, and the general subject about downloading files from JS seems to be a lot trickier than I thought :sweat_smile:)

pdesoyres-cc: i just need the style of a `cc-button`

florian-sanders-cc:
Tested again on staging-five, everything looks good! I didn't check the code modifications since there are no feedback commits but I have no worries about that, they were mostly straightfoward nitpicks :+1: 

Great job, especially focus handling, it works really great and I find the keyboard experience very fluid.

src/components/cc-cellar-object-list/cc-cellar-object-list.ctrl.js
```diff
@@ -0,0 +1,313 @@
...
+        const signedUrl = await this.#cellarClient.getObjectSignedUrl(this.#bucketName, objectKey, 30);
+        this.#updateState(
+          /** @param {CellarObjectListStateLoaded} state */ (state) => {
+            state.details = { state: 'idle', ...bucketDetails, signedUrl: signedUrl.url };
+          },
+        );
+      };
+
+      await refreshSignedUrl();
+      this.#signedUrlInterval = setInterval(refreshSignedUrl, 20_000);
```
roberttran-cc, line 174: **❔ question:** (from Claude Code review)

> Signed URL refresh interval not cleared on navigation. `showDetails()` creates a `setInterval` at line 174, but `changeBucket()`, `navigateTo()`, `filter()`, `previousPage()`, and `nextPage()` don't call `#clearSignedUrlInterval()`. If the user navigates while the details drawer is open, the interval keeps firing orphaned API calls every 20s. Only `hideDetails()` and `abort()` clear it.

pdesoyres-cc: > If the user navigates while the details drawer is open, the interval keeps firing orphaned API calls every 20s

This cannot happen because drawer is a modal dialog

PR #1665, opened 2026-02-03
fix(cc-logs-instances): support null commitId

pdesoyres-cc:
# Context

PHP apps with FTP/SFTP deployment mode generate deployments without commit id.

# Problem

`cc-logs-instances` component does not support this nullity.

# Solution

add nullity check in `cc-logs-instances` when rendering the commit id. When commit id is null, just renders nothing.


florian-sanders-cc (approved):
All good, I was a little bit confused at first because looking at the diff I thought it was exactly the same as before but after searching where `_renderCommit` is called I got it 👍

PR #1666, opened 2026-02-10
feat(cc-addon-*.smart-cellar): init

HeleneAmouzou:
## What does this PR do?

- Adds Cellar dashboard to `index.html`,
- Declines `cc-addon-header`, `cc-addon-info` and `cc-addon-credentials-beta`  for Cellar product.

## How to review?

- Check the commits ,
- Check the code,
- Play with `demo-smart`.

The `.md` documentation files has been  coded by IA, even though I have checked and adjusted afterwards, please pay attention to it.

### DIFF
--- a/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
+++ b/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
@@ -48,7 +48,7 @@ export class CcAddonCredentialsBeta extends LitElement {
       tabs: {
         default: {
           content: [],
-          docLink: { text: fakeString(10), href: null },
+          docLink: { text: fakeString(10), href: undefined },
         },
       },
     };
@@ -130,11 +130,12 @@ export class CcAddonCredentialsBeta extends LitElement {
               ></cc-toggle>
             `
           : ''}
-        ${this.state.type === 'loaded' || this.state.type === 'loading'
+        ${this.state.type === 'loaded' || this.state.type === 'loading' || this.state.type === 'waiting'
           ? html`
               <cc-addon-credentials-content
                 .credentials="${activeTab.content}"
                 .skeleton="${skeleton}"
+                ?disabled="${this.state.type === 'waiting'}"
                 slot="content"
               ></cc-addon-credentials-content>
             `
--- a/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.smart-cellar.js
+++ b/src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.smart-cellar.js
@@ -1,0 +1,216 @@
+import { ONE_SECOND } from '@clevercloud/client/esm/with-cache.js';
+import { notifyError, notifySuccess } from '../../lib/notifications.js';
+import { sendToApi } from '../../lib/send-to-api.js';
+import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-smart-container/cc-smart-container.js';
+import { CcAddonCredentialsBetaClient } from './cc-addon-credentials-beta.client.js';
+import './cc-addon-credentials-beta.js';
+
+/** @type {AddonCredentialsBetaStateLoading} */
+const LOADING_STATE = {
+  type: 'loading',
+  tabs: {
+    default: {
+      content: [
+        {
+          code: 'host',
+          value: 'fake-skeleton',
+        },
+        {
+          code: 'key-id',
+          value: 'fake-skeleton',
+        },
+        {
+          code: 'key-secret',
+          value: 'fake-skeleton',
+        },
+        {
+          code: 'download-file',
+          value: 'fake-skeleton',
+        },
+      ],
+    },
+  },
+};
+const PROVIDER_ID = 'cellar-addon';
+
+/**
+ * @import { CcAddonCredentialsBeta } from './cc-addon-credentials-beta.js'
+ * @import { CellarCredentials, AddonCredentialsBetaStateLoaded, AddonCredentialsBetaStateLoading } from './cc-addon-credentials-beta.types.js'
+ * @import { AddonCredential } from '../cc-addon-credentials-content/cc-addon-credentials-content.types.js'
+ * @import { ApiConfig } from '../../lib/send-to-api.js'
+ * @import { OnContextUpdateArgs } from '../../lib/smart/smart-component.types.js'
+ */
+
+defineSmartComponent({
+  selector: 'cc-addon-credentials-beta[smart-mode="cellar"]',
+  params: {
+    apiConfig: { type: Object },
+    addonId: { type: String },
+    ownerId: { type: String },
+  },
+  /**
+   * @param {OnContextUpdateArgs<CcAddonCredentialsBeta>} args
+   */
+  async onContextUpdate({ context, updateComponent, onEvent, signal }) {
+    const { apiConfig, addonId, ownerId } = context;
+    const api = new Api({ apiConfig, ownerId, addonId, signal });
+
+    updateComponent('state', LOADING_STATE);
+
+    api
+      .getAllCredentials()
+      .then((credentials) => {
+        updateComponent(
+          'state',
+          /** @param {AddonCredentialsBetaStateLoaded|AddonCredentialsBetaStateLoading} state */
+          (state) => {
+            state.type = 'loaded';
+            state.tabs.default.content = credentials;
+          },
+        );
+      })
+      .catch((error) => {
+        console.error(error);
+        updateComponent('state', { type: 'error' });
+      });
+
+    onEvent('cc-addon-credentials-renew-secret', async () => {
+      updateComponent('state', (state) => {
+        state.type = 'waiting';
+      });
+
+      try {
+        await api.renewSecret();
+      } catch (error) {
+        console.error(error);
+        notifyError(i18n('cc-addon-credentials.renew-secret.error'));
+        updateComponent('state', (state) => {
+          state.type = 'loaded';
+        });
+        return;
+      }
+
+      try {
+        const credentials = await api.getAllCredentials();
+        notifySuccess(i18n('cc-addon-credentials.renew-secret.success'));
+        updateComponent(
+          'state',
+          /** @param {AddonCredentialsBetaStateLoaded|AddonCredentialsBetaStateLoading} state */
+          (state) => {
+            state.type = 'loaded';
+            state.tabs.default.content = credentials;
+          },
+        );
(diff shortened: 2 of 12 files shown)


src/translations/translations.fr.js
```diff
@@ -189,7 +189,9 @@ export const translations = {
   'cc-addon-credentials.field.port': `Port`,
   'cc-addon-credentials.field.url': `URL`,
   'cc-addon-credentials.field.user': `Utilisateur`,
-  'cc-addon-credentials.loading-error': `Une erreur est survenue pendant le chargement des informations de connexion.`,
+  'cc-addon-credentials.loading-error': `Une erreur est survenue pendant le chargement des informations de connexion`,
+  'cc-addon-credentials.renew-secret.error': `Une erreur est survenue pendant le renouvellement du secret.`,
```
pdesoyres-cc: nitpick: trailing dot not needed

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+defineSmartComponent({
+  selector: 'cc-addon-header[smart-mode="cellar"]',
+  params: {
+    apiConfig: { type: Object },
+    ownerId: { type: String },
+    addonId: { type: String },
+    cellarId: { type: String },
+    explorerUrlPattern: { type: String },
```
pdesoyres-cc: fix: the link to "open Cellar explorer" depends on a console feature flag. We may want to show this link only if the feature flag is enabled.

We could make this property optional, and if `null`, we don't setup the `openLinks` in the component state.

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
```diff
@@ -48,7 +48,7 @@ export class CcAddonCredentialsBeta extends LitElement {
       tabs: {
         default: {
           content: [],
-          docLink: { text: fakeString(10), href: null },
+          docLink: { text: fakeString(10), href: undefined },
```
florian-sanders-cc, line 51: **question:** why did you need this change? 🤔

src/components/cc-addon-credentials-content/cc-addon-credentials-content.js
```diff
@@ -56,6 +59,7 @@ export class CcAddonCredentialsContent extends LitElement {
   static get properties() {
     return {
       credentials: { type: Array, attribute: 'credentials' },
+      disabled: { type: Boolean },
```
florian-sanders-cc, line 62: **suggestion:** could we showcase this in a story?

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
```diff
@@ -0,0 +1,120 @@
...
+  },
+
+  /** @param {OnContextUpdateArgs<CcAddonHeader>} args */
+  onContextUpdate({ context, updateComponent, signal }) {
+    const { apiConfig, ownerId, addonId, cellarId } = context;
+    const api = new Api({ apiConfig, ownerId, addonId, signal }, cellarId);
+    let explorerUrl = '';
+    updateComponent('state', {
+      type: 'loading',
+      logsUrl: fakeString(15),
```
florian-sanders-cc: **fix (very small one):** remove since cellar has no logs once loaded

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
```diff
@@ -0,0 +1,120 @@
...
+    ownerId: { type: String },
+    addonId: { type: String },
+    cellarId: { type: String },
+    explorerUrlPattern: { type: String },
+  },
+
+  /** @param {OnContextUpdateArgs<CcAddonHeader>} args */
+  onContextUpdate({ context, updateComponent, signal }) {
+    const { apiConfig, ownerId, addonId, cellarId } = context;
+    const api = new Api({ apiConfig, ownerId, addonId, signal }, cellarId);
```
florian-sanders-cc: **nitpick:** make `cellarId` part of the object instead of using a positioned arg

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
```diff
@@ -0,0 +1,120 @@
...
+    let explorerUrl = '';
```
florian-sanders-cc: **suggestion:** I don't think we need the `let` here since `explorerUrl` is only used in the `then` of the `getCellarInfoWithZone`, it can be computed as a `const` within the `then` body (or even inlined directly in the state object).

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
```diff
@@ -0,0 +1,120 @@
...
+ * @import { OnContextUpdateArgs } from '../../lib/smart/smart-component.types.js'
+ */
+
+defineSmartComponent({
+  selector: 'cc-addon-header[smart-mode="cellar"]',
+  params: {
+    apiConfig: { type: Object },
+    ownerId: { type: String },
+    addonId: { type: String },
+    cellarId: { type: String },
```
florian-sanders-cc: **question:** is `cellarId` like a `realId`? If so, why don't we fetch this info from API like we typically do for other add-ons? (I'm okay if you want to keep this as a context param, this is a question but not a request for change, I think it makes sense to use info that you already have but that's not what we did for others and I'm not sure why 🤔)

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.smart-cellar.js
```diff
@@ -0,0 +1,208 @@
...
+          );
+        })
+        .catch((error) => {
+          console.error(error);
+          notifyError(i18n('cc-addon-credentials.renew-secret.error'));
+          updateComponent('state', (state) => {
+            state.type = 'loaded';
+          });
+        });
+    });
```
florian-sanders-cc: **question:** do we want to handle cases where the renewal succeeds but the get fails afterwards? Right now if this happens, we're going to show an error saying the credential hasn't been updated (when in fact it has), and show the old secret.

We can consider this is too much of an edge case to handle this, or we can discriminate the errors (if renewal failed => toast about this, if get failed => toast about the fact that info couldn't be updated in UI even though renewal was successful and invite to refresh the page?)

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
HeleneAmouzou: Because in `Tabs` type, `href` is a `string` and not `string | null`, but maybe it is not the right approch ?
```
type Tabs = {
  [key in TabName & string]?: {
    content: Array<AddonCredential>;
    docLink?: {
      text: string;
      href: string;
    };
  };
};
```

src/components/cc-addon-header/cc-addon-header.smart-cellar.js
HeleneAmouzou: Yes it is ! Thanks for the feeback, I have made some changes in all the Cellar smarts to fetch it from the API to match what we did for the other addons.

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.smart-cellar.js
HeleneAmouzou: Thanks, I have handled this edge case, tested it and added some error messages, let me know what you think :eyes:

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.js
florian-sanders-cc: It's mostly a TypeScript config issue, we should use `strictNullCheck` and handle this another way but I think it's fine for now :+1:

src/components/cc-addon-header/cc-addon-header.smart-cellar.md
```diff
@@ -38,7 +37,7 @@ interface ApiConfig {
 | Method   | URL                                                        | Cache?   |
 |----------|------------------------------------------------------------|----------|
 | `GET`    | `/v2/organisations/${ownerId}/addons/${addonId}`           | 1 second |
-| `GET`    | `/v4/cellar/organisations/${ownerId}/cellar/${cellarId}`   | 1 second |
+| `GET`    | `/v4/cellar/organisations/${ownerId}/cellar/${realId}`     | 1 second |
```
florian-sanders-cc, line 40: **question:** is it `realId` or `addonId`? I thought we were using the `addonId` to retrieve the `realId` 🤔

src/translations/translations.en.js
```diff
@@ -178,6 +178,7 @@ export const translations = {
   'cc-addon-credentials.field.port': `Port`,
   'cc-addon-credentials.field.url': `URL`,
   'cc-addon-credentials.field.user': `User`,
+  'cc-addon-credentials.get-credentials.error': `An error occurred while updating the information, please refresh the page.`,
```
florian-sanders-cc: **nitpick:** remove the trailing dot 😇

src/components/cc-addon-credentials-beta/cc-addon-credentials-beta.smart-cellar.js
```diff
@@ -0,0 +1,216 @@
...
+      },
+    ];
+  }
+
+  /**
+   * @return {Promise<string>}
+   */
+  async renewSecret() {
+    const { realId: cellarId } = await this._getAddon();
+    return renewSecret({ ownerId: this._ownerId, cellarId }).then(sendToApi({ apiConfig: this._apiConfig }));
```
florian-sanders-cc: **suggestion/nitpick:** (not important at all, just a personal preference 😇 )


```suggestion
    const { realId } = await this._getAddon();
    return renewSecret({ ownerId: this._ownerId, cellarId: realId }).then(sendToApi({ apiConfig: this._apiConfig }));
```

florian-sanders-cc (approved):
A few nitpicks but overall it looks good to me! 👍

src/components/cc-addon-header/cc-addon-header.smart-cellar.md
HeleneAmouzou: Yes we use the `addonId` on `/v2/organisations/${ownerId}/addons/${addonId}`  to retrieve the `realId` and then use the `realId` to retreive the cellar info on `/v4/cellar/organisations/${ownerId}/cellar/${realId}`

florian-sanders-cc: ahhh yes sorry it makes sense this is just the endpoint list. All good! :+1:

PR #1667, opened 2026-02-10
Clever Client: add some helper to allow sharing the new client instance for CC API calls

florian-sanders-cc:
## What does this PR do?

- Adds a `getCcApiClient` utility (src/lib/cc-api-client.js) that returns a shared `CcApiClient` instance configured from an `apiConfig`:
  - Clients are cached per `apiConfig` reference using a `WeakMap`, so all smart components sharing the same config share the same client instance,
  - The `WeakMap` allows automatic garbage collection when the `apiConfig` is no longer referenced.
- Adds unit tests for the caching behavior (`test/cc-api-client.test.js`),
- Updates the smart component contributing docs (`docs/contributing/smart-component.md`):
  - Adds a new "API Clients: New vs Legacy" section explaining when to use `getCcApiClient...` vs `sendToApi`,
  - Includes usage examples for both approaches,
  - Clarifies that existing smart components using `sendToApi` do not need to be migrated right away.

**Note:** the idea of using a `WeakMap` and storing the `apiConfig` object ref in it was definitely not mine, it was suggested by Claude when discussing the best ways to share the same client instance across components. I don't have much experience with that but it seemed way better than anything I could think of so I went with that.

## How to review?

- Check the code,
- Check the docs changes in `docs/contributing/smart-component.md` -- focus on whether the "New vs Legacy" section is clear and provides enough guidance for contributors,
- 2-3 reviewers should be enough, specifically @pdesoyres-cc  and @hsablonniere since they've worked on the new client.

### DIFF
--- a/docs/contributing/smart-component.md
+++ b/docs/contributing/smart-component.md
@@ -18,19 +18,29 @@ Before defining your smart component, import the necessary dependencies for conn
 import './cc-my-component.js';
 import '../cc-smart-container/cc-smart-container.js';
 import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
+import { getCcApiClientWithOAuth, getCcApiClientWithToken } from '../../lib/cc-api-client.js';
 import { sendToApi } from '../../lib/send-to-api.js';
 import { notify, notifyError, notifySuccess } from '../../lib/notifications.js';
 ```
 
-You'll also likely need to import specific API functions from the `@clevercloud/client` package:
+You'll also likely need to import API commands from the `@clevercloud/client` package.
+For the new client, use the command classes from `cc-api-commands/`:
 
 ```js
-import { someApiFunction } from '@clevercloud/client/esm/api/v2/some-resource.js';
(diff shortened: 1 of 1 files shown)


pdesoyres-cc (commented):
Thank you @florian-sanders-cc  for taking this subject.

I've got some remarks, nothing blocking.

- you made two caches (two weak maps) but I think we could have only one. The cache key is an ApiConfig object instances in both caches, and I don't see how the keys could clash between the oauth and the apiKey auth strategy.
- Finally, with the ApiConfig object instance as a cache key, we leave the responsability of the memoization to the instance of the config. I'm wondering if we could have another strategy:
  - the cache key is a string
  - when asking for a client from an ApiConfig, we generate a unique string of the given config (a JSON stringify can be fine, or a string concatenation of every properties(key=value))
  - we use this string as a cache key
  - No need of a WeakMap here, a Map will be enough

Maybe a sync with @hsablonniere would be more appropriate to discuss about that

PR #1668, opened 2026-02-13
cc-range-selector: various fixes and updates

roberttran-cc:
## Context
Fixes multiple issues in cc-range-selector related to form value synchronization, validation, and event naming. Also improves component usability by exposing internal state, customizing padding, and cleaning up unused code.

## Changes
- **Breaking**: Rename custom select event to follow convention (`cc-range-selector:select-custom` → `cc-range-selector:custom-select`)
- **Fix**: Sync form value when mode changes (ensures proper state when switching between modes)
- **Fix**: Ensure validation re-runs when mode changes
- **Fix**: Only clear relevant property on custom select (prevent unintended state resets)
- **Fix**: Remove unnecessary CSS custom property on cc-range-selector-option
- **Improvement**: Expose custom option active state as public property for better component introspection
- **Enhancement**: Make cc-range-selector-option padding customizable via CSS variable
- **Enhancement**: Reduce padding on separator element for better component density
- **Cleanup**: Remove unused range input refs

## Breaking changes
The custom select event has been renamed from `cc-range-selector:select-custom` to `cc-range-selector:custom-select` to follow the event naming convention.

### DIFF
--- a/src/components/cc-range-selector/cc-range-selector.js
+++ b/src/components/cc-range-selector/cc-range-selector.js
@@ -1187,7 +1187,7 @@ export class CcRangeSelector extends CcFormControlElement {
           align-items: center;
           color: var(--cc-color-text-weak, #404040);
           display: inline-flex;
-          padding-inline: 0.25em;
+          padding-inline: 0.125em;
           visibility: hidden;
         }
 


src/components/cc-range-selector/cc-range-selector.js
florian-sanders-cc, line 1: **question:** Your parent never writes the `isCustomActive` property value right? The `cc-range-selector` modifies its own public property without dispatching an event to warn parents that this property value has changed so my comment is mostly to make sure you've considered this and decided it was no use. 
If your parent is only reading this prop value and never writing it, then I think we're fine 👍

roberttran-cc: Actually, there's a case where the parent could write this property value: as the component can be reused for both single and range modes, we have to be able to restore this state from the parent when we're toggling mode.

Also, the `CcRangeSelectorSelectCustomEvent` event is dispatched when `isCustomActive` is set to `true`.

Anyway, I don't like how I implemented this part and I guess it could largely be improved.

florian-sanders-cc: Thanks for your explanation, sorry I missed that, it is indeed dispatched 🙈 
I think it's ok as it is? 👍

PR #1669, opened 2026-02-13
feat(cc-homepage-onboarding): init component

clara-layus-cc:
# What does this PR do?

- Adding a component with a set of cards for user onboarding

# How to review?

- Check the commits ,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/homepage-onboarding/index.html?path=/docs/%F0%9F%9B%A0-homepage-cc-homepage-onboarding--docs)


src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+import { classMap } from 'lit/directives/class-map.js';
+import {
+  iconRemixAddCircleLine,
+  // eslint-disable-next-line camelcase
+  iconRemixBankCard_2Line,
+  iconRemixBuildingLine,
+  // eslint-disable-next-line camelcase
+  iconRemixChat_4Line,
+  // eslint-disable-next-line camelcase
+  iconRemixKey_2Line,
```
HeleneAmouzou: **refactor**: you can rename all those icons' name so you don't have to use the `eslint-disable-next-line camelcase` rule. For example : `iconRemixBankCard_2Line as iconBankCard` .

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+      description: i18n('cc-homepage-onboarding.card.secure.description'),
+      icon: iconRemixLockLine,
+      iconColor: 'purple',
+      buttonText: i18n('cc-homepage-onboarding.card.secure.button'),
+      event: () => new CcHomepageOnboardingSecureEvent(),
+    },
+    sshKeys: {
+      title: i18n('cc-homepage-onboarding.card.ssh-keys.title'),
+      description: i18n('cc-homepage-onboarding.card.ssh-keys.description'),
+      // eslint-disable-next-line camelcase
```
HeleneAmouzou: **refactor**: see my comment above

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+      description: i18n('cc-homepage-onboarding.card.new-organisation.description'),
+      icon: iconRemixBuildingLine,
+      iconColor: 'dark-purple',
+      buttonText: i18n('cc-homepage-onboarding.card.new-organisation.button'),
+      event: () => new CcHomepageOnboardingNewOrganisationEvent(),
+    },
+    configPayment: {
+      title: i18n('cc-homepage-onboarding.card.config-payment.title'),
+      description: i18n('cc-homepage-onboarding.card.config-payment.description'),
+      // eslint-disable-next-line camelcase
```
HeleneAmouzou: **refactor**: see my comment above

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+      icon: iconRemixBankCard_2Line,
+      iconColor: 'orange',
+      buttonText: i18n('cc-homepage-onboarding.card.config-payment.button'),
+      event: () => new CcHomepageOnboardingConfigPaymentEvent(),
+    },
+    support: {
+      title: i18n('cc-homepage-onboarding.card.support.title'),
+      description: i18n('cc-homepage-onboarding.card.support.description'),
+      // eslint-disable-next-line camelcase
```
HeleneAmouzou: **refactor**: see my comment above

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+    icon: iconRemixTerminalLine,
+    iconColor: 'blue',
+    buttonText: 'Lorem ipsum',
+  },
+  {
+    title: 'Lorem ipsum dolor',
+    description: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit sed.',
+    icon: iconRemixBuildingLine,
+    iconColor: 'orange',
+    buttonText: 'Lorem ipsum',
```
HeleneAmouzou: **fix**: you can use the `fakeString()` function instead of Lorem ipsum

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+    this.state = { type: 'loading' };
+
+    /** @type {string} */
+    this._selectedOrganisation = '';
+  }
+
+  render() {
+    if (this.state.type === 'error') {
+      return html`<cc-notice intent="warning" message="${i18n('cc-homepage-onboarding.error')}"></cc-notice>`;
+    }
```
HeleneAmouzou, line 204: **question**: I wonder if we should not display a title even in error state ? I am curious to know what you and the others think about that.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+                this._selectedOrganisation = e.detail;
+              }}
+            ></cc-select>`
+          : ''}
+        <cc-button class="card-button" ?skeleton=${skeleton} @cc-click=${() => this._onCardButtonClick(card)}
+          >${card.buttonText}</cc-button
+        >
+      </div>
+    `;
+  }
```
HeleneAmouzou, line 247: **suggestion**: maybe we could extract this in a card component ? Like we did for the `cc-article-card` or `cc-doc-card` ? I'm interested by your and the others opinions on that one.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+   */
+  _onCardButtonClick(card) {
+    if (card.event != null) {
+      if (card.select != null) {
+        this.dispatchEvent(card.event(this._selectedOrganisation));
+      } else {
+        this.dispatchEvent(card.event());
+      }
+    }
+  }
```
HeleneAmouzou, line 275: **nitpick**: move this before the render

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+          flex-direction: column;
+          gap: 1.25em;
+          min-height: 24em;
+          padding: 2em;
+          text-align: left;
+        }
+
+        .blue-card {
+          background: linear-gradient(135deg, #3d5fb4 0%, #6d1cf0 100%);
+          color: white;
```
HeleneAmouzou: **suggestion**: use token

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+        }
+
+        .blue-card cc-select {
+          color: black;
```
HeleneAmouzou: **suggestion**: use token

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+      event: (/** @type {string} */ orgId) => new CcHomepageOnboardingNewResourceEvent(orgId),
+    },
+    newProject: {
+      title: i18n('cc-homepage-onboarding.card.new-project.title'),
+      description: i18n('cc-homepage-onboarding.card.new-project.description'),
+      icon: iconRemixAddCircleLine,
+      iconColor: 'white',
+      buttonText: i18n('cc-homepage-onboarding.card.new-project.button'),
+      event: () => new CcHomepageOnboardingNewProjectEvent(),
+    },
```
HeleneAmouzou, line 51: I found it confusing that the translations for this 'project card' is about creating organization, but maybe David will suggest new wording !

src/components/cc-homepage-onboarding/cc-homepage-onboarding.stories.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+export const defaultStory = makeStory(conf, {
+  items: [
+    {
+      /** @type {HomepageOnboardingStateLoaded} */
+      state: {
+        type: 'loaded',
+        userType: 'already-user',
+        cardIds: ['newProject', 'secure', 'configPayment', 'support'],
```
HeleneAmouzou, line 48: **suggestion**: add 'newOrganisation'

HeleneAmouzou (changes requested):
GG @clara-layus-cc ! 👏 
I have left some comments and reflexions below, feel free to question it and/or wait for the others opinions !

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+      css`
+        :host {
+          display: block;
+          width: 100%;
+        }
+
+        .container {
+          align-items: center;
+          border: solid 1px var(--cc-color-border-neutral-weak);
+          border-radius: var(--cc-border-radius-default);
```
HeleneAmouzou: **fix**: add a fallback value

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+          grid-template-columns: repeat(auto-fit, minmax(max(18em, calc((100% - 3 * 1.3em) / 4)), 1fr));
+          width: 100%;
+        }
+
+        .card {
+          border-radius: 1em;
+          border-top: 2px solid #f3f4f6;
+          box-shadow:
+            0 4px 6px -4px #0000001a,
+            0 10px 15px -3px #0000001a;
```
HeleneAmouzou, line 386: **suggestion**: use color tokens

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+
+          --cc-color-text-weak: white;
+          --cc-select-label-color: white;
+        }
+
+        .icon {
+          border-radius: 0.5em;
+          box-shadow:
+            0 4px 6px -4px #0000001a,
+            0 10px 15px -3px #0000001a;
```
HeleneAmouzou, line 415: **suggestion**: use color tokens

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+        .color-dark-orange {
+          background: linear-gradient(135deg, #cc2121 0%, #e67151 100%);
+        }
+
+        .color-white {
+          background: rgb(255 255 255 / 20%);
+        }
+
+        .card-title {
+          font-size: 1.5rem;
```
HeleneAmouzou: **fix**: use `em`

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+        ${!skeleton && card.select
+          ? html`<cc-select
+              .label=${card.select.title}
+              .placeholder=${card.select.placeholder}
+              required
+              value=""
+              .options=${this.state.type === 'loaded' ? this.state.organisationOptions : []}
+              @cc-select=${(/** @type {CustomEvent<string>} */ e) => {
+                this._selectedOrganisation = e.detail;
+              }}
```
HeleneAmouzou: **fix**: extract this in a dedicated method `onOrganisationSelected` (for example)

HeleneAmouzou (changes requested):
Sorry some of my review was not sent, here is the rest 🙈

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+  CcHomepageOnboardingNewResourceEvent,
+  CcHomepageOnboardingSecureEvent,
+  CcHomepageOnboardingSshKeysEvent,
+  CcHomepageOnboardingSupportEvent,
+} from './cc-homepage-onboarding.events.js';
+
+/**
+ * @returns {Record<string,HomepageOnboardingCard>}
+ */
+function getCards() {
```
florian-sanders-cc, line 30: **suggestion:** not sure but we typically avoid using i18n outside of the class, I just cannot remember why? I would move this function to the class as a private method, but I cannot really give you a good reason why 🤔

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  constructor() {
+    super();
+
+    /** @type {HomepageOnboardingState} Set the state of the component. */
+    this.state = { type: 'loading' };
+
+    /** @type {string} */
+    this._selectedOrganisation = '';
```
florian-sanders-cc: **suggestion:** I would remove this internal prop and rely on a `ref` set on the select to retrieve the value when the button is clicked. You could even place the button + select in a `form` and retrieve `formData` on submit (either by using the `formSubmit` directive or manually). 
We can discuss in sync if it's unclear (not easy to explain in a PR comment since it affects different parts of your code)

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+              value=""
+              .options=${this.state.type === 'loaded' ? this.state.organisationOptions : []}
+              @cc-select=${(/** @type {CustomEvent<string>} */ e) => {
+                this._selectedOrganisation = e.detail;
+              }}
+            ></cc-select>`
+          : ''}
+        <cc-button class="card-button" ?skeleton=${skeleton} @cc-click=${() => this._onCardButtonClick(card)}
+          >${card.buttonText}</cc-button
+        >
```
florian-sanders-cc: **fix:** in most cases, these should be `<cc-link mode="button">` because semantically links go somewhere and buttons do something.

This has a fairly big impact on your code, it means you component should accept URLs and most of the events should be removed.
You should keep the button & event for the specific case of the `newResource` case when the `userType` is `new-user` 👍

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+        ? this.state.userType === 'new-user'
+          ? i18n('cc-homepage-onboarding.title.new-user')
+          : i18n('cc-homepage-onboarding.title.already-user')
+        : SKELETON_TITLE;
+    const description =
+      this.state.type === 'loaded'
+        ? this.state.userType === 'new-user'
+          ? i18n('cc-homepage-onboarding.description.new-user')
+          : i18n('cc-homepage-onboarding.description.already-user')
+        : SKELETON_DESCRIPTION;
```
florian-sanders-cc: **suggestion:** when you end up with nested ternaries I'd say it's good to reconsider the approach, possibly move to some private methods to make it clearer.

For instance a `getDescription(state.type, state.userType)` that would have a clear `if` / `switch` branching.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+        ${this._renderIcon(card.icon, card.iconColor, skeleton)}
+        <h3 class=${classMap({ 'card-title': true, skeleton })}>${card.title}</h3>
+        <p class=${classMap({ 'card-description': true, skeleton })}>${card.description}</p>
+        ${!skeleton && card.select
+          ? html`<cc-select
+              .label=${card.select.title}
+              .placeholder=${card.select.placeholder}
+              required
+              value=""
+              .options=${this.state.type === 'loaded' ? this.state.organisationOptions : []}
```
florian-sanders-cc: **suggestion:** could compute this at the beginning of the `_renderCard` method to make the template lighter / easier to read

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+   * @param {HomepageOnboardingCard} card
+   * @param {boolean} skeleton
+   * @private
+   */
+  _renderCard(card, skeleton) {
+    return html`
+      <div class=${classMap({ card: true, 'blue-card': card.iconColor === 'white', skeleton })}>
+        ${this._renderIcon(card.icon, card.iconColor, skeleton)}
+        <h3 class=${classMap({ 'card-title': true, skeleton })}>${card.title}</h3>
+        <p class=${classMap({ 'card-description': true, skeleton })}>${card.description}</p>
```
florian-sanders-cc: **nitpick:** 

```suggestion
      <div class="card ${classMap({ 'blue-card': card.iconColor === 'white', skeleton })}">
        ${this._renderIcon(card.icon, card.iconColor, skeleton)}
        <h3 class="card-title ${classMap({ skeleton })}">${card.title}</h3>
        <p class="card-description ${classMap({ skeleton })}">${card.description}</p>
```

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+    `;
+  }
+
+  /**
+   * @param {HomepageOnboardingCard} card
+   * @private
+   */
+  _onCardButtonClick(card) {
+    if (card.event != null) {
+      if (card.select != null) {
```
florian-sanders-cc: **suggestion:** you can regroup these conditions instead of nesting `if`

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
clara-layus-cc: No token matches this color because there is a high degree of transparency.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
clara-layus-cc: No token matches this color because there is a high degree of transparency.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.stories.js
clara-layus-cc: The purpose of the component is to display a maximum of four cards, as agreed in the design. I may clarify this in the description to make it clearer.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  static get styles() {
+    return [
+      skeletonStyles,
+      // language=CSS
+      css`
+        :host {
+          display: block;
+          width: 100%;
```
florian-sanders-cc: **fix:** this is not supposed to be necessary and if it is, it shouldn't be set from this component (`display: block` already makes a content span through the whole width, if it's not doing that in your usecase, it might be because you're using flex or grid on the parent, in which case the issue should be handled from the parent and not here).

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+          font-size: 1em;
+          line-height: 1.3;
+          margin: 0;
+          max-width: 50em;
+        }
+
+        .cards {
+          display: grid;
+          gap: 1.3em;
+          grid-template-columns: repeat(auto-fit, minmax(max(18em, calc((100% - 3 * 1.3em) / 4)), 1fr));
```
florian-sanders-cc: **suggestion:** the current rule works but could be simplified considering the context.
I feel like we're always going to have at most 4 cards (this is an asumption, maybe I'm wrong) so we don't have to support cases where there are more than 4 cards.

If that's so, then we can remove the `max` & `calc`, a simpler `minmax` will do the trick and will be easier to maintain.
If you really need to support more than 4 cards, then I would introduce custom props for the gap. This way if we change the value, we don't break the grid. (`--gap: 1.3em` used by both `gap` and your calc).

In any case, I also think we could handle responsive differently (more of a design concern than a code concern):
- if we have enough space, 4 cards on the same line,
- at a given breakpoint 2 cards per line,
- at a given breakpoint 1 card per line.

I find this more elegant for this type of layout but I'll let you and @marion-izd discuss / decide:
(currently we have this)

Image

(I suggest we move to this)

Image

@clara-layus-cc this is a great usecase for [`subgrid`](https://developer.mozilla.org/fr/docs/Web/CSS/Guides/Grid_layout/Subgrid), we can discuss all of this in sync if you want 😉

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
florian-sanders-cc: This is tricky, I'm not sure what to do, we need to take into account our customized consoles in the wild. 

Hardcoded colors (even line 330 & 331) will make it impossible for custom themes to customize this component. 
Are we ok with this tradeoff? We need to check the current custom theme we have and whether this component will look weird or not in their design.

Let's discuss this subject with the team when we can, maybe we should introduce new tokens for this 🤷 (I'm not sure what to do because this is for the legacy console anyway and this is supposed to be a quick win).

florian-sanders-cc (changes requested):
Nice component @clara-layus-cc my main concern is that we should use links instead of buttons in most cases + a form to contain the select + button (and also benefit from errors when no value is selected)

As usual do not hesitate to ping me if you want to discuss some feedback because some stuff is hard to explain in PR comments 🙈

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
florian-sanders-cc: Not so sure about this feedback actually, maybe wait for other reviews :sweat_smile:

pdesoyres-cc: most of the time we don't do that... but why not

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
pdesoyres-cc: yes, we should avoid nested ternaries. it's too hard to read. In that case, you can create a method with early returns

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
pdesoyres-cc: I agree. This is a lot of events that could be replaced by URLs

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,402 @@
...
+          display: flex;
+          flex-direction: column;
+          gap: 1.25em;
+          min-height: 24em;
+          padding: 2em;
+          text-align: left;
+        }
+
+        .blue-card {
+          background: linear-gradient(135deg, #3d5fb4 0%, #6d1cf0 100%);
```
pdesoyres-cc, line 398: With this color, the `<cc-button>` focus ring doesn't fit well. I don't know how we could fix that 🤔

pdesoyres-cc (commented):
Well done @clara-layus-cc

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
clara-layus-cc: Personally, I don't really know, but I'm also curious to hear other opinions.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
clara-layus-cc: Apart from the code for a card, the rest is fairly straightforward, so I don't necessarily see the point of splitting it into two components, but I'm curious to hear other people's opinions.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
clara-layus-cc: Good point, I'll discuss it with the team.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  /**
+   * @param {HomepageOnboardingCard} card
+   * @param {FormDataMap} formData
+   */
+  _onCardFormSubmit(card, formData) {
+    const orgId = /** @type {string} */ (formData.organisation);
+    window.location.href = card.href.replace('{orgId}', orgId);
+  }
```
florian-sanders-cc, line 199: **fix:** component should not update the location themselves, this is the console's job. The pattern replace is good I think but the format is usually this: 

```
          openScalabilityLink: scalabilityUrlPattern.replace(':id', phpAppId),
```

In any case, you should dispatch an event and let the console redirect

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
florian-sanders-cc: Sorry my feedback wasn't clear: should remove `width: 100%;` instead of `display: block;`

src/translations/translations.en.js
```diff
@@ -977,6 +898,40 @@ export const translations = {
...
+  'cc-homepage-onboarding.card.new-resource.description': `Add applications, add-ons, databases, storage...`,
+  'cc-homepage-onboarding.card.new-resource.select.placeholder': `-- Select --`,
+  'cc-homepage-onboarding.card.new-resource.select.title': `Choose an organisation`,
+  'cc-homepage-onboarding.card.new-resource.title': `Launch new resource`,
+  'cc-homepage-onboarding.card.secure.button': `Activate 2FA`,
+  'cc-homepage-onboarding.card.secure.description': `Enhance your account security with 2FA, highly recommended to protect your access.`,
+  'cc-homepage-onboarding.card.secure.title': `Secure your account`,
+  'cc-homepage-onboarding.card.ssh-keys.button': `Add SSH key`,
+  'cc-homepage-onboarding.card.ssh-keys.description': `Secure your connections and simplify server access with SSH key authentication.`,
+  'cc-homepage-onboarding.card.ssh-keys.title': `Configure SSH Keys`,
```
roberttran-cc: **🤓 nitpick:** 


```suggestion
  'cc-homepage-onboarding.card.ssh-keys.title': `Configure SSH keys`,
```

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
roberttran-cc: Yeah, displaying nothing is maybe a bit much, especially on this component which will have a significant exposure.

On the other hand, it's the policy we chose for many other components (based on the assumption that network errors should not happen often).

I guess it's okay to leave it as is, to save time, for simplicity and homogeneity.

src/components/cc-homepage-onboarding/cc-homepage-onboarding.js
roberttran-cc: Good intention.

But I think it's mostly useful when the card implementation is complex or when it would be used among several components. And it does not look like it's the case here (I may be wrong though).

So I would leave it as is for the sake of simplicity and extract it later if needed.

PR #1671, opened 2026-02-16
feat(cc-homepage-video): init component

clara-layus-cc:
# What does this PR do?

 - Adding a component to display a video

# How to review?

- Check the commits ,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/homepage-video/index.html?path=/story/%F0%9F%9B%A0-homepage-cc-homepage-video--default-story)

### DIFF
--- a/src/components/cc-homepage-video/cc-homepage-video.js
+++ b/src/components/cc-homepage-video/cc-homepage-video.js
@@ -1,0 +1,181 @@
+import { css, html, LitElement } from 'lit';
+import { createRef, ref } from 'lit/directives/ref.js';
+import { iconRemixPlayFill } from '../../assets/cc-remix.icons.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-block/cc-block.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-link/cc-link.js';
+
+/**
+ * A component displaying a YouTube video thumbnail with a play button overlay.
+ *
+ * Clicking the play button replaces the thumbnail with an embedded YouTube iframe.
+ *
+ * @cssdisplay block
+ */
+export class CcHomepageVideo extends LitElement {
+  static get properties() {
+    return {
+      channelUrl: { type: String, attribute: 'channel-url' },
+      videoUrl: { type: String, attribute: 'video-url' },
+      _playing: { type: Boolean, state: true },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {string} YouTube video URL (e.g. https://youtu.be/xxx or https://www.youtube.com/watch?v=xxx) */
+    this.videoUrl = '';
+
+    /** @type {string} URL to the YouTube channel */
+    this.channelUrl = '';
+
+    /** @type {boolean} */
+    this._playing = false;
+
+    /** @type {import('lit/directives/ref.js').Ref<HTMLIFrameElement>} */
+    this._iframeRef = createRef();
+  }
+
+  async _onPlay() {
+    this._playing = true;
+    await this.updateComplete;
+    this._iframeRef.value.focus();
+  }
+
+  /**
+   * Extracts the video ID from a YouTube URL.
+   * Supports youtu.be/ID, youtube.com/watch?v=ID, and youtube.com/embed/ID formats.
+   * @param {string} url
+   * @returns {string}
+   */
+  _extractVideoId(url) {
+    try {
+      const parsed = new URL(url);
+      if (parsed.hostname === 'youtu.be') {
+        return parsed.pathname.slice(1);
+      }
+      if (parsed.hostname.includes('youtube.com')) {
+        if (parsed.pathname.startsWith('/embed/')) {
+          return parsed.pathname.split('/')[2];
+        }
+        return parsed.searchParams.get('v');
+      }
+    } catch {
+      // not a valid URL
+    }
+    return 'xxxxxxxxxxx';
+  }
+
+  render() {
+    const videoId = this._extractVideoId(this.videoUrl);
+    const thumbnailUrl = `https://img.youtube.com/vi/${videoId}/maxresdefault.jpg`;
+    const embedUrl = `https://www.youtube.com/embed/${videoId}?autoplay=1`;
+
+    return html`
+      <cc-block class="block">
+        <div slot="header-title" class="header">${i18n('cc-homepage-video.title')}</div>
+        <cc-link slot="header-right" href=${this.channelUrl}>${i18n('cc-homepage-video.link')}</cc-link>
+        <div slot="content" class="video-container">
+          ${this._playing
+            ? html`
+                <iframe
+                  ${ref(this._iframeRef)}
+                  tabindex="-1"
+                  src=${embedUrl}
+                  title=${i18n('cc-homepage-video.iframe-title')}
+                  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; fullscreen"
+                ></iframe>
+              `
+            : html`
+                <button class="thumbnail" @click=${this._onPlay}>
+                  <img src=${thumbnailUrl} alt="" />
+                  <div class="play-button">
+                    <cc-icon .icon=${iconRemixPlayFill} size="xl" a11y-name=${i18n('cc-homepage-video.play')}></cc-icon>
+                  </div>
+                </button>
+              `}
+        </div>
+      </cc-block>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: block;
+        }
+
+        .block {
+          border: solid 1px var(--cc-color-border-neutral-weak);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          display: block;
+          padding: 1em;
+        }
+
+        .video-container {
+          aspect-ratio: 16 / 9;
+          margin-top: -0.6em;
+          position: relative;
+        }
+
+        .video-container iframe {
+          border: none;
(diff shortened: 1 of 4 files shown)


marion-izd:
Great work 👍 
 I don't know if it's my browser or not but it's seems the font size is big :face_with_monocle:
It seems that the title is not the same size as that of the blocks, blocks details: normally it is 19px and here I see the preview at 22px. The same goes for the link: in my preview, the link is 19px, whereas the link component is more around 15 or 16px

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,183 @@
...
+      }
+      if (parsed.hostname.includes('youtube.com')) {
+        if (parsed.pathname.startsWith('/embed/')) {
+          return parsed.pathname.split('/')[2];
+        }
+        return parsed.searchParams.get('v');
+      }
+    } catch {
+      // not a valid URL
+    }
```
HeleneAmouzou, line 67: **question:** should we handle this error more precisely ?

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,183 @@
...
+            `
+          : html`
+              <button class="thumbnail" @click=${this._onPlay}>
+                <img src=${thumbnailUrl} alt=${i18n('cc-homepage-video.thumbnail-alt')} />
+                <div class="play-button">
+                  <cc-icon .icon=${iconRemixPlayFill} size="xl" a11y-name=${i18n('cc-homepage-video.play')}></cc-icon>
+                </div>
+              </button>
+            `}
+      </div>
```
HeleneAmouzou: **suggestion:** maybe we could use the `cc-block` here instead of divs and get rid of some css ?

src/components/cc-homepage-video/cc-homepage-video.types.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface HomepageVideoProperties {
+  videoUrl: string;
+  channelUrl: string;
+}
```
HeleneAmouzou: **chore:** unused type

HeleneAmouzou (changes requested):
GG @clara-layus-cc ! 👏 

I left few feedbacks and I have two questions : 
- ~what is happening if the fetch to the video URL fails ? I was wondering how the component would react in error state, but also in loading while the page is not ready yet~ => I think I got my anwser, if the fetch fails, it will dislay an error from Youtube right ? 
- Is this component suppose to only handle Youtube videos or is it going to evolve later ?

src/components/cc-homepage-video/cc-homepage-video.js
florian-sanders-cc: :+1: @HeleneAmouzou 's suggestion should also fix the heading color (it's not exactly black, should be the "blue strongest" color).

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        .link {
+          font-size: 1em;
+        }
+
+        .video-container {
+          aspect-ratio: 16 / 9;
+          background-color: var(--cc-color-bg-strong, #000);
+          border-radius: var(--cc-border-radius-default, 0.25em);
```
florian-sanders-cc: **suggestion:** the `border-radius` is higher in the wireframe, you could hardcode it and we'll add a new design token later on as discussed this morning

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,183 @@
...
+          display: block;
+          height: 100%;
+          padding: 0;
+          position: relative;
+          width: 100%;
+        }
+
+        .thumbnail:focus-visible {
+          outline: var(--cc-focus-outline, #000 solid 2px);
+          outline-offset: -2px;
```
florian-sanders-cc: **suggestion:** I find the focus barely visible, in this case, I'd rather use the regular offset `--cc-focus-outline-offset`

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,183 @@
...
+          ? html`
+              <iframe
+                src=${embedUrl}
+                title=${i18n('cc-homepage-video.iframe-title')}
+                allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; fullscreen"
+              ></iframe>
+            `
+          : html`
+              <button class="thumbnail" @click=${this._onPlay}>
+                <img src=${thumbnailUrl} alt=${i18n('cc-homepage-video.thumbnail-alt')} />
```
florian-sanders-cc: **fix:** this alt will be part of the button name, making the final button name: `YouTube video thumbnail Play video`

For images in interactive elements (buttons, links, etc.), here's how you can decide what to do:
- your interactive element needs a clear accessible name that tells users what it does,
- the accessible name is computed by aggregating all accessible names found in the content of the interactive element (text content, `alt` values, `aria-label`, `aria-labelledby` etc.),
- so you need to provide either:
  - a text (which you can hide visually with the `visually-hidden` class) if you don't want it to be visible,
  - the `alt` value of your image, 
  - the `aria-label` value on an SVG tag (this is what the `cc-icon` does when you provide an `a11y-name` value).

Since you chose to provide an `a11y-name` for the `cc-icon` that clearly tells users about the button purpose, the `alt` attribute of your image should be left empty (do not remove the `alt` attribute itself, just its value, you should have `alt=""` so that screen readers do not try to read the filename, this tells them "this image is decorative, don't bother").

If any of this is unclear, ping me and we'll discuss in sync 😉 

```suggestion
                <img src=${thumbnailUrl} alt="" />
```

src/translations/translations.en.js
```diff
@@ -977,6 +898,12 @@ export const translations = {
   'cc-heptapod-info.storage-bytes': /** @param {{storage: number}} _ */ ({ storage }) => formatBytes(storage, 1),
   'cc-heptapod-info.storage-description': `Storage size`,
   //#endregion
+  //#region cc-homepage-video
+  'cc-homepage-video.iframe-title': `YouTube video`,
+  'cc-homepage-video.link': `See our Youtube channel`,
```
roberttran-cc: **🤓 nitpick:** 

```suggestion
  'cc-homepage-video.link': `See our YouTube channel`,
```

src/translations/translations.fr.js
```diff
@@ -989,6 +910,12 @@ export const translations = {
   'cc-heptapod-info.storage-bytes': /** @param {{storage: number}} _ */ ({ storage }) => formatBytes(storage, 1),
   'cc-heptapod-info.storage-description': `Stockage utilisé`,
   //#endregion
+  //#region cc-homepage-video
+  'cc-homepage-video.iframe-title': `Vidéo YouTube`,
+  'cc-homepage-video.link': `Voir notre chaîne Youtube`,
```
roberttran-cc: **🤓 nitpick:** 

```suggestion
  'cc-homepage-video.link': `Voir notre chaîne YouTube`,
```

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,169 @@
...
+              `}
+        </div>
+      </cc-block>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
```
roberttran-cc, line 107: **🔨 issue:** 

```suggestion
      css`
        :host {
          display: block;
        }
```

roberttran-cc (approved):
LGTM! A little issue with the style but I'll approve nonetheless.

src/components/cc-homepage-video/cc-homepage-video.js
clara-layus-cc: The case where the URL is invalid is handled by the return at the end, as is the case where the URL is valid but not a YouTube URL.

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -0,0 +1,173 @@
...
+
+    /** @type {string} URL to the YouTube channel */
+    this.channelUrl = '';
+
+    /** @type {boolean} */
+    this._playing = false;
+  }
+
+  _onPlay() {
+    this._playing = true;
```
florian-sanders-cc, line 42: **fix:** so sorry, I didn't spot this issue during the first review but it should be "quick" to fix

Since the `button` is removed from the DOM upon playing the video, we need to move the `focus` manually so that it's not "lost" (the element being focused is removed, the focus needs to go somewhere).

To do so:
1. add a `tabindex="-1"` on the `<iframe>` tag (if you remember our sync on a11y, it means: "this element may receive focus but it's not in the tab order"),
2. add a `ref` on the `<iframe>` tag,
3. after setting `this._playing = true` here, move the focus to the iframe.


```suggestion
    this._playing = true;
    this._iframeRef.value.focus();
```

florian-sanders-cc (changes requested):
One last fix, sorry @clara-layus-cc this is my mistake 😉 
Apart from that, everything looks good!

src/components/cc-homepage-video/cc-homepage-video.js
florian-sanders-cc: Didn't think about this when I suggested this change: the iframe is not in the DOM directly, we need to wait for the next render before trying to focus it


```suggestion
    this._playing = true;
    await this.updateComplete;
    this._iframeRef.value.focus();
```

PR #1718, opened 2026-04-08
`cc-network-group-member-*`: support `DELETED` resources

florian-sanders-cc:
## What does this PR do?

When Clever Cloud resources are deleted, they are not cleared from Network Groups.
This means we need to support members referring to deleted resources.

This PR adds a new type of member card to show members referring to deleted resources and allow users to unlink these "ghost" members.

## How to review?

- Check the commits,
- Check the stories,
- Small QA:
  1. From your prod console Create an app / addon in `ACME BAR`,
  2. From your local demo-smart, link your newly created resource to the Network Group used for testing,
  3. From your prod console, delete the resource,
  4. Refresh your local demo-smart: you should see the ghost member as last member with different look (opacity, sentence + plain danger button),
  5. Delete the ghost member: you should get no confirm dialog, directly into waiting, then success toast. (no need to test other stuff since the only difference with other members is we skip the dialog, the rest of the unlink process / code is exactly the same).

### DIFF
--- a/src/components/cc-network-group-member-card/cc-network-group-member-card.events.js
+++ b/src/components/cc-network-group-member-card/cc-network-group-member-card.events.js
@@ -1,14 +1,16 @@
 import { CcEvent } from '../../lib/events.js';
 
+/** @import { NetworkGroupMember } from './cc-network-group-member-card.types.js'; */
+
 /**
  * Dispatched when the user requests to unlink a member.
  * The parent component should handle the confirmation dialog.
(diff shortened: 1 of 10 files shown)


pdesoyres-cc (approved):
LGTM

- the source code seems fine
- I did the QA with Firefox: all good.

Galimede (approved):
Checked the code and done the QA, LGTM! 💪

HeleneAmouzou (approved):
All good 👍

PR #1719, opened 2026-04-09
`cc-network-group-list`: support unlinking

florian-sanders-cc:
## What does this PR do?

- Adds `unlink` buttons to the `cc-network-group-list` component:
  - this allows users to unlink a Network Group right from this screen instead of having to access the members and peers screen to do so.

## How to review?

- Check the commits,
- Play with the new feature using the `demo-smart` locally, try to block a few `delete` requests just to make sure we're good :wink:

### DIFF
--- a/src/components/cc-network-group-list/cc-network-group-list.events.js
+++ b/src/components/cc-network-group-list/cc-network-group-list.events.js
@@ -13,3 +13,17 @@ export class CcNetworkGroupLinkEvent extends CcEvent {
     super(CcNetworkGroupLinkEvent.TYPE, detail);
   }
 }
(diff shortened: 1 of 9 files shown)


pdesoyres-cc (dismissed):
- I played with the demo smart and everything works fine 
- The smart now uses `DeleteNetworkGroupMemberCommand` so we need to change the `smart.md` file accordingly.

Galimede (approved):
LGTM, thanks Florian! 💪

HeleneAmouzou (approved):
Reviewed and tested on demo-smart and everything works fine !

PR #1720, opened 2026-04-09
feat(cc-network-group-member-list): only allow specific providers

florian-sanders-cc:
## What does this PR do?

- Network Groups are not supported for all add-on providers (although the API lets you set them up for any add-on provider, they don't do anything for Matomo, Otoroshi, Metabase, Keycloak, Cellar, FSBucket, Jenkins, etc.).
- `dev` plans also should not support Network Groups,
- This means we need to filter out irrelevant resources from our select in the UI,
- Ideally, we would rely on the API to get the list of supported providers or resources but it's not possible at the moment so we need a temporary solution: we rely on the client to provide this temporary solution with helpers that filter out the list.
- Note that we also still expose underlying resources from operators (apps, add-ons) because it could make sense for some users to add a NG to one of those (in order to communicate with them through an external peer).

## How to review?

- Check the commit,
- Play with the component in `demo-smart`, you should only see relevant add-ons (`'es-addon', 'mongodb-addon', 'mysql-addon', 'postgresql-addon', 'redis-addon'`),

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "25.1.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@clevercloud/client": "^11.5.0",
+        "@clevercloud/client": "^12.3.0",
         "@lit-labs/motion": "^1.0.7",
         "@lit-labs/virtualizer": "^2.0.14",
         "@shoelace-style/shoelace": "^2.18.0",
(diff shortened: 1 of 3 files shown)


HeleneAmouzou (commented):
Maybe I am missing something, but on `demo-smart` I can still see operators on the list, shouldn't there be filter out ? 🤔

HeleneAmouzou (approved):

roberttran-cc (approved):

PR #1725, opened 2026-04-17
Visual tests : fix stories order and move to prod CDN

florian-sanders-cc:
## What does this PR do?

- Fixes the visual tests report menu to always display the `defaultStory` stories first,
- Makes the report rely on the prod CDN instead of the preview now that components for the report are available in prod (components were part of the PR that initially added the report so we had to rely on the preview at first).

## How to review?

- Check the commits,

### DIFF
--- a/src/components/cc-visual-tests-report-menu/cc-visual-tests-report-menu.js
+++ b/src/components/cc-visual-tests-report-menu/cc-visual-tests-report-menu.js
@@ -73,9 +73,14 @@ export class CcVisualTestsReportMenu extends LitElement {
         return componentCompare;
       }
 
-      const storyCompare = a.storyName.localeCompare(b.storyName);
-      if (storyCompare !== 0) {
-        return storyCompare;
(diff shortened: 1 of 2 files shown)


roberttran-cc (approved):
LGTM although I have a question.

Any reason to pin `?version=25.1.0` explicitly (both for the component and the style)? Omitting the `version` parameter (so the CDN falls back to the latest) seems simpler, unless I'm missing something. 🤔

florian-sanders-cc:
@roberttran-cc I hesitated but it means we don't have to think about breaking change adaptations this way :shrug: 
Though it's true that the report itself is part of this codebase so if we were to create a breaking change on low level components that it relies on, our analysis would probably catch it and fix it directly? 

Do you have a preference @roberttran-cc? I'm not sure, I find both acceptable :shrug:

roberttran-cc:
I see what you mean: considering that, a pinned version is less troublesome after all. In that case, we should add a small comment to indicate it's a deliberate decision.

PR #1729, opened 2026-04-22
feat(cc-network-group-list): show notice if add-on plan is dev

florian-sanders-cc:
## What does this PR do?

- For add-ons on a dev plan, Network Groups are unsupported so we show a notice inviting people to migrate to a paid plan.
- Why don't we show specific notice for other "unsupported" add-ons (operators, materia, etc.) ?
  - the console won't show the Network Groups tab for such add-ons,
  - If users reach the route by entering the URL manually:
    -  worst case scenario it doesn't break stuff it's just useless to add a Network Group to such resources,
    -  the API should not allow any interaction (right now it might, but should be fixed someday).

## How to review?

- Check the new story,
- Check `demo-smart` with the `Network Group (add-on with dev plan)` context button,

### DIFF
--- a/src/components/cc-network-group-list/cc-network-group-list.js
+++ b/src/components/cc-network-group-list/cc-network-group-list.js
@@ -25,7 +25,7 @@ import '../cc-select/cc-select.js';
 import { CcNetworkGroupLinkEvent, CcNetworkGroupUnlinkEvent } from './cc-network-group-list.events.js';
 
 /**
- * @import { NetworkGroupListState, NetworkGroupLinkFormState, NetworkGroup } from './cc-network-group-list.types.js'
+ * @import { NetworkGroupListState, NetworkGroup } from './cc-network-group-list.types.js'
  * @import { Ref } from 'lit/directives/ref.js'
  * @import { Option } from '../cc-select/cc-select.types.js'
  * @import { CcLink } from '../cc-link/cc-link.js'
@@ -47,9 +47,9 @@ import { CcNetworkGroupLinkEvent, CcNetworkGroupUnlinkEvent } from './cc-network
 export class CcNetworkGroupList extends LitElement {
   static get properties() {
     return {
-      linkFormState: { type: Object, attribute: 'link-form-state' },
-      listState: { type: Object, attribute: 'list-state' },
       resourceId: { type: String, attribute: 'resource-id' },
+      state: { type: Object },
+      _isUnlinkDialogOpen: { type: Boolean, state: true },
       _networkGroupIdToUnlink: { type: String, state: true },
     };
   }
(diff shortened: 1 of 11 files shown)


HeleneAmouzou (approved):
LGTM ! 👏 I was just wondering if the `cc-notice` should have a `warning` intent instead of an `info` one ? 🤔

florian-sanders-cc:
@HeleneAmouzou i hestitated, I don't know :sweat_smile: I'll ask Marion for her opinion :+1:

src/components/cc-network-group-list/cc-network-group-list.js
```diff
@@ -114,50 +127,63 @@ export class CcNetworkGroupList extends LitElement {
...
   /** @param {PropertyValues<CcNetworkGroupList>} changedProperties */
   willUpdate(changedProperties) {
-    if (changedProperties.has('listState')) {
-      const previousListState = changedProperties.get('listState');
-      const wasUnlinking = previousListState?.type === 'unlinking';
-      const isNotUnlinking = this.listState.type !== 'unlinking';
+    if (changedProperties.has('state')) {
+      const previousState = changedProperties.get('state');
+      const wasUnlinking = previousState?.type === 'loaded' && previousState?.listState?.type === 'unlinking';
+      console.log(this.state);
```
roberttran-cc: **🧹 chore:** wandering `console.log`

src/components/cc-network-group-list/cc-network-group-list.types.d.ts
```diff
@@ -1,60 +1,63 @@
...
+  type: 'loading';
+}
+
+export interface NetworkGroupListStateError {
+  type: 'error';
+}
+
+export interface NetworkGroupListStateUnsupported {
+  type: 'unsupported';
+  addonMigrationScreenUrl: string;
```
roberttran-cc, line 20: **❓ question:** not marked as optional when optional in `src/components/cc-network-group-list/cc-network-group-list.smart.js`

src/components/cc-network-group-list/cc-network-group-list.smart.js
```diff
@@ -118,18 +165,21 @@ defineSmartComponent({
...
       try {
-        await ccApiClient.send(new DeleteNetworkGroupMemberCommand({ memberId: resourceId, networkGroupId, ownerId }), {
-          signal,
-        });
+        await ccApiClient.send(
+          new DeleteNetworkGroupMemberCommand({ memberId: resolvedResourceId, networkGroupId, ownerId }),
+          {
+            signal,
+          },
+        );
```
roberttran-cc, line 175: **❓ question:** why passing the signal here and not for `CreateNetworkGroupMemberCommand`? Not really an issue but not consistent.

roberttran-cc (approved):
LGTM! A very minor feedback and 2 questions, therefore the approval.

src/components/cc-network-group-list/cc-network-group-list.types.d.ts
florian-sanders-cc: Yes it's confusing and it's smell but the smart might be used for an app or an add-on.

If it's an app:
- it doesn't have a migration screen,
- but it cannot be unsupported anyway (at least for now).

If it's an add-on:
- it can be unsupported,
- if it is, we need the migrationScreenUrl for sure.

The types can be revisited if one day some apps become unsupported, a refactoring will be necessary for such case anyway (the error message wouldn't be the same).

src/components/cc-network-group-list/cc-network-group-list.smart.js
florian-sanders-cc: This is a mistake :+1: shouldn't pass signal for delete, only `get`, nice catch :wink:

PR #1732, opened 2026-04-22
fix(cc-homepage-template-project): open project links in a new tab

clara-layus-cc:
# What does this PR do?

- Force links to open in new tabs

# How to review?

- Check the commit,
- Check the [preview]()
- 1 reviewers should be enough.

### DIFF
--- a/src/components/cc-homepage-template-project/cc-homepage-template-project.js
+++ b/src/components/cc-homepage-template-project/cc-homepage-template-project.js
@@ -65,7 +65,7 @@ export class CcHomepageTemplateProject extends LitElement {
   _renderProjectRow(project) {
     return html`
       <li>
-        <a href="${project.href}" class="project-row">
+        <a href="${project.href}" target="_blank" rel="noopener noreferrer" class="project-row">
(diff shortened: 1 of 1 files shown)


src/components/cc-homepage-template-project/cc-homepage-template-project.js
```diff
@@ -65,7 +65,7 @@ export class CcHomepageTemplateProject extends LitElement {
   _renderProjectRow(project) {
     return html`
       <li>
-        <a href="${project.href}" class="project-row">
+        <a href="${project.href}" target="_blank" rel="noopener noreferrer" class="project-row">
```
florian-sanders-cc, line 68: **nitpick/note:** `noopener` is not necessary when using `target=_blank` (it used to be but nowadays browsers automatically handle this subject) but it doesn't harm to set it so I don't mind if it says 😉 

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener

florian-sanders-cc (approved):
LGTM!

PR #1733, opened 2026-04-27
feat(cc-search-bar): init component

clara-layus-cc:
# What does this PR do?

- Creating a new search bar component

# How to review?

- Check the commit,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/search-bar/index.html?path=/story/%F0%9F%A7%AC-atoms-cc-search-bar--default-story)

### DIFF
--- a/src/components/cc-search-bar/cc-search-bar.js
+++ b/src/components/cc-search-bar/cc-search-bar.js
@@ -1,0 +1,391 @@
+import { css, html, LitElement, nothing } from 'lit';
+import {
+  iconRemixArrowRightSLine as iconArrowRight,
+  iconRemixExternalLinkLine as iconExternalLink,
+  iconRemixSearchLine as iconSearch,
+} from '../../assets/cc-remix.icons.js';
+import { isExternalUrl } from '../../lib/utils.js';
+import { i18n } from '../../translations/translation.js';
+import '../cc-badge/cc-badge.js';
+import '../cc-dialog/cc-dialog.js';
+import '../cc-icon/cc-icon.js';
+import '../cc-input-text/cc-input-text.js';
+
+/**
+ * @import { SearchBarItem, SearchBarItemType, SearchBarSection } from './cc-search-bar.types.js'
+ * @import { BadgeIntent } from '../cc-badge/cc-badge.types.js'
+ * @import { CcInputEvent } from '../common.events.js'
+ */
+
+const KEYWORD_TOKEN_REGEX = /^is:./;
+
+/** @type {Record<SearchBarItemType, { label: string, intent: BadgeIntent }>} */
+const ITEM_TYPE_BADGE = {
+  app: { label: 'APP', intent: 'success' },
+  addon: { label: 'ADDON', intent: 'warning' },
+  'network-group': { label: 'NG', intent: 'neutral' },
+  cke: { label: 'KUBE', intent: 'neutral' },
+};
+
+/**
+ * A search bar dialog that displays categorized results.
+ *
+ * ## Details
+ *
+ * The component wraps a search input and a list of sections inside a `cc-dialog`.
+ * Items can have optional badges and external link indicators.
+ *
+ * @cssdisplay contents
+ *
+ */
+export class CcSearchBar extends LitElement {
+  static get properties() {
+    return {
+      open: { type: Boolean, reflect: true },
+      sections: { type: Array },
+      value: { type: String },
+    };
+  }
+
+  constructor() {
+    super();
+
+    /** @type {boolean} Displays or hides the search bar dialog. */
+    this.open = false;
+
+    /** @type {SearchBarSection[]} The sections to display, each containing a label, icon, and items. */
+    this.sections = [];
+
+    /** @type {string} The current search input value. */
+    this.value = '';
+  }
+
+  /** Opens the search bar dialog by setting the `open` property to true. */
+  show() {
+    this.open = true;
+  }
+
+  /** Closes the search bar dialog by setting the `open` property to false. */
+  hide() {
+    this.open = false;
+  }
+
+  /**
+   * Filters `this.sections` based on `this.value`, using a token-based syntax.
+   *
+   * Tokens are split on whitespace and partitioned into:
+   * - keyword tokens (`is:<value>`): an item passes only if every keyword token
+   *   is present in its derived matchers (`is:<itemType>` and explicit `matchers`).
+   * - text tokens: an item passes only if every text token is `includes`'d in its
+   *   lowercased label or its lowercased id.
+   *
+   * @returns {SearchBarSection[]}
+   */
+  _getFilteredSections() {
+    const query = this.value.trim().toLowerCase();
+    if (query === '') {
+      return [];
+    }
+    const tokens = query.split(/\s+/);
+    const keywordTokens = tokens.filter((token) => KEYWORD_TOKEN_REGEX.test(token));
+    const textTokens = tokens.filter((token) => !KEYWORD_TOKEN_REGEX.test(token));
+
+    return this.sections
+      .map((section) => ({
+        ...section,
+        items: section.items.filter((item) => {
+          const itemMatchers = [...(item.itemType != null ? [`is:${item.itemType}`] : []), ...(item.matchers ?? [])];
+          const keywordsMatch = keywordTokens.every((keyword) => itemMatchers.includes(keyword));
+          if (!keywordsMatch) {
+            return false;
+          }
+          const label = item.label.toLowerCase();
+          const id = item.id?.toLowerCase() ?? '';
+          return textTokens.every((token) => label.includes(token) || id.includes(token));
+        }),
+      }))
+      .filter((section) => section.items.length > 0);
+  }
+
+  _onDialogClose() {
+    this.open = false;
+  }
+
+  /** @param {CcInputEvent} e */
+  _onInput(e) {
+    this.value = e.detail;
+  }
+
+  render() {
+    const filteredSections = this._getFilteredSections();
+    const hasItems = filteredSections.length > 0;
+    return html`
+      <cc-dialog ?open="${this.open}" @cc-close="${this._onDialogClose}">
+        <h1 slot="heading" class="heading">${i18n('cc-search-bar.heading')}</h1>
+        <div class="search-bar">
+          ${this._renderSearchInput()}
+          ${hasItems
+            ? html`<div class="sections">${filteredSections.map((section) => this._renderSection(section))}</div>`
+            : this._renderEmpty()}
+        </div>
+      </cc-dialog>
+    `;
+  }
+
+  _renderEmpty() {
+    const isInitial = this.value.trim() === '';
+    return html`
+      <div class="empty">
+        <cc-icon class="empty-icon" .icon="${iconSearch}" size="xl"></cc-icon>
+        <p class="empty-title">
+          ${isInitial ? i18n('cc-search-bar.initial.title') : i18n('cc-search-bar.no-result.title')}
+        </p>
+        <p class="empty-description">
+          ${isInitial ? i18n('cc-search-bar.initial.description') : i18n('cc-search-bar.no-result.description')}
+        </p>
+      </div>
+    `;
+  }
+
+  _renderSearchInput() {
+    return html`
+      <div class="input-wrapper">
+        <div class="input-field">
+          <cc-input-text
+            label="${i18n('cc-search-bar.label')}"
+            placeholder="${i18n('cc-search-bar.placeholder')}"
+            value="${this.value}"
+            ?autofocus="${true}"
+            @cc-input="${this._onInput}"
+          ></cc-input-text>
+          <cc-icon class="search-icon" .icon="${iconSearch}" size="sm"></cc-icon>
+        </div>
+      </div>
+    `;
+  }
+
+  /** @param {SearchBarSection} section */
+  _renderSection(section) {
+    return html`
+      <div class="section">
+        <h2 class="section-header">
+          <cc-icon .icon="${section.icon}" size="md"></cc-icon>
+          <span class="section-header-label">${section.label}</span>
+        </h2>
+        <ul class="section-items">
+          ${section.items.map((item) => this._renderItem(item))}
+        </ul>
+      </div>
+    `;
+  }
+
+  /** @param {SearchBarItem} item */
+  _renderItem(item) {
+    const isExternal = isExternalUrl(item.href);
+    const badge = item.itemType != null ? ITEM_TYPE_BADGE[item.itemType] : null;
+    const title = isExternal ? i18n('cc-search-bar.external-link.title', { linkText: item.label }) : nothing;
+    return html`
+      <li>
+        <a
+          class="item"
+          href="${item.href}"
+          target="${isExternal ? '_blank' : nothing}"
+          rel="${isExternal ? 'noreferrer' : nothing}"
+          title="${title}"
+        >
+          <span class="item-label">${item.label}</span>
+          ${badge != null ? html` <cc-badge intent="${badge.intent}" weight="dimmed">${badge.label}</cc-badge> ` : ''}
+          ${isExternal
+            ? html`
+                <cc-icon
+                  class="external-link-icon"
+                  .icon="${iconExternalLink}"
+                  size="md"
+                  a11y-name="${i18n('cc-search-bar.external-link.name')}"
+                ></cc-icon>
+              `
+            : html`<cc-icon class="hover-chevron" .icon="${iconArrowRight}" size="md"></cc-icon>`}
+        </a>
+      </li>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: contents;
+        }
+
+        cc-dialog::part(dialog) {
+          margin-block: 1em auto;
+          overflow: hidden;
+        }
+
+        .search-bar {
+          display: flex;
+          flex-direction: column;
+        }
+
+        .heading {
+          font-size: 1.125em;
+          margin: 0;
+        }
+
+        .input-wrapper {
+          display: flex;
+          flex-direction: column;
+          gap: 0.35em;
+          padding: 0.5em 0.5em 1em;
+        }
+
+        .input-field {
+          position: relative;
+        }
+
+        .input-field cc-input-text {
+          display: block;
+          font-size: 0.8em;
+          width: 100%;
+        }
+
+        .search-icon {
+          bottom: 0.8em;
+          color: var(--cc-color-text-default, #000);
+          pointer-events: none;
+          position: absolute;
+          right: 0.67em;
+          transform: translateY(50%);
+        }
+
+        .sections {
+          flex: 1;
+          max-height: calc(100dvh - 16em);
+          overflow-y: auto;
+          padding: 0 0.5em;
+        }
+
+        .empty {
+          align-items: center;
+          color: var(--cc-color-text-default, #000);
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
+          padding: 1.5em 1em;
+          text-align: center;
+        }
+
+        .empty-icon {
+          color: var(--cc-color-text-primary-strongest, #000);
+          margin-bottom: 0.25em;
+        }
+
+        .empty-title {
+          font-size: 0.78em;
+          font-weight: bold;
+          margin: 0;
+        }
+
+        .empty-description {
+          color: var(--cc-color-text-weak, #666);
+          font-size: 0.78em;
+          line-height: 1.5;
+          margin: 0;
+        }
+
+        .empty-description code {
+          background-color: var(--cc-color-bg-neutral, #f5f5f5);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          font-family: var(--cc-ff-monospace, monospace);
+          padding: 0.1em 0.3em;
+        }
+
+        .section:not(:last-child) {
+          border-bottom: solid 1px var(--cc-color-border-neutral-weak, #e7e7e7);
+          padding-bottom: 1em;
+        }
+
+        .section:not(:first-child) {
+          padding-top: 0.5em;
+        }
+
+        .section-header {
+          align-items: center;
+          color: var(--cc-color-text-weak, #666);
+          display: flex;
+          font-size: 1em;
+          font-weight: normal;
+          gap: 0.4em;
+          line-height: 1.3;
+          margin: 0;
+          padding: 1em 0;
+        }
+
+        .section-header-label {
+          font-size: 0.75em;
+        }
+
+        .section-items {
+          display: flex;
+          flex-direction: column;
+          gap: 0.5em;
+          list-style: none;
+          margin: 0;
+          padding: 0.33em 0;
+        }
+
+        .item {
+          align-items: center;
+          border-radius: 0.5em;
+          color: var(--cc-color-text-default, #000);
+          display: flex;
+          font-weight: normal;
+          gap: 0.5em;
+          letter-spacing: -0.15px;
+          padding: 0.5em;
+          text-decoration: none;
+        }
+
+        .item:hover {
+          background: var(--cc-color-bg-neutral, #f5f5f5);
+        }
+
+        .item:focus-visible {
+          border-radius: 0.625em;
+          outline: var(--cc-focus-outline);
+          outline-offset: var(--cc-focus-outline-offset, 2px);
+        }
+
+        .item-label {
+          flex: 1;
+          font-size: 0.875em;
+          max-width: 80%;
+          min-width: 0;
+          overflow-wrap: anywhere;
+        }
+
+        .item-label + * {
+          margin-left: auto;
+        }
+
+        .external-link-icon {
+          color: var(--cc-color-text-primary-strongest, #012a51);
+          flex-shrink: 0;
+        }
+
+        .hover-chevron {
+          color: var(--cc-color-text-primary, #1a51b3);
+          display: none;
+          flex-shrink: 0;
+        }
+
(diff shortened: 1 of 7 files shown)


src/components/cc-search-bar/cc-search-bar.events.js
```diff
@@ -0,0 +1,16 @@
+import { CcEvent } from '../../lib/events.js';
+
+/**
+ * Dispatched when the search input value changes.
+ * @extends {CcEvent<string>}
+ */
+export class CcSearchBarInputEvent extends CcEvent {
+  static TYPE = 'cc-search-bar:input';
```
pdesoyres-cc: fix: we usually do not use colon separator in event names.

```suggestion
  static TYPE = 'cc-search-bar-input';
```

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+    `;
+  }
+
+  /** @param {SearchBarSection} section */
+  _renderSection(section) {
+    return html`
+      <div class="section">
+        <h3 class="section-header">
+          <cc-icon .icon="${section.icon}" size="sm"></cc-icon>
+          <span>${section.label}</span>
```
pdesoyres-cc: I find the icon and the text a bit too small

pdesoyres-cc (changes requested):
Nice work @clara-layus-cc

src/components/cc-search-bar/cc-search-bar.types.d.ts
```diff
@@ -0,0 +1,16 @@
+import { BadgeIntent } from '../cc-badge/cc-badge.types.js';
+import { IconModel } from '../common.types.js';
+
+export interface SearchBarItem {
+  label: string;
+  href: string;
+  badge?: string | null;
+  badgeIntent?: BadgeIntent;
```
florian-sanders-cc: **suggestion:** you could expose an "itemType" prop that would be a union of `"app" | "addon" | "network-group" | "cke"` so that the component owns the "how it should look" part (maps each itemType to a badge style) and the console only owns the "what am I dealing with / categorization" part.

It can remain as it is, it's not bad, but it's opening to unrealistic / impossible states (you could set a `badgeIntent` but not a `badge` or the opposite) and it's making the console responsible for a small part of UI concern mixed in the data concern. (this is definitely not a big issue, I'll let you decide)

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+ *
+ * ## Details
+ *
+ * The component displays a search input and a list of sections, each containing items.
+ * Items can have optional badges and external link indicators.
+ *
+ * @cssdisplay block
+ *
+ * @fires {CcSearchBarInputEvent} cc-search-bar-input - Fires when the search input value changes.
+ * @fires {CcCloseEvent} cc-close - Fires when the search bar is closed.
```
florian-sanders-cc: **nitpick:** we don't really set these anymore because everything is handled through the events-map.
Setting these here means they might end up out of sync with the actual event names used (if you refactor the event name, we may easily forget to update them since we don't use these anymore).

I think we can remove them, although if you prefer to keep them, it's not that big of a deal 😉

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+          cursor: pointer;
+          display: flex;
+          height: 1.67em;
+          justify-content: center;
+          padding: 0;
+          width: 1.67em;
+        }
+
+        .close-button:hover {
+          background: var(--color-grey-10, #f5f5f5);
```
Galimede: issue: we should not use a color choice token, we should either use a decision token that already exist or create a new one.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+
+  /** @param {SearchBarItem} item */
+  _renderItem(item) {
+    return html`
+      <li>
+        <a
+          class="item"
+          href="${item.href}"
+          target="${item.externalLink ? '_blank' : nothing}"
+          rel="${item.externalLink ? 'noopener noreferrer' : nothing}"
```
Galimede: nitpick: we can drop `noopener` as we have `target: _blank`

src/translations/translations.en.js
```diff
@@ -1843,6 +1825,16 @@ export const translations = {
...
   //#endregion
+  //#region cc-search-bar
+  'cc-search-bar.close': `Close`,
+  'cc-search-bar.empty.description': () =>
+    sanitize`Start to search, try different keywords or check the spelling.<br>You can search across organizations, resources (applications, add-ons, etc.), pages, and documentation`,
+  'cc-search-bar.empty.title': `No result found`,
+  'cc-search-bar.external-link.new-tab': `(opens in a new tab)`,
+  'cc-search-bar.heading': `Search bar`,
+  'cc-search-bar.label': `What are you looking for?`,
+  'cc-search-bar.placeholder': `No value yet...`,
```
Galimede: nitpick: I'd either drop this placeholder or we could be more precise by adding a scope. Maybe somehting like `Search for apps, add-ons...`

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+
+  /** @param {SearchBarSection} section */
+  _renderSection(section) {
+    return html`
+      <div class="section">
+        <h3 class="section-header">
+          <cc-icon .icon="${section.icon}" size="md"></cc-icon>
+          <span>${section.label}</span>
+        </h3>
+        <ul class="section-items" aria-label="${section.label}">
```
florian-sanders-cc: **fix:** `aria-label` on a `ul` will probably not be announced, `aria-label` / `aria-labelledby` attributes are only compatible with specific roles (interactive roles like buttons & form controls, region, nav, and a few others).

It's not a big issue since it has no consequences but it's better to avoid using `aria` unless we need to. The actual issue you are trying to fix with this attribute is already fixed by the use of `<h3>` heading above.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+
+/**
+ * @import { SearchBarItem, SearchBarSection } from './cc-search-bar.types.js'
+ * @import { CcInputText } from '../cc-input-text/cc-input-text.js'
+ * @import { CcInputEvent } from '../common.events.js'
+ * @import { Ref } from 'lit/directives/ref.js'
+ */
+
+/**
+ * A search bar / command palette component that displays categorized results.
```
Galimede: question: this isn't a command palette right? If so it's quite misleading.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+        <p class="empty-title">${i18n('cc-search-bar.empty.title')}</p>
+        <p class="empty-description">${i18n('cc-search-bar.empty.description')}</p>
+      </div>
+    `;
+  }
+
+  _renderHeader() {
+    return html`
+      <div class="header">
+        <h2 class="heading">${i18n('cc-search-bar.heading')}</h2>
```
florian-sanders-cc: **fix:** headings are a tricky subject for clever components because they are not supposed to know in which context they're gonna be used (in fact this is very visible here since the dialog is not even part of the component).

We used to avoid headings because we thought we'd come up with a solution to define "context aware headings" that would adapt but we never got the chance to work on that.

We can make an exception here I guess but then it should be `h1` since it's gonna be displayed within a `dialog` which makes the rest of the page inactive. This means this will, be the main heading of the content being viewed by the user => h1 (and the `h3` below would become `h2`).

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  _renderSearchInput() {
+    return html`
+      <div class="input-wrapper">
+        <span class="input-label">${i18n('cc-search-bar.label')}</span>
+        <div class="input-field">
+          <cc-input-text
+            hidden-label
+            label="${i18n('cc-search-bar.label')}"
```
Galimede, line 155: thought: I'm wondering if the label for a11y is precise enough here.

src/components/cc-search-bar/cc-search-bar.js
Galimede: thought:  In real life usage, if you'd click or trigger this search bar you would likely know the context. 
Thus said, I wonder if we need to have a heading with `Search bar` here. 
Plus, if we don't need it we could drop it and gain some space.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+          ${item.badge != null
+            ? html` <cc-badge intent="${item.badgeIntent ?? 'neutral'}" weight="strong"> ${item.badge} </cc-badge> `
+            : ''}
+          ${item.externalLink
+            ? html`
+                <span class="visually-hidden">${i18n('cc-search-bar.external-link.new-tab')}</span>
+                <cc-icon class="external-link-icon" .icon="${iconExternalLink}" size="sm"></cc-icon>
+              `
+            : ''}
+        </a>
```
Galimede, line 208: thought: I understand why we're mostly rewritting the `cc-link` here but maybe we could extract some helper from it and use it here; `cc-link` handles externalLink, focus etc. by itself and we don't take advantage of it here.

src/components/cc-search-bar/cc-search-bar.stories.js
```diff
@@ -0,0 +1,174 @@
...
+  title: '🧬 Atoms/<cc-search-bar>',
+  component: 'cc-search-bar',
+};
+
+const conf = {
+  component: 'cc-search-bar',
+  css: `
+    cc-search-bar {
+      max-width: 500px;
+    }
```
HeleneAmouzou, line 29:
```suggestion
    cc-search-bar {
      max-width: 500px;
      margin-inline: auto;
    }
```

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+      <li>
+        <a
+          class="item"
+          href="${item.href}"
+          target="${item.externalLink ? '_blank' : nothing}"
+          rel="${item.externalLink ? 'noopener noreferrer' : nothing}"
+        >
+          <span class="item-label">${item.label}</span>
+          ${item.badge != null
+            ? html` <cc-badge intent="${item.badgeIntent ?? 'neutral'}" weight="strong"> ${item.badge} </cc-badge> `
```
Galimede: note: As a user, checking the stories, I can't quite get why badges are a certain color or not. It felt weird to me. I mean, I understand it corresponds to a group, but we don't use these grouping colors anywhere else don't we?

src/components/cc-search-bar/cc-search-bar.js
florian-sanders-cc, line 208: **fix:** for external links we tend to use the icon (link you did 👍 ) + a `title` attribute. Though it's an imperfect solution, it does have the advantage of showing a small tooltip on desktop. 
I think we should stick with that, to be consistent overall. You can find a good implementation example in the `cc-link` component I think.
The visually-hidden text could be the `a11y-name` of the icon but it can remain like that, the important thing is adding a `title`.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+          : this._renderEmpty()}
+      </div>
+    `;
+  }
+
+  _renderEmpty() {
+    return html`
+      <div class="empty">
+        <cc-icon class="empty-icon" .icon="${iconSearch}" size="xl"></cc-icon>
+        <p class="empty-title">${i18n('cc-search-bar.empty.title')}</p>
```
Galimede: thought/nitpick: When opening the search bar at first you'll see `No result found` . However,  you won't have made a search yet, which is a bit misleading.

src/components/cc-search-bar/cc-search-bar.js
Galimede: nitpick: you can remove these from the documentation, we don't need these anymore as they're documented in the events file directly.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+        .item-label {
+          flex: 1;
+          min-width: 0;
+          overflow: hidden;
+          text-overflow: ellipsis;
+          white-space: nowrap;
+        }
+
+        .external-link-icon {
+          color: var(--cc-color-text-weak, #888);
```
Galimede: nitpick/question: this is a codebase wide issue but we don't fallback to the same value for everything, for the other `--cc-color-text-weak` you put the fallback to `#666` when here it's `#888`. Is there a reason? 
If not, maybe we could put `#666` here too.

src/components/cc-search-bar/cc-search-bar.types.d.ts
Galimede: thought: maybe we could add a comment to say that the default is `neutral` if ommited.

Galimede (changes requested):
Hey Clara, overall mostly nitpicks and thoughts but LGTM, GG! 👏

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        .search-bar {
+          background: var(--cc-color-bg-default, #fff);
+          border: 1px solid var(--cc-color-border-neutral-weak, #e7e7e7);
+          border-radius: var(--cc-border-radius-default, 0.25em);
+          box-shadow: 0 8px 30px rgb(0 0 0 / 12%);
+          display: flex;
+          flex-direction: column;
+          max-height: 80vh;
```
florian-sanders-cc: **fix:** setting `max-height` + box-shadow is not this components job, this should be handled by the `cc-dialog` that will contain this.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,364 @@
...
+        .header {
+          align-items: center;
+          display: flex;
+          justify-content: space-between;
+          padding-bottom: 0.5em;
+        }
+
+        .heading {
+          color: var(--cc-color-text-primary-strongest, #000);
+          font-size: 0.89em;
```
HeleneAmouzou: **question :** why chosing this `font-size` value ? With it the heading and the input-label are the same size

florian-sanders-cc (changes requested):
Good job @clara-layus-cc the component looks nice!

I have a few questions: 
- will there be async loading involved? The console will need the list of all the resources, it already has it in it's summary so I assumed you're going to rely on this? (this would mean no loading I think so it would be fine)
- why not using the `cc-dialog` component here? Do you plan to use it from the console directly?
- I think we could use the `cc-dialog` and let it handle the main heading + close button. Since the wireframes seem to have a smaller padding than the default dialog, we could use the padding custom props to customize these and match the wireframes while still using the `cc-dialog` component. We can discuss all of that in sync if needed 😉 
- the margin / gap between the section label and the border above it does not match the wireframes (margin after each section is fine, but margin before each section is too narrow)

src/translations/translations.en.js
florian-sanders-cc: Note that the placeholder cannot contain more info than the label, so what you're suggesting doesn't work @Galimede (placeholder is not necessarily read by screen readers, not contrasted enough and disappears when you start typing) so the choice is really either a useless placeholder or no placeholder 😂 
Could discuss with Marion ;)

src/components/cc-search-bar/cc-search-bar.js
florian-sanders-cc: **fix:** I think the icon is dark blue in the wireframes so you'll need to set a different color 😉

HeleneAmouzou (changes requested):
GG Clara ! I only left question/suggestion. ~I was also wondering if we should add some tests ? I would be curious to have other opinions on this.~

src/translations/translations.en.js
Galimede: Didn't know about that, thanks for the clarification and the info :smile:

src/components/cc-search-bar/cc-search-bar.js
clara-layus-cc: I agree with you, but I'd like to hear @marion-izd's opinion

src/components/cc-search-bar/cc-search-bar.js
clara-layus-cc: That's a good idea

src/components/cc-search-bar/cc-search-bar.js
clara-layus-cc: Good question, I'd like to hear @marion-izd's opinion

src/components/cc-search-bar/cc-search-bar.js
clara-layus-cc: It's a design choice, but I'd like to hear @marion-izd's opinion

src/translations/translations.en.js
marion-izd: We’re going to try to go beyond just search apps and add-ons, so I don’t want to be too specific with the label. I agree with Florian’s explanation about the placeholder, but we could still make it a bit more user-friendly and clear without getting too specific: ‘start searching’, ‘search by keywords’ ,... :)

src/components/cc-search-bar/cc-search-bar.js
marion-izd: No, it’s definitely a deliberate choice to assign a colour to a group. There’s no precedent, so there’s no correspondence with other parts of the console at the moment. I used the existing colours from the badge component to avoid having to ‘rebuild’ it. But I know there are limitations: as the colours in cc-badges are limited, if a new category appears we’ll be stuck (> but the idea is to have a component with the new system design) 

P.S.: @clara-layus-cc, as we discussed during our meeting, let’s use the ‘dimmed’ colours for the badge (rather than the ‘strong’ variants)

src/components/cc-search-bar/cc-search-bar.js
marion-izd: Yes, it’s tricky to find a text that works for both an unsuccessful search and the start of the search. 
Two solutions: we could duplicate the screen with an ‘empty’ state and an ‘unfound’ state, each with more specific text. Or we could leave it as it is, with the description below being more general and including both recommendations to start the search and suggestions to change the keywords.
Is it difficult for you, @clara-layus-cc , at this stage to add a variant just for the text?

marion-izd:
Well done, Clara! 
I think I’ve addressed the previous comments and questions, so here are my own comments and questions:
- As Hélène pointed out, the title is a bit small; we should base it either on the title of the cc-dialog or on the title of a cc-block. For my part, I’ve based it on the title of a cc-block, so 18px
- The headings for the results categories: these should be clearer/smaller to distinguish them from the results themselves (in the mock-up, they are 12px in light grey, whilst the results are 14px in the default colour)
- The icons in the categories need to be made slightly larger (your component is 12px > mock-up is 16px)
- Same for icons for external link (your component is 9px > mock-up is 16px)
- As mentioned earlier, the colour of the badges in the dimmed version
- I have a few comments about the hover/focus effect: In terms of alignment, when the mouse hovers over or focuses on it, it should be aligned with the category icon (There is currently a slight spacing). And the corner radius is small in hover (your component is 2px > mock-up il 8px) and 10px for focus.

I don't think I've left anything out. 💪 Good work, and please don't hesitate to ask me or drop me a line if you have any further questions.

src/components/cc-search-bar/cc-search-bar.js
clara-layus-cc: No problem on my end, but I’d be happy for you to suggest the alternative

HeleneAmouzou (approved):
All good for me !

roberttran-cc (approved):
LGTM, nice component! No additional comments beyond those already raised.

marion-izd:
I’ve just had another look, particularly at the text in the two versions —‘empty’ and ‘no result’— and I agree with your suggestions. 
- Just a quick note on the ‘no result’ version: you could remove the full stop at the end (something David would probably point out 😉).
- I have a question about the ‘empty’ section. I like the examples, but what about ‘is:app’? Will that be replaced automatically? Does it also work with keywords to find apps?

Great work again 💪

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,411 @@
...
+        <span class="input-label">${i18n('cc-search-bar.label')}</span>
+        <div class="input-field">
+          <cc-input-text
+            hidden-label
+            label="${i18n('cc-search-bar.label.a11y')}"
+            placeholder="${i18n('cc-search-bar.placeholder')}"
+            value="${this.value}"
+            @cc-input="${this._onInput}"
+            ${ref(this._inputRef)}
+          ></cc-input-text>
```
florian-sanders-cc: **fix:** the visible text should be the label (sorry if I missed it in the first review), I don't think we need a hidden-label for a11y, the visible label of an element should be the accessible name of the element unless we have no other choice (we can discuss it if you think it's needed 😉 )

florian-sanders-cc: oh maybe you did this to place the icon within the text input? If that's so we're facing a limit of the current `cc-input-text` and we should just use a custom input instead 👀 (or improve the `cc-input-text` component)

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,411 @@
...
+
+        .item:hover .hover-chevron {
+          display: inline-block;
+        }
+
+        .item:focus-visible .item-label {
+          overflow: visible;
+          text-overflow: clip;
+          white-space: normal;
+        }
```
florian-sanders-cc: **fix:** let's discuss text overflow strategies in sync, it's a tricky subject

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,403 @@
...
+      </li>
+    `;
+  }
+
+  static get styles() {
+    return [
+      // language=CSS
+      css`
+        :host {
+          display: contents;
```
florian-sanders-cc, line 218: **question:** why `display: contents` here?

florian-sanders-cc (changes requested):
A few a11y issues but almost there 👍

florian-sanders-cc (approved):
All good for me, great job @clara-layus-cc 😎

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,391 @@
...
+
+/**
+ * A search bar dialog that displays categorized results.
+ *
+ * ## Details
+ *
+ * The component wraps a search input and a list of sections inside a `cc-dialog`.
+ * Items can have optional badges and external link indicators.
+ *
+ * @cssdisplay contents
```
roberttran-cc, line 38: **💭 thought:** this value is now out of sync, as the `:host` style has been removed.

But I'm not sure what we should do in this case:
- default `display` value is `inline`
- but should we enforce a default value in the CSS?
- if yes, what value should it be?

Actually, from my understanding, `:host { display: content; }` made sense because of the `cc-dialog` internal implementation. Would love @florian-sanders-cc on that.

roberttran-cc (approved):
Only a minor doc-related feedback for me, LGTM otherwise, hence the approval.

src/components/cc-search-bar/cc-search-bar.js
florian-sanders-cc: Agreed, the `display: contents;` makes sense here :+1:.
My comment was merely a question to explain the choice. 
The initial choice did make sense and I was just too dumb to see it :smile:.
Sorry for the confusion :wink:

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,395 @@
...
+        if (!keywordsMatch) {
+          return false;
+        }
+        const label = item.label.toLowerCase();
+        const id = item.id?.toLowerCase() ?? '';
+        return textTokens.every((token) => label.includes(token) || id.includes(token));
+      }),
+    }))
+    .filter((section) => section.items.length > 0);
+}
```
hsablonniere: nitpick: this could be a private method of the class `_getFilteredSections()` and the method already knows about the sections and current value with `this`.

hsablonniere (changes requested):
Well done @clara-layus-cc, I added a small comment, feel free to apply it or not and here are some changes I want :

- The main bloc should be aligned to the top to prevent the input to move while typing.
- The hover with the ">" shifts the content, I think we need to prevent this.

Notes :

- I was wondering if I needed a `clear()` method but assigning `value=""` is enough.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -0,0 +1,411 @@
...
+  _onInput(e) {
+    this.value = e.detail;
+  }
+
+  /** @param {Map<string, unknown>} changedProperties */
+  updated(changedProperties) {
+    if (changedProperties.has('open') && this.open) {
+      this._inputRef.value?.focus();
+    }
+  }
```
florian-sanders-cc: **fix:** remove this and use `.autofocus="${true}" on the `cc-input-text` instead. 
The `cc-dialog` already supports autofocusing the input on open thanks to this attribute / property.

We'll simplify to use the plain autofocus attribute instead of the property in a serparate PR (see #1759)

florian-sanders-cc (changes requested):
A small change to make but apart from that it's good to me!

pdesoyres-cc (approved):
LGTM. Well done Clara

florian-sanders-cc (approved):
All good!

PR #1735, opened 2026-04-29
fix: cc-network-group-* translations update

davlgd:


florian-sanders-cc (approved):
All good, thanks for handling this @davlgd ! 🙌

PR #1740, opened 2026-05-07
refactor(cc-cellar-explorer): drop cellarProxyUrl smart context property

pdesoyres-cc:
## Context

`cc-cellar-explorer` originally talked to a separate Cellar proxy backend (`cellarProxyUrl`) for
bucket and object operations. Those operations are now exposed directly on the Clever Cloud v4
cellar API, so the proxy URL is dead weight in the smart context — every consumer would have to
pass a value that no code path actually reads.

## Changes

- Remove `cellarProxyUrl` from the smart component's context definition.
- Drop the corresponding entry from the sandbox harness so it stops advertising a parameter that
  is no longer wired up.
- Refresh `cc-cellar-explorer.smart.md` to describe the v4 cellar endpoints actually used and
  remove the "Cellar Proxy API" section.

## How to review

1. Start with `src/components/cc-cellar-explorer/cc-cellar-explorer.smart.js` — the context shape
   change is the load-bearing edit.
2. Cross-check `cc-cellar-explorer.smart.md` against the endpoints the component actually calls
   to confirm the documented table is accurate.
3. Run the sandbox (`cc-cellar-explorer`) locally to verify bucket/object listing still works
   without the proxy URL.

### DIFF
--- a/sandbox/cc-cellar-explorer/cc-cellar-explorer-sandbox.js
+++ b/sandbox/cc-cellar-explorer/cc-cellar-explorer-sandbox.js
@@ -37,7 +37,6 @@ class CcCellarExplorerSandbox extends LitElement {
 
   render() {
     const context = {
-      cellarProxyUrl: 'http://localhost:8082',
       ownerId: 'orga_540caeb6-521c-4a19-a955-efe6da35d142',
       addonId: this._addonId,
     };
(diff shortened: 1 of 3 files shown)


HeleneAmouzou (approved):
LGTM !

PR #1741, opened 2026-05-11
refactor(cc-homepage-template-project): use `cc-block` as wrapper and drop `adapt-height`

clara-layus-cc:
# What does this PR do?
  
  - Replaces the custom `<div class="wrapper">` (border, rounded corners, manual title) by `<cc-block>`. The title is now provided through the `header-title` slot and the list/notice through the `content` slot, so styling (border, header typography, paddings) comes from `cc-block` instead of being duplicated.
  - Removes the `adapt-height` mode and its associated story. `cc-block`'s internal grid layout doesn't propagate  `height: 100%` to its content, so the previous "stretch rows to fill the parent height" behavior wouldn't work cleanly through `cc-block`. The component now sizes naturally to its content.

# How to review?
  
  - Check the commit,
  - Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/fix/cc-homepage-template-project/index.html)


hsablonniere (approved):
LGTM 👍

PR #1746, opened 2026-05-20
ci(browserslist-db): automate monthly update

### LINKED ISSUE #1655: Automate monthly browserslist-db update via scheduled GitHub Actions workflow
## Context

We should automate the monthly update of the browserslist database to keep our browser support data current.

## Proposed Solution

Create a GitHub Actions scheduled workflow that:
1. Runs on the 1st of each month (cron: `0 0 1 * *`)
2. Executes `npx update-browserslist-db@latest` (only modifies `package-lock.json`)
3. Creates a PR with a chore commit (e.g., `chore(deps): update browserslist-db`)
4. Assigns the team as reviewers

## Implementation Notes

- Use `peter-evans/create-pull-request` action for PR creation (Note: Let's avoid using an action for this, it should be easy to do by ourselves really).
- Will need to use a PAT or GitHub App token (not `GITHUB_TOKEN`) if we want CI workflows to trigger on the created PR
- The existing `CI_TOKEN` secret could potentially be reused

## References

- [update-browserslist-db](https://github.com/browserslist/update-db)
- [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request)

/cc @galimede @florian-sanders-cc

HeleneAmouzou:
## What does this PR do?

Fixes #1655

- Introduces a new workflow to monthly update the browserslist database,
- Adds on `yml` file to describe the workflow,
- Adds a `manage-pull-request.js` script with `create` and `assign-reviewers` actions,
- Adds `reviewers.json` file with the list of reviewers to add to the PR.

### About the issue 

In the [issue](https://github.com/CleverCloud/clever-components/issues/1655) it was suggested to use a ready-made action (`peter-evans/create-pull-request`) to create a PR, but I have decided to match what we already did for managing comment and handle it without a third party action.

### Multiple uses of `if` condition

I have try several ways to refactor the workflow to simplify all the uses of `if` by grouping the steps together, but I decided to leave it as is because I thought it is more convenient to see all the steps in detail in case one of them doesn't work. Feel free to question/challenge this decision, maybe there is one way of doing I did not thought of.

### Reviewers

I decided to put all possible reviewers in a separate `json` file that could be updated when needed. Claude suggested another solution that would be to do a `reviewer team` on GitHub. I did not know which solution could be the best, I chose the `json` one because I thought it could be reused in another case on the repository.

### Test

To test the workflow, I have copy the `clever-components` repository on my account and started the action directly from GitHub and it all worked. The repo in private, but if needed I could pass it in public so reviewers could tested it. That does however require a few changes to the workflow file to adapt to the repository. 

## How to review?

- Check the commit.

2 or 3 reviewers should be enough.

### DIFF
--- a/.github/scripts/manage-pull-request.js
+++ b/.github/scripts/manage-pull-request.js
@@ -1,0 +1,313 @@
+import { Octokit } from '@octokit/rest';
+import { appendFile, readFile } from 'node:fs/promises';
+
+/**
+ * @typedef {import('@octokit/rest').RestEndpointMethodTypes} RestEndpointMethodTypes
+ */
+
+const ACTIONS = /** @type {const} */ (['create', 'assign-reviewers']);
+
+/**
+ * GitHub Client for creating pull requests
+ */
+class GithubPullRequestClient {
+  /**
+   * @param {string} token - GitHub API token
+   */
+  constructor(token) {
+    this.octokit = new Octokit({ auth: token });
+  }
+
+  /**
+   * Create a new pull request
+   * @param {string} owner - Repository owner
+   * @param {string} repo - Repository name
+   * @param {string} title - Pull request title
+   * @param {string} head - Source branch
+   * @param {string} base - Target branch
+   * @param {string} [body] - Pull request body (optional)
+   * @returns {Promise<RestEndpointMethodTypes["pulls"]["create"]["response"]["data"]>}
+   */
+  async createPullRequest(owner, repo, title, head, base, body) {
+    try {
+      const response = await this.octokit.rest.pulls.create({
+        owner,
+        repo,
+        title,
+        head,
+        base,
+        body,
+      });
+      console.log(`✓ Created PR #${response.data.number} at ${response.data.html_url}`);
+      return response.data;
+    } catch (err) {
+      console.error('✗ Error creating pull request:', err instanceof Error ? err.message : 'unknown');
+      process.exit(1);
+    }
+  }
+
+  /**
+   * Request reviews on a pull request
+   * @param {string} owner - Repository owner
+   * @param {string} repo - Repository name
+   * @param {number} prNumber - Pull request number
+   * @param {string[]} reviewers - GitHub handles to request reviews from
+   */
+  async assignReviewers(owner, repo, prNumber, reviewers) {
+    try {
+      const response = await this.octokit.rest.pulls.requestReviewers({
+        owner,
+        repo,
+        // eslint-disable-next-line camelcase
+        pull_number: prNumber,
+        reviewers,
+      });
+      console.log(`✓ Requested review from ${reviewers.join(', ')} on ${owner}/${repo}#${prNumber}`);
+      return response.data;
+    } catch (err) {
+      console.error('✗ Error requesting reviewers:', err instanceof Error ? err.message : 'unknown');
+      process.exit(1);
+    }
+  }
+}
+
+/**
+ * Command-line interface manager for GitHub pull request operations
+ */
+class PullRequestActionManager {
+  constructor() {
+    this.client = null;
+  }
+
+  /**
+   * @param {string} action
+   * @returns {action is typeof ACTIONS[number]}
+   */
+  isValidAction(action) {
+    return ACTIONS.includes(/** @type {typeof ACTIONS[number]} */ (action));
+  }
+
+  printUsage() {
+    console.log(`
+Usage: node manage-pull-request.js <action>
+
+Actions:
+  create            Create a new pull request
+  assign-reviewers  Request reviews on an existing pull request
+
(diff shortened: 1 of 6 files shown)


.github/scripts/manage-pull-request.js
```diff
@@ -0,0 +1,313 @@
+import { Octokit } from '@octokit/rest';
+import { appendFile, readFile } from 'node:fs/promises';
+
+/**
+ * @typedef {import('@octokit/rest').RestEndpointMethodTypes} RestEndpointMethodTypes
```
florian-sanders-cc: **question:** is this a case where we can't use `@import`? (I remember there are some cases where it makes our tooling fail)

.github/scripts/manage-pull-request.js
```diff
@@ -0,0 +1,313 @@
...
+    const requiredEnvVars = this.#actionConfig[action];
+
+    for (const envVar of requiredEnvVars) {
+      if (!process.env[envVar]) {
+        errors.push(`${envVar} is required`);
+      }
+    }
+
+    return errors;
+  }
```
florian-sanders-cc, line 177: **question:** this is a useful method on its own, that could be reused for other actions (I believe we have very similar code in the manage-comments.js file). 
Maybe we could extract that in a common helper? (input = required env vars, output = array of errors or we even make the helper log the required env vars itself I'm not sure? 🤔 

Note that this could be tracked in an issue and be done in a separate PR, your code is really fine 👍

florian-sanders-cc (approved):
All good for me, well done @HeleneAmouzou this is awesome 🔥 

About the repeated `if`, I think I would split this in two separate jobs, even if it means repeating some steps but arguably this is a worse solution compared to yours. 
I think I'm fine with what you chose 👍

.github/workflows/browserslist-database-update.yml
```diff
@@ -0,0 +1,83 @@
+name: Browserslist Database Update
+
```
roberttran-cc, line 2: **👀 suggestion:** from Claude

> **No concurrency group**. `workflow_dispatch` + the cron (or two manual runs) can overlap, both reach `git push` / `gh pr create` and produce duplicate PRs or a push race.
>
> Add concurrency: `{ group: browserslist-db-update, cancel-in-progress: false }` at the workflow level.


```suggestion

concurrency:
  group: browserslist-db-update
  cancel-in-progress: false
```

.github/workflows/browserslist-database-update.yml
```diff
@@ -0,0 +1,83 @@
...
+        if: steps.check.outputs.has_changes == 'true'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: '[Versioning] Create branch'
+        id: branch
+        if: steps.check.outputs.has_changes == 'true'
+        run: |
+          BRANCH_NAME=chore/update-browserslist-db-$(date -u +%Y-%m)
```
roberttran-cc, line 53: **👀 suggestion:** From Claude

> **Branch name collision on re-run within the same month.**
>
> `BRANCH_NAME=chore/update-browserslist-db-$(date -u +%Y-%m)` → re-running the workflow in the same calendar month (manual retry after a failure, or testing) makes `git checkout -b` fail (local) or `git push` fail (remote).
>
> Options:
> - (a) detect existing local/remote branch and `checkout` it + force-update,
> - (b) append a short SHA or `%Y-%m-%d-%H%M%S` suffix,
> - (c) early-exit with a friendly message when the PR already exists.

.github/workflows/browserslist-database-update.yml
```diff
@@ -0,0 +1,83 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: '[Prepare] Checkout'
+        uses: actions/checkout@v5
+        with:
+          token: ${{ secrets.CI_TOKEN }}
+      - name: '[Prepare] Setup Node.js'
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: 'package.json'
```
roberttran-cc, line 24: **👀 suggestion:** to match other workflows


```suggestion
          node-version-file: 'package.json'
          cache: 'npm'
```

roberttran-cc (approved):
I made a review with Claude which submitted a few feedback: as they look non-blocking (and as I'm not proficient with CI), I'll approve anyway! GG!

.github/scripts/manage-pull-request.js
HeleneAmouzou: No it is not, I have use `@import`, thanks !

pdesoyres-cc (approved):
Well done Hélène

roberttran-cc (approved):
LGTM! Looking forward Monday's first run.

Issue #1621, opened 2025-11-25
`cc-domain-management`: delete button a11y name should include the path prefix


florian-sanders-cc:
## Context

In the `cc-domain-management` component, the delete buttons only contain the hostname and not the path prefix.

This means that the delete button for `example.com` and the one for `example.com/path` both have the same accessible name: `delete - example.com`.

The accessible name should be: `delete - [hostname+pathPrefix` so `delete - example.com` and `delete - example.com/path`

PR #1747, opened 2026-05-20
fix(cc-domain-management): include path prefix in delete and mark-as-primary buttons a11y-name

clara-layus-cc:
#1621

# What does this PR do?
  
  - Includes the path prefix in the accessible name of the `delete` and `mark as primary` buttons in
  `cc-domain-management`.
  
# How to review?

  - Check the commit,
  - Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/fix/cc-domain-management/delete-btn-a11y-name/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story)


roberttran-cc (approved):

HeleneAmouzou (approved):
LGTM !

florian-sanders-cc (approved):
Thanks for handling that 😎

PR #1748, opened 2026-05-21
feat(cc-logs): add clear button with clear-and-continue overflow option

### LINKED ISSUE #1467: cc-logs: add a button to clear current logs view
Sometimes we just want to clear the current view. It could be great to have a button (or a way) to empty the local logs components

hsablonniere:
Closes #1467

## Summary

- Add an opt-in `clearable` property to `cc-logs-loading-progress-beta` that renders an eraser icon button next to play/pause
- Wire the `cc-logs-loading-clear` event through all three smart controllers (`cc-logs-app-runtime`, `cc-logs-addon-runtime`, `cc-logs-app-access`)
- Add `clearLogs()` to `LogsStream` and `resetProgress()` to `LogsProgress` to clear displayed logs and reset counters without disconnecting the SSE stream
- The clear button is shown in all modes (live and cold logs)
- During an overflow pause, clearing does NOT resume the stream — a new "Clear and continue" option is available in the overflow dialog alongside "Continue" and "Stop"

## Behavior

| State | Clear action |
|---|---|
| Running | Reset progress to 0, clear buffer, clear display |
| Paused (user) | Reset progress to 0, clear buffer, clear display, stay paused |
| Paused (overflow) | Reset progress to 0, clear buffer, clear display, stay paused (transition to user pause) |
| Completed / Idle | No effect |

### Overflow dialog options

| Option | Action |
|---|---|
| Continue | Resume stream (existing) |
| Clear and continue | Clear logs + resume stream (new) |
| Stop | Discard overflow, stop stream (existing) |

## Review

- [`<cc-logs-loading-progress-beta>`](https://clever-components-preview.cellar-c2.services.clever-cloud.com/clear-logs/index.html?path=/story/%F0%9F%9A%A7-beta-%F0%9F%9B%A0-logs-cc-logs-loading-progress-beta--with-clearable)
- [`<cc-logs-app-runtime-beta>`](https://clever-components-preview.cellar-c2.services.clever-cloud.com/clear-logs/index.html?path=/story/%F0%9F%9A%A7-beta-%F0%9F%9B%A0-logs-app-cc-logs-app-runtime-beta--default-story)
  - same for app access logs and add-on logs
- [next-console](https://next-console.cleverapps.io/organisations/orga_540caeb6-521c-4a19-a955-efe6da35d142/applications/app_457bd0d5-23c5-48c3-93bc-881ab5e4492c/logs)

### DIFF
--- a/src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
+++ b/src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
@@ -2,6 +2,7 @@ import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import {
   iconRemixInformationFill as iconInfo,
+  iconRemixEraserLine,
   iconRemixPauseLine,
   iconRemixPlayLine,
   iconRemixAlertFill as iconWarning,
@@ -10,7 +11,9 @@ import { i18n } from '../../translations/translation.js';
 import '../cc-button/cc-button.js';
 import '../cc-icon/cc-icon.js';
 import {
+  CcLogsLoadingClearEvent,
   CcLogsLoadingOverflowAcceptEvent,
+  CcLogsLoadingOverflowClearAndContinueEvent,
   CcLogsLoadingOverflowDiscardEvent,
   CcLogsLoadingPauseEvent,
   CcLogsLoadingResumeEvent,
@@ -28,6 +31,7 @@ import {
 export class CcLogsLoadingProgress extends LitElement {
   static get properties() {
     return {
+      clearable: { type: Boolean },
       limit: { type: Number },
       state: { type: Object },
     };
@@ -35,6 +39,9 @@ export class CcLogsLoadingProgress extends LitElement {
 
   constructor() {
     super();
+
+    /** @type {boolean} Whether to display the clear button. */
+    this.clearable = false;
 
     /** @type {LogsLoadingProgressState} The state of the component */
     this.state = {
@@ -79,6 +86,14 @@ export class CcLogsLoadingProgress extends LitElement {
     this.dispatchEvent(new CcLogsLoadingOverflowAcceptEvent());
   }
 
+  _onClear() {
+    this.dispatchEvent(new CcLogsLoadingClearEvent());
+  }
+
+  _onClearAndContinue() {
+    this.dispatchEvent(new CcLogsLoadingOverflowClearAndContinueEvent());
+  }
+
   _onDiscardOverflow() {
     this.dispatchEvent(new CcLogsLoadingOverflowDiscardEvent());
   }
@@ -100,7 +115,7 @@ export class CcLogsLoadingProgress extends LitElement {
     return html`
       <div class="wrapper ${classMap({ warning: shouldAskForOverflowDecision })}">
         <div class="content inline">
-          ${this._renderPlayPauseButton()}
+          ${this._renderPlayPauseButton()}${this._renderClearButton()}
           ${
             shouldAskForOverflowDecision
               ? html`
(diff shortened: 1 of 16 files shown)


src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
```diff
@@ -28,6 +30,7 @@ import {
 export class CcLogsLoadingProgress extends LitElement {
   static get properties() {
     return {
+      clearButton: { type: Boolean, attribute: 'clear-button' },
```
pdesoyres-cc: nitpick: I'm not sure about the name of the property. The fact that it will be a button is more an internal implementation detail. (`can-clear`, `allow-clear`)

src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
```diff
@@ -154,6 +164,21 @@ export class CcLogsLoadingProgress extends LitElement {
...
+  _renderClearButton() {
+    if (!this.clearButton) {
+      return null;
+    }
+
+    return html`
+      <cc-button
+        .icon=${iconRemixEraserLine}
+        hide-text
+        a11y-name=${i18n('cc-logs-loading-progress.control.clear')}
```
pdesoyres-cc, line 184: It doesn't say that in some case it will resume the stream

src/lib/logs/logs-stream.js
```diff
@@ -155,6 +155,38 @@ export class LogsStream {
...
+   */
+  clearLogs() {
+    if (this.#streamState.type !== 'running' && this.#streamState.type !== 'paused') {
+      return;
+    }
+
+    const wasPausedByUser = this.#streamState.type === 'paused' && this.#streamState.reason === 'user';
+    const wasPausedByOverflow = this.#streamState.type === 'paused' && this.#streamState.reason === 'overflow';
+
+    this.#logsBuffer.flush();
```
pdesoyres-cc: fix: We should not flush the buffer.
- buffer content is not taken into account in the current progression
- buffer content has never been displayed to the user
- so clearing it here will make the content never appear. User may loose some important log line

src/lib/logs/logs-stream.js
```diff
@@ -155,6 +155,38 @@ export class LogsStream {
...
+        reason: 'user',
+        progress: this.#progress.getProgress(),
+        overflowing: this.#progress.isOverflowing(),
+      });
+    } else {
+      this._updateStreamState(this.#buildRunningState());
+    }
+
+    if (wasPausedByOverflow) {
+      this.#logsStream.resume();
```
pdesoyres-cc: nitpick: I'm ok with that, but I'm just not sure.
- Maybe, the clear button should always just clear and that's it.
- When paused by overflow, we ask for two choices. Maybe we should ask for three choices:
  - Continue
  - Clear and continue
  - Stop
- When paused by overflow, we could also not show the clear button at all but it's ok if it's there too

If @marion-izd said it's ok, let's go for yous solution.

pdesoyres-cc (changes requested):
Well done Hubert. I've got several feedback, and also I'm wondering why you choose not to display it for non-live streams. It could still be useful when paused on overflow, even in non-live mode.

HeleneAmouzou (approved):
It works fine for me. The eraser icon is nice and I have no opinion on adding it only in live mode.

And I agree with @pdesoyres-cc  on this : 

> When paused by overflow, we ask for two choices. Maybe we should ask for three choices:
>
>   Continue
>   Clear and continue
>   Stop

src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
hsablonniere: Done — renamed to `clearable` (attribute: `clearable`).

src/components/cc-logs-loading-progress/cc-logs-loading-progress.js
hsablonniere: Resolved — `clearLogs()` no longer resumes the stream. The label is now always accurate.

src/lib/logs/logs-stream.js
hsablonniere: Good catch — replaced `flush()` with `clear()` so buffered logs are discarded without being emitted.

src/lib/logs/logs-stream.js
hsablonniere: Agreed — clear now just clears (never resumes). Added a "Clear and continue" option in the overflow dialog alongside "Continue" and "Stop". The clear button is also now available in all modes (not just live).

PR #1755, opened 2026-05-28
refactor(metrics): migrate metrics tiles and logsmap to cc-api-client

pdesoyres-cc:
## Context

Continues the migration away from the legacy `@clevercloud/client` Warp 10 access path
toward the new typed `cc-api-client`. Once every consumer of `sendToWarp` is gone, we can
drop the Warpscript plumbing and the `WARP_10_HOST` config surface, which is what this
branch achieves for the metrics tiles and the logs map.

## Changes

- `cc-tile-metrics`: smart binding now uses `GetMetricsCommand`. The new response is
  already keyed by metric name with normalized timestamps, so the local `extractMetric`
  helper and the `RawMetric` type are gone. Doc fixed from `/v4/metrics/...` to
  `/v4/stats/...` to match the endpoint.
- `cc-tile-status-codes`: replaces the two-step Warp 10 token + Warpscript exec with a
  single `GetStatusCodeDistributionCommand`. Client-side filtering of non-standard codes
  is dropped — the new client exposes `excludeNonStandardStatusCodes`.
- `cc-tile-requests`: adds the smart binding it never had, backed by
  `GetStatusCodeDistributionCommand`, so it can be wired into the demo and console
  without going through Warpscript.
- `cc-logsmap`: adds a smart binding using `GetRequestsCommand` for `heatmap` mode and
  `StreamRequestsCommand` for the live `points` mode.
- `send-to-api`: removes `sendToWarp`, the `Warp10ApiConfig` type, and the
  `WARP_10_HOST` plumbing in the storybook auth plugin. The API config surface now only
  carries what is actually used.
- Build: `cc-api-client.js` is added as a rollup main entry so consumers can import it
  without pulling in a specific component chunk.

### Implementation notes

For `cc-logsmap` heatmap mode, the response is byte-identical within an hour-aligned
bucket, so the current hour is cached until the next hour boundary plus a 2-minute
margin — same cadence the previous Warpscript query had, but expressed at the client
level instead of inside the script.

For `cc-tile-status-codes`, dropping the local filter changes a subtle behavior: codes
outside the standard ranges used to be silently dropped after fetch. The new flag
asks the client to omit them upstream, so the tile shows exactly what the client returns.

## How to review

1. Start with `src/lib/send-to-api.js` and the storybook plugin to confirm nothing else
   in the tree still references `sendToWarp` / `WARP_10_HOST`.
2. Then walk the four `*.smart.js` files in this order:
   `cc-tile-metrics` → `cc-tile-status-codes` → `cc-tile-requests` → `cc-logsmap`.
   Each one is independent; reading them in that order matches increasing complexity
   (single command → two commands + streaming).
3. The `.smart.md` docs next to each component were updated in lockstep — worth a pass
   to check the example payloads match the new command outputs.
4. Local smoke test: run the storybook smart demos for the four components against a
   real account and confirm the tiles render the same data as on `master`.

**1-2 approvals** from anyone on the team.

### DIFF
--- a/src/components/cc-logsmap/cc-logsmap.smart.js
+++ b/src/components/cc-logsmap/cc-logsmap.smart.js
@@ -1,0 +1,216 @@
+import { GetHeatMapCommand } from '@clevercloud/client/cc-api-commands/metrics/get-heat-map-command.js';
+import { StreamRequestsCommand } from '@clevercloud/client/cc-api-commands/metrics/stream-requests-command.js';
+import { getCcApiClientWithOAuth } from '../../lib/cc-api-client.js';
+import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
+import '../cc-smart-container/cc-smart-container.js';
+import './cc-logsmap.js';
+
+/**
+ * @import { CcLogsmap } from './cc-logsmap.js'
+ * @import { HeatmapPoint, MapModeType } from '../common.types.js'
+ * @import { ApiConfig } from '../../lib/send-to-api.types.js'
+ * @import { OnContextUpdateArgs, UpdateComponentCallback } from '../../lib/smart/smart-component.types.js'
+ * @import { RequestsStream } from '@clevercloud/client/cc-api-commands/metrics/stream-requests-command.js'
+ * @import { CcApiClient } from '@clevercloud/client/cc-api-client.js'
+ */
+
+// Spread the dots of a live batch over this duration so they don't all blink at once.
+const POINTS_SPREAD_DURATION = 3000;
+// How long a live dot stays on the map before fading out (outlives the spread so dots don't vanish mid-batch).
+const POINTS_DELAY = POINTS_SPREAD_DURATION + 2000;
+
+// The heat map aggregates requests into hourly buckets, so its response is byte-identical for the rest of the current
+// hour. We cache it until the next hour boundary, plus this margin to let the just-completed bucket be fully ingested
+// before we read it.
+const HEATMAP_CACHE_INGESTION_MARGIN = 2 * 60 * 1000;
+
+defineSmartComponent({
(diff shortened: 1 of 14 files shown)


src/components/cc-logsmap/cc-logsmap.smart.js
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  _loadHeatmap() {
+    const generation = this._generation;
+    this._heatmapAbortController = new AbortController();
+
+    this._updateComponent('error', false);
+    this._updateComponent('loading', true);
+    // Drop the previous result: a stale empty array would make `<cc-map>` show its "no points" message while loading.
+    this._updateComponent('heatmapPoints', null);
```
florian-sanders-cc, line 130: **note:** this makes me think the UI component should expose a `state` prop but I get that it would lead to a bigger refactoring that might not be worth it 👍

src/components/cc-tile-requests/cc-tile-requests.smart.js
```diff
@@ -0,0 +1,66 @@
...
+      ownerId,
+      applicationId: appId,
+      to,
+    }),
+    { signal },
+  );
+
+  return data.byDate.map((entry) => [
+    new Date(entry.date).getTime(),
+    new Date(entry.date).getTime() + 1000 * 60 * 60 * 24 - 1,
```
florian-sanders-cc: **fix:** small issue spotted by an LLM in review

Image

should be 
```js
    new Date(entry.date).getTime() + 1000 * 60 * 60 - 1,
```

florian-sanders-cc (changes requested):
A small fix but apart from that, LGTM! 👍

florian-sanders-cc (approved):

PR #1757, opened 2026-05-29
fix(cc-homepage-video): drop cc-block and size the video via container queries

hsablonniere:
## Context

The `cc-homepage-video` component has been through multiple sizing strategies, each with tradeoffs:

### State 1: Before commit 1c27a3f (no cc-block wrapper)
- Not using `<cc-block>` inside and thus incorrect border styling
- Homepage row layout: the video imposes its height on neighboring sibling (`cc-homepage-template-project`), breaking the height of the list items

image

- Homepage column layout: works correctly

image

### State 2: Commit 1c27a3f (cc-block wrapper)
- Attempted to fix by using `cc-block` wrapper to enforce a bordered container
  - Forgot to remove `cc-block` selector with wrong border color overriding the default one
- Implementation became complex: `position: absolute` for the `.content-wrapper`, flex centering, hardcoded `top: 5em`
- Homepage row layout: `cc-homepage-template-project` does cannot impose its height on the video

image

- Homepage column layout: awkward spacing around the video, especially at the bottom

image

## Proposal

Drop `cc-block` and use a direct **CSS Grid layout** with **container queries** for sizing the video. This allows the component to adapt to its parent's constraints (both width and height) without complex positioning or forced dimensions.

### Key changes
- Remove `cc-block` wrapper → simpler DOM and CSS
- Reuse existing CSS custom props for border, title font size and color
- Root `:host` uses CSS Grid (`grid-template-columns: 1fr auto; grid-template-rows: auto 1fr`) to position header and video area separately
- Video area constrained via **container queries** with a variable aspect ratio
- Expose sizing behavior through new CSS custom property: `--cc-homepage-video-container-type`
  - Default: `inline-size` (width-driven) — video fills available width, height follows 16/9 ratio
  - Opt-in: `size` (contained) — video fits within both width and height constraints, centered

### How container queries work

The video always preserves its 16/9 aspect ratio via CSS:
```css
width: min(100cqw, 100cqh * var(--video-aspect-ratio))
```

This ensures:
- `100cqw`: never wider than the container
- `100cqh * (16/9)`: width that keeps the video within the available height

**Why `inline-size` is the safe default:**
- Only the width is a query reference; height resolves against the viewport
- Safe even without a definite parent height (typical in single-column layouts)
- The second term of `min()` is huge and inert in this mode

**When to use `size`:**
- Set on the component when it has an imposed height (e.g., a grid cell with fixed or flex height)
- Both axes become query references, allowing the video to shrink if the parent is narrow and tall (or wide and short)
- Requires coordinating with the parent's layout

### Test cases
- **Width-driven (default)**: row and column layouts in the homepage work correctly
  - Row: the video adapts to its container's width; siblings can impose their own heights
  - Column: no more awkward spacing, clean integration with the header

Homepage row layout:

Capture d’écran du 2026-05-29 20-12-21

Homepage column layout: 

Capture d’écran du 2026-05-29 20-12-17

## Breaking changes

The `cc-homepage-video` component is not a breaking change in terms of markup. However:

- **Homepage opt-in required:** When using the component in a row layout (where both width and height are constrained), set `--cc-homepage-video-container-type: size` on the component instance
  - We will do this in the `new-menu` branch of the menu refactor

### DIFF
--- a/src/components/cc-homepage-video/cc-homepage-video.js
+++ b/src/components/cc-homepage-video/cc-homepage-video.js
@@ -2,7 +2,6 @@ import { css, html, LitElement } from 'lit';
 import { createRef, ref } from 'lit/directives/ref.js';
 import { iconRemixPlayFill } from '../../assets/cc-remix.icons.js';
 import { i18n } from '../../translations/translation.js';
-import '../cc-block/cc-block.js';
 import '../cc-icon/cc-icon.js';
 import '../cc-link/cc-link.js';
 
@@ -11,7 +10,19 @@ import '../cc-link/cc-link.js';
  *
  * Clicking the play button replaces the thumbnail with an embedded YouTube iframe.
  *
- * @cssdisplay block
+ * ## Sizing
+ *
+ * The embedded video always keeps a 16/9 ratio. Two layouts are supported:
+ *
+ * - **Width-driven (default):** give the component a width and let its height grow on its own. The video
+ *   fills the width and its height follows the 16/9 ratio. Typical in a single-column layout.
+ * - **Contained:** give the component both a width and a height (e.g. a grid cell whose height is imposed
+ *   by a sibling) and set `--cc-homepage-video-container-type: size`. The 16/9 video is then scaled to fit
+ *   within both dimensions and centered below the header.
+ *
+ * @cssdisplay grid
+ *
+ * @cssprop {normal|size|inline-size} --cc-homepage-video-container-type - Sizing mode of the video area. Defaults to `inline-size` (width-driven). Set to `size` when the component has an imposed height, to contain the 16/9 video within both dimensions.
  */
 export class CcHomepageVideo extends LitElement {
   static get properties() {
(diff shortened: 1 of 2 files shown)


clara-layus-cc (approved):
LGTM !

src/components/cc-homepage-video/cc-homepage-video.js
```diff
@@ -11,7 +10,19 @@ import '../cc-link/cc-link.js';
...
+ *
+ * The embedded video always keeps a 16/9 ratio. Two layouts are supported:
+ *
+ * - **Width-driven (default):** give the component a width and let its height grow on its own. The video
+ *   fills the width and its height follows the 16/9 ratio. Typical in a single-column layout.
+ * - **Contained:** give the component both a width and a height (e.g. a grid cell whose height is imposed
+ *   by a sibling) and set `--cc-homepage-video-container-type: size`. The 16/9 video is then scaled to fit
+ *   within both dimensions and centered below the header.
+ *
  * @cssdisplay block
```
HeleneAmouzou:
```suggestion
 * @cssdisplay grid
```

HeleneAmouzou (approved):
Just a small suggestion, otherwise LGTM !

pdesoyres-cc (approved):

src/components/cc-homepage-video/cc-homepage-video.js
hsablonniere: Done in 128c392a, thanks!

hsablonniere:
Thanks a lot @clara-layus-cc, @HeleneAmouzou and @pdesoyres-cc for the reviews! 🙏

PR #1761, opened 2026-06-10
fix(cc-search-bar): add tabindex to sections and widen dialog

clara-layus-cc:
# What does this PR do?

- Widens the cc-search-bar dialog from 38em (the cc-dialog default) to 45em via the --cc-dialog-width custom property.
- Adds tabindex="-1" to the scrollable .sections container so it can be focused and scrolled with the keyboard without entering the tab order.

# How to review?

- Check the commits,
- Check the preview


florian-sanders-cc (approved):
LGTM thanks for handling this @clara-layus-cc 👍

roberttran-cc (approved):
Tested and looking good! 🚀

PR #1763, opened 2026-06-15
ci: fix visual tests & preview workflows

florian-sanders-cc:
## What does this PR do?

- The move from `npm` to `pnpm` has bumped:
  - `playwright` from `1.55.0` to `1.60.0`,
    - We need to keep the Playwright docker image version in sync with the actual `playwright` version (playwright throws an error when it's out of sync).
  - `@aws-sdk/client-s3` from `3.29.0` (2021 :grimacing: ) to `3.1049.0` (latest),
    - Since v3.729.0, @aws-sdk/client-s3 enables integrity checksums by default (WHEN_SUPPORTED). 
    - This triggers errors with Cellar as documented in the Clever Cloud docs: https://www.clever.cloud/developers/doc/addons/cellar/#using-sdks (Node.js tab)
- When we refactored the `manage-comment.js` file to set up a common way to check for required env vars, we accidentally broke it (the initial design I implemented was too fragile / too easy to break).
  - This PR replaces our custom `validate-env` script that broke during our last refactoring with the `@clevercloud/reglage` package (slightly more verbose but a lot more robust).

## How to review?

- Let the CI do its job :robot:,


.github/scripts/config-utils.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Print the issues from a reglage validation error and exit.
+ *
+ * @param {InvalidConfigError} error
+ * @returns {never}
+ */
+export function exitWithConfigErrors(error) {
+  console.error('Environment validation errors:');
+  const issues = /** @type {InvalidConfigError} */ error.issues ?? [];
```
roberttran-cc: **🔨 issue:** type cast probably off (non-blocking)

.github/workflows/manage-visual-tests.yml
roberttran-cc: **🔨 issue:** version should be up bumped here too

Consider as a follow-up to hoist it to one common source.

florian-sanders-cc: nice catch and agreed about the common source. Not sure how we can do that but I'll ask around (and if I don't find any good solution right now I'll merge without common source + track in an issue) :+1:

.github/scripts/config-utils.js
florian-sanders-cc: yes it's obsolete, thanks :+1:

roberttran-cc (approved):
LGTM!

PR #1766, opened 2026-06-17
fix(cc-plan-picker): prevent detail icon from shrinking when description wraps

roberttran-cc:
## Context

In `cc-plan-picker`, each option-footer detail row is a flex row (`display: inline-flex`) containing a leading `cc-icon` and a description `<span>`. When the description wraps to multiple lines, the icon shrinks (flex item default `flex-shrink: 1`) and gets distorted.

## Fix

Expose the detail icon as a CSS part (`option-footer--detail-icon`) and set `flex: 0 0 auto` so it keeps its intrinsic size regardless of how the description wraps.

## How to test

Check the `cc-plan-picker` stories in Storybook (`defaultStory` / `readonly`): the `DEV` plan now ships long, wrapping detail descriptions — the leading icons stay full-size and undistorted.

> Only one reviewer is required for this PR.

### DIFF
--- a/src/components/cc-plan-picker/cc-plan-picker.js
+++ b/src/components/cc-plan-picker/cc-plan-picker.js
@@ -233,7 +233,7 @@ export class CcPlanPicker extends CcFormControlElement {
       <ul part="option-footer">
         ${details.map((detail) => {
           return html`<li part="option-footer--detail">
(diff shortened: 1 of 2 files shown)


florian-sanders-cc (approved):
LGTM! 👍

PR #1768, opened 2026-06-17
feat(cc-addon-admin): emit cc-addon-tags-was-changed after tags save

hsablonniere:
## Context

The new console sidebar groups apps and add-ons into `project:` sections driven by **tags**, which only live on `/v2/summary` — so a tag edit must trigger a full summary refresh.

For **apps** the console owns the save and refreshes right after the PUT resolves. For **add-ons**, `cc-addon-admin`'s own smart saves the tags but — unlike the name field — emitted no post-save success event. The console only saw the input intent (`cc-addon-tags-change`) and worked around it with a fragile `setTimeout(…, 1000)`:

- save **fails** → console refetches the **old** tags, sidebar stays stale
- save **> 1 s** → same stale outcome
- save **< 1 s** → needless wait

## Proposal

Add a `cc-addon-tags-was-changed` event, dispatched by the smart **only after the tag save resolves successfully**, mirroring the existing `cc-addon-name-was-changed` flow:

- payload `{ id, tags }` (consistent with `cc-addon-name-was-changed`'s `{ id, name }`)
- not emitted on error → the error keeps going through `notifyError`, as for the name field
- the input intent event `cc-addon-tags-change` is unchanged

### Files

- `cc-addon-admin.events.js` — new `CcAddonTagsWasChangedEvent`
- `cc-addon-admin.smart.js` — dispatch in the tags-save `.then()`
- `lib/events-map.types.d.ts` — regenerated (`pnpm components:events-map-generate`)
- `cc-addon-admin.smart.md` — added to the "Events fired" table

### Design decisions

- **Payload `{ id, tags }`** rather than `{ tags }`: mirrors `cc-addon-name-was-changed` for consistency across the component's post-save events.
- **No tests added**, intentionally aligned with the reference flow: `cc-addon-name-was-changed` has no test, and the repo has no harness for testing smart components against a mocked API. The event fires only from the smart, so neither the stories nor the dumb component can exercise it.

### DIFF
--- a/src/components/cc-addon-admin/cc-addon-admin.events.js
+++ b/src/components/cc-addon-admin/cc-addon-admin.events.js
@@ -70,3 +70,16 @@ export class CcAddonNameWasChangedEvent extends CcEvent {
     super(CcAddonNameWasChangedEvent.TYPE, detail);
   }
 }
+
+/**
+ * Dispatched when an add-on's tags have been changed successfully.
+ * @extends {CcEvent<{id: string, tags: Array<string>}>}
+ */
+export class CcAddonTagsWasChangedEvent extends CcEvent {
+  static TYPE = 'cc-addon-tags-was-changed';
+
+  /** @param {{ id: string, tags: Array<string> }} detail */
+  constructor(detail) {
+    super(CcAddonTagsWasChangedEvent.TYPE, detail);
+  }
+}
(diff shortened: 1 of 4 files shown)


florian-sanders-cc (approved):
LGTM!

PR #1769, opened 2026-06-18
refactor(cc-logs): migrate log virtualizer from @lit-labs to @tanstack

pdesoyres-cc:
## Context

`cc-logs` virtualizes an effectively unbounded, continuously-streaming log list. It was built
on `@lit-labs/virtualizer`, which is experimental and behaves poorly under a sustained append
load — it has been the source of a string of recurring bugs: a Firefox-only ResizeObserver
memory leak, overlapping log lines, and duplicate-key crashes from its keyed rendering.

On top of that, "follow the bottom" was hand-rolled on raw scroll geometry. During a streaming
burst the virtualizer's own programmatic re-pinning scrolls were indistinguishable from a user
scrolling up, so `follow` flickered off mid-append and fired spurious `cc-logs-follow-change`
events.

## Changes

- Migrate the virtualizer from `@lit-labs/virtualizer` to `@tanstack/lit-virtual`, which is
  actively maintained and exposes the low-level control (explicit element measurement, manual
  cache and rendering management) needed to keep memory bounded and avoid the streaming bugs.
- Drive "follow" from the virtualizer's native bottom-anchoring (`anchorTo: 'end'`) instead of
  inferring it from scroll deltas. A single `scrollToEnd()` per append does the pinning, and
  `follow` is derived only when `scrollState` confirms no programmatic scroll is in flight —
  giving an exact user-vs-virtualizer discriminator where the old heuristic only approximated.
- Add `cc-logs.follow.test.js` and `logs-controller.test.js` covering empty-list append,
  throttled streaming, user scroll-up / re-follow, and at-limit front-trim.
- Add a bulk "Add +1000 logs" sandbox button to make the streaming / bottom-pinning paths
  easy to trigger by hand.

### Implementation notes

The list is rendered with **flow rendering** (rows in document flow, not absolutely positioned
in a translated window) so concurrent reconciles can no longer overlap two rows.

Rows are **index-keyed**, not id-keyed: under `@tanstack/virtual` 3.17 id-keying leaks unbounded
size/element caches over an endless stream. The cost is that FIFO front-trimming (once `limit`
is reached) shifts every surviving log's index down while the index-keyed size cache does not
follow, so cached heights would apply to the wrong rows. `LogsController.consumeIndexSpaceChange()`
reports the per-render front-trim count (and full-rebuild flag on filter/clear) so the host can
realign that cache instead of dropping it.

## How to review

1. Start with `bdc8c8d1` (the bare migration) then read the three follow-up commits in order —
   each commit message explains the specific failure mode it addresses.
2. Read `src/components/cc-logs/cc-logs.js` (the virtualizer wiring + follow derivation) and
   `logs-controller.js` (the `consumeIndexSpaceChange()` index-space accounting).
3. Run the new tests: `cc-logs.follow.test.js` and `logs-controller.test.js`.
4. Exercise it by hand in the sandbox: stream with the new "Add +1000 logs" button, scroll up
   mid-stream and confirm follow turns off cleanly, then scroll back to the bottom to re-follow.
   Check Firefox memory stays bounded over a long stream.

**1 approval** from a front-end reviewer.

### DIFF
--- a/src/components/cc-logs/cc-logs.follow.test.js
+++ b/src/components/cc-logs/cc-logs.follow.test.js
@@ -1,0 +1,174 @@
+import { expect, fixture, nextFrame } from '@open-wc/testing';
+import { html } from 'lit';
+import './cc-logs.js';
+
+/** @param {number} length @param {number} offset */
+function generateLogs(length, offset = 0) {
+  return Array.from({ length }, (_, i) => {
+    const id = offset + i;
+    return {
+      id: `log-${id}`,
+      date: new Date(1600000000000 + id * 1000),
+      message: `Message ${id} lorem ipsum dolor sit amet`,
+      metadata: [{ name: 'instance', value: `instance-${id % 4}` }],
+    };
+  });
+}
+
+/** @param {number} ms */
+function wait(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+/** @param {import('./cc-logs.js').CcLogs} el */
+function snap(el) {
+  const c = el.shadowRoot.querySelector('.logs_container');
+  const v = el._getVirtualizer();
+  return {
+    follow: el.follow,
+    scrollTop: Math.round(c.scrollTop),
+    distToBottom: Math.round(v.getTotalSize() - c.clientHeight - c.scrollTop),
+  };
+}
+
+describe('cc-logs native follow', function () {
+  // The "ResizeObserver loop completed with undelivered notifications" warning is benign and already filtered by the
+  // project's web-test-runner config; swallow it here so it doesn't fail these strict, churn-heavy tests.
+  before(function () {
+    const original = window.onerror;
+    window.onerror = function (message, ...rest) {
+      if (typeof message === 'string' && message.includes('ResizeObserver loop')) {
+        return true;
+      }
+      return original ? original.call(this, message, ...rest) : false;
+    };
+  });
+
+  it('follows the initial big append', async function () {
+    const el = await fixture(html`<cc-logs-beta follow style="display:block; height:300px;"></cc-logs-beta>`);
+    await el.updateComplete;
+    el.appendLogs(generateLogs(2000));
+    for (let i = 0; i < 8; i++) {
+      await nextFrame();
+    }
+    const s = snap(el);
+    expect(s.follow, 'follow stays on').to.equal(true);
+    expect(s.scrollTop, 'pinned to bottom').to.be.greaterThan(0);
+    expect(s.distToBottom, 'at bottom').to.be.lessThan(8);
+  });
+
+  it('keeps following through a throttled stream then silence', async function () {
+    const el = await fixture(html`<cc-logs-beta follow style="display:block; height:300px;"></cc-logs-beta>`);
+    await el.updateComplete;
+    const transitions = [];
+    el.addEventListener('cc-logs-follow-change', (e) => transitions.push(e.detail));
+    for (let b = 0; b < 25; b++) {
+      el.appendLogs(generateLogs(200, b * 200));
+      await nextFrame();
+    }
+    await wait(300);
+    await nextFrame();
+    const s = snap(el);
+    expect(s.follow, `follow stayed on (transitions=${transitions})`).to.equal(true);
+    expect(s.distToBottom).to.be.lessThan(8);
+  });
+
+  it('unfollows on user scroll-up and re-follows at the bottom', async function () {
+    const el = await fixture(html`<cc-logs-beta follow style="display:block; height:300px;"></cc-logs-beta>`);
+    await el.updateComplete;
+    el.appendLogs(generateLogs(2000));
+    for (let i = 0; i < 8; i++) {
+      await nextFrame();
+    }
+    const c = el.shadowRoot.querySelector('.logs_container');
+
+    // User scrolls up.
+    c.scrollTop = 100;
+    c.dispatchEvent(new Event('scroll'));
+    await nextFrame();
+    expect(el.follow, 'unfollowed after scroll-up').to.equal(false);
+
+    // New logs arrive: view must NOT jump to bottom.
+    el.appendLogs(generateLogs(200, 2000));
+    for (let i = 0; i < 4; i++) {
+      await nextFrame();
+    }
+    expect(el.follow, 'stays unfollowed while reading history').to.equal(false);
+    expect(c.scrollTop, 'view stayed up').to.be.lessThan(2000);
+
+    // User scrolls back to the bottom.
+    c.scrollTop = c.scrollHeight;
+    c.dispatchEvent(new Event('scroll'));
+    await nextFrame();
+    expect(el.follow, 're-followed at bottom').to.equal(true);
+  });
+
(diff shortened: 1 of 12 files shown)


src/components/cc-logs/cc-logs.follow.test.js
```diff
@@ -0,0 +1,124 @@
...
+    };
+  });
+
+  it('follows the initial big append', async function () {
+    const el = await fixture(html`<cc-logs-beta follow style="display:block; height:300px;"></cc-logs-beta>`);
+    await el.updateComplete;
+    el.appendLogs(generateLogs(2000));
+    for (let i = 0; i < 8; i++) {
+      await nextFrame();
+    }
```
roberttran-cc, line 53: **💭 thought:** (from Claude, see fix suggestion in my main message)

> RAF/timer cadence under CI load isn't guaranteed, so "throttled stream then silence" and "capped at limit (front-trim)" are the most likely to flake.
> Prefer polling a settle condition (e.g. loop until `v.getTotalSize()` stabilises and `scrollState == null`) over magic frame counts.

src/components/cc-logs/cc-logs.follow.test.js
```diff
@@ -0,0 +1,124 @@
...
+  // project's web-test-runner config; swallow it here so it doesn't fail these strict, churn-heavy tests.
+  before(function () {
+    const original = window.onerror;
+    window.onerror = function (message, ...rest) {
+      if (typeof message === 'string' && message.includes('ResizeObserver loop')) {
+        return true;
+      }
+      return original ? original.call(this, message, ...rest) : false;
+    };
+  });
```
roberttran-cc, line 45: **🔨 issue:** (from Claude, see fix suggestion in my main message)

> `before()` installs a `window.onerror` that swallows `ResizeObserver loop` messages but there is no `after()` restoring `original`. Web Test Runner shares the browser context across files, so this override leaks into subsequent suites and could mask real errors.
> Add `after(() => { window.onerror = original; })`.
Also worth checking whether it's needed at all: the project already filters that message centrally (`test/helpers/*.js` via `@lit-labs/virtualizer/.../resize-observer-errors.js`).

src/components/cc-logs/logs-controller.js
roberttran-cc: **🫣 nitpick:** (adjacent observation from Claude, pre-existing so out of scope)

> `if (this._focusedIndex > logsToRemoveFiltered.length)` should be `>=`. At the exact boundary `f === N` the focused log *survives* (becomes index 0) but its focus is cleared instead of moved.

src/components/cc-logs/logs-controller.js
```diff
@@ -88,10 +98,29 @@ export class LogsController {
...
     this._selection.clear();
+    this._pendingRebuild = true;
     this._updateList();
     this.clearFocus();
   }
 
+  /**
+   * Reports — and resets — how the filtered list's index space changed since the last call.
+   *
+   * The virtualizer caches each rendered log's measured height in a map keyed by index (see the `getItemKey` note in
```
roberttran-cc: **🫣 nitpick:** `getItemKey` mentioned does not exist anymore.

src/components/cc-logs/cc-logs.js
```diff
@@ -343,6 +349,93 @@ export class CcLogs extends LitElement {
...
+      paddingEnd: this._horizontalScrollbarHeight,
+      measureElement: (element) => {
+        // Measure the real rendered height of a log line.
+        // We guard against a 0 height: a `ResizeObserver` notification can report a 0 height transiently when an
+        // element is being detached or reused (which happens a lot when logs are appended by buckets). Caching a 0
+        // would make several lines share the same vertical offset and overlap. Falling back to the estimate keeps a
+        // sane height until the next valid measurement.
+        const height = Math.round(element.getBoundingClientRect().height);
+        return height > 0 ? height : ESTIMATED_LOG_LINE_HEIGHT;
+      },
```
roberttran-cc: **👀 suggestion:** (from Claude)

A minor optimization I guess.


```suggestion
      measureElement: (element, _entry, instance) => {
        // Measure the real rendered height of a log line.
        // We guard against a 0 height: a `ResizeObserver` notification can report a 0 height transiently when an
        // element is being detached or reused (which happens a lot when logs are appended by buckets). Caching a 0
        // would make several lines share the same vertical offset and overlap. We keep the row's last measured
        // height when we have one (so an already-measured row does not shrink to the estimate for a frame), and
        // fall back to the estimate only for a row that has never been measured.
        const height = Math.round(element.getBoundingClientRect().height);
        if (height > 0) {
          return height;
        }
        const index = Number(element.dataset.index);
        return instance.itemSizeCache.get(index) ?? ESTIMATED_LOG_LINE_HEIGHT;
      },
```

roberttran-cc (commented):
Wow, that's some heavy lifting. Great job.

I did the QA, LGTM. Claude did the code review with some feedback, overall non-blocking but maybe worth noting and treating it separately.

---

Below is Claude's suggestion for fixing `src/components/cc-logs/cc-logs.follow.test.js` both feedback:

```js
import { expect, fixture, nextFrame, waitUntil } from '@open-wc/testing';
import { isResizeObserverLoopErrorMessage } from '@lit-labs/virtualizer/support/resize-observer-errors.js';
import { setupIgnoreIrrelevantErrors } from '../../../test/helpers/story-testing-utils.js';

const AT_BOTTOM_PX = 8; // mirrors FOLLOW_SCROLL_THRESHOLD
const isPinnedToBottom = (el) => { const s = snap(el); return s.follow && s.distToBottom < AT_BOTTOM_PX; };

describe('cc-logs native follow', () => {
  // Fixes finding #3 too: restores window.onerror in after() + uses the canonical predicate.
  setupIgnoreIrrelevantErrors(before, after, isResizeObserverLoopErrorMessage);

  it('follows the initial big append', async () => {
    const el = await fixture(html`<cc-logs-beta follow style="display:block;height:300px;"></cc-logs-beta>`);
    el.appendLogs(generateLogs(2000));
    await waitUntil(() => isPinnedToBottom(el), 'settles pinned to bottom', { timeout: 3000 });
    expect(snap(el).scrollTop).to.be.greaterThan(0);
  });

  it('keeps following through a throttled stream then silence', async () => {
    const el = await fixture(/* … */);
    const unfollowed = [];
    el.addEventListener('cc-logs-follow-change', (e) => e.detail === false && unfollowed.push(e.detail));
    for (let b = 0; b < 25; b++) { el.appendLogs(generateLogs(200, b * 200)); await nextFrame(); }
    await waitUntil(() => isPinnedToBottom(el), 're-settles after stream', { timeout: 3000 });
    expect(unfollowed, 'no spurious unfollow during stream').to.be.empty;
  });

  it('keeps following once the list is capped at the limit (front-trim)', async () => {
    const el = await fixture(html`<cc-logs-beta follow limit="500" …></cc-logs-beta>`);
    for (let b = 0; b < 15; b++) { el.appendLogs(generateLogs(200, b * 200)); await el.updateComplete; await nextFrame(); }
    await waitUntil(() => isPinnedToBottom(el), 'pinned at limit', { timeout: 3000 });
    expect(el._logsCtrl.listLength).to.equal(500);
  });
});
```

florian-sanders-cc (approved):
Did not look at the code but tested with all options in prod and it works really well 🔥

PR #1770, opened 2026-06-19
feat(cc-search-bar): add item types and restore generated type definitions

hsablonniere:
## Context

While integrating `cc-search-bar` in the console, three gaps surfaced:

- only 4 `itemType` values were supported (`app`, `addon`, `network-group`, `cke`), missing OAuth consumers and add-on providers;
- the badge labels were hardcoded in the component (bypassing i18n), and the "OAuth consumer" / "add-on provider" terms were translated inconsistently across the app;
- the autodocs "Type Definitions" section was missing for this component — and, after investigation, for 4 others.

## Proposal

Four focused commits:

**`fix(cc-search-bar)` — stop translating oauth consumer and add-on provider**
Move the badge labels into the translation files (literal keys so `check-i18n` still detects them statically) and freeze a single untranslated English form for "OAuth consumer" / "add-on provider" everywhere (previously "consumer OAuth" / "fournisseur", with shifting casing). Also renames the `addon` badge `ADDON` → `ADD-ON`.

**`docs` — move @import blocks above class**
Fix the placement of the `@import` JSDoc block on 5 components (see Design decisions).

**`feat(cc-search-bar)` — add oauth-consumer and addon-provider item types**
Two new `itemType` values with their badges (both `neutral` intent):
- `oauth-consumer` → `OAUTH CONSUMER`
- `addon-provider` → `ADD-ON PROVIDER`

Labels now mirror the `itemType` (and the `is:` filter value) — addresses @florian-sanders-cc's review note.

**`docs(cc-search-bar)` — move story from Atoms to Navigation**

### Design decisions

The CEM `support-typedef-jsdoc` plugin only reads the `@import` tags **attached to the component class**. When the `@import` block is separated from `export class` by `const` declarations, TypeScript attaches it to the first `const` instead, so no type is resolved and the generated "Type Definitions" section is empty.

Moving the `@import` block right above the class declaration restores the auto-generated section. Same root cause fixed on 5 components:
- `cc-search-bar`
- `cc-network-group-dashboard`
- `cc-homepage-template-project`
- `cc-cellar-object-list`
- `cc-cellar-bucket-list`

No runtime / public API / consumer-side type change — only the generated Storybook docs are affected.

## How to test

1. `npm run storybook:dev`
2. Open `🛠 Navigation/<cc-search-bar>` docs → the **Type Definitions** section now lists `SearchBarSection`, `SearchBarItem`, `SearchBarItemType` (incl. the 2 new types) and `IconModel`.
3. In the `default` story, the new `oauth-consumer` / `addon-provider` items show `OAUTH CONSUMER` / `ADD-ON PROVIDER` badges.
4. Same Type Definitions section is back on the 4 other components.

## Out of scope

- Full `is:` discoverability (tooltip & autocomplete so users can guess valid filter values) — discussed separately, left for a follow-up.
- `cc-smart-container` and `cc-map-marker-dot` also have a misplaced `@import` but no constructor type, so no generated section is expected — left untouched.


src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -25,6 +25,8 @@ const ITEM_TYPE_BADGE = {
   addon: { label: 'ADDON', intent: 'warning' },
   'network-group': { label: 'NG', intent: 'neutral' },
   cke: { label: 'KUBE', intent: 'neutral' },
+  'oauth-consumer': { label: 'OAUTH', intent: 'neutral' },
+  'addon-provider': { label: 'PROVIDER', intent: 'neutral' },
```
florian-sanders-cc: **nitpick (non blocking & unrelated to this PR):** should label match the item type used for filtering? 
As a user it's hard to guess which will work with `is:` and for instance I expected `is: provider` to work since it's the label we display next to these items.

This is related to a bigger subject that we've already discussed (tooltip & autocomplete to help people) and can be dealt with later on 👍

florian-sanders-cc (approved):
All good, just a minor suggestion for later iterations 😉

src/components/cc-search-bar/cc-search-bar.js
hsablonniere: Good catch 🙏 The labels now line up with the item types (and their `is:` filter values): `OAUTH` → `OAUTH CONSUMER`, `PROVIDER` → `ADD-ON PROVIDER` (plus `ADDON` → `ADD-ON`). So the displayed badge mirrors `is:oauth-consumer` / `is:addon-provider`.

Full `is:` discoverability (tooltip + autocomplete to help users guess valid filters) stays out of scope here as you suggested — let's tackle it in a dedicated PR. Thanks!

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -19,14 +19,35 @@ import '../cc-input-text/cc-input-text.js';
...
+  switch (itemType) {
+    case 'app':
+      return i18n('cc-search-bar.badge.app');
+    case 'addon':
+      return i18n('cc-search-bar.badge.addon');
+    case 'network-group':
+      return i18n('cc-search-bar.badge.network-group');
+    case 'cke':
+      return i18n('cc-search-bar.badge.cke');
+  }
```
florian-sanders-cc, line 48: **praise:** I initially wanted to suggest having i18n keys even for words that translated the same in both EN & FR but figured it was out of scope. I like this change, we never know which language we have to support in the future, maybe these will need to be translated at some point in some language 🙌

florian-sanders-cc (approved):
All good! 😎

clara-layus-cc (approved):
LGTM !

HeleneAmouzou (approved):
LGTM !

PR #1773, opened 2026-06-25
fix(cc-network-group-member-list): prevent all members from being shown as deleted

florian-sanders-cc:
## Context

The network group API returns member `kind` in lowercase (e.g. "application"), but `GetNetworkGroupCommand` does not transform the payload and the kind-based switches in the smart component only matched uppercase values with no default case. As a result every member silently fell through to "DELETED", even when the underlying application or add-on still existed.

Normalize `member.kind` to uppercase before use, and add a default branch to the kind switches so an unexpected kind throws (surfacing an error state) rather than silently being treated as deleted.

## How to review?

- Check the demo-smart,


florian-sanders-cc:
Closing this in favor of a client PR since the transform would benefit both CLI and Components

PR #1776, opened 2026-06-26
fix(cc-domain-management): only treat real cleverapps.io domains as test domains

hsablonniere:
## Context

`isTestDomain()` used a bare suffix check (`hostname.endsWith('cleverapps.io')`), so any hostname *ending* in `cleverapps.io` matched — including unrelated domains such as `mycleverapps.io` or `notcleverapps.io`.

These were wrongly flagged as test domains and pushed into testing-only / HTTP-only behavior in `cc-domain-management`.

## Proposal

Match on the `.cleverapps.io` label boundary instead:

```js
return hostname.endsWith('.cleverapps.io');
```

This keeps every real app hostname (always a subdomain, e.g. `app-xxx.cleverapps.io`) while rejecting look-alike domains.

### Design decisions

- No `=== 'cleverapps.io'` apex case: a bare apex is never received as an app hostname in practice, so the leading-dot check is enough.

## How to test

```js
isTestDomain('app.cleverapps.io')   // true
isTestDomain('sub.main.cleverapps.io') // true
isTestDomain('mycleverapps.io')     // false (was true)
isTestDomain('notcleverapps.io')    // false (was true)
isTestDomain('example.com')         // false
```

### DIFF
--- a/src/lib/domain.js
+++ b/src/lib/domain.js
@@ -77,7 +77,7 @@ export function getDomainUrl(hostname, pathPrefix, isWildcard, isHttpOnly) {
  * @return {boolean}
  */
 export function isTestDomain(hostname) {
-  return hostname.endsWith('cleverapps.io');
+  return hostname.endsWith('.cleverapps.io');
 }
 
 /**


FlorentinDUBOIS:
We have to be able to provide a custom sets of domains that are corresponding to test or sandbox environment. 

For example for air gap or on-premise, we already have `*.<region>.<custom-test-domain>.` which implies a custom domain name per `region`. 

Besides, we must be able to configure it in console deployment.

src/lib/domain.js
```diff
@@ -77,7 +77,7 @@ export function getDomainUrl(hostname, pathPrefix, isWildcard, isHttpOnly) {
  * @return {boolean}
  */
 export function isTestDomain(hostname) {
-  return hostname.endsWith('cleverapps.io');
+  return hostname.endsWith('.cleverapps.io');
```
FlorentinDUBOIS, line 80: This is not air gap or on-premise compatible

FlorentinDUBOIS (changes requested):

hsablonniere:
Closing in favor of a new approach: instead of patching our local copy of the domain utils, we bump `@clevercloud/client` to 12.6.0 (which already ships the `.cleverapps.io` boundary fix in `isTestDomain`) and remove the duplicated `src/lib/domain.js`. New PR incoming.

PR #1777, opened 2026-06-29
refactor(cc-domain-management): use domain utils from @clevercloud/client

hsablonniere:
Replaces #1776.

## Context

`#1776` patched the `isTestDomain` boundary (`cleverapps.io` → `.cleverapps.io`) in `src/lib/domain.js`. That file turned out to be a verbatim copy of `@clevercloud/client/esm/utils/domains.js`, and the very same fix already ships upstream in `@clevercloud/client@12.6.0`. So instead of maintaining a diverging local copy, we consume the client.

## Changes

- `chore(deps)`: bump `@clevercloud/client` `^12.5.1` → `^12.6.0` (was resolving to 12.3.1).
- `refactor`: repoint `cc-domain-management.js` / `cc-domain-management.smart.js` imports to `@clevercloud/client/esm/utils/domains.js` (already the convention for `env-vars`/`payment` utils) and delete the duplicated `src/lib/domain.js`.

The `.cleverapps.io` boundary fix is therefore inherited from the client rather than duplicated locally.

## Validation

- `tsc -p tsconfig.ci.json` → 0 error (confirms `DomainInfo` structural compat with the client's `sortDomains` signature).
- `eslint` on both touched files → 0 error.


pdesoyres-cc (approved):
LGTM

PR #1783, opened 2026-06-30
feat(cc-feature-list): add select display mode

clara-layus-cc:
# What does this PR do?

- Add an optional `displayMode` property on `Feature` to render features as a `cc-select` instead of the default `cc-toggle`, useful when a feature has too many options for a toggle,
- Keep the select's mandatory label for accessibility while visually hiding it and compensating its reserved height so it stays vertically centered with the title,
- Add a `withSelect` story demonstrating the new display mode.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/cc-feature-list-select-mode/index.html?path=/story/%F0%9F%9B%A0-profile-cc-feature-list--with-select)


hsablonniere (changes requested):
Code is perfect 👍 
Can you update the stories so one of the features shows of this new display option?

src/components/cc-feature-list/cc-feature-list.js
```diff
@@ -193,6 +220,16 @@ export class CcFeatureList extends LitElement {
           font-weight: bold;
         }
 
+        /* The select's label is mandatory and kept for accessibility, but visually hidden to avoid duplicating the feature title. */
+        .feature-select {
+          --cc-form-label-gap: 0;
+          --cc-select-label-font-size: 0;
+
```
florian-sanders-cc: **fix:** unfortunately, setting `font-size` to 0 is not a reliable way to visually hide an element (screen readers may not read it even if technically, they should). 
This is why we have a `visually-hidden` class. 

This means we need to modify the `cc-select` element to support a `hidden-label` like `cc-input-*` elements do.

florian-sanders-cc (changes requested):
Well done @clara-layus-cc!

We have two issues about accessibility:
- all form controls must have a visible label. I think in this case, we can make an exception (I'm not sure WCAG / RGAA auditors would agree though),
- since we want to hide the label, we need to make sure it's done in a way that works for screen readers / assistive technologies (see my comment for more info).

florian-sanders-cc (approved):
All good! `hidden-label` breaks the `inline` mode of `cc-select` but it doesn't make sense to set `inline` with `hidden-label` so we're good 👍 
Gg @clara-layus-cc

PR #1789, opened 2026-07-21
feat(cc-search-bar): add keyboard navigation and footer

clara-layus-cc:
# What does this PR do?

- Add arrow-key navigation over the filtered results using a virtual focus: the real focus stays on the input while `ArrowUp`/`ArrowDown` move a looping visual selection,
- Activate the virtually focused item with `Enter`, mirroring a click on the corresponding link,
- Reset the virtual focus back to the input when the query changes or when the selected item disappears, and scroll the selection into view,
- Keep the default screen reader experience on purpose (no ARIA added),
- Add a footer displaying the keyboard hints (navigate / select / close).

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/cc-search-bar/keyboard-navigation/index.html?path=/story/%F0%9F%9B%A0-navigation-cc-search-bar--default-story)

### DIFF
--- a/src/components/cc-search-bar/cc-search-bar.js
+++ b/src/components/cc-search-bar/cc-search-bar.js
@@ -1,4 +1,5 @@
 import { css, html, LitElement, nothing } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
 import {
   iconRemixArrowRightSLine as iconArrowRight,
   iconRemixExternalLinkLine as iconExternalLink,
@@ -51,7 +52,9 @@ function getItemTypeBadgeLabel(itemType) {
 /**
  * @import { SearchBarItem, SearchBarItemType, SearchBarSection } from './cc-search-bar.types.js'
  * @import { BadgeIntent } from '../cc-badge/cc-badge.types.js'
+ * @import { IconModel } from '../common.types.js'
  * @import { CcInputEvent } from '../common.events.js'
+ * @import { PropertyValues } from 'lit'
  */
 
 /**
@@ -71,6 +74,7 @@ export class CcSearchBar extends LitElement {
       open: { type: Boolean, reflect: true },
       sections: { type: Array },
       value: { type: String },
+      _activeItemIndex: { type: Number, state: true },
     };
   }
 
@@ -85,6 +89,29 @@ export class CcSearchBar extends LitElement {
 
     /** @type {string} Sets the current search input value. */
     this.value = '';
+
+    /**
+     * @type {number} Index of the virtually focused item within the flattened list of filtered items.
+     * `-1` means no item is virtually focused: the real focus stays on the input.
+     */
+    this._activeItemIndex = -1;
+
+    /**
+     * @type {SearchBarSection[]} Cache of `_getFilteredSections()`, recomputed in `willUpdate` only when
+     * `value` or `sections` change (the filtering is otherwise run on every render, `updated` and keystroke).
+     */
+    this._filteredSections = [];
+
+    /** @type {SearchBarItem[]} Cache of the filtered items flattened into a single ordered list. */
+    this._flatItems = [];
+
+    // We listen to `keydown` in the capture phase because `cc-input-text` stops the propagation of keydown events
+    // during the bubbling phase (to avoid conflicts with shortcuts).
+    /** @type {{ handleEvent: (e: KeyboardEvent) => void, capture: boolean }} */
+    this._inputKeydownListener = {
+      handleEvent: (e) => this._onInputKeydown(e),
+      capture: true,
+    };
   }
 
   /** Opens the search bar dialog by setting the `open` property to true. */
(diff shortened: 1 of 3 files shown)


HeleneAmouzou (approved):
LGTM ! GG @clara-layus-cc ! 👏

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -134,33 +151,121 @@ export class CcSearchBar extends LitElement {
...
+    return html`
+      <div class="footer">
+        <span class="footer-hint">
+          <kbd class="footer-key">↑↓</kbd>
+          <span>${i18n('cc-search-bar.footer.navigate')}</span>
+        </span>
+        <span class="footer-hint">
+          <kbd class="footer-key">↵</kbd>
+          <span>${i18n('cc-search-bar.footer.select')}</span>
+        </span>
```
florian-sanders-cc: **suggestion:** since the screen reader nav is different, we could hide these with `aria-hidden="true"` on each `<span class="footer-hint">` so that screen reader users don't get this content? The arrows hint should be hidden I'd say but the select and close are still relevant 🤔

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -298,7 +409,8 @@ export class CcSearchBar extends LitElement {
 
         .sections {
           flex: 1;
-          max-height: calc(100dvh - 16em);
+          /* We reserve some room (compared to a footer-less layout) so the footer is never clipped by the dialog's \`overflow: hidden\`. */
+          max-height: calc(100dvh - 19em);
```
florian-sanders-cc: **suggestion:** arbitrary numbers which depend on other parts of the UI are usually brittle, they are easy to forget and annoying to debug (somebody changing the footer padding also needs to update this value that has no clear relationship with the footer padding).
It works but I think it's worth taking ~30 mins to:
- try to rethink the overall max-height on `.sections`, there might be other solutions by changing the DOM nesting and other CSS (haven't tried, it can be complex and maybe your current solution is the best so if you investigate, timebox this so you don't "waste" time),
- try to use CSS custom props instead of magic values if it makes sense (here I'm not sure we can do much more than what you did, except more complex `calc` usage),
- if no other solution is found, accept the current design but add some comment above the footer padding properties to warn that updating these values means updating values in `max-height` of `.sections` as well.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -417,9 +534,62 @@ export class CcSearchBar extends LitElement {
...
+          align-items: center;
+          background-color: var(--cc-color-bg-neutral, #f5f5f5);
+          border-top: solid 1px var(--cc-color-border-neutral-weak, #e7e7e7);
+          color: var(--cc-color-text-weak, #666);
+          display: flex;
+          flex-wrap: wrap;
+          gap: var(--cc-spacing-5, 1em);
+          justify-content: center;
+          margin-bottom: calc(-1 * var(--footer-dialog-padding));
+          margin-inline: calc(-1 * var(--footer-dialog-padding));
```
florian-sanders-cc, line 574: **note:** maybe our future dialog should offer a "footer" slot :shrug: (or maybe not, we'll see when we design it 😉 )

src/translations/translations.en.js
```diff
@@ -1874,6 +1874,9 @@ export const translations = {
   'cc-search-bar.external-link.name': `new window`,
   'cc-search-bar.external-link.title': /** @param {{linkText: string}} _ */ ({ linkText }) =>
     `${linkText} - new window`,
+  'cc-search-bar.footer.close': `Close`,
+  'cc-search-bar.footer.navigate': `Navigate`,
+  'cc-search-bar.footer.select': `Select`,
```
florian-sanders-cc: **nitpick (not sure):** could discuss with Marion but maybe "select" is too close to a dropdown behavior here, maybe we could be more explicit like "Access selected page" (but it's probably too long 😅)

florian-sanders-cc (approved):
Well done @clara-layus-cc most of my comments are suggestions but the component looks good and behaves as specified 👌

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -134,33 +151,121 @@ export class CcSearchBar extends LitElement {
...
+        <span class="footer-hint">
+          <kbd class="footer-key">↑↓</kbd>
+          <span>${i18n('cc-search-bar.footer.navigate')}</span>
+        </span>
+        <span class="footer-hint">
+          <kbd class="footer-key">↵</kbd>
+          <span>${i18n('cc-search-bar.footer.select')}</span>
+        </span>
+        <span class="footer-hint">
+          <kbd class="footer-key footer-key--text">ESC</kbd>
```
roberttran-cc, line 273: **👀 suggestion:** Consider a translation for `ESC` (`ECHAP` in french).

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -202,21 +308,26 @@ export class CcSearchBar extends LitElement {
...
+  _renderItem(item, index) {
     const isExternal = isExternalUrl(item.href);
     const badgeIntent = item.itemType != null ? ITEM_TYPE_BADGE_INTENT[item.itemType] : null;
     const title = isExternal ? i18n('cc-search-bar.external-link.title', { linkText: item.label }) : nothing;
+    const isActive = index === this._activeItemIndex;
     return html`
       <li>
         <a
-          class="item"
+          class="item ${isActive ? 'active' : ''}"
```
roberttran-cc: **👀 suggestion:** Consider using `classMap` (remove the trailing space when not active).

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -134,33 +151,121 @@ export class CcSearchBar extends LitElement {
...
+
+    if (e.key === 'ArrowDown') {
+      // We prevent the native behavior (moving the caret / scrolling).
+      e.preventDefault();
+      // Loop back to the first item when reaching the end.
+      this._activeItemIndex = this._activeItemIndex >= itemCount - 1 ? 0 : this._activeItemIndex + 1;
+    } else if (e.key === 'ArrowUp') {
+      e.preventDefault();
+      // Loop back to the last item when reaching the start (including from the input, index `-1`).
+      this._activeItemIndex = this._activeItemIndex <= 0 ? itemCount - 1 : this._activeItemIndex - 1;
```
roberttran-cc, line 199: **❓ question:** When first item is active and I press the up arrow, some users may expect to get the focus back on the text input.

But that would make the circling from the last to the first item on down arrow unexpected (and weird)? And also push the need for `HOME` and `END` support?

Yeah, that may be a bad idea after all but I typed all this text and I'm committed now. WDYT?

Would need @marion-izd and @florian-sanders-cc feedback on this one I guess.

src/components/cc-search-bar/cc-search-bar.js
```diff
@@ -134,33 +151,121 @@ export class CcSearchBar extends LitElement {
       .filter((section) => section.items.length > 0);
   }
 
+  /**
+   * Returns the filtered items flattened into a single ordered list, matching the order they are rendered in.
+   *
+   * @returns {SearchBarItem[]}
+   */
+  _getFlatItems() {
+    return this._getFilteredSections().flatMap((section) => section.items);
```
roberttran-cc: **👀 suggestion:** `_getFilteredSections` is run multiple times on keystroke, on direct call or via `_getFlatItems`, calls being `render`, `updated` and `_onInputKeydown`.

Consider cache filtered sections in `willUpdate` for a small optimization.

roberttran-cc (approved):
Overall LGTM, good job on this one! My feedback are minor and non-blocking.

And about the footer, my opinion is to ship as is and to open an issue to ask for a footer natively in `cc-dialog`.

src/components/cc-search-bar/cc-search-bar.js
florian-sanders-cc: @roberttran-cc I hesitated on this one as well, I find both "okay" in theory since the focus is technically always on the input itself (so my feedback is basically useless :smile:).

As for Home / End in this context I would argue against: these are already binded in a native input (start of the input / end of the input) for which we just removed the alternative (arrow up / down does the same as Home / End in a single line text input).

roberttran-cc: Hoooo ok, nevermind then, I didn't pay attention about the focus not leaving the input. :see_no_evil:

florian-sanders-cc (approved):

src/components/cc-search-bar/cc-search-bar.js
florian-sanders-cc: @clara-layus-cc not sure if you forgot this one but I think it's still relevant 👍

PR #1790, opened 2026-07-23
feat(cc-logs-app-runtime): add collapsible instances panel

KannarFr:
## Context — a suggestion, not a mandate 🙂

I'm coming at this as a **user** of `cc-logs-app-runtime` in the console, not as a maintainer of this component. On narrower screens the fixed-width **Instances** panel eats a lot of horizontal space, and most of the time I just want to read the log stream. So this is a small quality-of-life proposal — very happy for you to take it, tweak it, or close it if it doesn't fit your plans for the component.

## What it does

Adds a way to collapse/expand the left **Instances** panel so the log stream can use the full width:

- A toggle **button** in the logs header (next to the fullscreen button), using the `sidebar-fold` / `sidebar-unfold` Remix icons.
- A keyboard shortcut: **`Alt+I`** (I = Instances). I checked the existing shortcuts in `cc-logs` (`Ctrl+C`, `Ctrl+A`, arrows, `Home`/`End`, `Escape`, `Space`/`Enter`) — `Alt`+letter was a free namespace, so no collision.

## How it's built (kept minimal + consistent)

- **Mirrors the existing fullscreen toggle**: internal reactive state (`_instancesCollapsed`), no new public property, no new event — so no `events-map` change.
- The collapse simply drives the existing **`--instances-width`** custom property to `0`, so the "no logs / loading" overlays that already center off that variable stay correct for free.
- **Accessibility**: the button carries `aria-expanded` (via `a11y-expanded`) and a translated `a11y-name` (EN + FR added), which `cc-button` also surfaces as the hover `title`.
- **Motion**: the width transition is disabled under `@media (prefers-reduced-motion)`.
- The `Alt+I` handler is element-scoped (`@keydown` on the component root), matching the existing keyboard-handling approach — no new global listeners. `event.code === 'KeyI'` is used because `Alt`/`Option` mutates `event.key` on some layouts.
- Documented the shortcut in the component's class JSDoc, same as `cc-logs` documents its keyboard navigation.

## Checks

Ran the CI-gate pieces locally on the touched files: `eslint`, `stylelint`, `components:check-i18n` (EN + FR present), and `components:events-map-check` (up to date — no new events). Happy to add a collapsed-state **story** and/or the `run-visual-tests` label if you'd like that before merging — just say the word.

## Open questions for you

- Placement/icon of the toggle OK, or would you prefer it attached to the panel's own header?
- Should the collapsed state be **persisted** (the console already stores date-range/options via its SettingManager)? I left it ephemeral to keep the change small, but I can wire a public property/event if you want the console to remember it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
+++ b/src/components/cc-logs-app-runtime/cc-logs-app-runtime.js
@@ -4,6 +4,8 @@ import { createRef, ref } from 'lit/directives/ref.js';
 import {
   iconRemixFullscreenExitLine as fullscreenExitIcon,
   iconRemixFullscreenLine as fullscreenIcon,
+  iconRemixSidebarFoldLine as instancesCollapseIcon,
+  iconRemixSidebarUnfoldLine as instancesExpandIcon,
 } from '../../assets/cc-remix.icons.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-loader/cc-loader.js';
@@ -43,6 +45,11 @@ const CUSTOM_METADATA_RENDERERS = {
  */
 
 /**
+ * A component displaying the runtime logs of an application.
+ *
+ * ## Keyboard shortcuts
+ *
+ * * `Alt+I` to collapse or expand the instances panel (the left panel).
  *
  * @beta
  */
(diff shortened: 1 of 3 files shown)


hsablonniere:
Thanks a lot for this @KannarFr, and sorry for the delay.

As we discussed, we would rather get a well detailed issue that describes the problem, and let us evaluate the impacts, ask the right questions and find the right solution. That is easier on our side, where we manage the business concerns, the code and its history, and the rest of the roadmap.

The problem itself is a real one and we want to solve it: I have opened #1830 to track it, and referenced your proposal there. Closing this PR, the discussion continues in the issue.

PR #1796, opened 2026-08-05
feat(filter-sections): share the search and sidebar filter grammar

clara-layus-cc:
# Context

`cc-search-bar` and the console sidebar both narrow sections of items with the same token syntax, each implementing it separately — and they had drifted apart:

| | console sidebar | `cc-search-bar` |
|---|---|---|
| Keywords | `is:` and `project:` | `is:` only (`/^is:./`) |
| Matchers read | `item.matchers` | derived `is:<itemType>` + `item.matchers` |
| Text fields | `name`, `id`, `realId` | `label`, `id` |
| Empty query | keeps everything | returns nothing |

The first row is a user-visible bug: typing `project:<name>` in the global search is treated as free text, looked for literally in the label, and silently returns nothing — while the same filter works in the sidebar.

This extracts the grammar so both sides share one engine. Note the vocabulary was already shared (the console builds matchers once for both); what was duplicated is how a query is matched against it.

# What does this PR do?

- Add `lib/filter-sections.js`, holding the tokenizing and the matching semantics, which the console can import for its sidebar the way it already imports a dozen other lib modules,
- Keep what legitimately differs between call sites configurable: which fields hold the text, which prefixes mark a keyword, and whether an empty query yields everything or nothing,
- Make keyword prefixes an explicit list rather than a generic `<word>:` pattern, so free text keeping a colon (a URL, an id) still searches as text,
- Treat `project:<name>` as a keyword in `cc-search-bar`, matched against an item's matchers instead of a literal string looked for in its label,
- Cover the grammar with 22 unit tests,
- No breaking change: `is:<itemType>` is still folded into the matchers, an empty query still yields no result, and the component's API is untouched.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/share-filter-grammar/index.html?path=/story/%F0%9F%9B%A0-navigation-cc-search-bar--with-keyword-filter)

### DIFF
--- a/src/lib/filter-sections.js
+++ b/src/lib/filter-sections.js
@@ -68,8 +68,8 @@ export function matchesFilterQuery({ keywordTokens, textTokens }, matchers, text
  * @param {Array<S>} sections
  * @param {string} query
  * @param {object} [options]
- * @param {(item: any) => Array<string>} [options.getMatchers] - reads an item's matchers
- * @param {(item: any) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
+ * @param {(item: S['items'][number]) => Array<string>} [options.getMatchers] - reads an item's matchers
+ * @param {(item: S['items'][number]) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
  * @param {Array<string>} [options.keywordPrefixes]
  * @param {'all'|'none'} [options.emptyQuery] - what an empty query yields: every non-empty section, or
  * nothing. A sidebar shows everything unfiltered; a search bar shows no result until something is typed.
--- a/src/lib/filter-sections.js
+++ b/src/lib/filter-sections.js
@@ -2,10 +2,8 @@
  * Filter grammar shared by the resource lists that narrow sections of items with a text query:
  * `cc-search-bar` (global search) and the console's own sidebar.
  *
- * The two used to implement the grammar separately and had drifted apart — `project:` was a keyword on
- * one side and free text on the other, and they did not search the same fields. The tokenizing and the
- * matching semantics live here; what differs legitimately between call sites (which fields hold the
- * text, what an empty query means) stays configurable.
+ * The tokenizing and the matching semantics live here; what differs legitimately between call sites
+ * (which fields hold the text, what an empty query means) stays configurable.
  *
  * Grammar:
  * - the query is lowercased, trimmed, and split on whitespace into tokens,
@@ -74,7 +72,7 @@ export function matchesFilterQuery({ keywordTokens, textTokens }, matchers, text
  * @param {(item: any) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
  * @param {Array<string>} [options.keywordPrefixes]
  * @param {'all'|'none'} [options.emptyQuery] - what an empty query yields: every non-empty section, or
- *   nothing. A sidebar shows everything unfiltered; a search bar shows no result until something is typed.
+ * nothing. A sidebar shows everything unfiltered; a search bar shows no result until something is typed.
  * @returns {Array<S>} the sections holding at least one matching item
  */
 export function filterSections(sections, query, options = {}) {
--- a/src/lib/filter-sections.js
+++ b/src/lib/filter-sections.js
@@ -1,0 +1,103 @@
+/**
+ * Filter grammar shared by the resource lists that narrow sections of items with a text query:
+ * `cc-search-bar` (global search) and the console's own sidebar.
+ *
+ * The two used to implement the grammar separately and had drifted apart — `project:` was a keyword on
+ * one side and free text on the other, and they did not search the same fields. The tokenizing and the
+ * matching semantics live here; what differs legitimately between call sites (which fields hold the
+ * text, what an empty query means) stays configurable.
+ *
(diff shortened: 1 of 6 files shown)


src/lib/filter-sections.js
```diff
@@ -0,0 +1,103 @@
+/**
+ * Filter grammar shared by the resource lists that narrow sections of items with a text query:
+ * `cc-search-bar` (global search) and the console's own sidebar.
+ *
+ * The two used to implement the grammar separately and had drifted apart — `project:` was a keyword on
+ * one side and free text on the other, and they did not search the same fields. The tokenizing and the
+ * matching semantics live here; what differs legitimately between call sites (which fields hold the
+ * text, what an empty query means) stays configurable.
```
HeleneAmouzou:
```suggestion
 * The tokenizing and the matching semantics live here; what differs legitimately between call sites (which fields hold the text, what an empty query means) stays configurable.
```

src/lib/filter-sections.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @template {{ items: Array<any> }} S
+ * @param {Array<S>} sections
+ * @param {string} query
+ * @param {object} [options]
+ * @param {(item: any) => Array<string>} [options.getMatchers] - reads an item's matchers
+ * @param {(item: any) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
+ * @param {Array<string>} [options.keywordPrefixes]
+ * @param {'all'|'none'} [options.emptyQuery] - what an empty query yields: every non-empty section, or
+ *   nothing. A sidebar shows everything unfiltered; a search bar shows no result until something is typed.
```
HeleneAmouzou:
```suggestion
 * nothing. A sidebar shows everything unfiltered; a search bar shows no result until something is typed.
```

HeleneAmouzou (changes requested):
Not much to say, GG ! 
Now that we can use `project:<name>` in the `cc-search-bar`, we need to update its stories ; update `withKeywordFilter` doc + ~~add a story with `project:` keyword~~

src/lib/filter-sections.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Filters sections and their items with the grammar above, without mutating the input.
+ *
+ * @template {{ items: Array<any> }} S
+ * @param {Array<S>} sections
+ * @param {string} query
+ * @param {object} [options]
+ * @param {(item: any) => Array<string>} [options.getMatchers] - reads an item's matchers
+ * @param {(item: any) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
```
Galimede: **nitpick (non-blocking):** `any` in the JSDoc turns off typechecking on the accessors.

`@template {{ items: Array<any> }} S` plus `(item: any)` means the `getMatchers`/`getTexts` callbacks are unchecked. I verified with a throwaway probe — these both pass `pnpm typecheck` today:

```js
filterSections(sections, "x", {
  getMatchers: (item) => [`is:${item.itemTypoooo}`],
  getTexts: (item) => [item.labelTypoooo, item.id],
});
```

Switching the two params to `S['items'][number]` catches all of them (`TS2339: Property 'itemTypoooo' does not exist on type 'SearchBarItem'`), the internal defaults still compile (the indexed access degrades to `any` inside the generic body), and the real `cc-search-bar.js` call site passes unchanged.

```suggestion
 * @param {(item: S['items'][number]) => Array<string>} [options.getMatchers] - reads an item's matchers
 * @param {(item: S['items'][number]) => Array<string|null|undefined>} [options.getTexts] - reads an item's text fields
```

Filed as a nitpick, but it looks like the cheapest durable win here: two words, no behaviour change, and it restores checking on the exact callbacks the abstraction exists to parameterise. `cc-search-bar` is a `strict` + `checkJs` call site that goes unchecked today, so a rename in `SearchBarItem` would only show up at runtime.

src/lib/filter-sections.js
```diff
@@ -0,0 +1,103 @@
...
+ * - every token must pass (AND between tokens, OR across an item's text fields),
+ * - a section with no item is dropped, whether the query emptied it or it was already empty.
+ *
+ * The prefixes are an explicit list rather than a generic `<word>:` pattern on purpose: free text can
+ * legitimately contain a colon (a URL, an `app_xxx:yyy` id), and treating those as keywords would
+ * silently return nothing.
+ */
+
+/** @type {Array<string>} */
+export const DEFAULT_KEYWORD_PREFIXES = ['is:', 'project:'];
```
Galimede, line 23: **question (non-blocking):** now that `project:` is a first-class keyword, should `cc-search-bar.initial.description` mention it — or is that deliberately deferred?

Commenting here since the translation files are outside this diff. The shipped help text (`src/translations/translations.en.js:1882` and `translations.fr.js:1905`) — the one users read on the initial screen to discover the syntax — still advertises `is:` only:

> Start searching by keywords, id or filter (for example: `is:app`, `is:addon`…).

Galimede (approved):
Hey Clara, overall LGTM! 💪 

I agree with Hélène too for the update of the stories.

HeleneAmouzou (approved):
LGTM !

PR #1797, opened 2026-08-06
feat(cc-domain-management): dispatch cc-domain-list-change after add and delete

clara-layus-cc:
# What does this PR do?

- Dispatch a new `cc-domain-list-change` event from the smart component, once a domain addition or deletion has been acknowledged by the API,
- Unlike `cc-domain-add` and `cc-domain-delete`, which are intents dispatched before the call, this one lets consumers deriving something from the domain list refetch safely. `cc-domain-primary-change` only covered the mark-as-primary case, and deleting the primary domain never re-emits it,
- Use a single event carrying no detail, on purpose: it only says that the list changed, not how. Consumers are expected to refetch rather than patch their own copy, and the id of a freshly added domain is a client-side guess until the API returns real ids,
- Leave the mark-as-primary path untouched, since list membership is unchanged there,
- Document the event in the smart component page and regenerate the events map.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/domain-list-change-event/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story)

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.events.js
+++ b/src/components/cc-domain-management/cc-domain-management.events.js
@@ -63,3 +63,20 @@ export class CcDomainPrimaryChangeEvent extends CcEvent {
     super(CcDomainPrimaryChangeEvent.TYPE, detail);
   }
 }
+
+/**
+ * Dispatched when a domain addition or deletion has been taken into account.
+ *
+ * Unlike `cc-domain-add` and `cc-domain-delete`, which are intents dispatched before the API call, this one is
+ * dispatched once the API has answered. Consumers deriving something from the domain list can safely refetch
+ * when they receive it.
+ *
+ * @extends {CcEvent}
+ */
+export class CcDomainListChangeEvent extends CcEvent {
+  static TYPE = 'cc-domain-list-change';
+
+  constructor() {
+    super(CcDomainListChangeEvent.TYPE);
+  }
+}
--- a/src/components/cc-domain-management/cc-domain-management.smart.js
+++ b/src/components/cc-domain-management/cc-domain-management.smart.js
@@ -13,7 +13,7 @@ import { sendToApi } from '../../lib/send-to-api.js';
 import { defineSmartComponent } from '../../lib/smart/define-smart-component.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-smart-container/cc-smart-container.js';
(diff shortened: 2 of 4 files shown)


src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -114,6 +114,9 @@ defineSmartComponent({
               domainListState.domains.push(newDomain);
             },
           );
+
+          // Dispatch event to make the console refresh its UI
```
HeleneAmouzou: **nitpick:** even if the console is the main clever-components consumer, I think we should avoid mention it on the project as the clever-component can be used by other consumers.

src/components/cc-domain-management/cc-domain-management.smart.js
```diff
@@ -160,6 +163,11 @@ defineSmartComponent({
               domainListState.domains = domainListState.domains.filter((domain) => domain.id !== id);
             },
           );
+
+          // Dispatch event to make the console refresh its UI. Deleting the primary domain does not re-emit
```
HeleneAmouzou: **nitpick:** same here

HeleneAmouzou (approved):
Only nitpicks (that you can ignore), otherwise LGTM !

src/components/cc-domain-management/cc-domain-management.smart.js
Galimede: Agree 👍

src/components/cc-domain-management/cc-domain-management.events.js
```diff
@@ -63,3 +63,20 @@ export class CcDomainPrimaryChangeEvent extends CcEvent {
     super(CcDomainPrimaryChangeEvent.TYPE, detail);
   }
 }
+
+/**
+ * Dispatched when a domain addition or deletion has been taken into account.
+ *
+ * Unlike `cc-domain-add` and `cc-domain-delete`, which are intents dispatched before the API call, this one is
+ * dispatched once the API has answered. Consumers deriving something from the domain list (the console's "open"
```
Galimede: nitpick: we should avoid mentiong console there too.

Galimede (approved):
LGTM! 💪

PR #1799, opened 2026-08-07
Cc domain management/fix domain list order

### LINKED ISSUE #1798: cc-domain-management: display domains as a sorted single-column list so paths are readable on apps with many domains
## Context

In `cc-domain-management`, the domain list is rendered as an auto-fit grid:

```css
.domains {
  display: grid;
  gap: var(--cc-spacing-5, 1em);
  grid-template-columns: repeat(auto-fit, minmax(min(25em, 100%), 1fr));
}
```

(`src/components/cc-domain-management/cc-domain-management.js`, `.domains` rule)

Domains are already sorted with `sortDomains()` (reversed hostname + path prefix), which is good. But on a wide screen the grid lays out 2+ columns and flows **left → right, then wrap**, so that sort order is broken visually: consecutive entries of the same hostname end up on opposite sides of the screen.

Real example, an app with 13 domains (order as displayed):

| left column | right column |
| --- | --- |
| `staging-base-api.cleverapps.io/` (primary) | `api.clever-cloud.com/v2/providers/addon-ts` |
| `api.clever-cloud.com/v2/providers/kv` | `api.clever-cloud.com/v4/addon-providers/addon-ts` |
| `api.clever-cloud.com/v4/addon-providers/kv` | `api.clever-cloud.com/v4/iam` |
| `api.clever-cloud.com/v4/materia` | `api.clever-cloud.com/v4/monitoring` |
| `api.clever-cloud.com/v4/ts` | `api.clever-cloud.com/v4/warp10` |
| `buildcache.services.clever-cloud.com/` | `app-f92cdc74-….cleverapps.io/` |
| `axo.cleverapps.io/` | |

Reading which paths are exposed on `api.clever-cloud.com` means zig-zagging between columns, and the path prefixes never line up vertically, so you can't scan them. The more domains an app has, the worse it gets — and path-routed apps (API gateways, backends split by path) are exactly the ones with many.

## Proposal

Render the domains as a **single-column list** so the existing sort order reads straight down and hostnames/paths align vertically. Possible variations:

- always one column (simplest, and the cards are already full-width-ish content);
- or keep the auto-fit grid only below a domain-count threshold, and switch to one column above it;
- if multi-column must be kept for short lists, use column-major flow (`grid-auto-flow: column` with an explicit row count, or CSS `columns`) so each column reads top → bottom in sort order.

Nice-to-have on top of that, for readability on path-heavy apps:

- group consecutive domains sharing the same hostname (or at least de-emphasize the repeated hostname) so the path prefixes stand out;
- align path prefixes so they can be compared at a glance.

## Why

The domain list is the main way to answer "what paths does this app serve?". Today that answer requires reading the whole grid in a zig-zag; a sorted single-column list makes it a straight scan.

clara-layus-cc:
Fixes #1798

# What does this PR do?

- Fill the domain list grid column by column so the sort order reads as a straight top to bottom scan instead of zig-zagging between columns,
- Compute the row and column counts when rendering the list, `grid-auto-flow: column` requiring an explicit row count that CSS cannot derive,
- Keep the single column fallback on narrow widths through a container query, independently of the viewport,
- De-emphasize a hostname identical to the one displayed right above it, so the path prefixes stand out on applications routing by path.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/cc-domain-management/fix-domain-list-order/index.html?path=/story/%F0%9F%9B%A0-domains-cc-domain-management--default-story)

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -972,8 +972,10 @@ export class CcDomainManagement extends LitElement {
            prefixes are what stands out when scanning a column.
            --cc-color-text-weak is only one step below --cc-color-text-default, which reads as the
            same color, hence the raw palette color here. Contrast stays above AA on the card
-           background. */
-        .repeated-hostname {
+           background.
+           A waiting domain is left out: its whole name is already dimmed by an opacity, which no
+           de-emphasized color could go through while keeping an AA contrast. */
+        .domain:not(.waiting) .repeated-hostname {
           color: var(--color-grey-60, #737373);
         }
 
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -11,6 +11,7 @@ import { LitElement, css, html } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { createRef, ref } from 'lit/directives/ref.js';
 import { repeat } from 'lit/directives/repeat.js';
+import { styleMap } from 'lit/directives/style-map.js';
 import {
   iconRemixDeleteBinLine as iconDelete,
   iconRemixLockUnlockFill as iconHttpOnly,
@@ -546,6 +547,10 @@ export class CcDomainManagement extends LitElement {
     const hasDomains = domains.length > 0;
     const isOneRowMarkingPrimary = domains.filter((domain) => domain.type === 'marking-primary').length > 0;
     const hasHttpOnlyDomains = domains.filter((domain) => domain.isHttpOnly).length > 0;
+    // The grid is filled column by column, which requires an explicit row count.
+    // A single domain stays on one full width column.
+    const columnCount = domains.length > 1 ? 2 : 1;
+    const rowCount = Math.ceil(domains.length / columnCount);
 
     if (!hasDomains) {
       return html`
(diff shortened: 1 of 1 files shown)


src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -546,6 +546,12 @@ export class CcDomainManagement extends LitElement {
     const hasDomains = domains.length > 0;
     const isOneRowMarkingPrimary = domains.filter((domain) => domain.type === 'marking-primary').length > 0;
     const hasHttpOnlyDomains = domains.filter((domain) => domain.isHttpOnly).length > 0;
+    // The grid is filled column by column, which requires an explicit row count.
+    // A single domain stays on one full width column.
+    // The grid is filled column by column, which requires an explicit row count.
```
HeleneAmouzou: **fix:** duplicated comment

src/components/cc-domain-management/cc-domain-management.js
```diff
@@ -573,11 +611,11 @@ export class CcDomainManagement extends LitElement {
           `
         : ''}
 
-      <div class="domains">
+      <div class="domains" style="--domain-columns: ${columnCount}; --domain-rows: ${rowCount};">
```
HeleneAmouzou: **suggestion**: maybe we can use `styleMap` here

HeleneAmouzou (changes requested):
Just small things to fix

HeleneAmouzou (approved):
LGTM !

PR #1802, opened 2026-08-21
fix(cc-search-bar): virtually focus the first result so Enter opens it

clara-layus-cc:
# What does this PR do?

- Virtually focus the first filtered item instead of the input, so `Enter` opens a result right away — pasting a resource id and hitting `Enter` now jumps to its single result,
- Reset that virtual focus from `willUpdate()`, right after the filtered list is rebuilt, so it follows `value` and `sections` instead of being cleared on every keystroke by the input handler,
- Reset it on dialog close too, so reopening with the same query starts over from the first item rather than from wherever the arrow keys left the selection,
- Drop the stale-index guard in `updated()`, now that the index is derived from the filtered items every time they change.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/fix/search-bar-select-first-result/index.html?path=/story/%F0%9F%9B%A0-navigation-cc-search-bar--with-value)

### DIFF
--- a/src/components/cc-search-bar/cc-search-bar.js
+++ b/src/components/cc-search-bar/cc-search-bar.js
@@ -91,7 +91,8 @@ export class CcSearchBar extends LitElement {
 
     /**
      * @type {number} Index of the virtually focused item within the flattened list of filtered items.
-     * `-1` means no item is virtually focused: the real focus stays on the input.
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
LGTM! Thanks for handling this @clara-layus-cc 😎

PR #1804, opened 2026-08-21
feat(cc-logs-instances): select every instance of a deployment at once

KannarFr:
## Why

In cold mode the instances panel groups instances by deployment, but filtering the logs on a whole deployment meant ticking each of its instances one by one.

## What

The deployment legend now carries a checkbox that selects, or deselects, every instance listed in its group. It dispatches the existing `cc-logs-instances-selection-change` event with the full list, so nothing changes for `cc-logs-app-runtime` or its smart component.

Design choices, all borrowed from what the panel already does:

- **A checkbox, not a clickable row.** The panel's whole vocabulary for "this filters the logs" is a checkbox. The deployment one sits at the group's left edge, and the instance checkboxes were already indented `1.5em` under it — so the hierarchy draws itself, no new indentation or borders.
- **`<label>` wrapping checkbox + state icon + "Deployed X ago"**, mirroring `_renderInstance`'s label around checkbox + name, so the text is clickable too. The commit chip stays outside the label, the same way the instance index chip does.
- **Tri-state**: checked when the whole group is selected, indeterminate when partial. Clicking a partially selected deployment completes the selection rather than clearing it.
- **CSS**: the legend grid went `auto 1fr auto` → `1fr auto`, with the label carrying the inner rhythm at the same `--cc-spacing-3` gap as `.instance`. One new rule, no new tokens, no new colors.

Since `_renderInstancesGroupedByDeployment` also backs live mode's **Deleted instances** section, that section gets the toggle too. That is why the label reads "all the instances *listed for* this deployment" — there the group holds only the deployment's deleted instances.

## Notes

- The groups are now rendered through `repeat()` keyed on the deployment id. Lit only re-commits a bound property when its cached value changes, while a click mutates `checked`/`indeterminate` directly, so a group reorder concurrent with a click could otherwise strand a checkbox showing state that no longer matches `state.selection` — `indeterminate` being property-only, with no attribute to fall back on.
- The purpose is exposed to assistive tech through a `.visually-hidden` span inside the label, using the existing `accessibilityStyles` helper. A `title` on the label would not have worked: the label's subtree text is what names the associated input.
- New `coldModeWithPartialDeploymentSelection` story covers the indeterminate and fully-checked states side by side.

## Checks

`lint`, `stylelint`, `format:check`, `lit-analyzer`, `components:check-i18n` and `components:events-map-check` are clean; `test:group a11y:cc-logs-instances` passes 34/34.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/cc-logs-instances/cc-logs-instances.js
+++ b/src/components/cc-logs-instances/cc-logs-instances.js
@@ -1,4 +1,5 @@
 import { css, html, LitElement } from 'lit';
+import { repeat } from 'lit/directives/repeat.js';
 import {
   iconRemixToolsLine as iconBuildInstance,
   iconRemixCloseCircleFill as iconDeploymentCancelled,
@@ -16,6 +17,7 @@ import {
   iconRemixStopCircleLine as iconInstanceStopping,
 } from '../../assets/cc-remix.icons.js';
 import { groupBy } from '../../lib/utils.js';
+import { accessibilityStyles } from '../../styles/accessibility.js';
 import { i18n } from '../../translations/translation.js';
 import '../cc-datetime-relative/cc-datetime-relative.js';
 import '../cc-icon/cc-icon.js';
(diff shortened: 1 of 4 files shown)


hsablonniere:
Thanks @KannarFr! As discussed, closing in favour of #1836, which describes the problem and references your proposal.

PR #1809, opened 2026-09-01
feat(cc-orga-member-list): let consumers deny member management

### LINKED ISSUE #1807: cc-orga-member-list: allow a consumer to deny member management
## Context

We are deploying partner-specific Consoles where an organisation's membership is managed
exclusively through an external portal, never through the Console. The API now exposes, per
organisation, a flag saying that members must not be modified from the Console.

The Console handles the "members can't even be listed" case on its own, by hiding the whole Members
tab. The case that reaches this component is the other one: **the member list stays visible and
readable, but every write must go away** — no invite, no role change, no member removal, and no
leaving the organisation.

## What exists today

`CcOrgaMemberList` already has almost everything for this:

- `authorisations: { invite, edit, delete }`, all `false` by default via `INIT_AUTHORISATIONS`
  (`cc-orga-member-list.js:51-70`). It gates the invite form (line 345) and is forwarded to each
  card (lines 460-462).
- `cc-orga-member-card` renders its action buttons only when
  `hasAdminRights = authorisations.edit && authorisations.delete`
  (`cc-orga-member-card.js:259`, used line 280). The delete button doubles as the "Leave" button on
  the current user's card (`cc-orga-member-card.js:187`), so that leave path is already covered by
  the authorisations.

Two things are missing.

**1. The danger zone isn't gated.** `_renderDangerZone` (`cc-orga-member-list.js:485`) is rendered
as soon as the list is loaded (line 372), with no reference to `authorisations`. So with all
authorisations `false` the list is read-only *except* for a fully working "Leave the organisation"
button at the bottom.

**2. The smart component owns the authorisations.** `updateAuthorisations(role)` derives them from
the current user's role alone — `hasAdminRights = role === 'ADMIN' || role === 'MANAGER'`
(`cc-orga-member-list.smart.js:61-68`), called once the member list is loaded (line 303). The
declared params are only `apiConfig` and `ownerId` (lines 33-36), so a consumer has no way to
restrict what the role grants.

## Proposal

1. **Add a `leave` authorisation.** Extend `ListAuthorisations`
   (`cc-orga-member-list.types.d.ts`) with an optional `leave?: boolean`, and gate
   `_renderDangerZone` on it. `CcOrgaMemberCard` honours it too: on the current user's card, where
   the delete button doubles as the "Leave" button, the action buttons read `leave` instead of
   `delete`.
2. **Add an optional smart param** `readonly: { type: Boolean, optional: true }` (optional params
   are already used elsewhere, e.g. `cc-network-group-list.smart.js:40`). When it is `true`,
   `updateAuthorisations` denies everything whatever the role, `leave` included.
3. Update `cc-orga-member-list.smart.md` (params table) and the stories.

pdesoyres-cc:
## Context

Closes #1807.

Partner-specific Consoles manage an organisation's membership through an external portal only, and
the API now exposes a per-organisation flag saying members must not be modified from the Console.
The Console handles the "members can't even be listed" case on its own by hiding the whole Members
tab; the case that reaches this component is the other one — the member list stays visible and
readable, but every write must go away: no invite, no role change, no member removal and no leaving
the organisation.

Invite, edit and delete were already gated by `authorisations`. Leaving was not: `_renderDangerZone`
ran as soon as the list was loaded, so an otherwise read-only list kept a fully working "Leave the
organisation" button.

## Changes

- New optional `leave` authorisation on `cc-orga-member-list`, gating the danger zone.
- `cc-orga-member-card` honours `leave` on the current user's card, where the delete button doubles
  as the "Leave the organisation" button.
- New optional `readonly` param on the smart component: when `true`, every authorisation is denied
  whatever the current user's role.
- Stories for both denied cases, and the smart doc params table.

### Implementation notes

**Naming.** A single boolean named `readonly`, rather than the `memberManagementDenied` floated in
the issue. It says what the consumer gets — a member list that is readable and nothing else —
without leaking the API's flag vocabulary into the component. If those per-organisation flags grow
into something richer, that is a second, additive param rather than a rename of this one.

**`leave` on the current user's card.** The card honours it, as the issue suggested. It falls back
to the `delete` authorisation when unset, which is exactly what used to gate that button.

**`leave` defaults to `true` instead of joining the all-`false` `INIT_AUTHORISATIONS`.** Leaving is
the one action a plain member has always had, independently of any admin right, so it is the sane
default for the dumb component. The smart component still computes it: `leave: role != null`, since
`updateAuthorisations()` is called with no role once the current user has left the organisation.

**Backward compatibility.** That default is what keeps this change invisible to existing consumers:
`leave` is optional, defaults to `true` on the list and falls back to `delete` on the card, so a
consumer rendering `<cc-orga-member-list>` without setting `authorisations` keeps its danger zone.
The only consumer-visible addition is the `readonly` smart param, which defaults to `false`.

## How to review

1. Start with `cc-orga-member-list.smart.js` — `updateAuthorisations()` is the load-bearing change,
   and shows what `readonly` maps to.
2. Then the two gates it feeds: the danger zone in `cc-orga-member-list.js`, and `mayRemove` in
   `cc-orga-member-card.js`.
3. On the preview, check the two new stories: `dataLoadedWithMemberManagementDenied` on the list
   (readable list, no invite form, no action buttons, no danger zone) and
   `dataLoadedWithIsCurrentUserAndLeaveDenied` on the card (admin, current user, no action buttons).
4. Worth a second look given the default above: the untouched stories should be unchanged, danger
   zone included.

**1 approval**, per the branch protection rule on `master`.

### DIFF
--- a/demo-smart/index.html
+++ b/demo-smart/index.html
@@ -1024,6 +1024,14 @@
                 <button
                   class="ctx-btn"
                   title="ownerId, currentUserId"
+                  data-context='{"ownerId":"orga_3547a882-d464-4c34-8168-add4b3e0c135","currentUserId":"user_d3d02829-8847-40ea-ab76-ada4e2f026c0", "readonly": "true"}'
+                >
+                  <span class="ctx-label">ACME BAR readonly</span>
+                  <span class="ctx-subtitle">ownerId, currentUserId</span>
+                </button>
+                <button
+                  class="ctx-btn"
+                  title="ownerId, currentUserId"
                   data-context='{"ownerId":"orga_540caeb6-521c-4a19-a955-efe6da35d142","currentUserId":"user_d3d02829-8847-40ea-ab76-ada4e2f026c0"}'
                 >
                   <span class="ctx-label">ACME FOO</span>
--- a/src/components/cc-orga-member-card/cc-orga-member-card.js
+++ b/src/components/cc-orga-member-card/cc-orga-member-card.js
@@ -45,6 +45,7 @@ const BREAKPOINTS = [BREAKPOINT_TINY, BREAKPOINT_SMALL, BREAKPOINT_MEDIUM];
  * With the right authorisations:
  * - This component provides a way to delete the member from the organisation.
  * - This component provides a way to edit the role of the member within a given organisation.
+ * - On the current user's card, the delete button becomes a "leave the organisation" button, gated by the `leave` authorisation.
  *
  * ## Technical Details
  *
@@ -69,6 +70,7 @@ export class CcOrgaMemberCard extends LitElement {
     this.authorisations = {
       edit: false,
       delete: false,
+      leave: true,
     };
 
     /** @type {OrgaMemberCardState} Sets the state and data of the member. */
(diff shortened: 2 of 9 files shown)


florian-sanders-cc (approved):
LGTM! Tested in demo smart as well, a small nitpick: could add `readonly: true` to the context of some new button (like ACME BAR readonly) within the demo-smart so that we don't can test this usecase easily.

PR #1811, opened 2026-09-02
fix(cc-zone-input): separate selection visual from focus outline

### LINKED ISSUE #1810: cc-zone-input: double focus ring is confusing
## Context

The `cc-zone-input` component has a double focus ring. 
First focus ring:
- around the `.zone-list` (the scrollable area), to mark that you're focusing an item within this list. 
- it is only displayed when not hovering the list so it "appears" when your cursor moves out of the list which may be confusing for mouse users.

Second focus ring:
- around the selected item (radio button label),
- it looks like a focus ring but kind of acts as a selection marker (always visible, even when not focusing the item).

The current design mixes two separate concerns:
- focus ring,
- selected item.

## Possible solutions 

A. Design refactor
- Design an actual "selected" state that doesn't rely on the focus ring, 
- Use the focus ring as a focus marker.

This is the cleanest solution.
Note that this is kind of what we use in `cc-toggle` and `cc-search-bar` (background specific to selected state and focus outline for focus).

B. Code only refactor
- Use `:has(:focus-visible):focus-within)` so that the focus outline on the `.zone-list` only appears when using keyboard,

This solution is not the best because we still have a double focus (nested), but at least only for keyboard users which is a bit less confusing.

florian-sanders-cc:
Fixes #1810

## What does this PR do?

- Changes the style of selected items in `cc-zone-input`: we now use a darker grey background instead of the focus ring to highlight selected items,
- Focus ring is now only used to highlight focused items (list itself or items within the list).

## How to review?

- Check the `cc-zone-input` stories,
- Check the commit,

### DIFF
--- a/src/components/cc-zone-input/cc-zone-input.js
+++ b/src/components/cc-zone-input/cc-zone-input.js
@@ -309,6 +309,8 @@ export class CcZoneInput extends LitElement {
       css`
         :host {
           display: block;
+
+          --label-border-width: 2px;
         }
 
         .wrapper {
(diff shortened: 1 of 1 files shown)


Galimede (approved):
LGTM, thanks Florian! 💪

roberttran-cc (approved):

PR #1814, opened 2026-09-03
revert(cc-domain-management): drop the de-emphasis of repeated hostnames

### LINKED ISSUE #1798: cc-domain-management: display domains as a sorted single-column list so paths are readable on apps with many domains

clara-layus-cc:
Fixes [console3#1401](https://gitlab.corp.clever-cloud.com/clever-cloud/console3/-/work_items/1401)

# What does this PR do?

- Revert ba4d6839 and 668f9e4d, which closed #1798: a hostname identical to the one displayed right above it was dimmed, so that the path prefixes stood out on applications routing by path,
- In practice the two shades read as an unexplained difference between entries — nothing tells which state each color stands for, so the list looks inconsistent rather than structured, which is what the console3 issue reports,
- All hostnames are back to the default text color, no distinction left: `getRepeatedHostnameFlags`, the `isHostnameRepeated` parameter, the `repeated-hostname` class and its rule are all gone,
- `rowCount` stays, it is still needed for the column layout.

# How to review?

- Check that the diff is the exact inverse of ba4d6839 + 668f9e4d,
- Open the `cc-domain-management` stories on an app exposing several paths on the same hostname and check every hostname now has the same color,

### DIFF
--- a/src/components/cc-domain-management/cc-domain-management.js
+++ b/src/components/cc-domain-management/cc-domain-management.js
@@ -55,37 +55,6 @@ const DNS_DOCUMENTATION = getDocUrl('/administrate/domain-names');
  */
 
 /**
- * Tells, for each domain, whether its hostname repeats the one displayed right above it.
- * Domains are sorted by hostname, so repeats are always consecutive and de-emphasizing them makes
- * the path prefixes stand out on apps exposing many paths on the same hostname.
- *
- * The hostname is compared with its wildcard prefix included, because `*.example.com` and
- * `example.com` are two different hostnames from the user point of view even though they share the
- * same `hostname` value.
- *
- * The first domain of a column is never flagged: its hostname sits at the top of a column, so the
- * previous domain is not displayed above it. `rowCount` marks a column break only when the list is
- * actually laid out on two columns. On a single column it merely skips one de-emphasis, which is
- * why it errs on that side: a de-emphasized hostname always has its full-color twin right above it.
- *
- * @param {FormattedDomainInfo[]} domains - sorted domains, as displayed
- * @param {number} rowCount - number of rows of the domain list layout
- * @returns {boolean[]} one flag per domain, in the same order
- */
-function getRepeatedHostnameFlags(domains, rowCount) {
-  return domains.map((domain, index) => {
-    if (index === 0 || index === rowCount) {
-      return false;
-    }
-    const previousDomain = domains[index - 1];
-    return (
-      getHostWithWildcard(domain.hostname, domain.isWildcard) ===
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
All good! 👍

PR #1817, opened 2026-09-09
refactor: migrate everything off the legacy @clevercloud/client esm api

### LINKED ISSUE #1816: half the codebase still uses the legacy @clevercloud/client esm api
## Context

`clever-components` imports two different API clients, both from the `@clevercloud/client` package ([clever-client.js](https://github.com/CleverCloud/clever-client.js)).

- The new client is the command client, under `cc-api-commands/`, `cc-api-bridge-commands/`, `redis-http-commands/` and `utils/`.
- The legacy client is the service and function API, under `esm/`.

Counts on `master`: 89 imports of the new client across 33 files, 91 imports of the legacy client across 53 files.

## Problem

`clever-client.js` wants to release the command client as GA and deprecate the legacy `esm/` API. `clever-components` is one of the largest consumer of that legacy API. As long as it imports `esm/`, the legacy API has to stay supported, so the deprecation cannot start.

The mixed state also costs us inside this repo. Two clients mean two error models, two authentication paths and two caching strategies. `src/lib/send-to-api.js` hand-rolls URL prefixing, OAuth signing, caching and timeouts, which the new client already does on its own.

### Scope

| Area | What still uses the legacy client |
| --- | --- |
| Smart components | 37 files under `src/components/*/*.smart*.js` |
| Plumbing | `src/lib/send-to-api.js`, `src/lib/send-to-auth-bridge.js`, `src/lib/api-helpers.js` |
| Logs | `src/lib/logs/logs-stream.js`, `src/lib/logs/logs-stream.types.d.ts` |
| Tests | `test/logs/logs-stream.test.js` |
| Docs | `docs/contributing/smart-component.md` |

Most used legacy entry points: `esm/with-cache.js` (19 imports), `esm/api/v2/application.js` (9), `esm/api/v2/addon.js` (9), `esm/api/v2/organisation.js` (6), `esm/utils/env-vars.js` (5), `esm/api/v2/user.js` (5).

### Impact

The GA release of the new client is blocked. Contributors have to know which client a given smart component uses before touching it, and the contributing guide documents the legacy one.

## Solution considered

Migrate every remaining `esm/` import to the command client, then delete the legacy plumbing in `src/lib` once nothing calls it.

Work is in progress on the `refactor/new-client` branch.

## Blocked by

CleverCloud/clever-client.js#231, "Close the gaps blocking the move off the legacy client". Some commands are still missing on the `clever-client.js` side. They have to land before the last `esm/` imports can go.

pdesoyres-cc:
Closes #1816

## Context

`clever-components` imported two API clients from `@clevercloud/client`. The command client lives under `cc-api-commands/`, `cc-api-bridge-commands/`, `redis-http-commands/` and `utils/`. The legacy service and function API lives under `esm/`.

53 files still imported the legacy one.

## Problem

`clever-client.js` cannot deprecate its `esm/` API while `clever-components` remains one of its largest consumers. That blocks the GA release of the command client.

The mixed state also costs us here. Two clients mean two error models, two authentication paths and two caching strategies. `src/lib/send-to-api.js` hand-rolled URL prefixing, OAuth signing, caching and timeouts, all of which the command client already does.

## Proposal

Every remaining `esm/` import moves to the command client. The legacy plumbing is then deleted, since nothing calls it any more.

The branch now leaves zero `esm/` imports in `src/`, `test/` and `docs/`.

What lands, by area:

- **Smart components.** The 37 smart components that used `sendToApi` now build a client with `getCcApiClient()` or `getCcApiClientWithOAuth()` and send commands.
- **Plumbing.** `send-to-api.js` and `send-to-auth-bridge.js` are removed. `api-helpers.js` sends invoice and price system commands. `cc-api-client.js` gains `getCcApiBridgeClient()`, so the API token components reach the auth bridge through a real client instead of hand-built OAuth v1 headers.
- **Types.** `common.types.d.ts` dropped its hand-written copies of API payloads, several of which carried a `FIXME` asking for exactly that. `src/operators.types.d.ts` is deleted. Consumers import those shapes from the client package.
- **Fixtures.** Story fixtures were snake_case copies of raw API payloads. They now match what the commands return.
- **Dead code.** `xml-parser.js` and its test are gone, along with the two `.client.js` files that wrapped Cellar and Kubernetes calls.

### Design decisions

`fetchApp()` disappeared from `api-helpers.js` rather than being ported. Its two callers send `GetApplicationCommand` directly, which is one less indirection for one call.

`isAbortError()` in `src/lib/abortable.js` accepts two error shapes. The platform rejects with a `DOMException` named `AbortError`, and the client rethrows a `CcRequestError` carrying the `ABORTED` code. Both mean the same thing, so both are treated as an abort.

The `install-local:*` scripts used a glob to move a packed tarball. The glob matched the version number, so a leftover tarball could win. Both scripts now pack straight to a fixed file name with `pnpm pack --out`.

## How to review

Read the history, not the global diff. The 56 commits are ordered and each one carries a body explaining its own change. The global diff is mostly fixture churn.

Two commits deserve real attention:

- `98947ebf refactor(lib): adapt the logs stream to the new client` is the only place where behaviour changed rather than the call shape. `_createStream()` became asynchronous, which broke stream cancellation, `pause()` and the progression start. The commit body walks through the three fixes, and `test/logs/logs-stream.test.js` covers them.
- `7d4ef4a7 feat(lib): add a cached auth bridge client` introduces `getCcApiBridgeClient()` and moves the request config into `createClient()`.

The rest of the commits repeat the same pattern, one component at a time. Reading two or three of them is enough to know what the others do.

## How to test

The migration is meant to be invisible from the outside. Every smart component should behave as it did before.

1. Run `pnpm run test` and `pnpm run typecheck`.
2. Start Storybook with `pnpm run storybook:dev` and exercise a few smart components against a real API token.
3. The log components are worth a closer look, since their streaming logic changed. Open a log view, let it stream, then pause and resume it.


src/lib/api-helpers.js
```diff
@@ -1,30 +1,19 @@
...
- * @param {ApiConfig} options.apiConfig
- * @param {AbortSignal} options.signal
- * @param {string} options.ownerId
- * @param {string} options.appId
- * @return {Promise<{name: string}>}
- */
-export function fetchApp({ apiConfig, signal, ownerId, appId }) {
-  return getApp({ id: ownerId, appId }).then(sendToApi({ apiConfig, signal }));
-}
+const ONE_DAY = 1000 * 60 * 60 * 24;
```
florian-sanders-cc, line 16: **nitpick:** we still use `ONE_DAY` in quite a few cases, can't we export it from here rather than redefining it everywhere?

src/lib/api-helpers.js
```diff
@@ -60,53 +50,73 @@ export async function fetchInvoiceHtml({ apiConfig, signal, ownerId, invoiceNumb
...
  */
-async function formatInvoice(apiConfig, ownerId, rawInvoice) {
+function formatInvoice(apiConfig, ownerId, rawInvoice) {
   return {
-    number: rawInvoice.invoice_number,
-    emissionDate: rawInvoice.emission_date,
-    type: rawInvoice.type || 'INVOICE',
+    number: rawInvoice.invoiceNumber,
+    emissionDate: rawInvoice.emittedAt,
+    type: /** @type {Invoice['type']} */ (rawInvoice.kind) || 'INVOICE',
```
florian-sanders-cc, line 99: **question:** why do we need the cast + the fallback to `'INVOICE'`? (the type seems to point to non optional + it already fits `Invoice['type']`) 🤔

src/components/cc-addon-backups/cc-addon-backups.smart.js
```diff
@@ -23,148 +34,115 @@ defineSmartComponent({
...
-    default:
-      return null;
+    // The "es-addon" / "es-addon-old" distinction (and its password) can only be resolved through
+    // the Elasticsearch specific info endpoint (it depends on whether the Kibana service is enabled).
+    if (addon.provider.id === 'es-addon') {
+      const esInfo = await this.fetchElasticsearchInfo({ addonId });
+      const kibana = esInfo.services.find((service) => service.name === 'kibana');
+      providerId = kibana != null && kibana.isEnabled ? 'es-addon' : 'es-addon-old';
+      passwordForCommand = esInfo.config.password;
+    }
```
florian-sanders-cc, line 126: **question:** should this logic move to the client later on? I imagine we have similar needs in the CLI?

src/components/cc-cellar-bucket-list/cc-cellar-bucket-list.ctrl.js
```diff
@@ -149,11 +152,11 @@ export class BucketsListController {
         this.#getComponent().dispatchEvent(new CcCellarBucketCreatedEvent(bucketName));
       }
     } catch (error) {
-      if (isCellarExplorerErrorWithCode(error, 'clever.file-explorer-proxy.cellar.bucket-already-exists')) {
+      if (isCcHttpErrorWithCode(error, CREATE_CELLAR_BUCKET_ERROR_CODES.BUCKET_ALREADY_EXISTS)) {
         this.#updateCreateForm({ type: 'idle', error: 'bucket-already-exists' });
-      } else if (isCellarExplorerErrorWithCode(error, 'clever.file-explorer-proxy.cellar.bucket-name-invalid')) {
```
florian-sanders-cc, line 154: **question:** I tried to create a bucket with a name that already exists, didn't trigger the corresponding error message (got the typical error notice instead).
Same happens in prod so it's not a new subject, this is the error payload I got:
```
{
  "apiRequestId": "request_7db9792d-1547-47a3-a413-240f564c7de4",
  "code": "clever.cellar.bucket-already-exists",
  "context": { "kind": "bucket", "name": "bob-empty-bucket-1", "type": "Resource" },
  "error": "Bucket 'bob-empty-bucket-1' already exists"
}
```

src/components/cc-email-list/cc-email-list.smart.js
```diff
@@ -244,17 +245,17 @@ defineSmartComponent({
...
-function convertApiError(apiErrorId) {
-  if (apiErrorId === 550) {
+function convertApiError(apiErrorCode) {
+  if (apiErrorCode === CREATE_PROFILE_EMAIL_ADDRESS_ERROR_CODES.INVALID_FORMAT) {
     return 'invalid';
   }
-  if (apiErrorId === 101) {
+  if (apiErrorCode === CREATE_PROFILE_EMAIL_ADDRESS_ERROR_CODES.ALREADY_DEFINED) {
     return 'already-defined';
   }
```
florian-sanders-cc, line 257: **question:** just like with cellar, this doesn't seem to work, I get a generic toast instead of the specific error message
API Response:
```
{ "id": 4001, "message": "This email address already belongs to you", "type": "error" }
```

src/components/cc-network-group-list/cc-network-group-list.smart.js
```diff
@@ -201,3 +210,26 @@ defineSmartComponent({
...
+/**
+ * The client is older than the API it talks to, so it publishes a peer kind or an endpoint kind it does not know as
+ * an unknown variant carrying only the raw payload. A peer card needs an id, a public key and an address, none of
+ * which such a variant exposes, so those peers are left out of the list.
+ *
+ * @param {NetworkGroupPeer} peer
+ * @returns {peer is KnownNetworkGroupPeer}
+ */
+function isKnownNetworkGroupPeer(peer) {
+  return isKnown(peer) && isKnown(peer.endpoint);
```
florian-sanders-cc, line 223: **question:** I guess I missed this change in the client 🙈 Could you tell me more about why we have to filter out some peers? (mostly for my information)

src/components/cc-orga-member-list/cc-orga-member-list.smart.js
```diff
@@ -328,29 +337,21 @@ defineSmartComponent({
...
-    );
+    ccApiClient.send(new GetProfileCommand(), { signal }),
+    ccApiClient.send(new ListOrganisationMemberCommand({ organisationId: ownerId }), { signal }),
+  ]).then(([profile, memberList]) => {
+    return memberList.map((member) => ({
+      id: member.id,
+      avatar: member.avatar,
+      name: member.name,
+      jobTitle: member.jobTitle,
+      role: /** @type {OrgaMemberRole} */ (member.role),
```
florian-sanders-cc, line 350: **nitpick:** this cast could cause issues if roles evolve at some point (like they did recently).

We're not supposed to get role: 'NONE' here, but casting basically turns TS off for this line: if a role gets added to the client, it will go through the cast silently and the card will receive a role it doesn't handle.

It could be more robust to narrow instead, like skip NONE members in a flatMap (or throw if we'd rather know about it). This way, TS would flag any mismatch between the client roles and `OrgaMemberRole`.

src/components/cc-token-api-creation-form/cc-token-api-creation-form.smart.js
```diff
@@ -65,10 +66,7 @@ defineSmartComponent({
         .catch(
           /** @param {Error} error */
           (error) => {
-            const errorCode =
-              'responseBody' in error && typeof error.responseBody === 'object' && 'code' in error.responseBody
-                ? error.responseBody.code
-                : null;
+            const errorCode = isCcHttpError(error) ? error.code : null;
```
florian-sanders-cc, line 69: **fix:** in prod, if I enter a wrong password, I get the error message below the input field and I can try again.
With this branch, I see the error below the input but I get disconnected just after that. This makes me think the error is not swallowed properly? (didn't dig further)

florian-sanders-cc (changes requested):
Gg @pdesoyres-cc I really like the new client so I reallyyyyy like this PR 🙌 

I only have one small issue with the way errors are caught in `cc-token-api-creation-form` but the rest are just nitpicks!

PR #1819, opened 2026-09-10
fix(cc-logs): keep the viewport still while logs stream in

### LINKED ISSUE #1818: logs view: scroll position moves when new logs arrive
## Context

User report from the Clever Cloud console logs view (`cc-logs`), on clever-components 26.3.0.

## Problem

A user scrolls somewhere in the logs view to read older logs. New logs keep arriving in the stream. The scroll position moves instead of staying put. The expected behavior is a fixed viewport: as long as the user does not scroll, the logs on screen must not move.

### How to reproduce

1. Open a logs view fed by a live stream of logs.
2. Scroll up, away from the bottom, so the view stops following the tail.
3. Wait for new logs to arrive.
4. The displayed logs shift.

### Impact

Reading the log history while the stream is active is hard. The user loses the line being read every time a batch of logs arrives.

pdesoyres-cc:
Closes #1818

## Context

`cc-logs` renders its list with a TanStack virtualizer. The virtualizer keys its caches and its scroll anchor by index, not by log id. Once `limit` is reached, every append trims logs off the front of the list, so every surviving log changes index.

## Problem

A user who scrolled up to read history does not stay there. Two independent bugs move the view.

1. **The scroll offset ignores the front-trim.** The list slides up by the height of the trimmed logs, but the scroll offset stays where it was. The view is carried forward on every append.

2. **The size estimate for unrendered logs is wrong.** It was hardcoded to the line height and ignored the row padding, so every unrendered row was assumed shorter than it really is. The view slides by about two lines on every scroll, with no log appended at all.

## Proposal

The scroll offset now follows the trim. The height of the trimmed logs is measured before the trim is applied, then subtracted from the scroll offset. The correction is skipped while following, because following already pins the view to the bottom.

The size estimate is now read from the CSS of a rendered row. A hardcoded value could not be right anyway, because the row height is expressed in `em` and follows the host's font size. The old constant stays as a fallback until the first row is in the DOM.

## How to test

Automated tests live in `src/components/cc-logs/cc-logs.scroll-anchor.test.js`. They scroll the component, append logs, and assert that the log under the viewport top does not change.

Manually, open the sandbox at http://localhost:6006/sandbox/?component=cc-logs-app-runtime. Scroll up while the stream runs. The logs on screen must not move.


src/components/cc-logs/cc-logs.js
```diff
@@ -446,6 +464,47 @@ export class CcLogs extends LitElement {
...
+   * difference. A hardcoded estimate cannot be right either, because the row height is expressed in `em` and
+   * therefore follows the host's font size.
+   *
+   * A row is a flex line whose height is driven by its text: the line height of `.log--right` plus its vertical
+   * padding (the select button in the gutter is built to match). Reading those from the computed style gives the
+   * exact single-line height, in both wrap modes, without having to guess which rendered row happens to be one line
+   * tall.
+   */
+  _updateEstimatedLineHeight() {
+    const row = this._logsRef.value?.querySelector('.log .log--right');
```
roberttran-cc, line 482: **🔨 issue:** From Claude but non-blocking I think.

> The gutter is not built to match below ~15px host font size.
> `.select_button` uses `height: var(--gutter-size)`, and that `1.5em` resolves against the button's UA font size (13.333px), so the gutter is always 24px.
> With a smaller host font, the gutter sets the row height and the estimate is too small.
> I measured with `style="font-size: 14px"`: estimate 22, real row 24, and the view drifts 22px after a scroll (84px at 11px).
> Maybe take the max with the button's outer height, or set `font-size: inherit` on the button? A test at a small font size would catch it.

roberttran-cc (approved):
Tested locally and LGTM!

PR #1820, opened 2026-09-10
feat(cc-search-bar): match items on an optional aliases field

pdesoyres-cc:
## Problem

`cc-search-bar` filters its items on `label` and `id` only, so an item exposes a single searchable identifier. Add-ons need to be searchable by their `realId` (`postgresql_xxx`), while the `id` already holds the `addon_xxx` the console routes use.

## Proposal

Items accept an optional `aliases: string[]` field. Those texts are searched like the label and the id, but they are never displayed.

The search grammar does not change. The `is:` and `project:` tokens still match against matchers, and free text now matches `label`, `id` and `aliases`.

No breaking change. The field is optional, and an item without `aliases` behaves as before.

## How to test

1. Run Storybook and open `cc-search-bar`, story `withAliases`.
2. Type `postgresql` or `pulsar_9a84bf38`. The matching add-ons show up even though their label does not contain that text.
3. Check that aliases appear nowhere in the rendered output.

### DIFF
--- a/src/components/cc-search-bar/cc-search-bar.js
+++ b/src/components/cc-search-bar/cc-search-bar.js
@@ -127,7 +127,11 @@ export class CcSearchBar extends LitElement {
   /**
    * Filters `this.sections` based on `this.value`, with the grammar shared with the console sidebar
    * (see `lib/filter-sections.js`): `is:`/`project:` tokens are matched against an item's matchers,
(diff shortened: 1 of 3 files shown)


clara-layus-cc (approved):
LGTM, thanks for the improvement.

PR #1823, opened 2026-09-15
fix(cc-email-list): stop displaying the primary address among the secondary ones

clara-layus-cc:
# What does this PR do?

- Filter the primary address out of the secondary address list, as `GET /v2/self/emails` returns every address of the user including the primary one,
- Stop swapping the promoted address with the former primary one when marking an address as primary: the API only promotes the given address, so it is now removed from the secondary addresses,

# How to review?

- Check the commits,


pdesoyres-cc:
Closing since this is an API problem

PR #1824, opened 2026-09-16
feat(cc-logs): allow to set the initial log message filter

clara-layus-cc:
# What does this PR do?

- Turn the log message filter into a public `messageFilter` property on `cc-logs-app-runtime`, `cc-logs-app-access` and `cc-logs-addon-runtime`, so that a consumer can restore a previously saved filter,
- Add an optional `messageFilter` smart param to the three components and document it,
- Dispatch `cc-logs-message-filter-change` when inspecting a log resets the filter, so that consumers are notified of every filter change instead of keeping a stale one,
- Fix the `filter` property of `cc-logs-message-filter`, declared as `String` while it holds an object.

# How to review?

- Check the commits,
- Check the [preview](https://clever-components-preview.cellar-c2.services.clever-cloud.com/feat/logs-persist-message-filter/index.html)


src/components/cc-logs-addon-runtime/cc-logs-addon-runtime.js
```diff
@@ -117,11 +117,13 @@ export class CcLogsAddonRuntime extends LitElement {
    * @param {CcLogsMessageFilterChangeEvent} event
    */
   _onMessageFilterChange({ detail }) {
-    this._messageFilter = detail;
+    this.messageFilter = detail;
   }
 
   _onLogInspect() {
-    this._messageFilter = { value: '', mode: this._messageFilter?.mode ?? 'loose' };
+    this.messageFilter = { value: '', mode: this.messageFilter?.mode ?? 'loose' };
```
roberttran-cc, line 129: **🧹 chore:** we should add a guard against an empty filter value, so that it does not fire an event multiple times when the value didn't change (same for the other components).

roberttran-cc (approved):
LGTM! Got only a small feedback that could be helpful. I'll approve nonetheless.

src/components/cc-logs-addon-runtime/cc-logs-addon-runtime.js
clara-layus-cc: Good catch, fixed in 924892fb: `_onLogInspect` now returns early when the message filter is already empty (in the three `cc-logs-*` components).

